PR #227, opened 2015-01-03
Upgraded to latest Rust

bvssvni:
Closes https://github.com/PistonDevelopers/image/issues/226

Travis CI will fail because of broken dependencies.


nwin:
That was a lot…

PR #231, opened 2015-01-06
Use unboxed closures.

mbrubeck:
This fixes build errors with rustc nightly, now that boxed closures have been removed.


cmr:
Cool!

PR #232, opened 2015-01-07
Upgrade to latest Rust nightly

ruud-v-a:
This makes the crate compile again with the `ea6f65c5f 2015-01-06` Rust snapshot.


bvssvni:
Thanks!

PR #235, opened 2015-01-10
Fixes for 1.0 alpha

jameson-ernst:
Mostly around macro syntax and the {:?} format specifier. Also had to add old_impl_check to some of the impls until that can be reviewed in more detail.
Also did a sweeping change from int/uint to isize/usize. No consideration was given to whether fixed sized integers would be more appropriate in some contexts; that would take a more thorough audit. This is just a blanket change to the new names.

### DIFF
--- a/examples/opening.rs
+++ b/examples/opening.rs
@@ -19,10 +19,10 @@ fn main() {
     let im = image::open(&Path::new(file.clone())).unwrap();
 
     //The dimensions method returns the images width and height
-    println!("dimensions {}", im.dimensions());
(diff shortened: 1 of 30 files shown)


drmugg123:
Hah, how convenient
https://github.com/PistonDevelopers/image/issues/234

cybergeek94:
I can test this on `img_dup` in a bit and merge if it works. I'm in the middle of another project at the moment.

tomaka:
Can we merge this? There aren't any controversial change in the commit.

PR #236, opened 2015-01-10
Update range syntax

brookst:
Fixes warnings of the form:

    warning: use of unstable item: will be replaced by range notation, #[warn(unstable)] on by default

From rust 1.0.0-alpha update.


bvssvni:
Thanks!

PR #238, opened 2015-01-11
Slice notation

brookst:
A work in progress at the moment, but this fixes some of the `use of unstable item: will be replaced by slice syntax` warnings.

Some of the slicing in `src/jpeg/encoder.rs` looks odd. They can probably be exact indexes into the quantisation tables, but I don't want to muck them up (if I haven't already).

Also in `src/buffer.rs` I'm not sure what the `Container` trait implies, so that is not touched yet.

### DIFF
--- a/examples/fractal.rs
+++ b/examples/fractal.rs
@@ -29,7 +29,7 @@ fn main() {
 
         let mut i = 0;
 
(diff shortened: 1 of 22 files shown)


bvssvni:
Looks good!

Issue #197, opened 2014-12-13
black top-left border after resize with FilterType other than Nearest


pavlov-dmitry:
This is resizing code:

```rust
extern crate image;

use std::io::fs::{ File };

fn save_image( filename: &str, img: image::DynamicImage ) {
	let fout = File::create( &Path::new( filename ) ).unwrap();
    let _ = img.save(fout, image::PNG);
}

fn main() {
    let img = image::open(&Path::new("./test.png")).unwrap();

    save_image( "./nearest.png", img.resize_exact( 200, 200, image::Nearest ) );
    save_image( "./triangle.png", img.resize_exact( 200, 200, image::Triangle ) );
    save_image( "./catmullrom.png", img.resize_exact( 200, 200, image::CatmullRom ) );
    save_image( "./gaussian.png", img.resize_exact( 200, 200, image::Gaussian ) );
    save_image( "./lanczos3.png", img.resize_exact( 200, 200, image::Lanczos3 ) );
}
```

This is `test.png`:
![test image](http://i.imgur.com/oDFqpvE.png)

This is `nearest.png`
![nearest](http://i.imgur.com/FratMS5.png)

This is `triangle.png`
![triangle](http://i.imgur.com/Fn53nDu.png)

This is `catmullrom.png`
![catmullrom](http://i.imgur.com/JWwfJde.png)

This is `gaussian.png`
![gaussian](http://i.imgur.com/f47UErT.png)

This is `lanczos3.png`
![lanczos3](http://i.imgur.com/1d71apc.png)

nwin:
Fixed by #239

PR #239, opened 2015-01-15
fix black top-left border after resize (issue #197)

pavlov-dmitry:
fix black top-left border after resize (issue [#197](https://github.com/PistonDevelopers/image/issues/197))

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -170,11 +170,11 @@ fn horizontal_sample<P: Primitive + 'static, T: Pixel<P> + 'static, I: GenericIm
 
             let inputx = (outx as f32 + 0.5) * ratio;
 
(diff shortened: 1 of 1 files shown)


iirelu:
Haha, I was just about to report this annoying bug. I'd love to see this PR merged soon.

nwin:
Nice catch! But I suppose we should use i64 instead to avoid overflow. Could you please change that?

iirelu:
@nwin Why would you be using u32s in the first place if you were worried about images with dimensions larger than a few billion?

pavlov-dmitry:
I think it is almost impossible to overflow in real world, but
this bug is so annoying that I didn't want to argue about overflows. I will change it.

nwin:
@iirelu because u32 is the maximum size supported by all image formats we support.

Thanks, pavlov!

PR #246, opened 2015-01-17
Use associated types

### LINKED ISSUE #240: Use associated types, remove old_orphan_check
Many type parameters — for example `T` on `Pixel<T>`, should actually be associated types. I suspect that it will fix the `old_orphan_check` warnings when implemented.

I am working on this, see https://github.com/ruud-v-a/image/tree/assoc-types.

ruud-v-a:
This still needs some cleanup, but this is going to be quite a big change, so I’m posting it here for early feedback. This resolves #240. There are a lot of breaking changes, so we might want to bump the version number when this can be merged.

- This uses associated types where appropriate. The most important changes are that the type parameters of `Pixel` and `GenericImage` are now associated types. `ImageBuffer` requires less type params.
- This resolves the `old_orphan_check` and `old_impl_check` warnings.

Associated types feel like the “right” way to go. In many places it simplifies the API, but naming the subpixel type has become uglier: for `I: GenericImage`, the subpixel type is `<I::Pixel as Pixel>::Subpixel`, instead of a `T` param that was often used before. However, the `T` param was often unconstrained, and there is no need for it with associated types.

A few issues:
- The return type of `into_dynamic` cannot be `ImageBuffer` any more. I removed the function.
- Many functions now require a `'static` bound on `I: GenericImage`. How will this impact users of the library, and is there a better way to do it?
- Should `Pixel::Subpixel` and `GenericImage::Pixel` have `'static` bounds? I had hoped that adding those would enable omitting the `where P::Subpixel: 'static` in most places, but unfortunately this is not the case.
- The pixel type was sometimes allowed to be unsized. What is the use of that?

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -484,7 +484,7 @@ impl<P: Pixel + 'static> ImageBuffer<P, Vec<P::Subpixel>>
 }
 
 /// Provides color conversions for whole image buffers.
-pub trait ConvertBuffer<T: ?Sized> {
+pub trait ConvertBuffer<T> {
     /// Converts `self` to a buffer of type T
     ///
     /// A generic impementation is provided to convert any image buffer to a image buffer
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1,5 +1,4 @@
 use std::slice::{ Chunks, ChunksMut };
-use std::any::Any;
 use std::ops::{ Index, IndexMut };
 use std::num::Int;
 use std::intrinsics::TypeId;
@@ -147,14 +146,13 @@ impl<'a, P: Pixel> DoubleEndedIterator for Pixels<'a, P::Subpixel, P>
 }
 
 /// Iterate over mutable pixel refs.
-pub struct PixelsMut<'a, T: 'static, PixelType: ?Sized> {
-    chunks: ChunksMut<'a, T>
+pub struct PixelsMut<'a, P: Pixel + 'a> where P::Subpixel: 'a {
+    chunks: ChunksMut<'a, P::Subpixel>
 }
 
-impl<'a, P: Pixel> Iterator for PixelsMut<'a, P::Subpixel, P>
-    where P::Subpixel: Primitive {
-
+impl<'a, P: Pixel + 'a> Iterator for PixelsMut<'a, P> where P::Subpixel: 'a {
     type Item = &'a mut P;
+
     #[inline(always)]
     fn next(&mut self) -> Option<&'a mut P> {
         self.chunks.next().map(|v| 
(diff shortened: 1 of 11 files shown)


nwin:
The Sized bound can go. It is just an artifact introduced during development.

src/buffer.rs
```diff
@@ -362,29 +365,14 @@ where Container: ArrayLike<T>, T: Primitive + 'static, PixelType: Pixel<T> + 'st
...
-    #[experimental]
-    pub fn into_dynamic(self) -> ImageBuffer<Container, T, &'static Any> {
-        ImageBuffer {
-            data: self.data,
-            width: self.width,
-            height: self.height,
-            type_marker: self.type_marker
-
-        }
-    }
```
nwin: This was useless anyway since I did not implement the accompanying code yet. But I think it should work when implementing Pixel for Any…

src/buffer.rs
```diff
@@ -406,54 +395,62 @@ where Container: ArrayLike<T>, T: Primitive + 'static, PixelType: Pixel<T> + 'st
...
 // concrete implementation for `Vec`-baked buffers
-impl<T, PixelType> ImageBuffer<Vec<T>, T, PixelType>
-where T: Primitive + 'static, PixelType: Pixel<T> + 'static {
-    /// Creates a new image buffer based on a `Vec<T>`.
-    pub fn new(width: u32, height: u32) -> ImageBuffer<Vec<T>, T, PixelType> {
+// TODO: I think that rustc does not "see" this impl any more: the impl with
+// Container meets the same requirements. At least, I got compile errors that
+// there is no such function as `into_vec`, whereas `into_raw` did work, and
+// `into_vec` is redundant anyway, because `into_raw` will give you the vector,
+// and it is more generic.
```
nwin: I'm a bit worried about that. Is this a compiler bug or something I did wrong? Does `rustc` still find `new`? If not this is bad. How is this supposed to work? Afaik one can’t just add a generic constructor to the ArrayLike trait because static methods are not object safe. 

PS: I just added `into_vec`  etc. to match the stdlib.

src/buffer.rs
ruud-v-a: Yes, in that case it would work.

src/buffer.rs
ruud-v-a: I don’t know yet whether it does not break anything, building in debug mode causes an ICE, but at least the library compiles in release mode, so `new` still works. I think that we should not have two identical methods, but removing `into_vec` does not belong in this PR either way.

nwin: Did you file a bug to rustc about that?

ruud-v-a: The report is [here](https://github.com/rust-lang/rust/issues/20797). I’m not working around it, it’s just that it prevents me from running the test suite until it is fixed.

nwin: I see… :/

TyOverby:
I was just about to start writing a change that was very similar to this.  Is there an update on being able to build the library?  I'd very much like to see this merged!

ruud-v-a:
I’m sorry, I have been away for a few days. The blocking issue on rustc has been closed, but there were some breaking changes as well. I’ll try to get it to compile again as soon as possible.

ruud-v-a:
This compiles with the latest nightly, and the test suite compiles and passes as well. There is still a bit of cleanup to be done.

nwin:
Great work! I'll merge it for now, the cleanup we could do later.

bvssvni:
@ruud-v-a You're awesome!

PR #253, opened 2015-01-21
TypeId has moved to 'any' module

AngryLawyer:
TypeId now lives in 'any', rather than 'intrinsics' - this simple PR fixes it.

PR #254, opened 2015-01-21
Update to latest Rust.

toqueteos:
`std::intrinsics::TypeId` stabilized and is now available at `std::any::TypeId`.

Waiting for Travis.


toqueteos:
Duplicate of #253.

PR #259, opened 2015-01-23
Update to Rust nightly 2015-01-23

mbrubeck:
Fixes build error in latest Rust nightly.


nwin:
Thanks.

nwin:
Ok. I checked, the compiler does not complain. As long as the type inferfence works, I would prefer to leave it like this.

The other changes where already merged by @tomaka.

Thanks for your contribution.

PR #263, opened 2015-01-25
Add lzw to tiff

nwin:
Adds LZW support for TIFF. This lib is now able to read screenshots taken by OSX and the GMARBLES.tiff test image.

LZW function have been moved to a separate module since the functionality is shared between gif and tiff. Note that giff and tiff use different algorithms. The two implementations are specialized by a macro.


bvssvni:
:balloon:

nwin:
Now I just need these PackBits and this obscure fax compression scheme… :)

PR #271, opened 2015-01-28
Add convenient save method on ImageBuffer

Ryman:
Also standardise on the spelling of Gray across the repo. I chose Gray as the TIFF spec has tags with the American spelling.

### DIFF
--- a/examples/fractal.rs
+++ b/examples/fractal.rs
@@ -1,5 +1,5 @@
 //!An example of generating julia fractals.
-#![allow(unstable)]
+#![feature(path, io)]
 
 extern crate num;
 extern crate image;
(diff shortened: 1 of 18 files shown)


nwin:
I think Gray is a good choice, we are using AE anyway (--> color).

Could you please remove the unsafe block from the saving method that is not necessary. If you have borrowing problem you have to save the width and height to a variable first.

Ryman:
It's not a borrowing problem, it was a coherence thing, but I finally figured out the required syntax, so it's unnecessary now :) Pushed updated commits, pending on travis.

nwin:
Can you please rebase to the latest rustc change?

Ryman:
@nwin Rebased and removed some warnings (last commit). I didn't add feature flags for rustc_private as it' probably something that might actually be worth investigating.

nwin:
Omg…what happened to rust?

Thank you for your contribution!

PR #287, opened 2015-02-08
Fix IndexMut implementation for rustc 0b6dbbc

robo9k:
[`rustc` 0b6dbbc9cfb747df1db646bba16561c022704056](https://github.com/rust-lang/rust/commit/0b6dbbc9cfb747df1db646bba16561c022704056) changed how `IndexMut` works; the `type Output = ..` bit is now redundant and can be removed from implementations.

With this change in place, `cargo build && cargo test` work again for:
```
$ rustc --version
rustc 1.0.0-nightly (74b874071 2015-02-08 00:24:03 +0000)
```

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -214,7 +214,6 @@ impl<T: Primitive> Index<usize> for $ident<T> {
 }
 
 impl<T: Primitive> IndexMut<usize> for $ident<T> {
-    type Output = T;
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


nwin:
Hmm, I closed a similar PR this morning, interesting…

robo9k:
You mean #284? I didn't see that one due to lack of proper description.
Either way, the `rustc` nightly mentioned in above is the current one.

nwin:
I just care about travis and it did not work this morning.

Issue #199, opened 2014-12-13
Handle transparency in PNG decoding


ruud-v-a:
I encountered this in `src/png/decoder.rs`:

                // (b"tRNS", HavePLTE) => {
                //     TODO: handle transparency
                // }

I am not familiar enough with the codebase to fix this myself yet, but will add the issue number to the todo in the code.

retep998:
I recently ran into some images that image was unable to decode the transparency for.
Is this issue related to this image?
![](http://puu.sh/er04j/a8b3960239.png)
When I try working with that image it gives me an opaque black background instead of a transparent background as it should.

kmcallister:
Unfortunately this is blocking adoption in Servo, because a `tRNS` chunk is used in Acid2. The base64 for that PNG file is:

```text
iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAABnRSTlMAAAAAAABupgeRAAAABmJL
R0QA/wD/AP+gvaeTAAAAEUlEQVR42mP4/58BCv7/ZwAAHfAD/abwPj4AAAAASUVORK5CYII=
```

> The most strait forward solution would probably be to add an alpha channel to the image and selectively set the value to 0. But it's a bit tricky because we would have to lie the outside an use the real color type internally.

This would be ideal from Servo's perspective. We convert RGB8 to RGBA8 in any case. At this stage we are not too worried about the overhead of copying the data.

nwin:
Maybe I should just hack it in. We have to rewrite the decoder anyway to support streaming at some point.

PR #290, opened 2015-02-12
Add color quantizer

nwin:
Ported the NeuQuant color quantizer to Rust. As expected, better results than gimp. It is the default for gif now.

Full support for alpha channel. Just our png encoder doesn’t write indexed colors yet so we don’t save any space.

Us:
lena_nq
GIMP:
lena_gimp


ruud-v-a:
Nice!

bvssvni:
@nwin You should post this on /r/rust!

nwin:
I wanted to wait until at least the png decoder can read and write such images images…see also #199

PR #293, opened 2015-02-13
Use Vec instead MemWriter

Potpourri:


nwin:
I think we have to check the code at some point when 1.0 is released. I hope that & mut[T] will implement reader soon as well…

PR #296, opened 2015-02-15
ICE fix attempt

tomaka:
This is just a wild guess for the cause of the ICE as I don't have the latest nightly.


nwin:
The compiler knows about the trait impl that is not the problem.

tomaka:
There are a lot of problems when you use traits that are in private modules.

PR #298, opened 2015-02-15
Save one allocation per line in the PNG encoder

nwin:


mbrubeck:
I think this introduced a bug.  See #305 for details.

PR #301, opened 2015-02-20
ImageBuffer::from_fn doesn't need a Box

retep998:
This also managed to fix some really weird lifetime errors I was suddenly getting due to a Rust upgrade.


bvssvni:
Thanks!

PR #306, opened 2015-02-26
Fix regressions in PNG filter

### LINKED ISSUE #305: Filter regression in PNG encoding
#298 created a regression that causes some images to be encoded incorrectly as PNG.  This following test case demonstrates the problem.  The PPM file is correct, but every other row of the PNG file is incorrect.

```rust
let white = image::Pixel::from_channels(255,255,255,255);
let black = image::Pixel::from_channels(0,0,0,255);
let img = image::ImageRgba8(image::ImageBuffer::from_fn(100, 100, move |x, y| {
    if (x + y) % 2 == 0 { white } else { black }
}));
img.save(&mut File::create(&Path::new("test.png")).unwrap(), image::PNG);
img.save(&mut File::create(&Path::new("test.ppm")).unwrap(), image::PPM);
```

Reverting #298 fixes the bug.

Looking at that patch, I think that [this line](https://github.com/PistonDevelopers/image/commit/e01fecbf75eff4a8939e68c49abfb6ab7ed7bce4#diff-e1a0e41840ecaefa415299d6f179b679R90) and [this line](https://github.com/PistonDevelopers/image/commit/e01fecbf75eff4a8939e68c49abfb6ab7ed7bce4#diff-e1a0e41840ecaefa415299d6f179b679R99) are incorrect, since `current[i - bpp]` is not the same as `orig[i - bpp]` in the previous code; it is a value that may have been changed by the preceding `for` loop.

mbrubeck:
Fixes #305.  This partially reverts #298, but keeps the correct parts.

### DIFF
--- a/src/png/filter.rs
+++ b/src/png/filter.rs
@@ -1,5 +1,4 @@
 use std::num::SignedInt;
-use std::iter::range_step_inclusive;
 
 #[derive(FromPrimitive, Debug)]
 pub enum FilterType {
@@ -72,7 +71,7 @@ pub fn filter(method: FilterType, bpp: usize, previous: &[u8], current: &mut [u8
     match method {
         FilterType::NoFilter => (),
         FilterType::Sub      => {
-            for i in range_step_inclusive(len-1, bpp, -1) {
+            for i in (bpp..len).rev() {
                 current[i] = current[i] - current[i - bpp];
             }
         }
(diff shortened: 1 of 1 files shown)


nwin:
Please remove the assertion and the clones. To patch the issue order of the two for loops in the Avg and  Paeth case has to be inverted. Then everything should be fine.

mbrubeck:
> Please remove the assertion and the clones. To patch the issue order of the two for loops in the Avg and Paeth case has to be inverted. Then everything should be fine.

Done.

bvssvni:
Thanks!

PR #307, opened 2015-02-27
Fix test build errors

mbrubeck:


nwin:
Thanks, but I’m not sure if it is a good idea yet to switch to crates.io. Can you please invert the comments in Cargo.toml?

nwin:
Never mind, let’s see how stable Rust got…

PR #308, opened 2015-02-27
Add some missing rev() calls

mbrubeck:
This is a follow-up to #306.  I forgot to keep these loops in reverse order.

This code needs tests; if I have time later I will write some.


mbrubeck:
Note: the Travis failure is fixed by #307.

nwin:
Thanks!

PR #310, opened 2015-03-01
Infer closure kinds

ruud-v-a:
The syntax was obsolete and generated a compiler error on the latest nightly. Also, in a few places a closure was not even required.


Potpourri:
Thanks!

PR #315, opened 2015-03-04
Fixed panics from arithmetic overflow

stjahns:
Fixes https://github.com/PistonDevelopers/image/issues/314. Overflow triggers panics in `rustc 1.0.0-nightly (fed12499e 2015-03-03) (built 2015-03-04)`, need to use `std::num::wrapping::Wrapping` when overflow is expected/intentional. See https://github.com/rust-lang/rand/commit/929ea771f2bd505bf4ce872bd43e3afc098241ae for a related fix.

Not sure how complete test coverage is in this lib, there could be more places this might blow up, but it's working for me now.

### DIFF
--- a/src/png/decoder.rs
+++ b/src/png/decoder.rs
@@ -8,6 +8,7 @@ use std::slice;
 use std::old_io::IoResult;
 use std::old_io::MemReader;
 use std::num::FromPrimitive;
+use std::num::wrapping::Wrapping;
 
 use image::{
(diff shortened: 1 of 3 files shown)


bvssvni:
Thanks!

bvssvni:
@stjahns btw, I recommend working against personal forks instead of pushing to a branch

stjahns:
@bvssvni - can do!

PR #318, opened 2015-03-08
Refactor math::nq module

hauleth:


src/math/nq.rs
```diff
@@ -48,35 +49,15 @@ const BETAGAMMA: f64 = BETA * GAMMA;
 const PRIMES: [usize; 4] = [499, 491, 478, 503];
 
 #[derive(Copy)]
-struct Neuron {
-    r: f64,
-    g: f64,
-    b: f64,
-    a: f64,
+struct Template<T> {
```
nwin: Better name? Maybe Quad?

src/math/nq.rs
```diff
@@ -31,16 +31,17 @@
...
 
 const RADIUS_DEC: i32 = 30; // factor of 1/30 each cycle
 
 const ALPHA_BIASSHIFT: i32 = 10;            // alpha starts at 1
-const INIT_ALPHA: i32 = 1<<ALPHA_BIASSHIFT; // biased by 10 bits
+const INIT_ALPHA: i32 = 1 << ALPHA_BIASSHIFT; // biased by 10 bits
 
 const GAMMA: f64 = 1024.0;
-const BETA: f64 = 1.0/1024.0;
+const BETA: f64 = 1024.0.recip();
```
nwin: I would leave it like it was, it is more obvious what it does. I also have the feeling that this function will either go away or change its name to reciproke.

hauleth: I have fixed this.

src/math/nq.rs
hauleth: It would be better.

nwin:
Can you please box the array? I had some bad experiences with stack sizes recently.

hauleth:
If it needed to be boxed type then `Vec ` would be fine, but IMHO there is no need for either.

nwin:
Well, `Vec` is fine. There is no gain in using an fixed size array. It only saves one allocation. Point is, that the struct is now 2kb big.

hauleth:
I've brought `Vec` back, but changed syntax in `new`.

nwin:
(y) the WIP still holds?

hauleth:
Still holds as I want to see if I can refactor sort and search. Probably some methods change their name.

bvssvni:
Looks good!

@hauleth The new io has some dependency issues and we might want to wait until those get fixed to check if it works.

bvssvni:
@hauleth @nwin Can we merge this now?

hauleth:
I have no idea for now. Maybe later I find way to cleanup it more (or even extract it to another crate).

bvssvni:
Nice!

PR #319, opened 2015-03-09
Fix integer overflow

rsaarelm:
Don't try to create the stride binding when you would be subtracting 1 from an unsigned 0 value.

Trying to write a PNG image panics with integer overflow here otherwise.


nwin:
Thanks, this change in overflow semantics just causes trouble… :(

PR #321, opened 2015-03-09
Move to new `io`

hauleth:
# WIP - DO NOT MERGE

Hell lot of compile errors. It is just for TravisCI reports.

### DIFF
--- a/src/utils/bitstream.rs
+++ b/src/utils/bitstream.rs
@@ -2,7 +2,8 @@
 
 use std::io::{
     self,
-    Read
+    Read,
+    Write
 };
 
 /// Bit reader
@@ -12,7 +13,7 @@ pub trait BitReader: Read {
 }
 
 /// Bit writer
-pub trait BitWriter: Writer {
+pub trait BitWriter: Write {
     /// Writes the next `n` bits.
     fn write_bits(&mut self, v: u16, n: u8) -> io::Result<()>;
 }
@@ -128,13 +129,13 @@ macro_rules! define_bit_writers {
 
 #[$doc]
 #[allow(dead_code)]
-              pub struct $name<'a, W> where W: Writer + 'a {
+              pub struct $name<'a, W> where W: Write + 'a {
                   w: &'a mut W,
                   bits: u8,
                   acc: u32,
               }
 
-              impl<'a, W> $name<'a, W> where W: Writer + 'a  {
+              impl<'a, W> $name<'a, W> where W: Write + 'a  {
                   /// Creates a new bit reader
                   #[allow(dead_code)]
                   pub fn new(writer: &'a mut W) -> $name<'a, W> {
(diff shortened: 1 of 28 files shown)


nwin:
I would wait with that. The docs still recommend using `old_io`. We also have to careful review this, because the semantics of `io` change. For example EOF is not an error any more and I don’t remember whether we exploited that somewhere.

src/utils/bitstream.rs
```diff
@@ -1,72 +1,75 @@
 //! This module provides a bit reader
 
-use std::old_io;
-use std::old_io::{IoResult};
+use std::io::{
+    self,
```
nwin: Is this new? I didn’t know you could do that.

hauleth:
It is preparing for change. I know that this take time to stabilize new `std::io`. But this will be hell lot of work to change it afterwards. We can do some of it aheado of time.

nwin:
I see your point. I just don’t want to force the people depending on `image` to use `io` as long as the docs recommend against using it. It’s good that you prepare everything but I don’t think we should merge before they remove that warning.

hauleth:
That's why there is big `WIP - DO NOT MERGE` at the top. I will make it bigger.

src/utils/lzw.rs
```diff
@@ -268,8 +270,8 @@ impl EncodingDict {
     }
 }
 
-pub fn encode<R, W>(mut r: R, mut w: W, min_code_size: u8) -> old_io::IoResult<()>
-where R: Reader, W: BitWriter {
+pub fn encode<R, W>(mut r: R, mut w: W, min_code_size: u8) -> io::Result<()>
```
nwin: This function has to be rewritten. `while let Ok(c) = ...` will be an infinite loop with `std::io`;

bvssvni:
@hauleth Thanks for working on this!

nwin:
Absolutely! 

One other remark it is probably the easiest to copy the read_le_u16() methods from old_io into a new extention trait and implement it for all `Read` like we did for `Primitive`. That will save a lot of pain.

hauleth:
@nwin, there is no need in creating that traits. There already is [podio](https://github.com/mvdnes/podio). We can use this.

jpernst:
Looks like the warning to continue using `old_io` has been removed, so this should probably be high priority. I imagine lots of people are going to take a look at rust during beta 1, particularly for game dev, so we should try to have this ready for the influx. I took a stab at doing this conversion a few weeks ago but it's such a mammoth change for a codebase I don't know very well that I quickly got buried in the avalanche of errors and just swapped it out for a C library binding. I'd prefer the all rust solution, but I'm not sure how to make the changes incrementally without the whole thing coming apart.

bvssvni:
I opened https://github.com/PistonDevelopers/image/issues/333

src/utils/bitstream.rs
bombless: @nwin It was added months ago.
See https://github.com/rust-lang/rust/pull/20365

bvssvni:
Fixed by https://github.com/PistonDevelopers/image/pull/339

PR #330, opened 2015-03-16
Use cargo features to conditionally compile codecs.

### LINKED ISSUE #311: Improve build times.
I was wondering if there was any ideas on how the build times could be improved for this crate. It currently takes about a 1minute to compiler for a none release build.

Some ideas:
- Move the image codecs into sub crates
- Add `feature` flags to disable codecs that are not used

huonw:
The default continues to have all codecs activated, but a user can
opt-in to only supporting specific ones, via features, e.g. depending on
`image`, while only supporting JPEG and GIF:

    [dependencies.image]
    version = "0.2"
    default-features = false
    features = ["jpeg", "gif"]

On my computer `cargo build` takes ~95s, while `cargo build --features
jpeg --no-default-features` (only JPEG) takes ~28s.

This patch also updates the travis file to also compile the library with
each feature, to ensure they continue to compile. These codec features
are all independent, so testing each one individually is sufficient.

Closes #311.


nwin:
Great thanks!

It’s still a bit puzzling why JPEG alone takes 30s…

bvssvni:
@huonw You are awesome!

kvark:
sweet, thanks!

PR #336, opened 2015-03-21
Rustup

tomaka:
Items like `b"GIF"` are now of type `[u8; 3]` instead of `[u8]` before.
This lead to the comparison not working.


bvssvni:
Thanks!

PR #339, opened 2015-03-24
Switch to new I/O

tomaka:
Three and a half hour of work, but it's there!

I started from scratch, as this approach was probably easier than rebasing #321 

It adds a dependency to the `byteorder` crate, which provides equivalents to `read_le_u16` and similar.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -5,6 +5,8 @@ use std::io;
 use std::slice;
 use std::iter::repeat;
 
+use byteorder;
+
 use color;
 use color::ColorType;
 use buffer::{ImageBuffer, Pixel};
@@ -59,6 +61,15 @@ impl fmt::Display for ImageError {
 impl FromError<io::Error> for ImageError {
     fn from_error(err: io::Error) -> ImageError {
         ImageError::IoError(err)
+    }
+}
+
+impl FromError<byteorder::Error> for ImageError {
+    fn from_error(err: byteorder::Error) -> ImageError {
+        match err {
+            byteorder::Error::UnexpectedEOF => ImageError::ImageEnd,
+            byteorder::Error::Io(err) => ImageError::IoError(err),
+        }
     }
 }
 
--- a/src/image.rs
+++ b/src/image.rs
@@ -1,7 +1,7 @@
 use std::error::FromError;
 use std::fmt;
 use std::mem;
-use std::old_io;
+use std::io;
 use std::slice;
 use std::iter::repeat;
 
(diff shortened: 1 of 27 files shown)


src/image.rs
```diff
@@ -56,12 +58,21 @@ impl fmt::Display for ImageError {
...
+impl FromError<io::Error> for ImageError {
+    fn from_error(err: io::Error) -> ImageError {
         ImageError::IoError(err)
     }
 }
 
+impl FromError<byteorder::Error> for ImageError {
+    fn from_error(err: byteorder::Error) -> ImageError {
+        match err {
+            byteorder::Error::UnexpectedEOF => ImageError::ImageEnd,
```
tomaka: I'm not sure about this. `ImageEnd` means "unexpected end", right?

bvssvni:
:balloon: @tomaka You are awesome!

src/image.rs
bvssvni: It seems it means the opposite. When it reads more data than allocated for the image. https://github.com/PistonDevelopers/image/blob/5df4d665ec2824b02e8ee0ac76c1d383a1e8e142/src/webp/decoder.rs#L112

@nwin Is this right?

nwin: This error is a bit fuzzy, I would say both cases are ok.

tomaka:
I'm using this branch for my local videogame project, and everything seems to work fine.

nwin:
Then let’s merge it, if something breaks we should think about better testing.

But before merging: Did we already push the latest old_io code to crates.io?

src/image.rs
tomaka: The underlying problem is that `ErrorKind::EndOfFile` no longer exists in new I/O.
If you want to read at least N bytes, but EOF happens before, it used to return this error.

That's why we need a replacement for it (here and at some other places). I picked `ImageEnd` but was not really sure about it.

tomaka:
The only thing that hasn't been uploaded to crates.io is #340 (touches only the README, doesn't really matter on crates.io)

I'd suggest bumping to 0.3 as it's a breaking change.

nwin:
although we are allowed to do anything we want in 0.x that sounds sensible

bvssvni:
Let's merge!

tomaka:
So what's the deal with the version number? 0.2.5 or 0.3.0?

bvssvni:
We can change it to 0.3.0.

tomaka:
I didn't understand this:
> although we are allowed to do anything we want in 0.x that sounds sensible

bvssvni:
@nwin I guess you meant that breaking changes are expected in 0.x, but since people could be having "0.2.*" in the Cargo.toml there is no reason not to go 0.3 with a such big change?

tomaka:
Wait, I'm stupid. For some reason I thought @nwin was against 0.3. Nevermind.

Let's publish 0.3.0 then?

bvssvni:
@tomaka If you got it working on your rustc then please do it!

tomaka:
Done

PR #345, opened 2015-03-26
PathBuf::new -> PathBuf::from

stjahns:
Fixes test compilation failures. One test still failing from an overflow on a shift operation, however..


nwin:
Hmm…this is weird. It compiles fine on my system…

tomaka:
This is a duplicate of #346.

PR #364, opened 2015-04-02
Resolve warnings for latest nightly

ruud-v-a:
The crate now compiles without warnings again. For the most part, this deals with `num::Int` et al. being deprecated.


tomaka:
Builds without warnings sometimes trigger the timeout on travis (10mn without any output on stdout).

ruud-v-a:
That is unfortunate, but I don’t think not resolving warnings is the right solution. We could add a `#[feature(supercalifragilisticexpialidocious)]` somewhere to deliberately generate a non-existing feature warning, but I am not sure whether that would break the silence at the right moment.

PR #365, opened 2015-04-03
Stop relying on implementation defined structure layout.

nwin:
Rust does not make any guarantees about how tuple structs are layed out in memory. Using C-type-structs mediates this potential future problem.


nwin:
Please do not merge wait for #367. That is more important.

PR #366, opened 2015-04-03
Remove unsafe blocks in png decoder

### LINKED ISSUE #362: Memory safety issues from mutable aliasing
`png/decoder.rs` contains this snippet copy-pasted a few times:

```rust
// Unsafe copy create two views into the vector
// This is unproblematic since it is only locally to this function and a &[u8]
let data = unsafe {
    let view: &mut [u8] = mem::transmute_copy(&buf);
    &view[..entries]
};
```

Are you *sure* this is safe? `unsafe` blocks [must avoid undefined behavior](http://doc.rust-lang.org/reference.html#behavior-considered-undefined) which includes aliasing of `&mut`, full stop. LLVM can optimize under the assumption that such aliasing does not happen.

The `MutPixels` iterator may have a similar issue. Using `transmute_copy` to circumvent aliasing rules seems to be a common pattern in the library.

I think it *would* be safe to split a slice into mutually-disjoint immutable slices, then transmute them to `&mut` slices. And you can provide a safe interface to this -- I've seen it done for data parallelism, in particular. I don't know if that covers all of your use cases, however.

nwin:
closes #362

### DIFF
--- a/src/png/decoder.rs
+++ b/src/png/decoder.rs
@@ -571,59 +571,56 @@ fn expand_pass(
     }
 }
 
-#[inline(always)]
-fn expand_packed<F>(buf: &mut [u8], channels: usize, bit_depth: u8, func: F)
-where F: Fn(u8, &mut[u8]) {
-    let entries = buf.len()/channels/(8/bit_depth as usize);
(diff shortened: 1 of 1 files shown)


bvssvni:
Nice!

PR #367, opened 2015-04-03
Remove some feature gates

tomaka:
Remaining features:

 - `core`: 154 compilation errors when I remove this one. Probably fixable, but that's a lot of work.
 - `std_misc`: several small things. Probably fixable too but I don't have time right now.
 - `step_by`: feature-gated function because added recently. Need to check what the plan is for this one. If it's not getting stable, it is possible to write a replacement.
 - `rustc_private`: used for the `flate` library. There is a replacement `flate2` on crates.io, however it's just bindings for miniz.c, which would force users to have a C compiler installed. I don't like this. Probably the most problematic feature.
 - `test`: not sure how to have access to the bencher without this one.

### DIFF
--- a/src/tiff/decoder.rs
+++ b/src/tiff/decoder.rs
@@ -423,17 +423,36 @@ impl<R: Read + Seek> ImageDecoder for TIFFDecoder<R> {
     }
 
     fn colortype(&mut self) -> ImageResult<ColorType> {
-        match (&*self.bits_per_sample, self.photometric_interpretation) {
-            // TODO: catch also [ 8,  8,  8, _] this does not work due to a bug in rust atm
-            ([ 8,  8,  8, 8],  PhotometricInterpretation::RGB) => Ok(ColorType::RGBA(8)),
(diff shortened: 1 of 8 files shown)


src/tiff/decoder.rs
```diff
@@ -418,17 +418,36 @@ impl<R: Read + Seek> ImageDecoder for TIFFDecoder<R> {
...
+                   self.photometric_interpretation == PhotometricInterpretation::WhiteIsZero)
+        {
+            Ok(ColorType::Gray(self.bits_per_sample[0]))
+
+        } else {
+            return Err(::image::ImageError::UnsupportedError(format!(
+                "{:?} with {:?} bits per sample is unsupported", self.bits_per_sample, self.photometric_interpretation
+            ))) // TODO: this is bad we should not fail at this point}
         }
     }
```
nwin: Is it possible to rewrite this as a match?

    PhotometricInterpretation::RGB if self.bits_per_sample == [8, 8, 8, 8] => ...

it is super unintuitive like this

tomaka: Fixed

nwin:
Great

PR #368, opened 2015-04-03
Update for changes to Neg and Copy

andersk:


andersk:
Depending on the response to my comments on rust-lang/num#76, we might not be required to replace `Ratio<u16>`, though future animation formats might want a larger type here anyway(?).

nwin:
Ratio<u16> covers the range of APNG and is well above that range of GIF. Everything that exceeds the range of Ratio<u16> is not relevant for human perception.

Potpourri:
Thanks!

PR #372, opened 2015-04-04
Upgrade to Rust nightly

bvssvni:
Still some issues left, but it is a start.


tomaka:
So what's the plan for the `flate` library?

bvssvni:
Didn't even get that far. It needs the `Reflect` trait for `Pixel`.

PR #373, opened 2015-04-05
Replace flate with flate2; remove #![feature(rustc_private)]

andersk:


andersk:
Closing in favor of another PR that I’m about to open.

PR #375, opened 2015-04-05
Replace flate with flate2; remove #![feature(rustc_private)]

andersk:


tomaka:
:'(

That's a very bad change. Is there really no pure-Rust implementation of flate/deflate?

andersk:
`flate2` binds the same miniz library that `flate` binds.

tomaka:
It's not a matter of binding. `flate2` requires `gcc` to be installed, which is a pain on Windows.

andersk:
I see.

Well, we could try to make it configurable, such that you can choose between `flate` if you’re on the nightly channel and are allowed to use `#![feature(rustc_private)]`, or `flate2` if you have `gcc`?

nwin:
@tomaka that would be great if you were motivated to do that. The easiest would probably be to port miniz to Rust. I had short look, well over 50% of the code is just bloat (it contains a PNG encoder, zip file reader etc…).

We just need these functions and it’s dependencies (copied from flate2):

    extern {
        pub fn mz_deflateInit2(stream: *mut mz_stream,
                               level: libc::c_int,
                               method: libc::c_int,
                               window_bits: libc::c_int,
                               mem_level: libc::c_int,
                               strategy: libc::c_int) -> libc::c_int;
        pub fn mz_deflate(stream: *mut mz_stream, flush: libc::c_int) -> libc::c_int;
        pub fn mz_deflateEnd(stream: *mut mz_stream) -> libc::c_int;
    
        pub fn mz_inflateInit2(stream: *mut mz_stream,
                               window_bits: libc::c_int) -> libc::c_int;
        pub fn mz_inflate(stream: *mut mz_stream, flush: libc::c_int) -> libc::c_int;
        pub fn mz_inflateEnd(stream: *mut mz_stream) -> libc::c_int;
    
        pub fn mz_crc32(crc: libc::c_ulong, ptr: *const u8,
                        len: libc::size_t) -> libc::c_ulong;
    }

I would tend to close this PR. Using flate2 is not an option.

tomaka:
It looks like deflate is [already implemented](https://github.com/PistonDevelopers/image/blob/master/src/png/deflate.rs).

nwin:
Indeed. If you consider porting it and wonder about the license just note that [CC0](http://directory.fsf.org/wiki/License:CC0) is probably a better choice than unlicense.

tomaka:
Note that with easter celebrations, I probably won't do that today.

nwin:
closing in favor of #378

Issue #376, opened 2015-04-05
ImageEnd reading many PNGs


OneSadCookie:
Attached image and 29 other PNGs in my project fail to load with

     "image 0.3.6 (git+https://github.com/PistonDevelopers/image)"

though they previously loaded with

    "image 0.2.0-alpha.10 (git+https://github.com/PistonDevelopers/image)"

They fail with `ImageEnd` on OS X (Yosemite, 10.10.3 beta)

    rustc --version
    rustc 1.0.0-nightly (f207ecbe0 2015-04-03) (built 2015-04-03)

![characterwalking1](https://cloud.githubusercontent.com/assets/37714/6996035/cd21161c-dbc2-11e4-81b5-6e808e6ee4bf.png)

OneSadCookie:
(reencoding the files with Preview's "Export" allows them to load correctly, too)

nwin:
I convinced myself that this has the same cause as #300. Closing as duplicate.

Issue #300, opened 2015-02-17
Unknown filter type


OneSadCookie:
OK, I figured out what I was doing dumb in the previous bug report, and it was that I just gave the wrong image :p I can reproduce the "unknown filter type" error with this png:

![metalfloor](https://cloud.githubusercontent.com/assets/37714/6236555/59318584-b750-11e4-952a-75e231deca65.png)

and this code:

    fn main() {
        let image_path = Path::new("/path/to/MetalFloor.png");
        let image = image::open(&image_path).unwrap();
    }

(on 64-bit OS X)

nwin:
Re-enconding helps. Obscure bug, my first guess would be a bug in the inflater but I have no clue…

nwin:
Bug is somewhere in the inflater…

OneSadCookie:
I see you closed #376 as a dup.  Whether or not the cause is similar to this one, #376 was introduced later — when I reported this one, the images that fail in #376 still loaded fine.

nwin:
Closed via #418.

PR #378, opened 2015-04-05
Remove the `rustc_private` feature

tomaka:
- Removes the `rustc_private` feature
 - Renames `deflate` to `inflate`
 - Adds `deflate` module
 - Adds `ZlibEncoder`
 - ~~Fixes a bug when reading uncompressed blocks in `inflate` (please review this, see `src/png/inflate.rs` in the second commit)~~ (extracted to #379)

The compressor doesn't actually compress anything. The deflate algorithm allows blocks of uncompressed data, so we just write everything in uncompressed data. This should of course be changed later.


nwin:
Nice workaround. ;) If we are lucky  #376 and #300 have the same origin as the bug you found.

tomaka:
I'll submit a different PR for the bugfix.

bvssvni:
@tomaka What was the strange thing you mentioned?

src/png/deflate.rs
```diff
@@ -1,431 +1,89 @@
...
-        self.h.byte_align();
+        assert!(self.buf.len() <= u16::max_value() as usize);
+        let len = self.buf.len() as u16;
+        let nlen = !len;    // one's complement
 
-        let len   = try!(self.h.receive(16));
-        let _nlen = try!(self.h.receive(16));
+        let writer = self.writer.as_mut().unwrap();
+        try!(writer.write_u8(if last { 0x29 } else { 0 }));       // bits and stuff
+                  // TODO: should be 1  ^, but a bug in the PNGDecoder can be bypassed with 0x29
```
tomaka: @bvssvni This

andersk:
`cargo bench` fails since you deleted `examples/fractal.png`.

tomaka:
Oops, my bad.

tomaka:
So.... let's merge this?

nwin:
I just don’t understand your hack with the magic 0x29 does it produce malformed streams or what?

tomaka:
According to the specs, the first bit indicates whether this is the last block, then the next two bits indicate the type of block (here we pass 0).
Then if the type of block is 0, the decoder must ignore the bits remaining in the current byte (so ignore the next 5 bits).

However image's PNGDecoder reads the next 3 bits and does a right shift depending on the value of these bits. This is clearly a bug in the decoder, but it's very specific so I guess there has to be a reason for it to do so.

To be compliant we must write `bXXXXX001` (where `X` can be anything), but in order for the PNGDecoder to decode it we must write `bXX101001`.

nwin:
ok

bvssvni:
@tomaka You are awesome!

PR #380, opened 2015-04-06
Replace many traits using num crate

andersk:


bvssvni:
@andersk You are awesome!

PR #381, opened 2015-04-07
Fix PNG decoder with uncompressed blocks

tomaka:
Follow-up of #379 

Stupid me didn't realize this was just a typo, and replacing `bits` by `num_bits` is the correct behavior.


bvssvni:
:sake:

PR #382, opened 2015-04-07
Replace #[derive(FromPrimitive)] with the enum_primitive crate

andersk:
The built-in `#[derive(FromPrimitive)]` only works with `std::num::FromPrimitive`, which is still unstable.  This significantly reduces our reliance on `#![feature(core)]`.


tomaka:
:heart:

PR #385, opened 2015-04-08
Finish making "image" compatible with beta

tomaka:
I copy-pasted the body of `std::slice::bytes::copy_memory`, the last unstable function that was still in use by "image".

I also added travis testing for the beta (in addition to the nightly). However `cargo test` isn't run for the beta since it uses the unstable bencher.


bvssvni:
Nice! :tada:

PR #391, opened 2015-04-21
Impl std::error::Error for ImageError

jpernst:
The exact purpose of the `description` string isn't entirely clear to me, but this seems to be roughly appropriate based on the documentation.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -2,6 +2,7 @@ use std::fmt;
 use std::mem;
 use std::io;
 use std::iter::repeat;
+use std::error::Error;
 
 use byteorder;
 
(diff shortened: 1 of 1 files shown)


nwin:
Thanks. I think I would use more human-friendly descriptions like "Image End" -> "Unexpected EOF occurred" but this will do it for now. I hope the `Error`-trait will get a `message`-Method at some point that can allocate a string to get a proper description.

jpernst:
I asked about this on IRC today, and the impression I got is that the `Display` impl is what should be used to generate the fully detailed error message, which is already the case here. There does seem to be a bit of general confusion about the best usage of `description`, so it's probably ok to just take a guess until a more clear consensus develops.

PR #393, opened 2015-04-22
Update the tested version on travis

tomaka:
The reason for this change is that byteorder is now broken on beta 1.

Ultimately this should be changed to `stable` after 1.0 is released, but this doesn't work yet.


bvssvni:
Thanks!

PR #397, opened 2015-05-02
Cleanup lib.rs

### LINKED ISSUE #396: Remove the crate_name and crate_type attributes
I don't think these are necessary, and people might want to compile a dynamic library when shipping multiple binaries.

Potpourri:
close #396


tomaka:
Why change the code style in the process? :huh:

Potpourri:
I know what change code style is bad. But this file did not have single style. Also code like `pub use foo::bar as bar` was strange for me.

PR #398, opened 2015-05-02
Use ImageResult instead Result<ImageResult> in return type of DynamicImage::save

Potpourri:


tomaka:
Breaking changes like these break half of the ecosystem :(

Image should at least get a minor version bump when this happens.

PR #405, opened 2015-05-08
Add BMP Decoder

pix64:
I decided to write a BMP decoder for Image. This is my first serious piece of Rust code. I tried to follow similar formats of the other decoders.

* Decodes 1, 4, 8, 16, 24, and 32-bit BMPs.
* Supports all Windows BMP Headers (BITMAPCOREHEADER, BITMAPINFOHEADER, BITMAPV2INFOHEADER, BITMAPV3INFOHEADER, BITMAPV4HEADER, BITMAPV5HEADER).
* No alpha support due to ambiguity. 32-bit BMPs can use the last byte in the 32-bits as an alpha, but this byte is usually ignored and is zero which would result in completely transparent images.
* Only supports well known color masks in BI_BITFIELDS mode. Windows NT is capable of rendering a BMP with any contiguous and non-overlapping color masks. This decoder only recognizes 555, 565, and 888 masks. (Formats which are savable in Photoshop and GIMP)
* Does not support RLE (Run Length Encoded) BMPs

For 16-bit BMPs, color data is saved in either 5 or 6 bits out of the 16 (X1-R5-G5-B5, or R5-G6-B5). These values have to be converted to 8-bit values. I used a look up table to achieve this. Since this is more general purpose it might be more suited to be moved out of the bmp decoder, but as far as I know its only needed for BMPs.


src/bmp/decoder.rs
```diff
@@ -0,0 +1,453 @@
...
+                    y += 1;
+                }
+            }
+        }
+
+        Ok(result)
+    }
+
+    fn read_palletized_pixel_data(&mut self) -> ImageResult<Vec<u8>> {
+        let mut pixel_data: Vec<u8> = repeat(0u8).take(self.width as usize * self.height as usize * 3).collect();
```
nwin: That would be more clear as `let mut pixel_data = vec![0; 3 * self.width as usize * self.height as usize];`

nwin:
Thank you! That look very good. Could you please also add a test case for it such that it doesn’t accidentally break? 

- Just take some small image which would we can legally add to the repo (public domain or create some 2x2 image yourself). 
- Add to  "tests/images/bmp/images" 
- Add "bmp" to [this line](https://github.com/PistonDevelopers/image/blob/master/tests/reference_images.rs#L18)
- run cargo test
- copy the corresponding image from tests/output to tests/reference.

pix64:
Okay. I added some test images. I'm not very familiar with git so hopefully everything is okay.

nwin:
Look very good. Unfortunately I cannot merge. You need to checkout the latest revision of image and and merge again.

Ideally you should have [rebased](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) but now it is too late ([merge vs rebase](https://www.atlassian.com/git/tutorials/merging-vs-rebasing)). Hope that two links help.

nwin:
Thanks, great work! Didn’t see your last merge attempt until now, github didn’t notify.

PR #412, opened 2015-05-12
JPG overflow fix. Fixes #409

### LINKED ISSUE #409: Panic in JPEG decoding: shift operation overflowed
[This JPEG](https://github.com/servo/servo/blob/master/tests/ref/test.jpeg) crashes decoding with

```
thread '<main>' panicked at 'shift operation overflowed', src/jpeg/entropy.rs:151
```

pix64:
On release builds overflows are not checked, on debug builds they are causing Image to panic on certain JPGs. "wrapping_shl" is unstable and not available in Rust Beta so instead this checks if the shift is too large and just sets the value to 0 if it is.

### DIFF
--- a/src/jpeg/decoder.rs
+++ b/src/jpeg/decoder.rs
@@ -308,8 +308,8 @@ impl<R: Read>JPEGDecoder<R> {
             )))
         }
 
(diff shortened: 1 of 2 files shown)


nwin:
Great, thanks what a relieve!

Issue #410, opened 2015-05-08
Resized images corrupting colors.


Xheotris:
Resizing an image turns the whole image pink. I've tested this with both jpegs and pngs. Size isn't a factor, I've used images from 2000x1500 to 300x300. Filter type doesn't seem to matter, I've used CatmullRom and Lanczos. Example:

![img_5874](https://cloud.githubusercontent.com/assets/9502592/7542931/e754b302-f57d-11e4-9683-8375a6d6d5d1.JPG)

The code that causes the issue is this: 
```
let selected_image = image::open(file.path()).unwrap();
let formatted_image = selected_image.resize(900, 600, image::FilterType::CatmullRom);
let ref mut fout = fs::File::create(new_file).unwrap();
let _ = formatted_image.save(fout, ImageFormat::PNG).unwrap();
```

Xheotris:
I just tried it with Triangle, Gaussian and Nearest for the sake of completeness. No joy. Filter is not relevant.

PR #417, opened 2015-05-25
Bump to 0.3.10.

huonw:
It'd be great to have a new version published now that #410 is fixed and the JPEG decoder doesn't overflow so much.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.3.9"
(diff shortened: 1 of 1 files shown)


tomaka:
Now that we're in stable Rust, I don't know if I'm still supposed to publish versions arbitrarily.
Therefore I suggested to open a PR instead of me publishing a version.

nwin:
Seems important enough to merge anyway but we should definitively discuss how to handle that now…

Issue #414, opened 2015-05-13
PNG inflate is sometimes missing last byte


kmcallister:
In some cases `inflate.rs` skips the last byte of the decompressed stream, e.g.:

```
input   02 ec 30 00
flate2  c3
piston

input   02 c8 00 30 00
flate2  30
piston

input   f2 84 00 80 00 0b 00 30
flate2  49 49 49 49 49 49 49 50
piston  49 49 49 49 49 49 49

input   da 30 7f f1 5d 80 00 ab 00 30
flate2  b0 9f a3 dd 78
piston  b0 9f a3 dd

input   da ff 80 30 30 82 00 80 00 cb 00 30
flate2  bf e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0
        e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0
        e0 e0 e0 e0 32 32 32 32 32 32 32 68
piston  bf e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0
        e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0 e0
        e0 e0 e0 e0 32 32 32 32 32 32 32

input   f2 00 01 80 00 13 00 30
flate2  48 48 48 48 48 10
piston  48 48 48 48 48
```

(I found these examples using [afl.rs](https://github.com/kmcallister/afl.rs/).)

I found some other discrepancies in the inflate implementation that I haven't categorized yet. I might hold off on that, if @nwin's new PNG decoder using libflate2 will be ready soon.

nwin:
Interesting finding. The new PNG decoder should be ready soon. I just need to implement some edge cases and support for 16 bit (which we don’t have right now anyway).

kmcallister:
Sounds great! There are some 16-bit images in Servo's test suite.

nwin:
it’s now passing the whole png testsuite I need to integrate it into image and we should be ready to roll at least on the png side…

nwin:
Closed via #418.

Issue #400, opened 2015-05-04
Again: using crates for the decoders


nwin:
I’m getting more and more the impression that it would be best to split image into standalone crates. Maybe it would be the best option to keep them in the same repository anyway? What’s your opinion?

I’m currently experimenting with replacements for the [gif](https://github.com/nwin/gif) and [png](https://github.com/nwin/png) de-/encoder (mainly to support streaming of content) and am quite happy with having these two guys as separate crates…

TyOverby:
I originally was on the side of using cargo features, but now I'm leaning this way.  

Seeing as we can put as many crates in the same cargo package, we should keep them all in the same repository, and in the same cargo `image` package.

kvark:
I might be missing something trivial, but there should be some code that selects the decoder based on the image format. How would this code know about external crates?

nwin:
By using features as we do already.

kvark:
@nwin ok, so what's the benefit you expect from having the decoder sub-crates?

bvssvni:
What about a small crate for the core and some sub crates, and then reexport all of them in an "image" crate?

TyOverby:
I think having `image`, `image-png (etc)` and `image-all` is more intuitive than `image-core`, `image-png`, and `image`.

ruud-v-a:
I am in favour of separate crates for the formats. Decoding and image processing are different things, it makes sense to separate them. It also makes sense to have a wrapper that just loads an image regardless of its format. In my opinion `image` depending on `image-png`, …, depending on `image-core` would be more intuitive. It has the advantage that it does not break anything for users of the `image` crate.

I wonder though, would it be possible to avoid `image-core` entirely? If you only use `image-png`, then for example the `ImageDecoder` trait is rather useless, the impl could be provided by `image`. There would be some duplication, but on the other hand it would make the encoders/decoders more useful in their own right.

nwin:
I think a high-level `ImageDecoder` should just be implemented in `image`. The different formats are so different that you need a custom de-/encoder type for every format anyway…

ruud-v-a:
Yes, that is what I meant.

bvssvni:
@nwin Are you working on this now?

nwin:
Not right now. I wanted to start when I get my new png decoder to work (where I'm almost done beside some edge cases)

PR #418, opened 2015-05-26
Use external PNG crate.

### LINKED ISSUE #390: Add compression to PNG
The ability to compress PNG images was reduced in order to make it work with Rust 1.0 beta that should be added back at some point…

nwin:
Image::png is now just a little layer over an external codec crate.

Closes #390 #300 #400 #414


nwin:
Note this is WIP. I will move the png crate to PistonDevelopers before merging.

tomaka:
According to one of the Servo devs (don't remember who), using C code defeats the purpose of using image in the first place.

nwin:
Servo already depends on flate2. At least @kmcallister seemed to be fine with it.

bvssvni:
@nwin You are awesome!

bvssvni:
@tomaka I opened https://github.com/PistonDevelopers/image/issues/419

LaylConway:
Having a pure rust implementation for most things would be preferable in my opinion.

bvssvni:
@LaylConway See https://github.com/PistonDevelopers/image/issues/419

nwin:
So the updated PNG crate now uses flate2 only for encoding.

PR #430, opened 2015-07-21
Use stable Rust instead of 1.0.0

tomaka:


ruud-v-a:
Instead of testing on stable _or_ 1.0, why not test both? As long as we do not use any of the new stuff, we can continue to support 1.0, right?

nwin:
@ruud-v-a Testing against both would increase the testing time even more.

I thing we should just keep supporting 1.0 unless we really need a feature from stable.

ruud-v-a:
Don’t the builds run in parallel? At least for this PR the logs show the same start time for every build.

frewsxcv:
Opened a new PR that addresses the comments in here #472

bvssvni:
Closed in favor of https://github.com/PistonDevelopers/image/pull/472

PR #432, opened 2015-07-24
Remove unnecessary use of `#[repr(packed)]`.

huonw:
This struct never seems to be used in a way that requires being packed.

The removal is good because there's some correctness issues with it, so
there may be breaking changes to it in future and removing it now will
avoid them all together. See
https://github.com/rust-lang/rust/issues/27060.


bvssvni:
Thanks!

Issue #434, opened 2015-08-08
Panic in JPEG decoding of large image: arithmetic operation overflowed


twoodford:
I get `thread '<main>' panicked at 'arithmetic operation overflowed', src/jpeg/decoder.rs:208` when loading the following image:

![20150728_0006](https://cloud.githubusercontent.com/assets/2175841/9151761/a09eaed4-3ddc-11e5-838e-2bf115b24c46.jpg)

Smaller images don't crash here.  It looks like `mcucount` needs to be a `u32` for images above 4194304 pixels.

nwin:
yes, that is correct

PR #435, opened 2015-08-10
Fix mcucount overflow when decoding large JPEG files

twoodford:
Fixes issue #434

### DIFF
--- a/src/jpeg/decoder.rs
+++ b/src/jpeg/decoder.rs
@@ -122,7 +122,7 @@ pub struct JPEGDecoder<R> {
     vmax: u8,
 
     interval: u16,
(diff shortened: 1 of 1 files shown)


bvssvni:
Looks good.

ruud-v-a:
Would it be a good idea to add the picture (or a smaller file that exhibits the same behaviour) to the test suite?

nwin:
Thank you for your contribution!

@ruud-v-a Yes ist would. I should also note that I have a new JPEG decoder in the pipeline which fixes many bugs of the old one including this bug. I just need to fix some bug in progressive JPEG decoding and need to clean up the code.

Closes #434

Issue #436, opened 2015-08-11
The std::old_io in the examples


prewk:
Hi, I'm completely new at Rust so yell at me if I'm doing something wrong, but I can't open an image due to the use of `std::old_io`.

````rust
extern crate image;

use std::old_io::File;
use image::GenericImage;

fn main() {
    let img = image::open(&Path::new("images/medium.jpg")).unwrap();

    println!("{:?}", img.dimensions());
}
````

"src/main.rs:3:5: 3:22 error: unresolved import `std::old_io::File`. Could not find `old_io` in `std`"

Furyhunter:
Have you tried switching `use std::old_io::File` to `use std::fs::File`? Also, it doesn't appear as though it's used in this example anyway, so you could just remove it as well.

nwin:
Thanks for bringing this up, the examples really need to be updated…

prewk:
Yeah, I ended up using `let img = image::open("images/medium.jpg").unwrap();` which worked (sort of, the JPEG support wasn't what I'd hoped for but that's for another day I guess.)

nwin:
Closed via #437.

PR #437, opened 2015-08-12
Fix issue #436: Update examples in README.md

twoodford:
I updated the opening and fractal examples in README.md to match examples/opening.rs (except that opening.rs uses a command line argument) and examples/fractal.rs.  I have confirmed that both examples work on stable rust (1.1.0).


bvssvni:
Thanks!

PR #438, opened 2015-09-02
Minor tweaks to the JPEG decoding code

huonw:
The major change is replacing `repeat(x).take(y).collect()` with `vec![x; y]` which seems to give a 5-10% speed-up for decoding a local file.


bvssvni:
Thanks!

PR #445, opened 2015-09-26
Fix flip_horizontal bug. Add tests for all affine.rs functions

theotherphil:
This is the first rust I've written, so please let me know if I'm doing anything stupid or unidiomatic.

I'd like to eventually add some quickcheck-y tests so that we can cover all small image sizes (probably more use for the filtering stuff than here). This tiny set of manual examples picked up a bug though, so they're already better than nothing.

### DIFF
--- a/src/imageops/affine.rs
+++ b/src/imageops/affine.rs
@@ -67,7 +67,7 @@ pub fn flip_horizontal<I: GenericImage + 'static>(image:  &I)
     where I::Pixel: 'static,
           <I::Pixel as Pixel>::Subpixel: 'static {
     let (width, height) = image.dimensions();
-    let mut out = ImageBuffer::new(height, width);
+    let mut out = ImageBuffer::new(width, height);
 
     for y in (0..height) {
         for x in (0..width) {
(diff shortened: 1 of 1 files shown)


src/imageops/affine.rs
```diff
@@ -97,3 +97,115 @@ pub fn flip_vertical<I: GenericImage + 'static>(image:  &I)
...
+    use buffer::{ImageBuffer,Pixel,GrayImage};
+    use image::{GenericImage};
+
+    macro_rules! assert_pixels_eq {
+        ($actual:expr, $expected:expr) => ({
+            let actual_dim = $actual.dimensions();
+            let expected_dim = $expected.dimensions();
+
+            if actual_dim != expected_dim {
+                panic!("dimensions do not match. \
```
xaviershay: I haven't written much rust either, so question I have is: why does this macro work inside `macro_rules!` but `assert!` doesn't?

src/imageops/affine.rs
```diff
@@ -97,3 +97,115 @@ pub fn flip_vertical<I: GenericImage + 'static>(image:  &I)
...
+                    actual: {:?}, expected: {:?}", actual_dim, expected_dim)
+            }
+
+            let diffs = pixel_diffs($actual, $expected);
+
+            if !diffs.is_empty() {
+                let mut err = "pixels do not match. ".to_string();
+
+                let diff_messages = diffs
+                    .iter()
```
xaviershay: `iter` or `into_iter`? I'm not clear on which is appropriate.

xaviershay:
LGTM, only comments I left are things I don't know myself :)

(Obviously I'm not the most qualified to review, bug figured it was better than nothing!)

src/imageops/affine.rs
theotherphil: As the iterator doesn't need ownership of the diffs I just used the regular iter(). It seems reasonable to me to use the least demanding iterator that works, but I'll change if anyone objects.

src/imageops/affine.rs
theotherphil: I'm defining a new macro here. It's invoked later as "assert_pixels_eq!(...)".

theotherphil:
Thanks for the review.

src/imageops/affine.rs
xaviershay: "this" in my question refering to `panic!`. Answering my own question from http://stackoverflow.com/questions/24718256/higher-order-macros

> It works fine when the nested macro has no argument itself, because there is no nested $... non-terminals and so the expansion phase doesn't ever hit the definition problem described above.

xaviershay:
@nwin looks like you typically collect the `CHANGES` file after the fact, so I'm going to merge this once it goes green.

src/imageops/affine.rs
theotherphil: Ah, sorry!

theotherphil:
A single build failed, but in the num crate. Is there a way for me to request a rerun?

Potpourri:
@theotherphil thanks!

PR #447, opened 2015-09-27
Add rotate_nearest function

theotherphil:
Any thoughts on the signature? It's a little clunky, but not too bad if the default is specified as Luma::White or similar. 

I've not exposed this on dynimage yet. Should I?

I'll add a bilinear interpolation variant soon. I might also add a "fast" version - I'm using f32::round to make this actually choose the nearest pixel in the source image, but this makes the benchmark take around 50% longer than just casting. Any thoughts?

I wanted this to be a single commit, but I've not used git before so managed to garble that. I'l fix up before merging if accepting this pull request would also make a mess of the main image repo.

### DIFF
--- a/src/imageops/affine.rs
+++ b/src/imageops/affine.rs
@@ -98,11 +98,70 @@ pub fn flip_vertical<I: GenericImage + 'static>(image:  &I)
     out
 }
 
+/// Rotate an image clockwise about center by theta radians by choosing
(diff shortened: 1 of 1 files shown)


theotherphil:
I'll need to add this to imageops/mod.rs as well. I'll do that when tidying up any other issues anyone has.

theotherphil:
I've added this to imageproc instead.

PR #450, opened 2015-10-07
Add ICO icon file decoding

mbrubeck:
This adds basic support for the [ICO file format][1]. This format is really just a container for PNG or BMP data.  This decoder is structured a little unusually so that it can delegate most of its work to either PNGDecoder or BMPDecoder.  It also requires some extra hooks in the BMPDecoder code.

A single ICO file may contain multiple images. The decoder currently just selects the highest-resolution image, but it could easily be extended to select images by size or other criteria.

[1]: https://en.wikipedia.org/wiki/ICO_%28file_format%29


xaviershay:
Thanks! Looks pretty good to me, I don't have any feedback.

src/ico/decoder.rs
```diff
@@ -0,0 +1,215 @@
...
+    Ok(entry)
+}
+
+/// Find the entry with the highest (color depth, size).
+fn best_entry(mut entries: Vec<DirEntry>) -> ImageResult<DirEntry> {
+    let mut best = try!(entries.pop().ok_or(ImageError::ImageEnd));
+    let mut best_score = (best.bits_per_pixel, best.real_width() + best.real_height());
+
+    for entry in entries.into_iter() {
+        let score = (entry.bits_per_pixel, entry.real_width() + entry.real_height());
```
samlh: I vaguely expected multiplication here, not that it should make a big difference in practice.

mbrubeck: Updated to use multiplication in the latest revision.

mbrubeck:
Updated to use multiplication when comparing sizes based on feedback from @samlh, and to support 32-bit RGBA bitmaps in ICO files, which is needed for many common website favicons like https://assets-cdn.github.com/favicon.ico

nwin:
Can you please a test image to the tests directory? Please create a new folder for `ico` in `test` and change `references_images.rs` accordingly.

mbrubeck:
@nwin Added a few test images (one BMP + AND mask, one 32-bit BMP with alpha channel, one PNG).

nwin:
Great thanks. Though I'm a bit unsure whether we are allowed to include these files in the repo.

mbrubeck:
The Rust and Mozilla logo artwork are available under free/open source licenses, and subject to the [Rust trademark policy](https://www.rust-lang.org/legal.html) and [Mozilla trademark policy](https://www.mozilla.org/en-US/foundation/trademarks/policy/).  The GitHub logo is available from [this official repo](https://github.com/logos), again with a trademark policy.

I don't believe that including the unmodified images in the test suite violates any of these policies or licenses, but I am not a lawyer. If you'd prefer, I can create some totally original test files.

mbrubeck:
Nevermind, I decided to just create new ICO test files by re-encoding one of the existing PNG test files.

nwin:
Thank you for contributing this new decoder! 

I think adding the logos would have been fine given their license.

PR #458, opened 2015-10-16
Fix panic on bad palette index

mbrubeck:
This fixes a panic in Servo on http://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html


mbrubeck:
Note: Firefox's BMP decoder solves this a different way, by always allocating 256 palette entries. (Any excess ones are simply zeroed.)  This avoids the need for a bounds check on each pixel.  Corrupt images are rendered with black for pixels with bad palette indices, rather than returning an error.

I'm happy to change this PR to implement Firefox's solution instead, if you prefer.

pix64:
> Firefox's BMP decoder solves this a different way, by always allocating 256 palette entries. (Any excess ones are simply zeroed.) This avoids the need for a bounds check on each pixel. Corrupt images are rendered with black for pixels with bad palette indices, rather than returning an error.

This seems to be how other browsers, Windows Photo Viewer, mspaint, etc behave so it's probably best to do so.

nwin:
@mbrubeck: I think allocating 256 palette entries would be the best approach. Lets just keep aligned with the common behavior.

Doesn’t server has it’s own BMP decoder? I heard of a plan to add it to Firefox? Couldn’t we use that instead if it is more battle-tested?

mbrubeck:
@nwin:

> I think allocating 256 palette entries would be the best approach. Lets just keep aligned with the common behavior.

Okay, I updated the pull request to allocate 256 palette entries.

> Doesn’t server has it’s own BMP decoder? I heard of a plan to add it to Firefox? Couldn’t we use that instead if it is more battle-tested?

No, Servo recently switched from the `stb_image` C library to this crate (PistonDevelopers/image) for BMP decoding: servo/servo#7933

pix64:
This looks good to me. What's the protocol for merging pull requests?

nwin:
@mbrubeck: Thanks.

@pix64 The protocol is leave it to the maintainer unless it is really urgent.

PR #460, opened 2015-10-17
add row iteration to the BMP decoder

froydnj:
While it makes the code look nicer through less manual indexing, it's not that impressive of an abstraction layer yet.  The real payoff from the row iterator will come with implementing more formats, I think, particularly the RLE compression ones.

This is my first "real" Rust code, so any style comments would be most appreciated.  It's also my first "real" pull request for this project, so any advice on common code that I ought to use would be appreciated as well!  There's some similar-looking code in `src/buffer.rs`, but that code appears to be working at a higher level than the code here.

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1,5 +1,6 @@
 use std::io::{Read, Seek, SeekFrom};
-use std::iter::repeat;
+use std::iter::{Iterator, repeat, Rev};
+use std::slice::ChunksMut;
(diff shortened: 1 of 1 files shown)


nwin:
Looks good. Please rebase onto master.


PS: Yes, it would be nice to have a row iterator in buffer at some point.

nwin:
Great thanks for contributing!

PS: You’ll need to post an update, github won’t send out a notification on rebasing.

PR #465, opened 2015-10-20
Bump to 0.3.16

bvssvni:
DO NOT MERGE YET!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.3.15"
+version = "0.3.16"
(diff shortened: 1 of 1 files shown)


bvssvni:
@nwin It looks like https://github.com/PistonDevelopers/image/pull/450 added enumm variants and https://github.com/PistonDevelopers/image/pull/453 has a new public reexport. If we follow strict semver then we should bump it to 0.4.0. What do you think?

nwin:
Ok fine with me.

bvssvni:
I'll close this and open up a new one.

PR #475, opened 2015-10-26
Fix panic on empty file

mbrubeck:
Fixes servo/servo#8093.


nwin:
Thanks for fixing that.

PR #478, opened 2015-10-30
run doc only once

juggle-tux:
why run cargo doc for every build?
cargo tests checks the doc-tests and the update to the gh-pages should be from a fix build with fix features


juggle-tux:
i added the repo to my rss feed and the build bot was flooding it

juggle-tux:
*oops* travis fails looks to me like the test results before was eaten by the cargo doc call after the test

have no idea how to resolve that sry

bvssvni:
Closing because outdated.

PR #483, opened 2015-11-13
add RLE support to the BMP decoder

froydnj:
The test images included in this commit are from the BMP Suite:

http://entropymine.com/jason/bmpsuite/


bvssvni:
Looks good. Merging.

PR #486, opened 2015-11-26
Update to byteorder 0.4

SimonSapin:
CC https://github.com/servo/servo/pull/8446

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.5.1"
+version = "0.5.2"
 license = "MIT"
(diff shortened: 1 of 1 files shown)


bvssvni:
Noticed that `byteorder::Error` is exposed in the API like [here](http://www.piston.rs/image/image/tiff/struct.TIFFDecoder.html). This change could break existing code. In this case, I think it's OK.

@tomaka Next time, please check quickly for breaking changes. We try to avoid breaking existing code, and when making exceptions other people should know.

SimonSapin:
What would have been a better way to do this? Bump to 0.6.0?

nwin:
Uh. That error should never have been exposed. 😳 Mea culpa.

PR #488, opened 2015-11-27
Do not expose `byteorder::Error`

bvssvni:
Closes https://github.com/PistonDevelopers/image/issues/487

- Hide all methods on `TIFFDecoder` except `new`
- Hide `tiff::ByteOrder` as it is not needed in public API
- Moved `ifd` and `stream` to submodules of `decoder` to solve internal
visibility issues


bvssvni:
@nwin Do these changes look OK?

nwin:
Yes, thanks.

bvssvni:
I think we could bump the patch version since there are likely no libraries using these methods.

nwin:
Yes.

PR #492, opened 2015-12-03
use chunks_mut iteration in bmp's read_palletized_pixel_data

froydnj:
I guess we could really use `zip` here, but getting rid of the explicit multiplications in the pixel data indexing is nice.


bvssvni:
Thanks!

PR #497, opened 2015-12-23
Fix JPEG encoder's quantization table

lgvz:
Pick up the pdf where the table was copied from and select some values in the table... they used white ones to align the numbers!!! So this reduces smudges.


nwin:
Good catch, thanks!

PR #499, opened 2015-12-24
Expose format-specific load functions

bfops:
Expose format-specific load functions, some of which don't have the `Seek` constraint that `load` does.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -542,7 +542,7 @@ pub fn open<P>(path: P) -> ImageResult<DynamicImage> where P: AsRef<Path> {
         )))
     };
 
-    load(fin, format)
+    load(BufReader::new(fin), format)
 }
 
 /// Saves the supplied buffer to a file at the path specified.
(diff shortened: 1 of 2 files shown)


nwin:
I’m not sure if this is the best solution. The public API has been more or less codec agnostic so far.

bfops:
Another option is to have the decoders do their own buffering internally, so they don't need `Seek`.

froydnj:
> Another option is to have the decoders do their own buffering internally, so they don't need `Seek`.

The model Firefox's image decoders use looks like:

```
[network] --passes available data to--> [imagelib] <-> [decoder]
```

Individual decoders tell `imagelib` how much data they need (e.g. so many bytes for the image header) and `imagelib` buffers the data received from the network as necessary.  When the amount of data reaches the threshold indicated by the decoder, `imagelib` passes the slice of data into `decoder`, which processes it and returns a status to `imagelib`: either "I'm done", "there was an error", or "I now need this much data."  There's also an option to pass data directly through, without buffering, for compressed formats like PNG and GIF.

I think this model has advantages for something like Servo (makes things more amenable to non-blocking I/O from the network), and also doesn't require parameterizing the decoders on `Read`...or `Seek`.  It'd be a much bigger change than this pull request, though.

nwin:
@froydnj Currently most decoders are not capable of working in this model. Only gif and png could be adapted for is as they are partially re-entrant. I’m currently working on a JPEG version but it is very much WIP…

Regarding the read/seek dependency we have the problem that there are decoders like TIFF which cannot be streamed and as the format itself is built around random access to the storage. As such we either need `Seek` or have to buffer the whole file in memory.

bvssvni:
Closing since this is probably outdated.

PR #501, opened 2016-01-05
impl unsafe_get/put_pixel to ImageBuffer

tafia:
Adds unsafe impl from GenericImage to ImageBuffer.
This leads to some speedups in imageproc.


bvssvni:
Thanks!

PR #502, opened 2016-01-05
Updated crates.io

bvssvni:
DO NOT MERGE YET!

- Bumped to 0.6.1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.6.0"
+version = "0.6.1"
 license = "MIT"
 description = "Imaging library written in Rust. Provides basic filters and decoders for the most common image formats."
 authors = [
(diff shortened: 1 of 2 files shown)


bvssvni:
The rand crate fails on Rust nightly https://github.com/rust-lang/rust/issues/30713.

`rustc 1.7.0-nightly (d5e229057 2016-01-04)`

tafia:
I think this is not a breaking change as these fn already had a default
impl through GenericImage.

bvssvni:
@tafia Perhaps the new methods should be moved [here](https://github.com/PistonDevelopers/image/blob/c5e7ca986139006ef77478cfe5cf1f7496401f3e/src/buffer.rs#L448)?

bvssvni:
@tafia I moved the methods to the `GenericImage` impl. Changed the version to 0.6.1. Had to change from `&P` to `P` when returning from `unsafe_get_pixel`.

bvssvni:
Merging.

PR #504, opened 2016-01-09
[Early prototype] Add bilinear sampling

TyOverby:
I'd find it really handy to sample images by using this crate.  Asking for pixel information at locations that aren't integers is incredibly handy when scaling images as well as other "non-picture" applications like using an image as a heightmap.

There are three main sampling methods used commonly
- Nearest Neighbor: Pick the closest pixel to the sampled point and return that.
- Bilinear: Linearly interpolate in the x and y directions using the 4 neighbors.
- Bicubic: Cubic interpolation that takes into account 16 neighboring pixels.

This PR currently implements bilinear interpolation, but I'd like suggestions as to what the API should be.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -441,6 +441,55 @@ pub trait GenericImage: Sized {
     Self::Pixel: 'static {
         SubImage::new(self, x, y, width, height)
     }
+
+    /// Samples the image at non-integer positions.
+    ///
+    /// Uses bilinear interpolation.
+    ///
+    /// X and Y coordinates must be in range.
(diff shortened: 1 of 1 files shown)


TyOverby:
Also I haven't added tests yet, but I plan on doing that before merging.

bvssvni:
I believe you can use the unsafe version of receiving a pixel for better performance, but be careful.

bvssvni:
Closing because of this filter is supported by https://github.com/PistonDevelopers/imageproc

PR #506, opened 2016-01-13
Correct luma coefficients

Mr4x:
Rec. 709 forms luma using RGB coefficients 0.2126, 0.7152, and 0.0722.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -252,9 +252,9 @@ impl<T: Primitive + 'static> FromColor<Rgba<T>> for Luma<T> {
     fn from_color(&mut self, other: &Rgba<T>) {
             let gray = self.channels_mut();
             let rgb = other.channels();
-            let l = 0.2125f32 * rgb[0].to_f32().unwrap() +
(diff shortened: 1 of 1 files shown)


nwin:
Where do the other almost identical constants come from? Looks like some rounding issue to me.

Mr4x:
@nwin you can find them [here](http://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.709-6-201506-I!!PDF-E.pdf) at page 4

nwin:
Yes, but where do the other values come from? They can also be found on the internet. Is is a copy&paste error or something rounding-related?

Mr4x:
@nwin yeah, I think it's the wrong rounding. I checked the older recommendations, but has not found other values

philipc:
The other values are from an older version, eg BT.709-1 https://www.itu.int/rec/R-REC-BT.709-1-199311-S/en

nwin:
Thanks. Sorry for taking so long to merge, I actually forgot about it.

PR #518, opened 2016-03-30
Make de-generics wrappers for open and save_buffer

bluss:
This makes the inner parts of these functions non-inlinable. It also
makes open_impl and save_buffer_impl be compiled once, as part of
compiling the crate, instead of having to compile them each time their
*uses* are compiled.

This was prompted by http://stackoverflow.com/questions/36317461/

Modifying the user library and recompiling took more than 20 seconds,
with this change, only < 0.3 s. This means that `open`'s implementation
may no longer be inlined, but it shouldn't matter.


bluss:
libstd uses this “pattern” in a few places too.

bluss:
`load` could use the same kind of transformation I think, but you that know the library well need to make the call there.

bvssvni:
Nice!

PR #521, opened 2016-04-03
tga/decoder: fix RLE read length

philipc:
The TGA decoder wasn't including the +1 when updating num_read for raw
byte sequences. Fix by using the length of pixel_data instead.


bvssvni:
Thanks!

PR #527, opened 2016-04-18
Removed use of deprecated function. Replaced with std::io::Write

HarryLovesCode:
Really minor update. Had no impact on performance when last tested, but I figured it's always better to keep current with Rust.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,6 +14,8 @@ extern crate enum_primitive;
 #[cfg(test)]
 extern crate test;
 
+use std::io::Write;
+
 pub use color::ColorType::{
     self,
     Gray,
@@ -134,16 +136,9 @@ mod animation;
 // Copies data from `src` to `dst`
 //
 // Panics if the length of `dst` is less than the length of `src`.
-// NOTE: this is a copy-paste of the unstable function `std::slice::bytes::copy_memory`.
 #[inline]
-fn copy_memory(src: &[u8], dst: &mut [u8]) {
+fn copy_memory(src: &[u8], mut dst: &mut [u8]) {
     let len_src = src.len();
     assert!(dst.len() >= len_src);
-    // `dst` is unaliasable, so we know statically it doesn't overlap
-    // with `src`.
-    unsafe {
-        std::ptr::copy_nonoverlapping(src.as_ptr(),
-                                      dst.as_mut_ptr(),
-                                      len_src);
-    }
+    dst.write(src).unwrap();
 }


bvssvni:
Old:

```
test buffer::test::bench_conversion        ... bench:   5,968,392 ns/iter (+/- 807,598) = 502 MB/s
test dynimage::bench::bench_conversion     ... bench:   5,980,635 ns/iter (+/- 560,672) = 501 MB/s
test imageops::sample::tests::bench_resize ... bench:   8,556,590 ns/iter (+/- 739,980) = 238 MB/s
```

New:

```
test buffer::test::bench_conversion        ... bench:   5,940,961 ns/iter (+/- 190,365) = 504 MB/s
test dynimage::bench::bench_conversion     ... bench:   5,938,701 ns/iter (+/- 507,953) = 505 MB/s
test imageops::sample::tests::bench_resize ... bench:   8,521,872 ns/iter (+/- 635,008) = 239 MB/s
```

bvssvni:
Nice!

tafia:
I don't think we even need the `Write` trait, `copy_from_slice` (unstable for now) should be enough:
```rust
dst[..len_src].copy_from_slice(src)
```

here is an example (inlined in order to see the llvm irs)
http://is.gd/aaVFkz

I'm not a llvm expert but it looks like it is doing exactly the same thing:
```llvm
; Function Attrs: noinline uwtable
define internal fastcc void @_ZN8rust_out11copy_memory17h891096968b2bcd54E(i8* noalias nocapture nonnull readonly, i8* nonnull) unnamed_addr #0 {
"_ZN11collections5slice29_$LT$impl$u20$$u5b$T$u5d$$GT$15copy_from_slice17h819ce68e108218c2E.exit":
  tail call void @llvm.memcpy.p0i8.p0i8.i64(i8* nonnull %1, i8* nonnull %0, i64 3, i32 1, i1 false)
  ret void
}

; Function Attrs: noinline uwtable
define internal fastcc void @_ZN8rust_out17copy_memory_write17h26d45df7c5e178a9E(i8* noalias nocapture nonnull readonly, i8* nonnull) unnamed_addr #0 personality i32 (i32, i32, i64, %"1.std::sys_common::libunwind::_Unwind_Exception"*, %"1.std::sys_common::libunwind::_Unwind_Context"*)* @rust_eh_personality {
"_ZN46_$LT$std..result..Result$LT$T$C$$u20$E$GT$$GT$6unwrap17he1d64214ddf1c7c8E.exit":
  tail call void @llvm.memcpy.p0i8.p0i8.i64(i8* nonnull %1, i8* nonnull %0, i64 3, i32 1, i1 false), !noalias !0
  ret void
}
```

HarryLovesCode:
@tafia Yepp they are doing the same thing AFAIK (just using `memcpy` under-the-hood). Here are the results from running the benchmarks:

## Using `std::io::Write`:

```
test buffer::test::bench_conversion        ... bench:   5,189,914 ns/iter (+/- 138,304) = 578 MB/s
test dynimage::bench::bench_conversion     ... bench:   5,196,098 ns/iter (+/- 608,945) = 577 MB/s
test imageops::sample::tests::bench_resize ... bench:   7,415,769 ns/iter (+/- 319,728) = 275 MB/s
```

## Using `copy_from_slice`:

```
test buffer::test::bench_conversion        ... bench:   5,196,551 ns/iter (+/- 142,362) = 577 MB/s
test dynimage::bench::bench_conversion     ... bench:   5,203,425 ns/iter (+/- 531,911) = 576 MB/s
test imageops::sample::tests::bench_resize ... bench:   7,429,126 ns/iter (+/- 135,504) = 274 MB/s
```

No real difference in performance as you probably guessed. If `copy_from_slice` is still unstable, then I think we should stick with `std::io::Write` because this crate is built against stable, beta, and nightly channels (according to the Travis CI configuration file).

PR #531, opened 2016-05-04
Update to png 0.5

mbrubeck:
I believe this can be a minor change because image does not expose any types from png in its public API.


bvssvni:
Thanks!

PR #532, opened 2016-05-04
Add support for encoding ICO files

mdsteele:
This change adds an `ICOEncoder` struct, and uses it in `DynamicImage` to support encoding ICO files.  The implementation is very simple, since `PNGEncoder` does all the heavy lifting.

If desired, `ICOEncoder` could be expanded in the future to support storing multiple images in the same ICO file, but for now it just encodes a single image (with the same API as `PNGEncoder`).


bvssvni:
cc @nwin

bvssvni:
Looks good! Merging.

Issue #477, opened 2015-10-27
Panic when decoding ICO


frewsxcv:
```rust
extern crate image;

use std::io::Cursor;

fn main() {
    let input = b"\x30\x30\x30\x30\x04\x00\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\xb6\x00\x00\x00\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x00\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x0c\x00\x00\x00\x30\x30\x30\x30\x30\x30\x08\x00".to_vec();
    let c = Cursor::new(input);
    let _ = image::ico::ICODecoder::new(c);
}
```
```
coreyf@frewbook-pro /t/imagee (master) [101]> cargo run
   Compiling imagee v0.1.0 (file:///private/tmp/imagee)
     Running `target/debug/imagee`
thread '<main>' panicked at 'index out of bounds: the len is 0 but the index is 0', ../src/libcollections/vec.rs:1110
Process didn't exit successfully: `target/debug/imagee` (exit code: 101)
```

found using [afl.rs](https://github.com/frewsxcv/afl.rs)

philipc:
Closed by #535

PR #535, opened 2016-05-28
bmp/decoder: ensure palette has correct length

philipc:
This is a minimal fix for #477.

I'm not sure this is the best fix though. What's the general philosophy regarding corrupt files? Should we fail if we detect corruption, or proceed as best as possible? A related concern is whether there should there be a limit on image width, height, palette size etc; we don't want to use gigabytes of memory due to a corrupt or malicious file.

Should I convert the test case into a file and add it under tests? This would be different from the existing tests, because I don't think there are any test cases for invalid files yet.

I also plan to audit other uses of read_to_end().

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -449,7 +449,8 @@ impl<R: Read + Seek> BMPDecoder<R> {
         let max_length = MAX_PALETTE_SIZE * bytes_per_color;
         let mut buf = Vec::with_capacity(max_length);
 
-        try!(self.r.by_ref().take(length as u64).read_to_end(&mut buf));
+        buf.resize(length, 0);
+        try!(self.r.by_ref().read_exact(&mut buf));
 
         // Allocate 256 entries even if palette_size is smaller, to prevent corrupt files from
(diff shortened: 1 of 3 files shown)


nwin:
We don’t have a policy regarding corrupted files yet. So far we are failing. It could be an option to make the decoder configurable such that it may continue decoding if the error is recoverable.

To understand it correctly: This error occurs if the file is truncated? Looks like a sensible fix

Big/corrupt images are actually an issue as valid file might actually take up several GiB of memory. PNG is especially problematic at it requires to hold an entire scanline in memory to decode the next so on-the-fly resizing is difficult. An option would be to compress the previous scanline…  😳

philipc:
This error could occur for both truncation of color palette or corruption of color palette size. This fix will make it try to continue decoding. We could make it fail immediately by using read_exact() instead of take().read_to_end().

philipc:
Changed the fix to fail if the image is truncated. Also added a test, and an additional fix in tga.

nwin:
Great, that looks good. Thank you for your contribution.

PR #536, opened 2016-05-31
ico/decoder: include all of metadata in data_offset

### LINKED ISSUE #534: Unexpected output from decoding some ICO files
Originally reported as servo/servo#10305: the attached ICO files are not decoded as expected.

[icons.zip](https://github.com/PistonDevelopers/image/files/280745/icons.zip)

philipc:
Closes #534


nwin:
It would be better to properly calculate the offset. At the moment this fix assumes that the cursor is at the right position after calling `read_metadata`. But as we have a test for it it’s ok. 

Thank you very much for contributing! :)

PR #538, opened 2016-06-05
bmp/decoder: 1/4/8 bit benchmarks and performance improvements

philipc:
I've added some benchmarks for reading 1/4/8 bit bmp images, and improved the performance a bit.

After this, I plan to see if I can avoid the bounds checks for setting pixels, but I want to fix the alpha support before attempting that.


nwin:
Thank you!

PR #541, opened 2016-06-06
Fix decoding V4/V5 bitmaps with palettes

### LINKED ISSUE #540: Decoding V4/V5 bitmaps with palettes fails
Test with pal8v4.bmp and pal8v5.bmp from https://github.com/jsummers/bmpsuite.

philipc:
Closes #540


nwin:
Thank you!

PR #547, opened 2016-06-11
bmp/decoder: extend support for bitfields

### LINKED ISSUE #539: Add alpha support for bmp decoder
The bmp decoder can read rgba32 images, but currently it discards the alpha information when we're not reading an ico file.

philipc:
Add support for bitfields with any valid mask, as well as alpha masks.

Closes #539


nwin:
I think your patch broke these test images you talkes about in #539. The resulting png is not just transparent, it doesn’t contain any data. All channels are 0 according to gimp, not just the alpha channel.

philipc:
If I open tests/reference/bmp/images/V3_A1_R5_G5_B5.bmp.dcd94d7c.png or tests/reference/bmp/images/V3_A4_R4_G4_B4.bmp.dcd94d7c.png in gimp and use the color picker, it can see the red/green/blue. How are you testing?

nwin:
True, I was looking at the channel view of Gimp, probably I was misinterpreting how it works. It doesn’t actually show the channels as you would expect, instead it shows the channel after multiplying it with the alpha channel.

nwin:
Thank you!

PR #550, opened 2016-06-17
Radiance HDR format support

red75prime:


src/hdr/hdr_decoder.rs
```diff
@@ -0,0 +1,799 @@
...
+    assert_eq!(split_at_first(&Cow::Owned("EXPOSURE= ".into()), "="), Some(("EXPOSURE"," ")));
+    assert_eq!(split_at_first(&Cow::Owned("EXPOSURE= =".into()), "="), Some(("EXPOSURE"," =")));
+    assert_eq!(split_at_first(&Cow::Owned("EXPOSURE== =".into()), "=="), Some(("EXPOSURE"," =")));
+    assert_eq!(split_at_first(&Cow::Owned("EXPOSURE".into()), ""), None);
+}
+
+// Reads inpult until b"\n" or EOF
+// Returns vector of read bytes NOT including end of line characters
+//   or return None to indicate end of file
+fn read_line_u8<R: Read>(r: &mut R) -> ::std::io::Result<Option<Vec<u8>>> {
```
philipc: `None` seems to always be treated as an error, so probably better to return a `Result<Vec<u8>>` instead. There's already an `io::ErrorKind` for EOF.

Also consider changing to require `R: BufRead` and use `read_until` instead.

src/hdr/hdr_decoder.rs
```diff
@@ -0,0 +1,799 @@
...
+pub struct HDRDecoder<R> {
+    r: R,
+    width: u32,
+    height: u32,
+    meta: HDRMetadata,
+}
+
+/// Refer to [wikipedia](https://en.wikipedia.org/wiki/RGBE_image_format)
+#[repr(C)] #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct RGBE8Pixel {
```
philipc: Was the requirement for `invert` the only thing preventing you from implementing this as a new `ColorType`?

philipc:
Do you have some suitable test images to include for the automated tests?

src/hdr/hdr_decoder.rs
red75prime: Rewriting `read_until` doesn't seem to be a good idea, yes. But in this case it is somewhat justified by the lack of specialization. I cannot provide different implementations of `HDRDecoder<R>` for `R: BufRead+Read+Seek` and `R: Read+Seek`, so I need to wrap `R` in `BufReader` in both cases. It is inefficient.

Returning `None` instead of EOF makes it easier to provide meaningful error message, while parsing image header. When I tested decoder on random files, generic errors like `ImageError::IoError(UnexpectedEof)` weren't very helpful.

src/hdr/hdr_decoder.rs
red75prime: No. RGBE8 format is just not well suited for image processing, due to quick precision loss, when performing operations on pixels with different exponent parts. It should to be converted into Rgb<f32> to be useful for `Pixel` trait. 

It could be done, but I decided to postpone it, to have time to discuss implementation.

src/hdr/hdr_decoder.rs
philipc: You do not need a `BufReader` to get `BufRead`. For example, it is implemented for `&[u8]`. You only need `BufReader` for the cases where it will make it more efficient. For example, when reading from a file, every call to `read_exact` is a system call. Buffered reading should be much faster. If you were able to read large blocks of data in a single call, then it may be debatable if it is needed, but that is not the case for this decoder.

red75prime:
I have test images. However, test suite should be modified to be able to work with HDR images. I didn't have time right now to finish that. I expect to be able to do that later this week.

src/hdr/hdr_decoder.rs
red75prime: I don't understand. I get `reader` from a user of the library either via `dynimage::load<R: Read+Seek>()` or via `HDRDecoder::new()`. I cannot add `BufRead` trait bound to `impl<R:Read> HDRDecoder<R> { ...`, because `dynimage::load` doesn't provide it. So I need to wrap provided `reader` in `BufReader` to use `BufRead` trait.

philipc: You're right that `dynimage::load` currently requires that, but in my opinion that is a flaw in `dynimage` that should be fixed there, rather than working around it in this decoder.

red75prime: I've seen your [PR #551](https://github.com/PistonDevelopers/image/pull/551). I'll update my code in 1 or 2 days.

nwin: I don’t see the issue here you can just create a new Buffered reader in `dynimage::load` as you do with the other formats. If you add that bound to `dynimage::load` you cannot pass a file to it any more.

red75prime: nwin, correct. I can create `BufReader` in `dynimage::load`, then I can pass it to `HDRDecoder<R: BufReader>::new`.

Downside is inconsistency in `dynimage::load` behavior. It's fine to pass `ButReader` into it, while `ImageFormat` is not HDR. 

Well, this can be described in documentation for now. When specialization lands, this special case can be removed.

red75prime:
Failed on documenting rayon on nightly

red75prime:
rustdoc seems to be slightly broken on nightly. Does it make sense to disable `cargo doc` for travis nightly pass?

nwin:
Looks good. Thank you very much for contributing. 

PS: Where did these spurious changes to `png.rs` etc. come from? That shouldn’t be in this patch. ;)

PR #551, opened 2016-06-19
Avoid using BufReader when loading from memory

philipc:
This may be wrong, because it is reversing a change from 3b99dc8, but I can't see why TIFF needed to be treated differently.


nwin:
It is not reversing the change from 3b99dc8, that commit added `Seek` not `BufRead`

philipc:
I meant the movement of `BufferedReader` from `open` into `load`, so that TIFF didn't use `BufferedReader`. Probably that change was because `BufferedReader` didn't support `Seek` back then?

So I think this is fine, as long as you are ok with adding `BufRead` to `load`. It makes it slightly less convenient to call `load` with a file, but it is still easy and most callers will use `open` anyway.

nwin:
Yes but I don't see why it is necessary. Nothing keep decoders from using BufRead if they want to.

philipc:
Moving the `BufReader` from `load` to `open` is not necessary to make the code work, but I think it can be more efficient. Using a `BufReader` when reading from memory is pointless. I haven't measured what difference it makes, but I'll do that if I rework this request.

Adding the `BufRead` trait to `load` isn't necessary for that change either, but calling the decoders without a buffered reader is definitely bad for performance, so it prevents callers making that mistake once `load` is no longer creating the `BufReader`.

It's a low priority though, and I'm happy to drop this if you think it isn't worth it.

nwin:
I'm pretty much sure that BufReader will be optimized away when loading from memory. I'm just hesitating changing a public interface if the gain is small. 

On the other hand the situation is now different than back then since Buffered reader now supports Seek. So maybe it's ok to change it.

philipc:
I fixed a bad interaction between BufReader and Seek in the bmp decoder. After that, the benefit from avoiding BufReader varies a bit between files, but in some cases it is quite significant. See the commit messages for benchmark results.

philipc:
@nwin I think you commented on #555 instead of here.

See a3a7851 for the performance change from removing the seek calls. It's about a 4-5X speedup when reading from a file. The reason is that `BufReader` empties its buffer whenever you seek.

nwin:
Oh, you’re right. Looks like a bug in `BufReader` to me.

philipc:
Unfortunately `BufReader` has been explicitly designed that way. I agree that fixing it in `BufReader` would be better. I'll try creating an issue in rust-lang to see if there is any chance of that happening.

nwin:
It would make `BufReader` a bit more complex because you have to check if it is possible to re-use the buffer instead of just emptying it. Maybe we can use this PR as an argument for the change. 

I'll merge anyway as it would take some time to get it changed, in case BufReader gets changes we'll just revert a3a7851457b5f08d506a4174307a7e05e72192bc. 

Thank you.

PR #555, opened 2016-06-28
ppm/decoder: Prep filtered slice before write_all

genkimarshall:
I noticed that saving an RGBA to a PPM file was taking ~620 times longer than an RGB.
It scaled linearly with # pixels. So this patch does all the filtering in one go and then writes the entire slice.

For a 2560x1700 pixel image:
```
rgb:      0.01888 seconds
old rgba: 11.6570 seconds
new rgba: 0.01891 seconds
```

Test code:
```
extern crate image;
extern crate time;

use std::fs::File;
use std::path::Path;

fn main() {
    let mut imgbuf = image::ImageBuffer::new(2560, 1700);
    for (x, y, pixel) in imgbuf.enumerate_pixels_mut() {
        *pixel = image::Rgba([x as u8, y as u8, x as u8, y as u8]);
    }

    let ref mut fout = File::create(&Path::new("rgba.ppm")).unwrap();
    let img = image::ImageRgba8(imgbuf);

    println!("Start save");
    let start = time::precise_time_s();
    let _ = img.save(fout, image::PPM);
    println!("Save: {} seconds", time::precise_time_s() - start);
}
```

Please me know if I did anything wrong; thanks for the great library otherwise.


philipc:
I suspect that using BufWriter would give similar improvements. We may want to use buffered writes for all encoders (similar to what #552 did for decoders).

genkimarshall:
Good point thank you. If I use BufWriter then my test from above runs in 0.146 seconds (half as fast as original patch). But if I change the 3 write_all() calls in the for-loop to be just one then  I get 0.06696 seconds (just about as fast). But nice side is that it should speed up Gray(8) as well. That alternate branch is here for reference: https://github.com/genkimarshall/image/tree/rgba-ppm-alt

Looking at the other encoders briefly, it seems only other one that has write_all() in a loop is jpeg.

philipc:
Thanks for testing that. You're right that jpeg seems to be the only other one that would benefit.
Also note that for your alternate branch the rgb slice needs to be `&x[..3]`.

nwin:
What is the rationale for removing the `seek` calls? Ist it really giving such a big speed up? The code is much more comprehensible compared to the version with the read_uxx calls.

genkimarshall:
@philipc  Hmm for the alternate branch, if my understanding is correct, I do believe that it should be `&x[..2]` because for PPM it doesn't have an alpha channel so we only want to grab the RGB from the RGBA.

philipc:
@genkimarshall The end of the range is not inclusive. so I think that will only give you two channels.

genkimarshall:
Ah right thank you.. was not thinking.. Force pushed change.

genkimarshall:
Hello, is there anything I could do to help move this PR along?
It's my impression that the travis build failed due to rayon not compiling successfully on nightly.

bvssvni:
Seems the original use case is fixed. Closing.

PR #557, opened 2016-06-30
Improve doc for DynamicImage::resize()

### LINKED ISSUE #556: Please support resize with height or width omitted
`DynamicImage::resize` takes a width and height, and resizes preserving aspect ratio.  However, sometimes the operation you want is "make the image this wide" or "make the image this tall", without caring about the other dimension other than preserving aspect ratio.  Rather than calculate this by hand, I'd like to have a resize method that takes two `Option<u32>` arguments for the sizes, which would allow calls like `.resize(Some(80), None, ...)` to say "make this 80 pixels wide and whatever height would preserve the aspect ratio".

philipc:
@joshtriplett Is this ok?

Closes #556

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -249,7 +249,8 @@ impl DynamicImage {
 
     /// Resize this image using the specified filter algorithm.
     /// Returns a new image. The image's aspect ratio is preserved.
-    /// ```nwidth``` and ```nheight``` are the new image's dimensions
+    /// The image is scaled to the maximum possible size that fits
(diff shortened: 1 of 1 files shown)


joshtriplett:
Perfect, thanks.

nwin:
I would leave out the last sentence. We should use `Option<u32>` if we really wanted to document that behavior.

nwin:
Thank you, gh didn’t notify about the change. Please leave a message if you squash commits etc.

PR #562, opened 2016-07-24
Make decoding result enum public

srdqty:
I was trying to compile a simple example using a jpeg decoder and noticed the `DecodingResult` enum is not public.

Example code I mentioned above: https://github.com/srdqty/rust-practice/blob/57cbc9e72d2429d6600c7adbdfc73bc6d027e1b5/image-hello-world/src/main.rs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,6 +38,7 @@ pub use image::{
     ImageDecoder,
     ImageError,
     ImageResult,
+    DecodingResult,
     SubImage,
     GenericImage,
     // Iterators


nwin:
How could that have happened? Can you please remove the test? That's a bug in rustc it should have given an error.

srdqty:
It seems like the problem is that there existed no tests trying to use that type outside of the private scope of the library. Before the change, the test does fail via compiler error.

I'm pretty new to rust so maybe I'm missing something related to the precise rules of visibility in rust. I'm not sure I understand why this would indicate a bug in rustc? Should the library itself (without the test) have failed to compile because the return type of a method on a public type was not also declared public?

nwin:
Exactly, rustc usually disallows to expose a private type via a public interface. As such, before your change, the library should have failed to compile. This is why we don't need to test for it separately. I'm currently trying to reproduce this behavior in form of a minimal example.

nwin:
Apparently I'm mistaken currently it should only emit a warning (which it didn’t) instead of failing. Maybe this is also a feature of rustc, but then I don’t know how we possibly could keep track if all types where properly re-exported.

srdqty:
Yeah, that does seem strange. Even if you want the type to be abstract and not export the definition, you would at least want the name to be public.

srdqty:
@nwin This PR should be good to merge?

nwin:
Yes, don’t have much time lately.

srdqty:
Awesome, thanks.

PR #570, opened 2016-08-13
Fix build failure when HDR feature is disabled

TheEnigmaBlade:
Currently, access to the `hdr` module is attempted when building the array of magic bytes even when the feature is disabled. This change inserts the bytes directly without accessing the module.


bvssvni:
Thanks!

PR #573, opened 2016-08-19
Add Format from Memory

TheNeikos:
I was missing this functionality as I would like to keep the original format from a file on files uploaded by users.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -658,6 +658,22 @@ pub fn load_from_memory_with_format(buf: &[u8], format: ImageFormat) -> ImageRes
     load(b, format)
 }
 
+/// Guess image format from memory block
+///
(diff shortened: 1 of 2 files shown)


nwin:
Thank you for contributing. Can you please give the function a more compelling name for example 'guess_format' and also use it in 'load_from_memory' as the patch is now we have  code duplication.

TheNeikos:
Alright, how does that look like?

TheNeikos:
Okay, so this job errored: https://travis-ci.org/PistonDevelopers/image/jobs/153595100

With no error :v so I'm a bit confused. 

Could someone restart that build?

nwin:
That looks good, thank you!

TheNeikos:
When do you think is there going to be a new release? Do you have some sort of cycle?

PR #574, opened 2016-08-23
Use a buffered writer for writing images

tbu-:
This uses less `write` syscalls.


crazymykl:
Can you give a rough benchmark of what the savings are?

nwin:
@tbu- thank you for adding this

PR #583, opened 2016-10-28
move benchmarking to a 'benchmarks' feature

chills42:
I tried out moving all of the testing code for benchmarking, which uses unstable features out into it's own feature.  This should allow the rest of the tests to run on stable and beta, and then the specialized tests can still be run on nightly.


nwin:
Thank you, looks good.

PR #584, opened 2016-10-29
Expose quality/compression when saving JPEGs

### LINKED ISSUE #581: Expose quality/compression controls when saving as JPG
When saving images as JPG, it would be useful for users to decide how much they want the image to be compressed.

0xbrt:
This implements a `save_jpeg_with_quality` function for `DynamicImage`s, that takes a `Write` and a quality value. The function constructs a new `JPEGEncoder` with said quality (using the recently implemented `JPEGEncoder::new_with_quality`) and encodes it to the `Write` - simple stuff!

I also added an example showing compression. Closes #581.

### DIFF
--- a/examples/compress.rs
+++ b/examples/compress.rs
@@ -1,0 +1,24 @@
+//! Example showing JPEG compression.
+extern crate image;
+
+use std::env;
+use std::fs::File;
+use std::path::Path;
+
+use image::GenericImage;
+
+fn main() {
+    let file = env::args().nth(1).expect("Please enter a file");
+    let qual = env::args().nth(2).expect("Please enter a quality");
+    let qual: u8 = qual.parse().expect(&format!("Invalid quality: {}", qual));
+
+    // Use the open function to load an image from a Path.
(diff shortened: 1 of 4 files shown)


nwin:
I would prefer having a more generic version, otherwise we would need a special convenience function for every image format we support.

agersant:
@nwin: Is there any work left to do before this can be merged?

cdekok:
Generic quality for compression can get a bit hairy as usually different values are used for example look at imagemagick https://www.imagemagick.org/script/command-line-options.php#quality
Which handles a generic quality option but it's sometimes confusing to what it does on PNG / TIFF / JPG2000 for example.

kaj:
An alternative might be for `DynamicImage::save(...)` to take an `Into<ImageOutputFormat>` rather than an `ImageFormat`.  Then `ImageOutputFormat` could be an enum containing image formats with options (and specific constructors to set the options as supported by each format), and an `impl Into<ImageOutputFormat> for ImageFormat` could provide default options.

PR #585, opened 2016-10-29
added stub for huerotate

RedRoosterMobile:
huerotate color just like webkit css filter does:

normal
test

huerotate(180)
test2

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -107,6 +107,35 @@ pub fn brighten<I, P, S>(image: &I, value: i32)
     out
 }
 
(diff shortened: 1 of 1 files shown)


RedRoosterMobile:
whoops. not ready yet

src/imageops/colorops.rs
```diff
@@ -107,6 +107,35 @@ pub fn brighten<I, P, S>(image: &I, value: i32)
     out
 }
 
+/// Hue roatate the supplied image.
```
nabijaczleweli: Typo

PR #586, opened 2016-10-30
added huerotate

RedRoosterMobile:
huerotate color just like webkit css filter does:

normal
test

huerotate(180)
test2

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -325,7 +325,7 @@ impl DynamicImage {
     }
 
     /// Hue rotate the supplied image.
-    /// ```value``` is the degrees to rotate each pixel by.
+    /// `value` is the degrees to rotate each pixel by.
     /// 0 and 360 do nothing, the rest rotates by the given degree value.
     /// just like the css webkit filter hue-rotate(180)
     pub fn hueroate(&self, value: i32) -> DynamicImage {
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -324,6 +324,10 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::brighten(p, value))
     }
 
+    /// Hue rotate the supplied image.
+    /// ```value``` is the degrees to rotate each pixel by.
+    /// 0 and 360 do nothing, the rest rotates by the given degree value.
+    /// just like the css webkit filter hue-rotate(180)
     pub fn hueroate(&self, value: i32) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::huerotate(p, value))
     }
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -324,6 +324,10 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::brighten(p, value))
     }
 
+    pub fn hueroate(&self, value: i32) -> DynamicImage {
+        dynamic_map!(*self, ref p => imageops::huerotate(p, value))
+    }
+
     /// Flip this image vertically
     pub fn flipv(&self) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::flip_vertical(p))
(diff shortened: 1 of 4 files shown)


RedRoosterMobile:
I'm new to rust and would welcome a proper code review

src/imageops/colorops.rs
```diff
@@ -106,6 +107,82 @@ pub fn brighten<I, P, S>(image: &I, value: i32)
...
+fn clamp_u8(num:f64) -> u8 {
+    if num < 0.0 {
+        return u8::min_value();
+    } else if num > 255.0 {
+        return u8::max_value();
+    } else {
+        return num as u8;
+    }
+}
+
```
nwin: You could just use `math::utils::clamp` instead.

src/imageops/colorops.rs
```diff
@@ -106,6 +107,82 @@ pub fn brighten<I, P, S>(image: &I, value: i32)
...
+    matrix[2] = 0.072 - cosv * 0.072 + sinv * 0.928;
+
+    matrix[3] = 0.213 - cosv * 0.213 + sinv * 0.143;
+    matrix[4] = 0.715 + cosv * 0.285 + sinv * 0.140;
+    matrix[5] = 0.072 - cosv * 0.072 - sinv * 0.283;
+
+    matrix[6] = 0.213 - cosv * 0.213 - sinv * 0.787;
+    matrix[7] = 0.715 - cosv * 0.715 + sinv * 0.715;
+    matrix[8] = 0.072 + cosv * 0.928 + sinv * 0.072;
+
```
nwin: You’re initializing the matrix as _I_3_ but you aren’t using it afterwards. You can just initialize it directly, then it doesn’t have to be `mut`. Or at least initialize it as `[0; 9]` otherwise that is confusing.

src/imageops/colorops.rs
```diff
@@ -106,6 +107,82 @@ pub fn brighten<I, P, S>(image: &I, value: i32)
...
+    for y in 0..height {
+        for x in 0..width {
+
```
nwin: Please use `for ... in image.enumerate_pixels_mut()` here.

src/imageops/colorops.rs
```diff
@@ -106,6 +107,82 @@ pub fn brighten<I, P, S>(image: &I, value: i32)
...
+            let p = image.get_pixel(x, y);
+            let (k1, k2, k3, k4) = p.channels4();
+            let vec: (f32, f32, f32, f32) = (
+                NumCast::from(k1).unwrap(),
```
nwin: You could as well use `f64` here that would save you some casts.

src/dynimage.rs
```diff
@@ -324,6 +324,14 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::brighten(p, value))
     }
 
+    /// Hue rotate the supplied image.
+    /// ```value``` is the degrees to rotate each pixel by.
+    /// 0 and 360 do nothing, the rest rotates by the given degree value.
```
nwin: `value` should be enough.

RedRoosterMobile:
thanks for checking. I applied your change requests

RedRoosterMobile:
guess it's ready to merge

RedRoosterMobile:
implementation is borrowed from https://github.com/WebKit/webkit/blob/master/Source/WebCore/platform/graphics/texmap/TextureMapperShaderProgram.cpp

where they have other stuff like a sepia filter.

nwin:
Thanks looks good now, can you please squash your commits?

RedRoosterMobile:
I don't really know how to do that, but it looks like you, as an administrator, can do it with some clicks:
https://github.com/blog/2141-squash-your-commits

nwin:
Ah, that is new. Thanks, I learned something. :)

FreeFull:
One problem, you misspelled `huerotate` as `hueroate` in `DynamicImage`'s impl

PR #588, opened 2016-11-02
Added a tint function that tints an image with a given color

RedRoosterMobile:
normal:

test

```
// tint with red
tint_rgb(0.5, 255,0,0); // OR
tint_hex(0.5, "#ACDCAD");
```

test2


theotherphil:
There's similar functionality here https://github.com/PistonDevelopers/imageproc/blob/master/examples/color_effects.rs with examples in the last few comments here: 
https://github.com/PistonDevelopers/imageproc/issues/96

nwin:
Thank you. Small review

 - `hex2rgb shouldn’t be added to `DynamicImage` it is a static utility function as such it should just go to `utils` (non-public) or maybe `color`. 
 - Please don’t pull in two more dependencies for such a small function. Hex colors are easy enough to parse by hand. First check if it starts by `#` then if it has 3 or 6 chars…
- The function should return `Option<(u8, u8, u8)>`

nagisa:
How about just `tint(..., rgb: image::Rgb<T>)` and then add something like a `FromStr` impl for `Rgb`?

PR #589, opened 2016-11-03
Avoid unnecessary allocations and copies in JPEG encoder

mbrubeck:
This splits the writing methods into a separate struct so they can be passed borrowed data from the encoder, instead of cloned data, to eliminate unnecessary allocations in the inner loop of `encode`. Credit goes to vfs on IRC for noticing the extra allocations.

### DIFF
--- a/src/jpeg/encoder.rs
+++ b/src/jpeg/encoder.rs
@@ -158,168 +158,19 @@ struct Component {
     dc_pred: i32
 }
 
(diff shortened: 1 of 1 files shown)


nwin:
Thank you! :)

tafia:
> FYI: In benchmarking this PR I've noticed about a 12-18% speed improvement.

crazy!

PR #590, opened 2016-11-03
Fix building of benchmarks.

vickenty:
When `benchmarks` feature is enabled in cargo, it be passed to rustc as `feature="benchmarks"` configuration option, not just `benchmarks`.


bvssvni:
Thanks!

PR #596, opened 2016-11-26
Regenerated Github auth token

bvssvni:
- Replaced old token with a new one
- Redirect stdout and stderr to /dev/null to hide output on error

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ rust:
     - nightly
 env:
     global:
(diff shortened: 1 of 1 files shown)


bvssvni:
The  command`cargo build -v` is failing on Rust nightly for some reason. Merging.

bvssvni:
Thanks to Ivan Vyshnevskyi for discovering the leak!

PR #600, opened 2016-12-22
dynimage: fix typo in huerotate method name

donachys:
minor fix to make the method name consistent with the imageops/colorops.rs huerotate function


bvssvni:
Thanks!

bvssvni:
I want to see if we can do something about https://github.com/PistonDevelopers/image/issues/419#issuecomment-105811998 before publishing.

PR #606, opened 2017-01-06
Published 0.12.0

bvssvni:
- Update to latest image-png
- Removed unused imports


bvssvni:
I think we're pure Rust now?

PR #611, opened 2017-01-16
Disable default features on num-rational.

Ms2ger:
In particular, this removes the dependency on num-bigint.


nwin:
Good catch, thanks!

PR #617, opened 2017-02-04
Add a basic PPM decoder

pedrocr:
Adds a basic decoder for 8bit PPM files following the normal setup of the other decoders.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -86,6 +86,7 @@ impl From<io::Error> for ImageError {
 pub type ImageResult<T> = Result<T, ImageError>;
 
 /// Result of a decoding process
+#[derive(Debug)]
 pub enum DecodingResult {
     /// A vector of unsigned bytes
     U8(Vec<u8>),
--- a/src/ppm/decoder.rs
+++ b/src/ppm/decoder.rs
@@ -37,7 +37,7 @@ impl<R: Read> PPMDecoder<R> {
           let mut buffer: [u8;1] = [0];
           try!(reader.read_exact(&mut buffer));
           match buffer[0] {
-              b'\n' | b' ' | b'\r' => {
+              b'\n' | b' ' | b'\r' | b'\t' => {
                   if bytes.len() > 0 {
                     break // We're done as we already have some content
                   }
--- a/src/ppm/decoder.rs
+++ b/src/ppm/decoder.rs
@@ -9,6 +9,7 @@ pub struct PPMDecoder<R> {
     reader: BufReader<R>,
     width: u32,
     height: u32,
+    maxwhite: u32,
 }
 
 impl<R: Read> PPMDecoder<R> {
@@ -18,11 +19,12 @@ impl<R: Read> PPMDecoder<R> {
         try!(PPMDecoder::read_next_string(&mut buf)); // Skip P6
         let width = try!(PPMDecoder::read_next_u32(&mut buf));
         let height = try!(PPMDecoder::read_next_u32(&mut buf));
-        try!(PPMDecoder::read_next_u32(&mut buf)); // Skip maxwhite
+        let maxwhite = try!(PPMDecoder::read_next_u32(&mut buf));
         Ok(PPMDecoder {
             reader: buf,
             width: width,
             height: height,
+            maxwhite: maxwhite,
         })
     }
 
@@ -64,7 +66,13 @@ impl<R: Read> ImageDecoder for PPMDecoder<R> {
     }
 
     fn colortype(&mut self) -> ImageResult<ColorType> {
-        Ok(ColorType::RGB(8))
+        if self.maxwhite < 256 {
+          Ok(ColorType::RGB(8))
+        } else if self.maxwhite < 65536 {
+          Ok(ColorType::RGB(16))
+        } else {
+          Err(ImageError::FormatError("Don't know how to decode PPM with more than 16 bits".to_string()))
+        }
     }
 
     fn row_len(&mut self) -> ImageResult<usize> {
@@ -76,12 +84,21 @@ impl<R: Read> ImageDecoder for PPMDecoder<R> {
     }
 
     fn read_image(&mut self) -> ImageResult<DecodingResult> {
-        let mut data = vec![0 as u8; (self.width*self.height*3) as usize];
+        let bytewidth = if self.maxwhite < 256 { 1 } else { 2 };
+        let mut data = vec![0 as u8; (self.width*self.height*3*bytewidth) as usize];
         match self.reader.read_exact(&mut data) {
             Ok(_) => {},
             Err(e) => return Err(ImageError::IoError(e)),
         };
 
-        Ok(DecodingResult::U8(data))
+        if bytewidth == 1 {
+            Ok(DecodingResult::U8(data))
+        } else {
+            let mut out = vec![0 as u16; (self.width*self.height*3) as usize];
+            for (o, i) in out.chunks_mut(1).zip(data.chunks(2)) {
+                o[0] = (i[0] as u16) << 8 | i[1] as u16;
+            }
+            Ok(DecodingResult::U16(out))
+        }
     }
 }
--- a/src/ppm/decoder.rs
+++ b/src/ppm/decoder.rs
@@ -33,18 +33,17 @@ impl<R: Read> PPMDecoder<R> {
 
   fn read_next_string(reader: &mut BufReader<R>) -> ImageResult<String> {
       let mut bytes = Vec::new();
-      loop {
-          let mut buffer: [u8;1] = [0];
-          try!(reader.read_exact(&mut buffer));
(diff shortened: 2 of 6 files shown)


src/ppm/decoder.rs
```diff
@@ -0,0 +1,87 @@
...
+            reader: buf,
+            width: width,
+            height: height,
+        })
+    }
+
+  fn read_next_string(reader: &mut BufReader<R>) -> ImageResult<String> {
+      let mut bytes = Vec::new();
+      loop {
+          let mut buffer = vec![0 as u8];
```
oyvindln: Wouldn't it be better to use a simple [u8] here? (e.g `let mut buffer = [0u8]`) Creating and dropping an 1-byte vec for each byte isn't going to be great for performance.

pedrocr: Sure, will push a fix but this is run for all of 20 bytes or so :)

nwin:
Can you please also add a little test like round-tripping en- and decoding a 1 px red image.

pedrocr:
I've added a round-trip test and while I was at it added 16bit decoding as well. While doing that I realized the 16bit encoding was broken and fixed it.

src/image.rs
```diff
@@ -86,6 +86,7 @@ impl From<io::Error> for ImageError {
 pub type ImageResult<T> = Result<T, ImageError>;
 
 /// Result of a decoding process
+#[derive(Debug)]
```
nagisa: Should this be here, or did you intend to only use this for debugging?

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+        })
+    }
+
+  fn read_next_string(reader: &mut BufReader<R>) -> ImageResult<String> {
+      let mut bytes = Vec::new();
+      loop {
+          let mut buffer: [u8;1] = [0];
+          try!(reader.read_exact(&mut buffer));
+          match buffer[0] {
+              b'\n' | b' ' | b'\r' => {
```
nagisa: Reading http://netpbm.sourceforge.net/doc/ppm.html suggests there may be more types of whitespace allowed, namely a `\t`.

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+            height: height,
+            maxwhite: maxwhite,
+        })
+    }
+
+  fn read_next_string(reader: &mut BufReader<R>) -> ImageResult<String> {
+      let mut bytes = Vec::new();
+      loop {
+          let mut buffer: [u8;1] = [0];
+          try!(reader.read_exact(&mut buffer));
```
nagisa: Why not an iterator over `reader.bytes()`?

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+        Ok(PPMDecoder {
+            reader: buf,
+            width: width,
+            height: height,
+            maxwhite: maxwhite,
+        })
+    }
+
+  fn read_next_string(reader: &mut BufReader<R>) -> ImageResult<String> {
+      let mut bytes = Vec::new();
```
nagisa: Seems a shame to allocate for each field, while you could read the whole header into a `Vec` in one go and only then parse. May not be worth the engineering effort though.

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+        Ok((self.width, self.height))
+    }
+
+    fn colortype(&mut self) -> ImageResult<ColorType> {
+        if self.maxwhite < 256 {
+          Ok(ColorType::RGB(8))
+        } else if self.maxwhite < 65536 {
+          Ok(ColorType::RGB(16))
+        } else {
+          Err(ImageError::FormatError("Don't know how to decode PPM with more than 16 bits".to_string()))
```
nagisa: Rather than not knowing how to decode, it is plain invalid according to http://netpbm.sourceforge.net/doc/ppm.html.

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    fn row_len(&mut self) -> ImageResult<usize> {
+        Ok((self.width*3) as usize)
```
nagisa: This should take into account the bit-depth.

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn read_scanline(&mut self, _buf: &mut [u8]) -> ImageResult<u32> {
```
nagisa: You probably *should* implement this. It should read a `row_len` bytes into the `_buf`, which seems pretty trivial to do. You could then "reuse" this to implement `read_image` like follows:

```rust
fn read_image(&mut self) -> ImageResult<DecodingResult> {
    let mut data = Vec::new();
    data.resize(self.row_len() * self.height, 0);
    for chunk in data.chunks_mut(self.row_len()) { 
         let ret = try!(self.read_scanline(chunk))
         debug_assert!(ret, self.row_len());
    }
}
```

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+            Ok(_) => {},
+            Err(e) => return Err(ImageError::IoError(e)),
+        };
+
+        if bytewidth == 1 {
+            Ok(DecodingResult::U8(data))
+        } else {
+            let mut out = vec![0 as u16; (self.width*self.height*3) as usize];
+            for (o, i) in out.chunks_mut(1).zip(data.chunks(2)) {
+                o[0] = (i[0] as u16) << 8 | i[1] as u16;
```
nagisa: This is wrong as it assumes the host machine is low endian, which is not always true.

Instead [`u16::from_be`](https://doc.rust-lang.org/stable/std/primitive.u16.html#method.from_be) should be used.

src/ppm/decoder.rs
```diff
@@ -0,0 +1,104 @@
...
+        let mut data = vec![0 as u8; (self.width*self.height*3*bytewidth) as usize];
+        match self.reader.read_exact(&mut data) {
+            Ok(_) => {},
+            Err(e) => return Err(ImageError::IoError(e)),
+        };
+
+        if bytewidth == 1 {
+            Ok(DecodingResult::U8(data))
+        } else {
+            let mut out = vec![0 as u16; (self.width*self.height*3) as usize];
```
nagisa: A copy could be avoided by allocating a vector of `u16` in the first place and then converting it to `Vec<u8>` with some unsafe code:

```rust
let bytedata = Vec::from_raw_parts(data.as_mut_ptr() as *mut u8, data.len(), data.capacity());
mem::forget(data);
```

It is up to you if you want to act on it. (also bear in mind that converting from `Vec<u8>` allocation to `Vec<u16>` is incorrect, one must to allocate `Vec<u16>` if they want to do this)

src/image.rs
pedrocr: My impression is that structs should generally implement Debug to make everyone's life easier. I used it while developing and left it in intentionally but it's not strictly needed.

src/ppm/decoder.rs
pedrocr: It's a 3 field header, performance-wise it's irrelevant and it makes the code simpler.

src/ppm/decoder.rs
pedrocr: Could probably work, didn't look at that method

src/ppm/decoder.rs
pedrocr: Probably makes sense to add it. Haven't found a case of it though.

src/ppm/decoder.rs
pedrocr: PPM is very loosely defined so I wouldn't be surprised someone has written a 32bit float or something strange. But yeah, it's supposedly invalid I just worded the error message as a surprise instead of a complaint :)

src/ppm/decoder.rs
pedrocr: Ah, right it's in bytes, thought it was pixels.

src/ppm/decoder.rs
pedrocr: I've done exactly that but it breaks the decode. I didn't see other formats implementing this so I suspect there's a bug on the read_scanline() using code.

src/ppm/decoder.rs
pedrocr: I thought of doing a transmute but then realized endian decoding was needed.

src/ppm/decoder.rs
pedrocr: No assumption is made. shifting always works the same way, be the machine big or little endian. But using the platform function is probably best indeed.

nwin:
Thank you!

src/image.rs
pedrocr: Actually I did use this but only in the PPM tests, so it is indeed strictly needed for this code.

Issue #618, opened 2017-02-08
"scan makes use of unset dc huffman table"


ArtemGr:
When I scale down the following image
http://images.icecat.biz/img/gallery/img_29375333_high_1482475603_0907_3862.jpg
and then try to load it back, I get the "scan makes use of unset dc huffman table" error
(and GraphicsMagick rejects the image with "Huffman table 0x00 was not defined").

This happens with a lot of ICECat images.

```
    let im = image::load_from_memory(&original)?;
    let ib = image::imageops::resize(&im, nw, nh, image::FilterType::Lanczos3);
    let mut downscaled = Vec::with_capacity(524288);
    {
        let mut j = image::jpeg::JPEGEncoder::new_with_quality(&mut downscaled, 90);
        j.encode(&ib, ib.width(), ib.height(), image::ColorType::RGBA(8))?;
    }
    image::load_from_memory(&downscaled)?; // Fails!
```

The problem vanishes when I downgrade the image crate to "0.10".

CPU: AMD Opteron(tm) Processor 3280.
Rust: rustc 1.17.0-nightly (0648517fa 2017-02-03).

ArtemGr:
Test case:

```
extern crate image;

use std::io::Read;

fn main() {
    let mut original = Vec::new();
    let mut file = std::fs::File::open("img_29375333_high_1482475603_0907_3862.jpg")
        .expect("!img_29375333_high_1482475603_0907_3862.jpg");
    file.read_to_end(&mut original).expect("!read");

    let im = image::load_from_memory(&original).expect("!load_from_memory 1");
    let ib = image::imageops::resize(&im, 100, 100, image::FilterType::Lanczos3);
    let mut downscaled = Vec::with_capacity(524288);
    {
        let mut j = image::jpeg::JPEGEncoder::new_with_quality(&mut downscaled, 90);
        j.encode(&ib, ib.width(), ib.height(), image::ColorType::RGBA(8)).expect("!encode");
    }
    image::load_from_memory(&downscaled).expect("!load_from_memory 2");
}
```

Fails on Windows 10, Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz, rustc 1.16.0-nightly (bf6d7b665 2017-01-15).

jdpage:
Does it fail if you don't resize? If so, it's probably a bug in the JPEG encoder introduced in 2c06a405ac7e3a04

ArtemGr:
> Does it fail if you don't resize?

[...]

PR #619, opened 2017-02-09
Make JPEG encoder output valid images again.

jdpage:
Fixes issue #618. It looks like commit 2c06a405ac7e3a04d139c1de16d07263c4a13573
stopped emitting the DC table by accident, resulting in JPEG files missing
huffman table 0.


nwin:
Thank yo. 🙂 Can you please also add a regression test? E.g. a reference 1px image or so?

jdpage:
Can do! What file do regression tests for encoders live in? (Are we putting them in with the encoders our out in the tests director?)

nwin:
Ok, thank you! :)

PR #630, opened 2017-02-26
Implement read_scanline() for PPMs

pedrocr:
I had tried implementing this like this before and it didn't work for some reason. Either something was fixed or I was doing something wrong. This version seems to work fine though.


nwin:
Thank you but I the read_scanline function should be removed on the long run as not every format supports it.

pedrocr:
Ok, I was just going for completeness. If this is the case wouldn't it be simple to just rip it out?

PR #637, opened 2017-04-22
dynimage: uncomment the catch-all arm

mathstuf:
When built without all of the features enabled, this arm is required.

---
Fixes the build following #617.


bvssvni:
Look good!

PR #646, opened 2017-05-10
Add simple BMP encoder.

brettfo:
Using the [Wikipedia page on BMP files](https://en.wikipedia.org/wiki/BMP_file_format) and the `JPEGEncoder` as a guide, I've implemented simple RGB BMP encoding.  RGBA images are simply stripped of their alpha channel as transparent bitmaps aren't widely supported.

In addition to the two unit tests, I've also manually validated the fractal example from the `README` works when `Luma` -> `Rgb`, `".png"` -> `".bmp"`, and `image::PNG` -> `image::BMP`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ https://docs.rs/image
 | PNG    | All supported color types | Same as decoding|
 | JPEG   | Baseline and progressive | Baseline JPEG |
 | GIF    | Yes | Yes |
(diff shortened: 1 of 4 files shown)


brettfo:
Shameless ping to @nwin and/or @ccgn for a review.

nwin:
Looks good, thank you!

PR #649, opened 2017-05-18
Enable encoding grayscale BMP images.

brettfo:
As a follow-up to #646, this adds the ability to encode grayscale BMP images.  Alpha channel handling is the same as for RGB BMP files; that is the alpha is stripped off because it's not a widely supported feature.

In addition to the unit tests, I've also manually verified the fractal example from the README works with luma-encoded BMP images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ https://docs.rs/image
 | PNG    | All supported color types | Same as decoding|
 | JPEG   | Baseline and progressive | Baseline JPEG |
 | GIF    | Yes | Yes |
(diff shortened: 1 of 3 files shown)


brettfo:
Looking at the Travis CI logs, the failures are all from the beta compiler which is a known issue: rust-lang/rust#41944.  Stable and nightly are green.

brettfo:
Beta compiler issue has been fixed, all tests are passing now.

brettfo:
Another shameless ping to @nwin for review.

nwin:
Thank you, looks good!

PR #651, opened 2017-05-27
Add fuzzing scripts

neosilky:
I've added fuzzing scripts to decode each format explicitly and one that runs the generic `load_from_memory` function.

I've got some crashes in incoming issues and PRs.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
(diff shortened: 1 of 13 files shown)


nwin:
Thank you, I’m not aware of this fuzzing framework. Is it something which could be automatically run?

neosilky:
Hey @nwin! The framework is [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz) and can easily be run by using `RUST_BACKTRACE=1 cargo fuzz run fuzzer_script_1 -- -max_len=10240`.

I've got a few crashes which I am just triaging... :)

neosilky:
ping @nwin

nwin:
ok, looks good, sorry for the delay :)

PR #652, opened 2017-05-28
Fix `illegal_floating_point_literal_pattern` warning

ordovicia:
This PR fixes `illegal_floating_point_literal_pattern` warning raised from `src/imageops/sample.rs`.

This warning will be a hard error in a future Rust (discussed in https://github.com/rust-lang/rust/issues/41620).


ordovicia:
083468d28ef34d2698658b98a5e5e598ab22290f
This commit has the same effect as that of this pull request.

PR #654, opened 2017-06-06
Fix issues indicated by clippy

neosilky:


nwin:
That’s a lot, thanks.

Issue #545, opened 2016-06-10
Run TIFF decoder through tiff test images


philipc:
Images at http://www.remotesensing.org/libtiff/images.html

nwin:
That’ll be nice but we would need to be very picky about the test images. We don’t support these fax encodings, no jpegs, etc. That I didn’t add a test image for `lzw` was my oversight though.

HeroicKatora:
Link is down. Potential replacement: https://download.osgeo.org/libtiff/

PR #659, opened 2017-06-16
Fix tiff Grey(16) support

### LINKED ISSUE #426: Add support for 16 bit GrayScale TIFF
It would be nice if support to read and write 16 bit GrayScale TIFFs is added.

khyperia:
Currently, `image` produces the (incorrect) error message somewhere along the lines of "Grey(16) is not supported for tiff". This is because of a wacky artifact of code - a DecodingResult::U8 is produced (due to the code edited in this PR), but Grey(16) is the image type, so a [match block falls through](https://github.com/khyperia/image/blob/f23fef81cf92325f1e793b7343f540bf8eaeb94e/src/tiff/decoder/mod.rs#L434) and an error is produced.

U8 is selected because of some strange logic to divide by 8 - I'm guessing that 16 bit images were always broken, and "n" is actually 1 in the case of 8 bit images - 2 in the case of 16 bit images (which is less than 8, so U8 is selected).

Tested manually. Seems like #545 or something similar should be implemented to make sure 16 bit tiffs aren't regressed again.

I guess this also resolves #426 ?

### DIFF
--- a/src/tiff/decoder/mod.rs
+++ b/src/tiff/decoder/mod.rs
@@ -463,10 +463,7 @@ impl<R: Read + Seek> ImageDecoder for TIFFDecoder<R> {
             self.width  as usize
             * self.height as usize
             * self.bits_per_sample.iter().count();
-        let mut result = match (self.bits_per_sample.iter()
-                                               .cloned()
-                                               .max()
-                                               .unwrap_or(8) as f32/8.0).ceil() as u8 {
+        let mut result = match self.bits_per_sample.iter().cloned().max().unwrap_or(8) {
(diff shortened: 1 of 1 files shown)


khyperia:
Note: The file I was using to test is this image: https://khyperia.com/space/2014_05_31/M17/telescope.5-31.1-42-2.tiff (if reviewers need a source of a 16 bit greyscale tiff - they're kind of rare)

nwin:
You’re right that really doesn’t make sense. Thank you for tracking this down. Can you please rebase to resolve the conflicts?

khyperia:
@nwin Rebased! 😊

bvssvni:
Looks good! Merging.

PR #661, opened 2017-07-02
Resize fixes

stevebob:
This fixes some problems I noticed with resizing images. I'll present some examples of images resized by this library prior to these changes, then show how they look after these changes. To make pixels visible, I've scaled each image up to 256x256 with no interpolation.

We'll start with this 16x16 image:
original

### Nearest neighbour scaling up also translates output

Double it's size using `FilterType::Nearest`:
```
let scaled = original.resize(32, 32, FilterType::Nearest);
```
Prior to this change, the output image was translated up and to the left by 1 pixel:
scale-broken

After this change it no longer translates:
scale-fixed

### Nearest neighbour scaling to original size produces artefacts

```
let scaled = original.resize(16, 16, FilterType::Nearest);
```
Prior to this change, each pixel's colour was influenced by its neighbour:
same-size-broken

After this change it looks like the original image:
same-size-fixed

### Nearest neighbour scaling down produces artefacts

```
let scaled = original.resize(8, 8, FilterType::Nearest);
```

Prior to this change, output pixels are influenced by input pixel neighbours:
reduce-broken

After this change no new colours are introduced:
reduce-fixed

### Scaling with triangle interpolation is asymetric

```
let scaled = original.resize(32, 32, FilterType::Triangle);
```

Before:
triangle-broken

After:
triangle-fixed

## Solutions

The triangle filter solution was easy - it was just missing a `.abs()`.

The problems with nearest-neighbour were indicative of deeper problems with image sampling. Specifically:
 - The rounding off of the left and right sides of the sample window (`left` and `right`), and the calculation of the value to pass to `filter.kernel` was off by 0.5.
 - When an output pixel position corresponded to the boundary of two input pixels, nearest neighbour would sample and average out both pixels. I changed sampling to break ties by always preferring pixels to the left and above the border.
 - The code used to scale up the filter radius when down-sampling. This produced artefacts with nearest neighbour when down-sampling, as it was allowing multiple pixels to influence the colour of each output pixel. I remove the code that implements this.

I also added detailed comments explaining the calculations that I changed, and removed some unnecessary duplication left over from when this code used SIMD.

I'm happy to explain my changes in more detail if anyone has questions. What's the best way to test that the more complex `FilterType`s still work after these changes?

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -109,7 +109,7 @@ pub fn catmullrom_kernel(x: f32) -> f32 {
 /// Also known as BiLinear sampling in two dimensions.
 pub fn triangle_kernel(x: f32) -> f32 {
     if x.abs() < 1.0 {
-        1.0 - x
+        1.0 - x.abs()
     } else {
         0.0
     }
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -147,15 +147,6 @@ fn horizontal_sample<I, P, S>(image: &I, new_width: u32,
 
         let ratio = width as f32 / new_width as f32;
 
-        // Scale the filter when downsampling.
-        let filter_scale = if ratio > 1.0 {
-            ratio
-        } else {
-            1.0
-        };
-
-        let filter_radius = filter.support * filter_scale;
-
         for outx in 0..new_width {
 
             // Find the point in the input image corresponding to the centre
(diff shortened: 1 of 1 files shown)


nwin:
@stevebob Thank you very much for your fix and the presentation of the errors. I have no good idea how to test the more complex filter types other than to add test cases like the images you presented in your PR.

PR #666, opened 2017-07-10
Add BGRA and BGR support.

### LINKED ISSUE #451: Support non-standard pixel types like Apple's BGRA
Apple (and possibly Windows' GDI) use a non-standard pixel format internally, [BGRA](http://iphonedevelopment.blogspot.com/2008/10/iphone-optimized-pngs.html). This means that all image buffers that come out of `CGDisplayCreateImage` / `CGImageGetDataProvider` / `CGImageGetDataProvider` / Apple's graphics APIs have the B and R pixel swapped. It would be very convenient to be able to tell `image` that a raw buffer is BGRA, rather than having to swap them manually. The information could also be stored, and used to avoid more byte-swapping, in case the Image is saved to a format that also uses BGR.

Edit: this format is commonly called [ARGB](https://en.wikipedia.org/wiki/RGBA_color_space#ARGB).
Edit 2: confusingly, OpenGL refers to this as [BGRA](https://en.wikipedia.org/wiki/RGBA_color_space#RGBA_byte_order). In any case, we should support it.

quadrupleslap:
Also allows TGA and PPM encoders to be used with them.

Resolves #451 (I think.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ All image format decoders implement the ```ImageDecoder``` trait which provides 
 ```image``` provides the following pixel types:
 + **Rgb**: RGB pixel
 + **Rgba**: RGBA pixel
++ **Bgr**: BGR pixel
++ **Bgra**: BGRA pixel
 + **Luma**: Grayscale pixel
 + **LumaA**: Grayscale with alpha
 
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -6,7 +6,7 @@ use std::io;
 use num_traits::Zero;
 
 use traits::Primitive;
-use color::{ Rgb, Rgba, Luma, LumaA, FromColor, ColorType };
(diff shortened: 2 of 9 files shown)


nwin:
I think it would be a better option to convert it to RGB for now. I'm hesitant adding new values to that enumeration since it changes a lot of public API.

quadrupleslap:
Fair enough, but can you at least check over the PPM encoder? It's short, but *very* strangely written, e.g. `let _ = try!(...);` is everywhere, and I'm not sure why it's calling `write_all` three times with an array of one element, instead of once with an array of three elements.

bvssvni:
The PPM encoder was changed in https://github.com/PistonDevelopers/image/pull/730, so I think this should be closed. Reopen if you want to make another attempt.

PR #671, opened 2017-07-21
new error class in more recent rust releases?

andrew-lucker:
Before Patch:

   Compiling image v0.14.0 (file:///Users/andrewjohnson/pd/image)

    error[E0001]: unreachable pattern
       --> src/dynimage.rs:641:9
        |
    641 |         _ => Err(image::ImageError::UnsupportedError(format!("A decoder for {:?} is not available.", format))),
        |         ^ this is an unreachable pattern
    
    error: aborting due to previous error
    
    error: Could not compile `image`.

After: Running with latest rust version, recognizes exhaustive matching with wildcard as error now.


andrew-lucker:
My bad, I was running rustc with version 1.15. This issue has been downgraded to a warning in later builds (closing).

PR #672, opened 2017-07-22
Update README.md

adrianN:
Remove unnecessary parentheses from iteration in the Julia set example to avoid a compiler warning.


bvssvni:
Thanks!

PR #675, opened 2017-07-28
Fixes #629 - PPM multiply with overflow

### LINKED ISSUE #629: PPMDecoder: multiply with overflow
Decoding a PPM image with huge dimensions

```
UDYKNTUyCjY2NjY2NjY1IDUK///////////////////////G/9j/////////+v//////////Cg==
```

fails with 

```
thread 'main' panicked at 'attempt to multiply with overflow', src/ppm/decoder.rs:87
```

when debug-assertions are enabled.

### Code

```
extern crate image;

use image::ImageDecoder;

fn main() {
    let f = ::std::fs::File::open("overflow.ppm").unwrap();
    let x = image::ppm::PPMDecoder::new(f).unwrap().read_image().unwrap();
}
```

---

Ideally decoder would return an Err instead of panicking.

tatref:
I made a little mess with the commits, please tell me if you want me to squash or correct them.

### DIFF
--- a/src/ppm/decoder.rs
+++ b/src/ppm/decoder.rs
@@ -84,7 +84,17 @@ impl<R: Read> ImageDecoder for PPMDecoder<R> {
     }
 
     fn read_image(&mut self) -> ImageResult<DecodingResult> {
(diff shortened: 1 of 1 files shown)


bvssvni:
Looks good! Merging.

PR #677, opened 2017-09-18
Use FnMut not Fn in most places

mgsloan:
Note, I'm quite new to rust, so this should be checked carefully by someone that knows what they're doing.  This is probably my 4th day playing with it (non consecutive).

I wanted to mutate a vector accumulating some other results while building an image buffer, and ran into `ImageBuffer::from_fn` wanting a `Fn` which does not allow mutation of the captured environment.  I believe this is the correct change by analogy to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map .

There are now just two occurrences of `Fn`:

* In `imageops::sample::Filter.kernel` (is this a reasonable way to identify that?), seems reasonable to have it be a `Fn` since the kernel should be as close to "pure" as possible.

* In `hdr_decoder::HDRDecoder.read_image_transform`, since it's using threadpools, shouldn't allow concurrent mutation of the closure environment, if I understand correctly.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -71,30 +71,30 @@ pub trait Pixel: Copy + Clone {
     fn to_luma_alpha(&self) -> LumaA<Self::Subpixel>;
 
     /// Apply the function ```f``` to each channel of this pixel.
-    fn map<F>(&self, f: F) -> Self where F: Fn(Self::Subpixel) -> Self::Subpixel;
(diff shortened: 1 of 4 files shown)


bvssvni:
cc @nwin

bvssvni:
It looks good to me! Since `FnMut` is a supertrait of `Fn` it should not break any existing code: https://doc.rust-lang.org/std/ops/trait.Fn.html

Merging.

PR #681, opened 2017-10-13
Add comment handling to PPM decoder

### LINKED ISSUE #628: PPMDecoder cannot read this PPM image generated by GIMP
```
UDYKIyBDUkVBVE9SOiBHSU1QIFBOTSBGaWx0ZXIgVmVyc2lvbiAxLjEKNSA1CjI1NQr/////////
////////////////////////////////////////////////////////////////////////////
//////////////8=
```

Is a base64 encoding of a PPM image exported by GIMP. The new PPMDecoder cannot decode it with error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: FormatError("Couldn\'t read preamble")', /checkout/src/libcore/result.rs:860
```

for code 

```
extern crate image;

use image::ImageDecoder;

fn main() {
    let f = ::std::fs::File::open("fuzz/ppm-corpus/test.ppm").unwrap();
    let x = image::ppm::PPMDecoder::new(f).unwrap().read_image().unwrap();
}
```

---

The cause most likely is because GIMP inserts something comment-like into the header of the file:

```
P6
# CREATOR: GIMP PNM Filter Version 1.1
5 5
255
```

where our decoder does not handle anything like this.

anthonynguyen:
This should fix #628

### DIFF
--- a/src/ppm/decoder.rs
+++ b/src/ppm/decoder.rs
@@ -31,35 +31,52 @@ impl<R: Read> PPMDecoder<R> {
         })
     }
 
-  fn read_next_string(reader: &mut BufReader<R>) -> ImageResult<String> {
-      let mut bytes = Vec::new();
-      for byte in reader.bytes() {
-          match byte {
-              Ok(b'\n') | Ok(b' ') | Ok(b'\r') | Ok(b'\t') => {
-                  if !bytes.is_empty() {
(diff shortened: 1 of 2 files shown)


bvssvni:
Looks good! Merging.

PR #684, opened 2017-10-17
Fix formatting warnings with commonmark enabled

mbrubeck:
This makes documentation work correctly with the new pulldown-cmark Markdown parser (rust-lang/rust#44229).


bvssvni:
Thanks!

PR #686, opened 2017-10-19
Add PackBits decoding to TIFF

aschampion:
This is my first Rust PR and I only have a few thousand lines experience with Rust, so let me know if there are obvious style/idiom/perf deficiencies.

The other stream decoders weren't documented so I didn't bother. A test is included. I've also manually tested end-to-end with both 8- and 16-bit TIFFs encoded with PackBits by a few different programs.


aschampion:
Anything I can do to help move this towards a merge?

bvssvni:
Looks good to me!

@nwin Thoughts?

nwin:
Thank your for the PR. Looks fine to me. Could you please change wrap to new? This is the convention in Rust. 

Furthermore please add a test file which is using packbits to prevent regressions.

aschampion:
Thanks, changes made/pushed and the image test also caught a bug which is fixed.

bvssvni:
Nice! Merging.

PR #687, opened 2017-10-22
Add crates.io categories

ijks:
I noticed this crate didn't have any categories on crates.io. I figured at least `Multimedia::Images` and `Multimedia::Encoding` would be relevant. I'm not sure if `Multimedia` itself is automatically implied if these categories are present.


bvssvni:
Thanks!

Issue #691, opened 2017-11-03
Various format issues with PPMDecoder


HeroicKatora:
The PPM Decoder does not validate the header of the read file. Consequently, it could wrongly try to read the currently supported other 5 formats (given by [magic constants][1] P1 through P5).

- [x] Whitespace fails to acknowledge some characters `A "white space" character is space, CR, LF, TAB, VT, or FF ` [as seen in the specification][2].

- [x] Comments do not end tokens and can appear in the middle of one. They are completely ignored and therefore can not delimit the header on their own [also in the specification][2] but can appear immediately after the last token of the header (before the whitespace that signals the end of the header).

- [x] ~~Characters are only ascii, decoding from utf-8 instead should therefore be considered an extension instead (referring to `String::from_utf8(bytes)` [in line 64][3]).~~ This is a non-issue as `parse` will do the appropriate checks. The error message could be clearer though as it currently overlaps with the completely equal error message in a previous line.

- [x] Not enough tests, neither positive nor negative ones.

I'll try to fix those and then provide a pull request (edit: it's here #692).

Some more work would be the following:
_The format specification allows more than one ppm image to be embedded within a ppm file._ (Not included in this pull request, will create a separate issue for this)
 

[1]: https://en.wikipedia.org/wiki/Netpbm_format
[2]: http://netpbm.sourceforge.net/doc/pgm.html
[3]: https://github.com/PistonDevelopers/image/blob/dbf2ea2579a3139ba76ec59ffa3528aeeac1a62e/src/ppm/decoder.rs#L64

HeroicKatora:
I just looked up the source code to netpbm and ImageMagick and noticed that their handling of whitespace and comments differs both from one another as well as from the specification. After some research to each of these topics, I'll gather some thoughts and results below:

This obviously raises the question which exact format should be supported.
_Comments_
- Issue: Comments appear to be according to the specification in netbpm but in ImageMagick, comments immediately after values are not treated as comments but rather the `#` is simply ignored. Also, comments can not be ended by `CR` but only by `LF`. Comments seem to end tokens.
- Thoughts & Fix: I'll implement this according to specification here as that would be a superset of the violation by ImageMagick and a lack of bug reports on ImageMagick seem to suggest that this format is seldomly relied upon.

_Whitespace_
[...]

PR #692, opened 2017-11-05
Improve support for ppm

HeroicKatora:
This fixes the following issues:
- Whitespace now includes all six character in the specification
- Comments are now handled correctly (fully ignored)
- Diagnostics when trying to read supported formats
  - Wrong magic value
  - `maxval` header parameter too large
  - Custom messages for several format issues  of the header
- Add a bunch of tests to PPMDecoder

Basically whatever is mentioned in #691


bvssvni:
Thanks!

Issue #693, opened 2017-11-10
Outdated examples in readme


leoschwarz:
Some of the examples in the README.md file are outdated, e.g. in section 6.1. `img.save` does only take one argument now and would not compile like that anymore.

Would it make more sense to move the examples into the examples directory, or should the ones in the README.md file just be updated? The advantage of the first course of action would be that bit rot could be prevented by running `cargo test`, however the later has the advantage of providing a quick reference which is probably the intended goal?

bvssvni:
Just update README.md. The API is less likely to change now, so bit rot is not a big problem.

PR #694, opened 2017-11-10
Update code samples.

leoschwarz:
I'm sorry I was a bit quick opening #693, the docs are in fact not as outdated as I made it sound in that issue, I just ended up combining the wrong code snippets without realizing there is a difference between `DynamicImage:save()` and `ImageBuffer::save()` but this should be clear when reading the docs more carefully.

So these are mostly cosmetic fixes, not particularily important.

### DIFF
--- a/examples/fractal.rs
+++ b/examples/fractal.rs
@@ -1,9 +1,8 @@
-//!An example of generating julia fractals.
+//! An example of generating julia fractals.
 extern crate num_complex;
 extern crate image;
 
 use std::fs::File;
-use std::path::Path;
 
 use num_complex::Complex;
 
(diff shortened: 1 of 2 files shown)


examples/fractal.rs
```diff
@@ -45,8 +44,8 @@ fn main() {
 
 
     // Save the image as “fractal.png”
-    let fout = &mut File::create(&Path::new("fractal.png")).unwrap();
```
bvssvni: This isn't really necessary, because of the `Into` trait.

leoschwarz: I'm not sure if I understand correctly, I deleted the `Path::new` since it is redundant. Are you saying you'd prefer to keep the `Path::new` there?

bvssvni: Sorry, I was mixing the deleted lines with added ones.

bvssvni:
Thanks! Merging.

PR #695, opened 2017-11-17
Add other pnm (pbm, pgm, pam) decoders

### LINKED ISSUE #664: PBM format
Add a new encoder/decoder for PBM format:

http://netpbm.sourceforge.net/doc/pbm.html

HeroicKatora:
Adds decoders and format recognition for the following formats from the netbpm package:
[Portable Bit Map][pbm], [Portable Gray Map][pgm] and [Portable Arbitrary Map][pam]

[pbm]: http://netpbm.sourceforge.net/doc/pbm.html
[pgm]: http://netpbm.sourceforge.net/doc/pgm.html
[pam]: http://netpbm.sourceforge.net/doc/pam.html

Details
---------
The pam format decoder recognizes the tuple types `BLACKANDWHITE`, `GRAYSCALE` and `RGB` and explicitely recognizes but rejects their `_ALPHA` variants (support for these might come in another patch).

Tests for the new formats are not yet included but will arrive in the next week, this should stay open until then.

This should deprecate the pam module, not yet part of this pull request. In case it is desired to explicitely test for the `ppm` format instead, the decoder class offers a method to query the specific image subtype after the header has been read and a wrapper for `ppm` specificially has been put into the place of the old decoder.

Related Issues
---------
This closes #664

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,13 +51,13 @@ num-complex = "0.1.32"
 glob = "0.2.10"
 
 [features]
-default = ["gif_codec", "jpeg", "ico", "png_codec", "ppm", "tga", "tiff", "webp", "bmp", "hdr"]
+default = ["gif_codec", "jpeg", "ico", "png_codec", "pnm", "tga", "tiff", "webp", "bmp", "hdr"]
 
 gif_codec = ["gif"]
 ico = ["bmp", "png_codec"]
 jpeg = ["jpeg-decoder"]
 png_codec = ["png"]
-ppm = []
+pnm = []
 tga = []
 tiff = []
 webp = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,13 +51,14 @@ num-complex = "0.1.32"
 glob = "0.2.10"
 
 [features]
-default = ["gif_codec", "jpeg", "ico", "png_codec", "pnm", "tga", "tiff", "webp", "bmp", "hdr"]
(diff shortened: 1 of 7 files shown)


bvssvni:
Tell me when you're ready.

HeroicKatora:
I'll add some more tests tomorrow and on the weekend, see if anything turns up in `pam` especially. Then this should be good to go probably on Saturday. One question, should `mod pnm` be deprecated right away or not? And if so, what's the preferred way?

bvssvni:
@nwin what do you think?

HeroicKatora:
Tests are currently done in `pnm` as far as `pam`, `pbm` and `pgm` are concerned and in `ppm` for `ppm`. Turned out I still had bugs before. All good on my end, ready to go.

nwin:
I think it would be best to deprecate `mod pnm` right away.

HeroicKatora:
This brings up something interesting, maybe one would want to assert the image format to be `ppm` specifically. There is currently no replacement for it when `ImageFormat::PPM` is removed. Maybe the `ImageFormat::PNM` could get the additional subtype parameter from the decoder struct so this can be queried/specified.

HeroicKatora:
Also, the encoder is still only written for `ppm`, so I've just deprecated the decoder for the moment.

bvssvni:
I think we can merge this now and make changes to `ImageFormat` later if needed.

HeroicKatora:
Possibly when the encoder has been transported to `pnm` or at `0.18`. Who knows, maybe both at once? But the former would be a good reason, as it would support specifying the intended output encoding in a uniform way with one implementation.

bvssvni:
Merging.

Issue #621, opened 2017-02-20
Support for DDS?


StelarCF:
Currently it's not supported and is a fairly common format for storing textures.

dobkeratops:
+1, hugely important for 3d

benoitryder:
I've implemented partial support of DDS, based on the DXT decoder implementation. (Basically, it parses the DDS header and creates a DXT decoder using parameters from the DDS header.) Therefore, only DDS files using DXT1, DXT3 or DXT5 are supported.
Should I open a PR with that, even if it is a limited support?

HeroicKatora:
Yes, there are a few other formats with only partial support and some success with incrementally improving them to full support.

Kanabenki:
This has been closed but the crate currently doesn't support most DDS texture formats, should this be reopened?

fintelia:
No, it would be better to create separate issues for specific formats / groups of formats. Though TBH the main bottleneck is someone coming along interested in implementing more formats so having the issues won't do a ton on their own

Issue #127, opened 2014-10-30
Encode/decode DXT compressed textures


bvssvni:
@dobkeratops Since you requested it the comment on Reddit http://www.reddit.com/r/rust/comments/2kp1di/presenting_rustimage_v01/clni7ci , would you care to elaborate on resources required to implement this?

sp-1234:
>Does the patent status prevent us simply writing an encoder/decode and distributing that on github?

there's no need to write decoder, it's done entirely by hardware.

CensoredUsername:
As the patents have expired DXT support has been merged.

PR #700, opened 2017-11-30
DXT encoding support

CensoredUsername:
This pull requests adds an Encoder and Decoder for the DXT1/3/5 lossy compression formats. It however does not implement any file format that uses DXT compression, like DDS at the moment. Due to this there is currently not a nice way of integrating testing of the decoding/encoding with the current testing infrastructure. I plan to add support for at least one container format to solve this, but since I'm not really familiar with the way the project works I've made the pull request now already to ask for advice on how to proceed with this.

Notes on the encoder/decoder
--------

The encoder is probably not the most performant possible implementation as it brute-forces the best fitting color table per block, but it does result in reasonably high-quality encoding. It's still reasonably fast, taking <0.3 sec to single-threaded encode a noisy 1024x1024 pixel RGBA image, which is the worst case for the algorithm. Images with large flats, small gradients, or limited colorspaces encode much faster.

Since the compression operates purely on a block basis, it would be trivial to speed it up using more threads.

Due to the way DXT blocks work the decoder's scanlines are actually 4 pixels tall.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ num-complex = "0.1.32"
 glob = "0.2.10"
 
 [features]
-default = ["gif_codec", "jpeg", "ico", "png_codec", "pnm", "ppm", "tga", "tiff", "webp", "bmp", "hdr"]
+default = ["gif_codec", "jpeg", "ico", "png_codec", "pnm", "ppm", "tga", "tiff", "webp", "bmp", "hdr", "dxt"]
 
 gif_codec = ["gif"]
 ico = ["bmp", "png_codec"]
@@ -64,5 +64,6 @@ tiff = []
 webp = []
 bmp = []
 hdr = ["scoped_threadpool"]
+dxt = []
 
 benchmarks = []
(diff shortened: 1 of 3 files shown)


CensoredUsername:
Previous discussion on DXT support can also be found in #127 and #621, although the patent discussion has become outdated since the patents in question expired last October.

bvssvni:
Can you rebase?

CensoredUsername:
It's rebased.

jeditobe:
Wikipedia says 
`Some (e.g. US 5956431 A) of the multiple USPTO patents on S3 Texture Compression expired on October 2, 2017.[6] At least one continuation patent, US6,775,417, however has a 165-day extension.`

CensoredUsername:
@jeditobe As far as I'm aware Mesa has already merged in [DXT support](https://www.phoronix.com/scan.php?page=news_item&px=S3TC-Lands-In-Mesa). Fedora is already shipping S3TC enabled code. There seems to be some confusion as [only a part of US6775417 is still protected](https://github.com/reactos/reactos/pull/6#issuecomment-335092453).

bvssvni:
It has passed 143 days since October 2, 2017, so in just 22 days we can ship it.

CensoredUsername:
Looks like the last patent has also passed now.

bvssvni:
OK, merging.

PR #701, opened 2017-12-08
Add a resize_to_fill method

### LINKED ISSUE #674: Add a "thumbnail" method
I would find it useful to have a method in `image` that takes an image and resizes to an exact width and height while preserving pixel aspect ratio by cropping any excess after resizing to achieve the desired dimensions.

shssoichiro:
Closes #674


bvssvni:
Thanks!

PR #707, opened 2017-12-23
Add `ExactSizeIterator` implementations to buffer structs.

samphippen:
`Pixels`, `PixelsMut`, `EnumeratePixels` and `EnumeratePixelsMut` are
all elligible to have `ExactsizeIterator` implementations added because
the underlying iterator (a `Chunks`) implements `ExactSizeIterator`.

I wasn't sure if I needed to do anything because of how many channels there are per pixel, but this is a first stab.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -119,6 +119,13 @@ impl<'a, P: Pixel + 'a> Iterator for Pixels<'a, P> where P::Subpixel: 'a {
     }
 }
 
+impl<'a, P: Pixel + 'a> ExactSizeIterator for Pixels<'a, P> where P::Subpixel: 'a {
+    fn len(&self) -> usize {
+        self.chunks.len()
+    }
+}
+
+
 impl<'a, P: Pixel + 'a> DoubleEndedIterator for Pixels<'a, P> where P::Subpixel: 'a {
 
     #[inline(always)]
@@ -142,6 +149,12 @@ impl<'a, P: Pixel + 'a> Iterator for PixelsMut<'a, P> where P::Subpixel: 'a {
         self.chunks.next().map(|v|
             <P as Pixel>::from_slice_mut(v)
         )
+    }
+}
+
+impl <'a, P: Pixel + 'a> ExactSizeIterator for PixelsMut<'a, P> where P::Subpixel: 'a {
+    fn len(&self) -> usize{
+        self.chunks.len()
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/buffer.rs
```diff
@@ -206,6 +225,12 @@ impl<'a, P: Pixel + 'a> Iterator for EnumeratePixelsMut<'a, P> where P::Subpixel
     }
 }
 
+impl <'a, P: Pixel + 'a> ExactSizeIterator for EnumeratePixelsMut<'a, P> where P::Subpixel: 'a {
```
sgrif: Can the entire where clause and all bounds on `P` be replaced with `where Self: Iterator`?

src/buffer.rs
```diff
@@ -180,6 +193,12 @@ impl<'a, P: Pixel + 'a> Iterator for EnumeratePixels<'a, P> where P::Subpixel: '
     }
 }
 
+impl <'a, P: Pixel + 'a> ExactSizeIterator for EnumeratePixels<'a, P> where P::Subpixel: 'a {
```
sgrif: Can the entire where clause and all bounds on `P` be replaced with `where Self: Iterator`?

src/buffer.rs
```diff
@@ -145,6 +152,12 @@ impl<'a, P: Pixel + 'a> Iterator for PixelsMut<'a, P> where P::Subpixel: 'a {
     }
 }
 
+impl <'a, P: Pixel + 'a> ExactSizeIterator for PixelsMut<'a, P> where P::Subpixel: 'a {
```
sgrif: Can the entire where clause and all bounds on `P` be replaced with `where Self: Iterator`?

src/buffer.rs
```diff
@@ -119,6 +119,13 @@ impl<'a, P: Pixel + 'a> Iterator for Pixels<'a, P> where P::Subpixel: 'a {
     }
 }
 
+impl<'a, P: Pixel + 'a> ExactSizeIterator for Pixels<'a, P> where P::Subpixel: 'a {
```
sgrif: Can the entire where clause and all bounds on `P` be replaced with `where Self: Iterator`?

src/buffer.rs
```diff
@@ -119,6 +119,13 @@ impl<'a, P: Pixel + 'a> Iterator for Pixels<'a, P> where P::Subpixel: 'a {
...
+    fn len(&self) -> usize {
+        self.chunks.len()
+    }
+}
+
+
```
sgrif: ✂️

src/buffer.rs
bvssvni: Looks like the right thing.

bvssvni:
Merging.

bvssvni:
Thanks!

Issue #706, opened 2017-12-23
Invalid code: expected 515 <= 511


fintelia:
I have been using image to load compressed TIFF images, but occasionally am running into the above error where decoding fails because of an invalid code. One such image is [here](https://edcintl.cr.usgs.gov/downloads/sciweb1/shared/gtc/downloads/WaterMask2010_UMD_individual/Hansen_GFC2013_datamask_40N_080W.tif.zip).  Other than some complaints about unrecognized GeoTIFF tags, the same image loads fine in GIMP.

Is the failure due to an unsupported feature, and if so, any pointers on how I could go about adding it?

nwin:
Sounds like an error in the lzw decoder.

It should be an easy fix by replacing the build-in decoder (utils.lzw), which is still used here, with the lzw crate.

fintelia:
I tracked down what is going on. The error was happening on the very last LZW code in one of the TIFF file strips. The code was intended to be 257, but the actual code written was bit short (100000001<sub>2</sub> instead of 0100000001<sub>2</sub>) because the decoder wrote out its final EndOfInformation code without checking whether it needed to increment the code size first.

The reason that LibTIFF (which is presumably what GIMP was using) doesn't run into this issue is because it doesn't bother waiting for EndOfInformation codes in the compressed stream and instead stops reading as soon as it has gotten enough bytes to fill its buffer. It probably makes sense to adopt the same behavior as LibTIFF, especially since the current version of the spec actually points out this exact edge case (_in italics!_) so it may be somewhat widespread...

nwin:
Thanks, but please do not attempt to fix it in this library, it won’t be merged. There are chances that this is already fixed in the lzw crate and if not, it should be fixed there.

nwin:
Beside that, such kind of things are really frustrating. TIFF is already off-by-one compared to gif… :(

PR #708, opened 2017-12-25
Switch to using lzw crate instead of using a built in version

fintelia:
This change also fixes issue #706.


nwin:
Thank you very much. Looks good to me, but can you please remove the utils mod altogether? It is not needed anymore.

nwin:
Do you, by any chance, have a smaller example image to hand, which exhibits the same issue as #706?

fintelia:
The utils module still contains `expand_packed` which is needed by buffer.rs. Although that function could probably be moved elsewhere if you prefer.

[test-images.zip](https://github.com/PistonDevelopers/image/files/1587775/test-images.zip) contains a couple smaller images that have the same issue as in #706. Inside the zip archive, good-lzw.tiff is the original (valid) TIFF file, while bad-lzw.tiff and bad-lzw2.tiff have been manually altered to have two different variants of the encoding bug.

fintelia:
Oops, that zip archive wasn't quite right. Use this one instead: [test-images-v2.zip](https://github.com/PistonDevelopers/image/files/1587784/test-images-v2.zip)

nwin:
Great thanks!

PR #711, opened 2017-12-29
bump byteorder requirement to 1.2.1

lovebug356:
Build failes because BigEndian::read_u16_into is used in pnm/decoder,
and is not available in byteorder 1.0.0.


nwin:
Thank you 🙂

PR #713, opened 2018-01-06
Allow using GenericImage trait object.

crlf0710:
This removes the `Sized` bound from `GenericImage`, and adds the nondefault `?Sized` bound to some structs so we can use `dyn GenericImage<Pixel=Rgba>`.


crlf0710:
Oops, sorry, will resubmit.

PR #718, opened 2018-01-31
A way to expose options when saving files

### LINKED ISSUE #581: Expose quality/compression controls when saving as JPG

kaj:
Provide a way to provide format-specific options (with default values) when saving an image.  Jpeg quality is used as an example.

This closes #581 in a different way than #584.


calum:
I prefer this approach. It keeps it generic. I think this should be merged,

bvssvni:
Merging.

PR #720, opened 2018-02-04
Improve downscale

kaj:
When scaling images down, the `support` pixels was counted before scaling, so only a small part of the pixels that should contribute actually did contribute, giving rough edges and moire aliasing problems.  [Here are some down-scaled sample images](https://rasmus.krats.se/tmp/imagescale/down.html) scaled before and after this change.

Also, the Gaussian method used a standard deviation of 1.0, where I think it should use a standard deviation of 0.5, giving a diameter of 1.0 (rather than a radius of 1.0) for the main contributing area. [Here are some up-scaled smaple images](https://rasmus.krats.se/tmp/imagescale/up.html) before and after, note the blurry images for gaussian scale before the change.

Also, I added an integer-only `thumbnail` method for fast down-scaling.

(Writing this message, I realize I should probably have made this three separate PRs.  Sorry.)


bvssvni:
Looks good to me.

bvssvni:
Merging.

bvssvni:
Thanks!

PR #721, opened 2018-02-05
Fix issue where resizing an image doesn't always give an exact width/height

### LINKED ISSUE #667: Resizing an image doesn't always give an exact width/height
I have been using `image` to resize images for the purpose of creating thumbnails. I noticed that if I resize an image with an automatic dimension (e.g. pass `std::u32::MAX` as height so the width is always exactly what I request), the dimension I do pass in is sometimes one lower than what I would expect.

For example, if I run this code on this [image](https://i.imgur.com/mSl1LZY.jpg):

```rust
fn resize_image(image: &DynamicImage) {
    let height = 880;
    let test_image = image.resize(::std::u32::MAX, height, FilterType::Lanczos3);
    println!("{}", test_image.height());
}
```

I would expect the resulting height of `test_image` to be 880, but instead it is 879.

shssoichiro:
Also refactors out common resizing logic from `resize` and `resize_to_fill`

It seems that this was due to insufficient precision in the `f32` type which was causing rounding errors in some cases. `f64` seems to have enough precision to handle this accurately, at least in the cases I tested.

Closes #667

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -725,6 +725,13 @@ pub fn guess_format(buffer: &[u8]) -> ImageResult<ImageFormat> {
     )
 }
 
+/// Calculates the width and height an image should be resized to.
+/// This preserves aspect ratio, and based on the `fill` parameter
+/// will either fill the dimensions to fit inside the smaller constraint
+/// (will overflow the specified bounds on one axis to preserve
+/// aspect ratio), or will shrink so that both dimensions are
+/// completely contained with in the given `width` and `height`,
+/// with empty space on one axis.
 fn resize_dimensions(width: u32, height: u32, nwidth: u32, nheight: u32, fill: bool) -> (u32, u32) {
     let ratio  = width as u64 * nheight as u64;
     let nratio = nwidth as u64 * height as u64;
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -260,19 +260,7 @@ impl DynamicImage {
                   nheight: u32,
                   filter: imageops::FilterType) -> DynamicImage {
 
-        let (width, height) = self.dimensions();
-
-        let ratio  = width as f32 / height as f32;
-        let nratio = nwidth as f32 / nheight as f32;
-
-        let scale = if nratio > ratio {
-            nheight as f32 / height as f32
-        } else {
-            nwidth as f32 / width as f32
-        };
-
-        let width2  = (width as f32 * scale) as u32;
-        let height2 = (height as f32 * scale) as u32;
+        let (width2, height2) = resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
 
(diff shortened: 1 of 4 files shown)


src/dynimage.rs
```diff
@@ -324,6 +302,25 @@ impl DynamicImage {
         }
     }
 
+    fn resize_dimensions(&self, nwidth: u32, nheight: u32, fill: bool) -> (u32, u32) {
```
calum: I think it would be beneficial to have a unit test for this function. However, I'm not sure where it should go. There aren't many tests written at the bottom of this file.

bvssvni: The bottom at the file is fine, I think. Could put it here: https://github.com/PistonDevelopers/image/blob/master/src/dynimage.rs#L764

shssoichiro:
Any update on this? I've added unit tests around the resizing logic.

bvssvni:
Looks good to me.

HeroicKatora:
Wouldn't it be better to use integer logic with `u64` instead? As in 
`iwidth / iheight < nwidht / nheight` <=> `iwhdth * nheight < nwidth*iheight`

`f64` with 53 bits of precision could be enough though. Compared to the old, the input arguments are surely represented without error and the division result is only be off by at most one `ulp`. Maybe this should be checked more thoroughly regardless.

shssoichiro:
I do like that better, if nothing else because integer math is generally faster. We still have to use float division to calculate the ratios, but can safely use u64's for calculating the new dimensions.

shssoichiro:
I also realized I could simplify two of the expressions e.g. `(width as u64 * nwidth as u64 / width as u64) as u32` to just `nwidth`. So I pushed another new commit to include that.

HeroicKatora:
Should we handle the case that calling with `fill = True` fails the cast to `u32`, i.e. the size got too big?

shssoichiro:
I agree. I've added tests around fill with and without overflow, and updated the code to handle the overflow case by bounding it to `u32::MAX`. I'm not sure that this code is the cleanest, but it does handle the edge cases.

shssoichiro:
I had to push a commit to fix a compile error if pnm feature is disabled; I was trying to pull this branch into a project where only png and jpeg are enabled, and encountered that. I realize it's not related to this issue, but I needed this branch to work.

src/dynimage.rs
```diff
@@ -747,6 +725,31 @@ pub fn guess_format(buffer: &[u8]) -> ImageResult<ImageFormat> {
     )
 }
 
+fn resize_dimensions(width: u32, height: u32, nwidth: u32, nheight: u32, fill: bool) -> (u32, u32) {
```
bvssvni: Can you add some comments? Explain e.g. `fill`.

bvssvni:
Merging.

PR #724, opened 2018-02-08
Make rayon an optional dependency

Diggsey:
This is a backwards compatible change, but may cause a drop in performance for anyone using `--no-default-features`. However, I do consider it to be an important change, because a dependency on rayon limits the targets on which `image` can be used.


bvssvni:
Thanks!

PR #727, opened 2018-02-21
Added into_frames implementation for gifs

### LINKED ISSUE #725: Zero frames in a gif
Hi there,

I'm struggling to get the frames from a gif image. I've tried a couple of different gifs and they all come out as just 1 frame. Am I doing something obviously wrong?

```rust
    let mut f = File::open(filename).expect("File not found");

    let mut decoder = gif::Decoder::new(f);

    let mut frames = decoder.into_frames().expect("error decoding gif");

    for (i, frame) in frames.enumerate() {
        let mut image = frame.into_buffer();

        // display the image:
        //render_image(image.convert(), width, height);

        println!("Frame: {}", i);
        thread::sleep(time::Duration::from_millis(100));
        println!("next frame...");
    }
```

Every time I run this, I only see
```
Frame: 0
next frame...
```

Thanks for any help!

Calum

calum:
This will resolve #725 by implementing the `into_frames` function for gifs

### DIFF
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -7,14 +7,20 @@
 //!
 
 extern crate gif;
+extern crate num_rational;
 
 use std::io::{Read, Write};
+use std::clone::Clone;
 
-pub use self::gif::Frame;
+pub use self::gif::{Frame, DisposalMethod};
 use self::gif::{SetParameter, ColorOutput};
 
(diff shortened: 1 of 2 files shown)


bvssvni:
Looks good.

bvssvni:
Merging.

PR #733, opened 2018-03-02
Fix multi image TIFF

aschampion:
TIFF files contain only a single header. In multi image TIFFs, the header should not be read for each image (because it's not there!), only the IFD.


nwin:
Thanks for contributing! :)

PR #734, opened 2018-03-05
Added jpeg_no_rayon feature

edwin0cheng:
It is an inverse approach of pr https://github.com/PistonDevelopers/image/pull/724 to allow no rayon dependency and without performance drop on default usage. 

The reason to this change is because rayon is not working on wasm target  right now, we should allow user to bypass it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
       - FEATURES=''
       - FEATURES='gif_codec'
       - FEATURES='jpeg'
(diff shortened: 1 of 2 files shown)


Diggsey:
Note that #724 does not cause a performance drop on default usage - it's only if people are using `--no-default-features`.

edwin0cheng:
Close this PR in favor to #724

PR #736, opened 2018-03-07
Fix off-by-one grayscale JPEG encoding

### LINKED ISSUE #735: Encoded jpeg has wrapping off-by-one in x
I've encountered a bug where when I encode a byte slice to JPEG, the first column is moved to the last column of the image. To demonstrate, here's the JPEG and then the same byte slice encoded as PNG (also through `image`) with the columns in the correct place. Look for the bright line on the right side of the JPEG, and the rest of the image being shifted left by one column.

JPEG:
dump
PNG:
dump

Here's a tiny rust program that produces these images and has the original byte slice:

https://github.com/aschampion/test_image_jpeg

If someone wants to fix this, feel free to use this image data in the regression test.

aschampion:
Will push a regression test for this tomorrow, since it looks like JPEG grayscale encoding may not be covered by tests.

Closes #735.

### DIFF
--- a/src/jpeg/encoder.rs
+++ b/src/jpeg/encoder.rs
@@ -708,4 +708,38 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn grayscale_roundtrip_sanity_check() {
+        // create a 2x2 8-bit image buffer containing a white diagonal
+        let img = [255u8, 0, 0, 255];
+
+        // encode it into a memory buffer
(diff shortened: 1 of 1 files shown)


aschampion:
It looks like there aren't reference image tests for encoding (only decoding), so I copied the existing JPEG encoding sanity unit test and adapted it for grayscale encoding. It fails on master and passes on this PR.

nwin:
Great, thanks!

PR #744, opened 2018-04-05
Avoid panic when blending two pixels with zero alpha.

lynaghk:
I ran into a situation where trying to `overlay` two rgba images led to a panic.
The root cause was a division by zero.
Here's a minimal reproducible example program run against image = "0.18.0":

```rust
extern crate image;
use image::{ImageBuffer, Rgba, RgbaImage};
use image::imageops::overlay;
fn main() {
    let mut a: RgbaImage = ImageBuffer::new(1, 1);
    a.put_pixel(0, 0, Rgba([0, 0, 0, 0]));

    let mut b: RgbaImage = ImageBuffer::new(1, 1);
    b.put_pixel(0, 0, Rgba([0, 0, 0, 0]));

    overlay(&mut a, &b, 0, 0);
}
```

I don't have a ton of graphics experience, so I'm not sure if there's a "correct" way to blend two completely transparent pixels in terms of retaining the color components.
This patch just uses the blended colors and leaves alpha at zero.

This patch is the simplest thing I could think of to fix, but I'd be happy to change to something more idiomatic (`num::CheckedDiv` trait and explicit handling of the None condition?)

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -377,7 +377,11 @@ impl<T: Primitive> Blend for LumaA<T> {
         let (bg_luma, bg_a) = (bg_luma.to_f32().unwrap() / max_t, bg_a.to_f32().unwrap() / max_t);
         let (fg_luma, fg_a) = (fg_luma.to_f32().unwrap() / max_t, fg_a.to_f32().unwrap() / max_t);
 
-        let alpha_final = bg_a + fg_a - bg_a * fg_a;
+        let mut alpha_final = bg_a + fg_a - bg_a * fg_a;
+        if 0.0 == alpha_final {
+            alpha_final = 1.0;
+        }
+
         let bg_luma_a = bg_luma * bg_a;
         let fg_luma_a = fg_luma * fg_a;
 
@@ -410,7 +414,10 @@ impl<T: Primitive> Blend for Rgba<T> {
         let (fg_r, fg_g, fg_b, fg_a) = (fg_r.to_f32().unwrap() / max_t, fg_g.to_f32().unwrap() / max_t, fg_b.to_f32().unwrap() / max_t, fg_a.to_f32().unwrap() / max_t);
 
         // Work out what the final alpha level will be
-        let alpha_final = bg_a + fg_a - bg_a * fg_a;
+        let mut alpha_final = bg_a + fg_a - bg_a * fg_a;
+        if 0.0 == alpha_final {
+            alpha_final = 1.0;
+        }
 
         // We premultiply our channels bu their alpha, as this makes it easier to calculate
         let (bg_r_a, bg_g_a, bg_b_a) = (bg_r * bg_a, bg_g * bg_a, bg_b * bg_a);


bvssvni:
It seems that the `final_alpha` is set to 1.0? Should this not be zero?

lynaghk:
Doh, I wasn't thinking and was only testing against images with completely opaque pixels. I'll amend the commit.

That said --- is there any reason to do anything with the color values at all when blending two transparent pixels?
If not, then presumably it'd be more efficient to return a transparent white/black pixel rather than doing calculations on the color components.

bvssvni:
I think blending is good, because it preserves the information in the image to neighboring pixels. For example, if you are super-sampling textures, white/black pixels can cause trouble.

lynaghk:
Thanks for the quick reply!

I did some reading, and I don't think this kind of compositing is well defined when both pixels have zero alpha.

The referenced [stack overflow answer](https://stackoverflow.com/questions/7438263/alpha-compositing-algorithm-blend-modes#answer-11163848) and [SVG spec](https://dev.w3.org/SVG/modules/compositing/master/#comp-op-src-over), defines compositing in terms of "premultiplied" color components (as defined in the [original Porter and Duff paper](https://keithp.com/~keithp/porterduff/p253-porter.pdf)).

But since alpha is in the domain [0, 1],

```rust
let alpha_final = bg_a + fg_a - bg_a * fg_a;
```

must always be greater than the alpha of either pixel.
Which means that it can only be zero when both pixel's alpha are zero.

But in that case, the "premultiplication" step

```rust
let (bg_r_a, bg_g_a, bg_b_a) = (bg_r * bg_a, bg_g * bg_a, bg_b * bg_a);
let (fg_r_a, fg_g_a, fg_b_a) = (fg_r * fg_a, fg_g * fg_a, fg_b * fg_a);
```

will set all of the color components to zero.

Or am I missing something in my logic there?

bvssvni:
Premultiplied color components do not conserve the color information in pixels when making them transparent. This is a different way of encoding images, which is not used here.

lynaghk:
The code I quoted wasn't hypothetical, it's from the RGBA blend implementation and has the comment "We premultiply our channels bu their alpha, as this makes it easier to calculate".

Or are you suggesting that we switch to a different blending algorithm in the case when both source alphas are zero?

bvssvni:
@lynaghk I messed this up, sorry.

bvssvni:
I was not aware of premultiplying colors being used here. Not sure if this is the right algorithm.

lynaghk:
@bvssvni No worries. I'm not too familiar w/ the tradeoffs for color blending and graphics stuff like this, so I'm not sure what the appropriate thing to do would be --- all I wanted from my changes was to avoid a panic.

If you do find another way through and want a hand, let me know.
But no worries if it'd be easier to just do the changes yourself too.

bvssvni:
Another issue that there is no conversion to sRGB before blending. This algorithm is very opinionated. However, that is a design decision that could be dealt with in a later version.

bvssvni:
Can you add a unit test that blends two transparent pixels and see what happens to the alpha channel? At least the resulting alpha channel should be zero.

bvssvni:
I think it should simply return instead of setting the alpha to `1.0`.

Closing in favor of https://github.com/PistonDevelopers/image/pull/759

Thanks for pointing this out!

lynaghk:
Sorry for the delay on unit tests, I was away from the Internet.

Returning makes sense, given that the blend operation mutates. Thanks for the fix and your work on this library!

PR #747, opened 2018-04-13
fix apply_with_alpha off by 1 error

d-dorazio:
I don't think `apply_with_alpha` is working correctly because it is not applying `g` to the alpha subpixel. The problem is that if `$alphas != 0` then it's `1`(as of now) and therefore the alpha index would be `$channels - 1` which is the correct index while now the alpha index would be `$channels - 1 - 1` which is `B` for `RGB` for example.


bvssvni:
Thanks!

PR #752, opened 2018-04-18
Make save() function consistant 

### LINKED ISSUE #742: save() function inconsistency
If you create a new image, you can save it using img.save(filename); but if you open an existing image, doing img.save(filename) doesn't work.
See https://users.rust-lang.org/t/can-create-save-an-image-but-cant-load-modify-save-one/16574

calum:
This PR makes the two `save` functions for `DynamicImage` and `ImageBuffer` the same and renames the old `DynamicImage` `save` to be ~`write`~ `write_to` instead.

This fixes #742.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -415,7 +415,7 @@ impl DynamicImage {
     }
 
     /// Encode this image and write it to ```w```
-    pub fn write<W: Write, F: Into<ImageOutputFormat>>(&self, w: &mut W, format: F) -> ImageResult<()> {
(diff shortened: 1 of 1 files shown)


HeroicKatora:
I am reluctant to choose the name `write` however consolidating on `save` for file operations is good for consistency and ergonomics. But in my head, `write` is too closely linked to `std::io::Write`, maybe `write_to` instead?

calum:
Thank you for the feedback. 

I see what you mean. I prefer `write_to`.

bvssvni:
Merging.

bvssvni:
Thanks!

Issue #753, opened 2018-04-18
Builds are failing silently due to `KTHS`


calum:
Builds seem to be failing with the same error:
```
error[E0433]: failed to resolve. Use of undeclared type or module `KTHS`
   --> src/imageops/sample.rs:140:34
    |
140 |     let max: f32 = NumCast::from(KTHS::max_value()).unwrap();
    |                                  ^^^^ Use of undeclared type or module `KTHS`
warning: unnecessary parentheses around function argument
   --> src/buffer.rs:493:23
    |
493 | /                       (width as usize
494 | |                       * height as usize
495 | |                       * (<P as Pixel>::channel_count() as usize)
496 | |                       )],
    | |_______________________^ help: remove these parentheses
    |
    = note: #[warn(unused_parens)] on by default
error: aborting due to previous error
error: Could not compile `image`.
Caused by:
[...]

PR #755, opened 2018-04-18
Let broken builds be marked as failures

kaj:
The problem is discovered and discussed in #753.

The way each build was make in three steps hid failures in the two earlier steps behind success in the last step.  Use shell `&&` rather than `;` in `.travis.yml` to break the build directly if one step fails.

I believe this is working properly now.

- [x] see a green build after changing `.travis.yml`.
- [x] introduce an error and see that properly failing a build
- [x] fix the error and see the build green again

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,11 +24,11 @@ env:
       - FEATURES='hdr'
 script:
     - if [ -z "$FEATURES" ]; then
-        cargo build -v;
-        if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then cargo test -v; fi;
+        cargo build -v &&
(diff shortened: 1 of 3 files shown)


kaj:
Seems I went directly to the second point; when checking properly, the build already fails due to two format-specific tests not being marked with their respective feature.

Fixing.

HeroicKatora:
`cargo doc` is run without the correct feature flags. This is also the cause for the feature related failure, since that would run & succeed with default features always.

calum:
This definitely needs to be merged 👍

HeroicKatora:
Merging!

Thank you for finding the problem and figuring out the broken tests.

PR #756, opened 2018-04-20
Fixed examples and tests

bvssvni:

PR #757, opened 2018-04-20
change out of date examples and tests to use write_to instead of save

calum:
I forgot to change the examples and tests in my last PR and now the builds are failing... 😳

I decided the quickest fix was to switch the `save`s to `write_to`s.

Sorry for the failed builds!


calum:
Closed in favour of #756

Issue #762, opened 2018-04-26
Feature request: SubImage::copy()


mark-summerfield:
The SubImage::new() function provides a view and so must take a mutable reference to the image.
However, I think it would be useful (certainly for my use cases) to be able to *copy* a sub image. So I'd like a function called, say SubImage::copy() which has the same API as SubImage::new() except that it takes a non-mutable image and returns a fresh image.

HeroicKatora:
It would also be possible to have a `decouple` on SubImage, producing an image with independent lifetime.

HeroicKatora:
As I just noted, `to_owned` already exists, as `to_image`. With the restrictive requirement of `SubImage::new` requiring `&mut I`, it might be best to have `copy` (or similar) on `GenericImage` instead of on `SubImage`. Otherwise, we would confuse the functionality of providing a view into an arbitrary, generic buffer with the functionality of cloning an image into a specific kind of predetermined buffer, `ImageBuffer<I::Pixel, Vec<<I::Pixel as Pixel>::Subpixel>>` as it exists in `SubImage::to_image`.

mark-summerfield:
I think I made a mistake in my original feature request. I don't need to copy the image data, I just need a read-only view on the data and to have the image as non-mut so that I can use a rayon .par_iter() to loop over a bunch of rectangles and examine each one inside the loop using, say, SubImage::read_only(&img, &rect).

HeroicKatora:
In the pull request, `SubImage` has been changed to accept generic reference types, any type implementing `Deref`to be exact, to work around the requirement for `&mut _`. The implementation of the `GenericImage` trait is only required when needing `SubImage` to be a `GenericImage`.

While this opens up many new possibilities, the current state of the rework might be a breaking change of the interface because I'm unsure if all previously possible instantiations still compile.

HeroicKatora:
The interface should be reverse compatible now, `SubImage::new(&mut image)` should now offer the exact same methods and trait implementations as before. Can I request a review of #764 @bvssvni ?

HeroicKatora:
This should be fixed as of #764 .

```
let image = …;
let subimage = image.view(x, y, w, h).to_image();
```

PR #764, opened 2018-04-30
Allow generic reference type for subimage

HeroicKatora:
As a proposed solution to #762 , this allows any reference type as the underlying storage to `SubImage`. This makes it possible to use the non-mutable `&ImageBuffer`. Only mutable reference types implement the `GenericImage` trait.


src/image.rs
```diff
@@ -495,33 +496,40 @@ pub trait GenericImage: Sized {
...
-        <Self::Pixel as Pixel>::Subpixel: 'static,
-        Self::Pixel: 'static,
-    {
+    fn sub_image(&mut self, x: u32, y: u32, width: u32, height: u32)
+    -> SubImage<&mut Self>
+    where 
+		Self: 'static, 
+		<Self::Pixel as Pixel>::Subpixel: 'static,
+    	Self::Pixel: 'static 
+	{
```
bvssvni: Can you move this `{` to the beginning of the line?

src/image.rs
```diff
@@ -495,33 +496,40 @@ pub trait GenericImage: Sized {
...
+        SubImage::new(self, x, y, width, height)
+    }
+
+    /// Returns an subimage that is an immutable view into this image.
+    fn view(&self, x: u32, y: u32, width: u32, height: u32)
+    -> SubImage<&Self>
```
bvssvni: This could be put after `)` on previous line.

src/image.rs
```diff
@@ -495,33 +496,40 @@ pub trait GenericImage: Sized {
...
+    	Self::Pixel: 'static 
+	{
         SubImage::new(self, x, y, width, height)
     }
 }
 
 /// A View into another image
-pub struct SubImage<'a, I: 'a> {
-    image: &'a mut I,
+pub struct SubImage <I> {
```
bvssvni: `SubImage<I>`

src/image.rs
```diff
@@ -495,33 +496,40 @@ pub trait GenericImage: Sized {
...
+    image:   I,
```
bvssvni: `image: I,`

src/image.rs
HeroicKatora: I just copied that style from `copy_from` directly above. I also think it is slightly more readable due to the visible space between the type bounds and the function implementation.

d-dorazio:
I like this implementation a lot, but I think we could do better here. The problem I see is that I'm basically unable to call any non mutating `GenericImage` functions on `SubImage` because we can only provide an implementation for `&mut I` since `GenericImage` also requires `get_pixel_mut` and `put_pixel`. This is a bit inconvenient imho because we'd be forcing a call to `to_image` to copy the whole view just to call some non mutating methods.

My proposal would be to create a new `GenericImageView` trait that contains only the non mutating methods and then `GenericImage` would also require `GenericImageView`. This way we could implement `GenericImageView` safely for `SubImage<&Self>` and in addition we could still implement `GenericImage` for `SubImage<&mut Self>`. I think that this shouldn't break any code, but I'm not totally sure.

We could also implement this in another PR if you think it's a bit out of scope in this PR since it requires changing `GenericImage`.

@HeroicKatora @bvssvni what do you think?

bvssvni:
@d-dorazio Sounds like a good idea. If this is a weakness in the current API then a breaking change is reasonable.

We won't publish yet, so this could be implemented in another PR.

bvssvni:
Merging.

bvssvni:
Thanks!

PR #766, opened 2018-05-03
Improve subimages

d-dorazio:
This PR mainly involves 3 changes:

1. remove the `'static` requirement for `Self` in `sub_image` and `view`. This allows to call `sub_image` and `view` not only on an image but also on `SubImage`s;
2. add a new associated type `InnerImage` to `GenericImage` which allows to specify the return type of `sub_image` and `view`. This allows to avoid some layers of indirections in the case of `SubImages` because we can reference the original image directly. Previously nested calls to `sub_image` would result in a type like `SubImage<SubImage<SubImage<&mut I>>>` while with this PR the type would always be `SubImage<&mut I>`. On top of this,  this also makes it easy to work with `SubImage`s recursively.
3. move non mutating methods from `GenericImage` to `GenericImageView`. This allows multiple views into the same image to coexist. It should also allow to use `rayon` to process each of the views concurrently.

Unfortunately 2 and 3 can break existing code because 2 added new methods to the traits and 3 moved the `Pixel` associated type into `GenericImageView`.

However, I think this PR improves the usability of `SubImage`s imho. Take a look at the following example:

https://github.com/d-dorazio/mattors/blob/8725a4bfb769484d142d6526ddd823d77d86c1bc/src/main.rs#L515-L551

The code makes a recursive use of `SubImage`s which doesn't compile without this changeset.

I also plan to make another change in `GenericImageView` that is change `get_pixel` to return a ref to the Pixel and not a copy, but I'll do it in another PR relatively soon.

### DIFF
--- a/examples/opening.rs
+++ b/examples/opening.rs
@@ -5,7 +5,7 @@ use std::env;
 use std::fs::File;
 use std::path::Path;
 
-use image::GenericImage;
+use image::GenericImageView;
 
 fn main() {
     let file = if env::args().count() == 2 {
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -512,6 +512,7 @@ where
     P::Subpixel: 'static,
 {
     type Pixel = P;
+    type InnerImage = Self;
 
     fn dimensions(&self) -> (u32, u32) {
         self.dimensions()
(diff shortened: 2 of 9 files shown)


bvssvni:
Looks good to me.

bvssvni:
@nwin Do you want to take a look?

bvssvni:
Merging.

bvssvni:
Thanks!

d-dorazio:
Thank you!

PR #776, opened 2018-05-24
Thumbnail

### LINKED ISSUE #767: Out of range panic with imageops::thumbnail
Repro: https://github.com/francesca64/thumbnail-repro

HeroicKatora:
Rewrite thumbnail to handle upscaling, fixes #767 

Upscaling images results in cases where there are pixels in the output to which no integer valued indices in the original image correspond. The current approach works around this by always always consuming at least one pixel but this result in a panic and would yield inaccurate result anyways.

The new algorithm instead special cases for these cases by instead weighting the two surrounding pixels and never forcing the window to shift.

Open questions are: 
 - What is the performance of the new algorithm?
 - Should the old behaviour be fully preserved for cases where we downscale in both directions, i.e. where the old window determination would not panic in the first place? If the performance has gotten worse, a single branch at the start of the method instead of branches within the loop might yield better results. The newer algorithm has different cache behaviour and uses more floating point arithmetic.

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -359,8 +359,10 @@ where
                     }
                 }
 
-                let n = <S::Larger as NumCast>::from((right - left) * (top - bottom)).unwrap();
-                let round = <S::Larger as NumCast>::from(n / NumCast::from(2).unwrap()).unwrap();
+                let n = <S::Larger as NumCast>::from(
+                    (right - left) * (top - bottom)).unwrap();
+                let round = <S::Larger as NumCast>::from(
+                    n / NumCast::from(2).unwrap()).unwrap();
                 (
                     S::clamp_from((sum.0 + round)/n),
                     S::clamp_from((sum.1 + round)/n),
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Travis is sometimes failing due to network errors. That is rather annoying

src/imageops/sample.rs
```diff
@@ -293,57 +293,184 @@ where
...
             );
 
-            let mut sum = (
-                S::Larger::zero(),
-                S::Larger::zero(),
-                S::Larger::zero(),
-                S::Larger::zero(),
-            );
-            for y in bottom..top {
+            let avg = if bottom != top && left != right {
```
d-dorazio: what do you think of using a `match` expression here to avoid repeating the clauses? I also think that we could separate the arms of the branch into separate functions to improve readability.

something like:

```rust
match (bottom != top, left != right) {
  (true, true) => get_avg_divergent(...),
  (true, false) => get_avg_divergent_hor(...),
  (false, true) => get_avg_divergent_ver(...),
  (false, false) => get_avg_convergent(...),
}
```

these are my 2 cents, take them for what they're worth :)

In general, this looks pretty good to me!

HeroicKatora: I chose not to do it in different method so that `sample_val` and `add_pixel` stay local to this method instead of being leaked into the outer scope. And then resorted to if-else instead of match because of a level of indentation and the fact that I find matching on bools much harder to read than conditionals.

But if those were factored out as separate methods then yeah, the matching would be cleaner.

HeroicKatora: Alright, I'll refactor the methods into a simple struct holding the sumed values, then we can have the best of both worlds.

HeroicKatora:
Okay, I'll provide documentation and maybe some benchmarking then merge this.

bvssvni:
Thanks!

PR #777, opened 2018-05-31
Fix ordering of interpolated parameters in TIFF decode error string

locriani:
Currently, the ordering of the parameters in the error string mean you get a result like this:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnsupportedError("[8] with RGBPalette bits per sample is unsupported")', libcore/result.rs:945:5
```

When really, it should look like this:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnsupportedError("RGBPalette with [8] bits per sample is unsupported")', libcore/result.rs:945:5
```

(Technically, the 'fixed' error message is still wrong, as that was actually a CIELab image with [8, 8, 8] encoding. However, since that's not supported yet, getting the error response to be less confusing is a step forward.)


bvssvni:
Thanks!

PR #782, opened 2018-06-17
Closes #781

### LINKED ISSUE #781: Readme example 6.1 behind new API
Using image v0.19.0 and wanted to build example 6.1. cargo build gives the error below

```
error[E0061]: this function takes 1 parameter but 2 parameters were supplied
  --> src/main.rs:21:9
   |
21 |     img.save(fout, image::PNG).unwrap();
   |         ^^^^ expected 1 parameter

```

According to [current src of save function](https://github.com/PistonDevelopers/image/blob/master/src/dynimage.rs#L455-L466), the format is inferred from the name of the file

petr-tik:
Remove the fs::File import and fout instance.

image.save() now handles the saving and only needs 1 argument

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,8 +183,6 @@ The image format is determined from the path's file extension.
 ```rust
 extern crate image;
 
-use std::fs::File;
-
 use image::GenericImage;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


bvssvni:
Thanks!

PR #790, opened 2018-07-04
Fix move of overlapping buffers in bmp/decoder

### LINKED ISSUE #789: Errors in reencoded images using this crate.
First: Thanks for this awesome crate. It's been really usefull for me and I appreciate all the work.

Today I tried a tool I wrote using this crate on my Elite Dangerous screenshot folder.
The tool basicly just converts images to pngs and saves some metadata to a json file.

I noticed some weired errors in the reencoded pngs when the original image was a "High Resolution Screenshot" (10240 x 5760).

Note: I scaled down the following images for the sake of this issue not taking years to load on slow connections, you can find links to the original images at the end of this issue.

Original:
original
Reencoded:
reencoded-skaled-down

I can reproduce this behavior in image v0.19.0 with what is basiclly example 6.1 from the readme:
```rust
extern crate image;

fn main() {
  let image = image::open("./original.bmp").unwrap();
  image.save("./reencoded.png").unwrap();
}
```

I also tried to encode this image in other formats using `DyanmicImage.write_to()` but got the same result.

If I can provide any more details that could help debug this, tell me.

[Original Image](http://cdn.kanaye.net/image-sources/original.bmp) [Reencoded Image](http://cdn.kanaye.net/image-sources/reencoded.png)

philipc:
Fixes #789 

I didn't add a test because that requires a test file size of at least 134MB.

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -167,21 +167,14 @@ fn extend_buffer(buffer: &mut Vec<u8>, full_size: usize, blank: bool) -> &mut [u
     let extend = full_size - buffer.len();
 
     buffer.extend(repeat(0xFF).take(extend));
-
-    let buffer_len = buffer.len();
+    assert_eq!(buffer.len(), full_size);
 
     let ret = if extend >= old_size {
         // If the full buffer length is more or equal to twice the initial one, we can simply
         // copy the data in the lower part of the buffer to the end of it and input from there.
-        {
-            let (lower, upper) = buffer.split_at_mut(old_size);
-            let upper_len = upper.len();
-            let lower_len = lower.len();
-
-            upper[upper_len - lower_len..].copy_from_slice(lower);
-        }
-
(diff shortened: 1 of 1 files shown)


philipc:
Also, by inspection, it looks like there's another bug with top-down RLE data. I think that `extend_buffer` shouldn't be called [here](https://github.com/PistonDevelopers/image/blob/d000240fc263fde294236ace79e1873eaa7a20ac/src/bmp/decoder.rs#L1057) for top-down RLE data, similar to what is done in [`with_rows`](https://github.com/PistonDevelopers/image/blob/d000240fc263fde294236ace79e1873eaa7a20ac/src/bmp/decoder.rs#L228-L254). @oyvindln what do you think?

bvssvni:
Is this ready for merging? Perhaps we could fix the bug in another PR.

philipc:
I looked into the RLE some more, and I think it's fine because RLE can't be top-down. So this PR is okay to merge if it has been reviewed.

PR #796, opened 2018-07-30
pnm/decoder: P4 decoding must return bits, not bytes

### LINKED ISSUE #794: Opening a PBM file returns wrong ColorType, thus trying to .save() panics
Something seems to be wrong inside `PNMDecoder` (using v0.19.0 of the crate). Trying to open a PBM file, and saving it back afterwards to a new PBM file (or a PNG file, did not try other formats though) panics:

```rust
extern crate image;

use image::GenericImage;

fn main() {
    let img = image::open("test.pbm").unwrap();
    println!("{:?}", img.color());
    println!("{:?}", img.raw_pixels().len());
    println!("{:?}", img.dimensions());
    img.save("out.pbm").unwrap();
}
```

The issue seems to be that the `ColorType` is recognized as a `Gray(8)` here, instead of `Gray(1)` which would be expected from a quick glance at [src/pnm/decoder.rs](https://github.com/PistonDevelopers/image/blob/master/src/pnm/decoder.rs).

Here is the small 10px x 10px test file used (attached as a PNG, because PBM is not allowed here, should be easy enough to convert back to PBM): test

philipc:
Fixes #794


philipc:
I've added some tests.

HeroicKatora:
BWBit from Portable Arbitrary Map also gets returned as `Gray(1)`. But there is the complication that its values are encoded as individual bytes nevertheless, i.e. we need to compress them into the expected format for DynamicImage where the bits are expanded again.

HeroicKatora:
Since fixing the `plain pbm` parsing and correct decoding of the values are related, should I rebase the fix for the decoder onto this pull request or the other way around?

philipc:
I haven't tested `P1`, but yes it looks wrong, and that plus `BWBit` are why I thought there were more colortype errors. My aim in this PR was to only fix `P4`. If you are doing work on the pnm decoder, then I prefer to merge this PR now and fix the rest in followups.

HeroicKatora:
Thanks for your help in improving this library!

The follow-up should be opened once #799 has been fixed.

PR #797, opened 2018-07-30
Implement Read + Seek for DXTDecoder

fintelia:
Still needs some tests/code polish, but shows the general idea.

### DIFF
--- a/src/dxt.rs
+++ b/src/dxt.rs
@@ -7,7 +7,8 @@
 //!
 //!  Note: this module only implements bare DXT encoding/decoding, it does not parse formats that can contain DXT files like .dds
 
-use std::io::{Read, Write};
+use std::io::{self, Read, Seek, SeekFrom, Write};
+use std::mem;
 
 use color::ColorType;
 use image::{DecodingResult, ImageDecoder, ImageError, ImageResult};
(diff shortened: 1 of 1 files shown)


src/dxt.rs
```diff
@@ -134,6 +142,82 @@ impl<R: Read> ImageDecoder for DXTDecoder<R> {
...
+    }
+}
+
+impl<R: Read + Seek> Seek for DXTDecoder<R> {
+    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
+        let row_len = self.row_len().unwrap();
+        let offset = match pos {
+            SeekFrom::Start(offset) => offset,
+            SeekFrom::Current(offset) => {
+                (((self.row - 1) as usize * row_len as usize + self.scanline_offset) as i64
```
fintelia: Watch out for underflow

bvssvni:
Does this need polish before being merged?

fintelia:
It could probably use a bit more polish before merging. Though more broadly it is worth thinking about whether we actually want to implement these traits on decoders at all.

Personally, I'd rather see image move to something like [this](https://github.com/PistonDevelopers/image/issues/798#issuecomment-414333504) where all decoder types could implement a common interface, without users having to worry about `unimplemented!()` methods or other edge cases (on some decoders `read_image`, `read_scanline` and `load_rect` can corrupt each other's state). I'd be willing to push ahead on implementation for that RFC, but I'd want some confirmation that it was actually a direction that the crate wanted to go first...

bvssvni:
@fintelia Sounds good! I'd like to see an RFC for that and hear people's responses.

Issue #799, opened 2018-07-31
Pbm decoder improperly relies on whitespace separator


HeroicKatora:
[`pbm` ascii] is different to the other ascii encoded pnm formats in that whitespace within the pixel raster is to be discarded instead of being used as a sample value separator. The current decoder works the same as for all other formats, thus trying to parse the amalgamated value as `u32`.

This can be reproduced by converting an image to *.pbm with gimp or NetPbm, here is an example image: 
![43282174-b76d2dea-9115-11e8-8639-5c84c2d831c5](https://user-images.githubusercontent.com/5550310/43487803-37c6ea18-9507-11e8-9652-a1647ff2b10e.png). Note that `convert` from ImageMagick will add whitespace but of course you can simply remove it manually.

[`pbm` ascii]: http://netpbm.sourceforge.net/doc/pbm.html#plainpbm

HeroicKatora:
Note: I will take the opportunity to improve performance of ascii reading. Currently, the read allocates a string for each read sample (`String::from_utf8`) instead of simply reinterpreting the buffer (`str::from_utf8`) which also had been validated as being purely ascii anyway. And I believe I have identified a potential infinite loop (with `Iterator::skip_while`) when the first byte of an expected sample is missing, such as when the file ends prematurely.

HeroicKatora:
Fixed in #800

PR #800, opened 2018-08-01
Fix and improve pbm decoding

HeroicKatora:
Addresses the issues found in #799 

* Io errors can no longer lead to infinite loops but will instead break early
* `pbm` decoding now works properly without separators, which are optional in contrast to the formats

### DIFF
--- a/src/pnm/decoder.rs
+++ b/src/pnm/decoder.rs
@@ -1,5 +1,5 @@
 use std::io::{self, BufRead, BufReader, Read};
-use std::str::FromStr;
+use std::str::{self, FromStr};
 use std::fmt::Display;
 
 use super::{ArbitraryHeader, ArbitraryTuplType, BitmapHeader, GraymapHeader, PixmapHeader};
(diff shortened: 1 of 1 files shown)


src/pnm/decoder.rs
```diff
@@ -622,22 +666,21 @@ impl Sample for BWBit {
...
-                "Sample value outside of bounds".to_string(),
-            )),
-        }
+    fn from_ascii(
+        _reader: &mut Read,
+        _width: u32,
+        _height: u32,
+        _samples: u32,
+    ) -> ImageResult<Vec<Self::T>> {
+        panic!("BW bits from anymaps are never encoded as ascii")
```
philipc: Should this be `unreachable!`?

HeroicKatora: Yes, for consistency.

HeroicKatora:
Since `String::from_utf8(Vec<u8>)` does not necessarily to an extra allocation as I had feared, the main benefit from `str::from_utf8(&[u8])` is being able to maybe in the future keep a decoding buffer between calls to avoid some allocations. This leaves us currently with one allocation for every call to `read_separated_asciii` and additional string allocations within the header reading in `HeaderReader::read_next_string`, which could theoretically use `BufRead` and return a `Cow<str>`. 

But for now we shouldn't overcomplicated the decoder until this becomes a performance concern, thus I'll only make the marginal change to `str::from_utf8` as a preparation.

PR #804, opened 2018-08-18
optimize image resizing by caching calls to filter.kernel

lukpank:
For resizing a (3008, 2000) JPEG photo to (320, 212) using Lanchos3 the resizing time drops
from arround 1.064s to arround 0.133s (8 times faster) on my machine.

The comments are moved together with the code, probably should be somehow modified.

The above results are with stable Rust compiler 1.28.0; on the other hand with 1.30.0-nightly compiler the optimized version takes arround 0.694s which seems to be caused by channels4 method being considerably slower by calling memcpy in copy_from_slice (from using valgrind).

I use the following program to do testing and timing. It takes input image as its first argument and saves it to tmp.jpg

```
extern crate image;

use image::GenericImageView;
use std::env;
use std::time::SystemTime;

fn main() {
    let args: Vec<_> = env::args().take(2).collect();

    let now = SystemTime::now();
    let img = image::open(&args[1]).expect("image open");
    if let Ok(elapsed) = now.elapsed() {
        println!(
            "loading:  {}.{:03} {:?}",
            elapsed.as_secs(),
            elapsed.subsec_millis(),
            img.dimensions()
        );
    }

    let now = SystemTime::now();
    let mini = img.resize(320, 212, image::imageops::Lanczos3);
    if let Ok(elapsed) = now.elapsed() {
        println!(
            "resizing: {}.{:03}",
            elapsed.as_secs(),
            elapsed.subsec_millis()
        );
    }

    let now = SystemTime::now();
    mini.save("tmp.jpg").expect("image save");
    if let Ok(elapsed) = now.elapsed() {
        println!(
            "saving:   {}.{:03} {:?}",
            elapsed.as_secs(),
            elapsed.subsec_millis(),
            mini.dimensions()
        );
    }
}
```

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -135,47 +135,51 @@ where
 {
     let (width, height) = image.dimensions();
     let mut out = ImageBuffer::new(new_width, height);
+    let mut ws = Vec::new();
 
     let max: f32 = NumCast::from(S::max_value()).unwrap();
     let ratio = width as f32 / new_width as f32;
     let sratio = if ratio < 1.0 { 1.0 } else { ratio };
     let src_support = filter.support * sratio;
 
-    for y in 0..height {
-        for outx in 0..new_width {
-            // Find the point in the input image corresponding to the centre
-            // of the current pixel in the output image.
-            let inputx = (outx as f32 + 0.5) * ratio;
+    for outx in 0..new_width {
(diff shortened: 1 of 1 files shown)


c-edw:
Any reason this hasn't been merged? Seems like a relatively straightforward change.

PR #809, opened 2018-09-11
Update num-rational dependency

Manishearth:
The ecosystem seems to be moving to v2, and I'd like to avoid duplication of this crate.

(Would be nice if this could be accompanied by a version bump / publish)

r? @bvssvni


bvssvni:
Thanks! Merging.

PR #810, opened 2018-09-18
Clippy lint pass

brookst:
I've implemented some suggestions from clippy-preview. There were two issues I had to ignore:
* `too_many_arguments` in `jpeg/encoder` since factoring out variables wouldn't improve this code clarity much.
* `while_let_loop` in `gif` since this would borrow the `reader` for the entire loop body. I'm not sure if this is something Clippy ought to be aware of or not.

Either way - this makes Image clean with Clippy on Rust stable.


bvssvni:
Thanks you! Merging.

PR #818, opened 2018-09-30
Get crate compiling on wasm32-unknown-emscripten

semtexzv:
This PR disables part of the functionality, but the crate correctly compiles on wasm32-unknown-emscripten, and is able to load images. 

Someone who understands the `gray_to_luma8` method should rewrite it.

This PR fixes https://github.com/PistonDevelopers/image/issues/817

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -665,7 +665,14 @@ pub fn decoder_to_image<I: ImageDecoder>(codec: I) -> ImageResult<DynamicImage> 
         None => Err(image::ImageError::DimensionError),
     }
 }
+// Compilation of this function failed on wasm32-unknown-emscripten
+// This is intented as temporary workaround
+#[cfg(target_os = "emscripten")]
+fn gray_to_luma8(_bit_depth: u8, _w: u32, _h: u32, _buf: &[u8]) -> Option<GrayImage> {
+    panic!("This feature is not supported on emscripten ");
+}
 
+#[cfg(not(target_os = "emscripten"))]
 fn gray_to_luma8(bit_depth: u8, w: u32, h: u32, buf: &[u8]) -> Option<GrayImage> {
     // Note: this conversion assumes that the scanlines begin on byte boundaries
     let mask = (1u8 << bit_depth as usize) - 1;


oyvindln:
It looks like it's a function to convert a higher than 8 bpp grayscale image to an 8bpp grayscale image. Probably not something that is used a lot.

HeroicKatora:
Also the opposite, converting lower than 8bpp images to 8bpp. Some decoders depend on this to load pure black-white bitmaps.

https://github.com/PistonDevelopers/image/blob/ad3ea2f89be96e118af67b308a3081ff1d33ff76/src/dynimage.rs#L659

philipc:
Here's a rewrite of the `p` calculation that seems to workaround the problem:
```rust
    let mut p = Vec::new();
    let mut i = 0;
    for v in buf {
        for shift in num_iter::range_step_inclusive(8i8-(bit_depth as i8), 0, -(bit_depth as i8)) {
            // skip the pixels that can be neglected because scanlines should
            // start at byte boundaries
            if i % (row_len as usize) < (w as usize) {
                let pixel = (v & mask << shift as usize) >> shift as usize;
                p.push(pixel * scaling_factor);
            }
            i += 1;
        }
    }
```
and in my opinion this is easier to read so it's not a bad thing to do. Note that this has all the same low level bit manipulation, I only removed the iterator adapter usage.

src/dynimage.rs
```diff
@@ -665,7 +665,14 @@ pub fn decoder_to_image<I: ImageDecoder>(codec: I) -> ImageResult<DynamicImage>
         None => Err(image::ImageError::DimensionError),
     }
 }
+// Compilation of this function failed on wasm32-unknown-emscripten
+// This is intented as temporary workaround
+#[cfg(target_os = "emscripten")]
+fn gray_to_luma8(_bit_depth: u8, _w: u32, _h: u32, _buf: &[u8]) -> Option<GrayImage> {
+    panic!("This feature is not supported on emscripten ");
```
bvssvni: I think using `unimplemented!` is better here.

bvssvni:
@philipc Do you think the workaround should be replacing the existing algorithm, or do you want it only for the wasm target?

philipc:
Replace the existing algorithm.

bvssvni:
@semtexzv If you replace the existing algorithm with the one proposed by @philipc , then we won't need to functions depending on the target.

bvssvni:
I think we should just replace the code. We can't wait forever on this getting fixed in the Rust compiler.

bvssvni:
Closed by https://github.com/PistonDevelopers/image/pull/844

Issue #798, opened 2018-07-30
RFC: Restructure ImageDecoder trait


fintelia:
I propose we break the `ImageDecoder` trait into several different traits targeting particular uses. These new traits consume self where appropriate to accommodate readers that can't seek, and `read_scanline` has been replaced with `into_box_read` and `read_image_with_progress`. I also introduce the concept of scanline_height so that decoders can express the minimum meaningful amount of data that they can decode without requiring internal buffering. Additionally, the new trait assumes that the image header has already been parsed, which means that metadata getter functions no longer need &mut self and cannot fail.

```rust
// This trait should be implementable for all image types, and should not require 
// `R: Seek` for any image types that don't need that bound for normal operation.
trait ImageDecoder {
    fn dimensions(&self) -> (u32, u32);
    fn colortype(&self) -> ColorType;
    fn row_bytes(&self) -> usize;
    fn read_image(self) -> ImageResult<DecodingResult>;

    // Returns minimum number of bytes that must be decoded at once to avoid buffering.
    // This could be as few as one byte or as many as the entire image.
    fn scanline_bytes(&self) -> usize;

    // Returns a reader that is intended to be called with slices of some multiple of 
    // scanline_bytes, but decodes into an internal buffer when required to read into
    // shorter slices.
    fn into_box_read(self) -> Box<Read>;
}

// This trait requires an underlying reader that has the ability to `Seek`.
trait ImageDecoderExt: ImageDecoder {
    // If possible for the image type, this function uses seek() to load only the
    // needed parts of the image into memory. Regardless, it then seeks back to the start
    // so `read_image` and friends still work.
    fn read_rect(&mut self, x: u32, y: u32, width: u32, height: u32, buf: &mut [u8]) -> ImageResult<()>;
}

// Non-object safe trait that lets users observe progress of image loading.
trait ImageDecoderIncremental {
    fn read_image_with_progress<F: Fn(f64)>(self, progress_callback: F) -> ImageResult<DecodingResult>;
}

// Only implemented for image types that support animation.
trait AnimationDecoder {
    fn into_frames(self) -> ImageResult<Frames>;
}
```

A typical decoder would then be implemented something like:
```rust
struct MyDecoder<R: Read> {}
impl<R: Read> ImageDecoder for MyDecoder<R> {}
impl<R: Read> ImageDecoderProgress for MyDecoder<R> {}
impl<R: Read + Seek> ImageDecoderExt for MyDecoder<R> {}
```

[...]

Issue #709, opened 2017-12-27
PNGDecoder::read_scanline returns row length instead of index 


fintelia:
Since the png crate doesn't not directly expose this information, the decoder should probably just track which row it is on itself. I'm also not certain, but a quick read through the code suggests that the read_scanline function will also silently return incorrect results for interlaced png images.

fintelia:
#819 Eliminates the `read_scanline` function making this no longer relevant

PR #819, opened 2018-10-01
Restructure ImageDecoder trait

fintelia:
## Overview
This pull request implements #798. The general motivation is that the existing `ImageDecoder` trait has some limitations which would be good to address before this crate reaches 1.0. This is necessarily a breaking change.

## Goals
* Address inconsistent naming: `load_rect` vs `read_image`.
* Fix lack of idempotence: typically only the first call to `read_image` returns the correct result (though this doesn't seem to be documented anywhere?)
   - These functions now consume self.
* Avoid unimplemented methods: many decoders currently leave `read_scanline` unimplemented.
   - No decoders panic on any of the methods for `ImageDecoder`. Implementions of `ImageDecoderExt` are still a work in progress, but will all have at least naive implementations.
* Eliminate inconsistent behavior: for instance, #709 
   - All methods now have clear semantics (I hope?)

## Trait Definitions
```rust
pub trait ImageDecoder: Sized {
    type Reader: Read;
    fn dimensions(&self) -> (u32, u32);
    fn colortype(&self) -> ColorType;
    fn into_reader(self) -> ImageResult<Self::Reader>;

    fn row_bytes(&self) -> usize { ... }
    fn total_bytes(&self) -> usize { ... }
    fn scanline_bytes(&self) -> usize { ... }
    fn read_image(self) -> ImageResult<Vec<u8>> { ... }
    fn read_image_with_progress<F: Fn(Progress)>(
        self, 
        progress_callback: F
    ) -> ImageResult<Vec<u8>> { ... }
}
```
```rust
// This trait requires an underlying reader that has the ability to `Seek`.
pub trait DecoderExt: ImageDecoder + Sized {
    // If possible for the image type, this function uses seek() to load only the
    // needed parts of the image into memory. Regardless, it then seeks back to the
    // start so `read_image` and friends still work.
    fn read_rect(
        &mut self, 
        x: u32, 
        y: u32, 
        width: u32, 
        height: u32, 
        buf: &mut [u8]
    ) -> ImageResult<Vec<u8>>;
}
```

## Outstanding questions
- [ ] Are the current choices for integer types correct? Specifically image sizes being `u32` and sizes of rows, scanlines, and images being `usize`? 
- [ ] Is getting rid of `DecodingResult` the right move?
- [ ] Should we take this opportunity to remove other deprecated functionality?

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1,13 +1,13 @@
 use std::cmp;
 use std::io;
-use std::io::{Read, Seek, SeekFrom};
+use std::io::{Cursor, Read, Seek, SeekFrom};
 use std::iter::{repeat, Iterator, Rev};
 use std::slice::ChunksMut;
 
 use byteorder::{LittleEndian, ReadBytesExt};
 
 use color::ColorType;
-use image::{DecodingResult, ImageDecoder, ImageError, ImageResult};
+use image::{ImageDecoder, ImageError, ImageResult};
 
 const BITMAPCOREHEADER_SIZE: u32 = 12;
 const BITMAPINFOHEADER_SIZE: u32 = 40;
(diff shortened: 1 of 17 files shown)


bvssvni:
Getting rid of `DecodingResult` seems fine. We can remove deprecated functionality through another PR.

I think we should just merge this.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.21.0"
+version = "0.20.1"
```
fintelia: This is a breaking change so we can't use a minor version bump...

philipc: This wasn't a minor version bump though, it was only changing it back to the version on master. Generally, this crate only bumps the version just before release. Also, the new version bump you added in https://github.com/PistonDevelopers/image/pull/819/commits/6a798c4235fff2618d58811a638822780a33d399 is invalid (it needs three numeric parts).

PR #822, opened 2018-10-13
Updated readme with a more colorful example for the Julia fractal

zumbalogy:
I am extremely new to Rust, and it took me a little while to figure out how to read image data with this library. I thought this updated version of the Julia fractal example maybe demonstrates that. 

The code runs and output fractal.png, but, as someone with no Rust experience, I would not be surprised if I am doing something in a non-idiomatic way or such. (As far as this specific fractal, I know all the work could be done in one loop, but I think demonstrating the get_pixel_mut method might also save someone a few minutes if that is what they need (like if they are just inverting the even rows of pixels or some such nonsense, or plotting a strange attractor).)

I tried to keep the example short enough to still fit all in one screen, but I am happy to add back the 'use' statements or whatever anyone thinks is best.

Feedback is welcome, but also feel free to ignore this whole thing if it is of little value.


README.md
```diff
@@ -203,45 +203,47 @@ fn main() {
...
+            let cy = x as f32 * scaley - 1.5;
 
-        let mut i = 0;
+            let c = num_complex::Complex::new(-0.4, 0.6);
+            let mut z = num_complex::Complex::new(cx, cy);
 
-        for t in 0..max_iterations {
-            if z.norm() > 2.0 {
-                break
+            let mut i = 0;
```
fintelia: Perhaps replace these lines with a while loop? Something like:

```rust
let mut i = 0;
while i < 255 && z.norm() <= 2.0 {
    z = z * z + c;
    i += 1;
}
```

zumbalogy: ah, i like that better too, will do.

bvssvni:
Merging.

bvssvni:
Thank you!

PR #825, opened 2018-10-16
sp

ibatra16:
sp


fintelia:
Care to explain why you'd like to add trailing whitespace there?

PR #828, opened 2018-10-31
Use Rust 1.24.1 as minimum supported version

### LINKED ISSUE #482: Travis CI fails for Rust 1.0.0
Currently the tests fails because an old bug in Cargo.

See https://github.com/rust-lang/cargo/pull/1714
See https://github.com/rust-num/num/pull/101

Thanks to @mbrubeck

francesca64:
Resolves #482 

I've also bumped the patch version; it would be super lovely if this could be released, since it's currently blocking the release of winit 0.18.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
     - linux
     - osx
 rust:
+    - 1.24.1
     - stable
     - beta
     - nightly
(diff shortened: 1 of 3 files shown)


bvssvni:
Published.

Issue #827, opened 2018-10-20
DynamicImage->ImageBuffer .. supportable types


dobkeratops:
http://www.piston.rs/image/image/struct.ImageBuffer.html

http://www.piston.rs/image/image/enum.DynamicImage.html

ok writing a texture loader from this.. at first , 24bit was fine. then generalising it to handle 24bit/32bit seemed slightly awkward (i might be a bit fussy about eliminating some cut-paste that happens because you need to match to get the buffer)

The effort will be worthwhile however if this wrapping(which seems to allow a type-parameter for the image buffer type) allows safe handling of a multitude of formats:

* 24,32bit - done
* 16bit:  1555,  565, 4444 - these do have some use-cases
* DXT compression .. 

I was thinking an easier way to make image buffers safe would be to use a 3D array as an intermediate: and moving images to & from **AI libraries** more frequently involves multi-dimensional array operations (especially consider 4D arrays for the feature detectors, and synergy between that and 'array texture' code, i.e. "16 slices of 4-channel RGBA 512x512 images" = u8[512][512][4][16]

.. but if this enum/match approach can handle different bit-level encodings, that would be great

fintelia:
Would using the [`ImageDecoder` trait](https://docs.rs/image/0.20.0/image/trait.ImageDecoder.html) directly work better for your use case?

dobkeratops:
@fintelia thanks for the info.. it's hard to tell, it seems to solve a different problem i.e. providing more interface to step through an image, whereas I just want to pass it to GL.
it also seems to decode then tell you a format via a variant meaning the usage code will look similar; I suspect in those kind of cases i'd want my code to ask for the image in a specific element format (i.e. "i'm going to do some image processing in HDR range fp32 ; give me the image as a 3D array of fp32 values" - because you've chosen to run on a machine that's biased to that format or you're doing somthing that needs that range for intermediates ).  (I was surprising impressed with how useable numpy was just throwing arrays around for these kind of use cases, and ditto *later* if i do move into AI-related image processing stuff i think thats what code would look like. I'm talking about 2 different things here.. just loading into an engine, and more advanced image processing)

dobkeratops:
to clarify the 'problem' here, not so bad but it will get worse scaling for more texture formats:-

```
fn create_texture_from(dimg:image::DynamicImage)->TextureHandle{

	if let Some(((us,vs),bfr,fmt))=
		(match dimg {
[...]

PR #829, opened 2018-11-08
Add a flat image buffer representation

197g:
As discussed in #827 , something like this is very likely to be helpful for interoperability. This is a rough draft of how it could look and needs:

- [x] Tests
- [x] Conversions from `ImageBuffer`, `DynamicImage`
- [x] Checked conversion to `ImageBuffer`

### DIFF
--- a/src/flat.rs
+++ b/src/flat.rs
@@ -164,7 +164,7 @@ impl SampleLayout {
         }
     }
 
-    /// Get the strides for indexing matrix-like [(c, w, h)].
+    /// Get the strides for indexing matrix-like `[(c, w, h)]`.
     ///
     /// For a row-major layout with grouped samples, this tuple is strictly
     /// increasing.
@@ -172,7 +172,7 @@ impl SampleLayout {
         (self.channel_stride, self.width_stride, self.height_stride)
     }
 
-    /// Get the dimensions (channels, width, height).
+    /// Get the dimensions `(channels, width, height)`.
     ///
     /// The interface is optimized for use with `strides_cwh` instead. The channel extent will be
     /// before width and height.
@@ -183,7 +183,7 @@ impl SampleLayout {
     /// Tuple of bounds in the order of coordinate inputs.
     ///
     /// This function should be used whenever working with image coordinates opposed to buffer
-    /// coordinates.
+    /// coordinates. The only difference compared to `extents` is the output type.
     pub fn bounds(&self) -> (u8, u32, u32) {
         (self.channels, self.width, self.height)
     }
@@ -248,7 +248,7 @@ impl SampleLayout {
             .and_then(|idx| idx.checked_add(1))
     }
 
-    /// Check if the buffer is large enough.
+    /// Check if a buffer of length `len` is large enough.
     pub fn fits(&self, len: usize) -> bool {
         self.min_length().map(|min| len >= min).unwrap_or(false)
     }
@@ -355,6 +355,10 @@ impl SampleLayout {
     }
 
     /// Check that the pixel and the channel index are in bounds.
+    ///
+    /// An in-bound coordinate does not yet guarantee that the corresponding calculation of a
+    /// buffer index does not overflow. However, if such a buffer large enough to hold all samples
+    /// actually exists in memory, this porperty of course follows.
     pub fn in_bounds(&self, channel: u8, x: u32, y: u32) -> bool {
         return channel < self.channels && x < self.width && y < self.height
     }
@@ -373,7 +377,8 @@ impl SampleLayout {
     /// Get the theoretical position of sample (channel, x, y).
     ///
     /// The 'check' is for overflow during index calculation, not that it is contained in the
-    /// image.
+    /// image. Two samples may return the same index, even when one of them is out of bounds. This
+    /// happens when all strides are `0`, i.e. the image is an arbitrarily large monochrome image.
     pub fn index_ignoring_bounds(&self, channel: usize, x: usize, y: usize) -> Option<usize> {
         let idx_c = (channel as usize).checked_mul(self.channel_stride);
         let idx_x = (x as usize).checked_mul(self.width_stride);
@@ -429,7 +434,7 @@ impl Dim {
 }
 
 impl<Buffer> FlatSamples<Buffer> {
-    /// Get the strides for indexing matrix-like [(c, w, h)].
+    /// Get the strides for indexing matrix-like `[(c, w, h)]`.
     ///
     /// For a row-major layout with grouped samples, this tuple is strictly
     /// increasing.
@@ -437,7 +442,7 @@ impl<Buffer> FlatSamples<Buffer> {
         self.layout.strides_cwh()
     }
 
-    /// Get the dimensions (channels, width, height).
+    /// Get the dimensions `(channels, width, height)`.
     ///
     /// The interface is optimized for use with `strides_cwh` instead. The channel extent will be
     /// before width and height.
@@ -448,7 +453,7 @@ impl<Buffer> FlatSamples<Buffer> {
     /// Tuple of bounds in the order of coordinate inputs.
     ///
     /// This function should be used whenever working with image coordinates opposed to buffer
-    /// coordinates.
+    /// coordinates. The only difference compared to `extents` is the output type.
     pub fn bounds(&self) -> (u8, u32, u32) {
         self.layout.bounds()
     }
@@ -780,7 +785,7 @@ impl<Buffer> FlatSamples<Buffer> {
         self.layout.min_length()
     }
 
-    /// Check if the buffer is large enough.
+    /// Check if a buffer of length `len` is large enough.
     pub fn fits(&self, len: usize) -> bool {
         self.layout.fits(len)
     }
@@ -806,6 +811,10 @@ impl<Buffer> FlatSamples<Buffer> {
     }
 
     /// Check that the pixel and the channel index are in bounds.
+    ///
+    /// An in-bound coordinate does not yet guarantee that the corresponding calculation of a
+    /// buffer index does not overflow. However, if such a buffer large enough to hold all samples
+    /// actually exists in memory, this porperty of course follows.
     pub fn in_bounds(&self, channel: u8, x: u32, y: u32) -> bool {
         self.layout.in_bounds(channel, x, y)
     }
@@ -820,7 +829,8 @@ impl<Buffer> FlatSamples<Buffer> {
     /// Get the theoretical position of sample (x, y, channel).
     ///
     /// The 'check' is for overflow during index calculation, not that it is contained in the
-    /// image.
+    /// image. Two samples may return the same index, even when one of them is out of bounds. This
+    /// happens when all strides are `0`, i.e. the image is an arbitrarily large monochrome image.
     pub fn index_ignoring_bounds(&self, channel: usize, x: usize, y: usize) -> Option<usize> {
         self.layout.index_ignoring_bounds(channel, x, y)
     }
@@ -849,9 +859,15 @@ impl<Buffer> FlatSamples<Buffer> {
 /// This is a nearly trivial wrapper around a buffer but at least sanitizes by checking the buffer
 /// length first and constraining the pixel type.
 ///
-/// Note that this does not eliminate panics as the `AsRef<[T]` implementation of `Buffer` may be
+/// Note that this does not eliminate panics as the `AsRef<[T]>` implementation of `Buffer` may be
 /// unreliable, i.e. return different buffers at different times. This of course is a non-issue for
 /// all common collections where the bounds check once must be enough.
+///
+/// # Inner invariants
+///
+/// * For all indices inside bounds, the corresponding index is valid in the buffer
+/// * `P::channel_count()` agrees with `self.inner.layout.channels`
+///
 #[derive(Clone, Debug)]
 pub struct View<Buffer, P: Pixel> 
 where 
@@ -867,6 +883,14 @@ where
 /// library endorsed normalized representation is still `ImageBuffer`. Also, the implementation of
 /// `AsMut<[P::Subpixel]>` must always yield the same buffer. Therefore there is no public way to
 /// construct this with an owning buffer.
+///
+/// # Inner invariants
+///
+/// * For all indices inside bounds, the corresponding index is valid in the buffer
+/// * There is no aliasing of samples
+/// * The samples are packed, i.e. `self.inner.layout.sample_stride == 1`
+/// * `P::channel_count()` agrees with `self.inner.layout.channels`
+///
 #[derive(Clone, Debug)]
 pub struct ViewMut<Buffer, P: Pixel> 
 where 
@@ -891,7 +915,7 @@ pub enum Error {
 
     /// The represented image can not use this representation.
     ///
-    /// The normalized form that would be accepted.
+    /// Has an additional value of the normalized form that would be accepted.
     NormalFormRequired(NormalForm),
 
     /// The color format did not match the channel count.


src/flat.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub horizontal_stride: usize,
+
+    /// Add this to an index to get to the next sample in y-direction.
+    pub vertical_stride: usize,
+
+    /// Add this to an index to get to the sample in the next channel.
+    pub channel_stride: usize,
+
+    /// The width of the represented image.
+    pub width: u32,
```
fintelia: Should this be a u64 for greater generality? I know most of the library currently uses 32 bit ints for sizes, but given there are already checks before converting to ImageBuffer we don't necessary need to stick with that decision

HeroicKatora: This is consistent with `GenericImageView::dimensions()` and guarantees that `width*height` does not overflow `usize` on 64-bit systems. Badly extrapolating current development in image capturing (x2 every five years), this has 17-18 more bits than currently required lasting some ~30-40 years.

fintelia: Fair enough. Much larger image sizes can definitely come up when working with other sorts of things like datasets but still probably won't run into the 32-bit size limit for quite some time. 

Do you know if there are any areas where it is common to store 1D data in image files? There it could be an issue much faster because a 4B x 1 image file would only be ~4GB which isn't hard to imagine. 

Regardless, I tend to err on the side of using 64-bits for everything unless have a strong reason not to, so it is probably fine just to ignore this concern.

HeroicKatora: You are right, you might want to quickly reshape and still be able to represent all the intermediate formats. If we encourage that usage, then `u64` is better than `usize` because the compressed image representations from `stride = 0` make it possible to have images larger than memory, possibly *just* for extracting a subimage.

src/buffer.rs
```diff
@@ -375,6 +376,54 @@ where
...
         <P as Pixel>::from_slice(&self.data[index..index + no_channels])
     }
+
+    /// Return the raw sample buffer with its stride an dimension information.
+    ///
+    /// The returned buffer is guaranteed to be well formed in all cases. It is
+    /// layed out by colors, width then height, meaning `channel_stride <=
+    /// vertical_stride <= horizontal_stride`. All strides are in numbers of
+    /// elements but those are mostly `u8` in which case the strides are also
+    /// byte strides.
```
fintelia: Based on `get_pixels` I think you have this reversed. vertical_stride should be larger than horizontal_stride since changing x by one results in index changing by no_channels whereas changing y by 1 results in index changing by no_channels*width.

src/buffer.rs
```diff
@@ -375,6 +376,54 @@ where
...
+    pub fn flattened(self) -> FlatSamples<P::Subpixel, Container> 
+        where Container: AsRef<[P::Subpixel]> 
+    {
+        // None of these can overflow, as all our memory is addressable.
+        let cstride = 1usize;
+        let vstride = <P as Pixel>::channel_count() as usize;
+        let hstride = cstride*self.width as usize;
```
fintelia: Shouldn't this be `vstride*self.width as usize;`?

src/buffer.rs
```diff
@@ -375,6 +376,54 @@ where
...
+    /// Return a view on the raw sample buffer.
+    ///
+    /// See `flattened` for more details.
+    pub fn as_flattened(&self) -> FlatSamples<P::Subpixel, &[P::Subpixel]>
+        where Container: AsRef<[P::Subpixel]> 
+    {
+        // None of these can overflow, as all our memory is addressable.
+        let cstride = 1usize;
+        let vstride = <P as Pixel>::channel_count() as usize;
+        let hstride = cstride*self.width as usize;
```
fintelia: See above

src/buffer.rs
HeroicKatora: You're right, I messed naming. `vertical_stride` is probably interpreted as going into the vertical/`y` direction. This then also implies a correction that is necessary in `FlatSamples::index`. I really should do tests soon.

HeroicKatora: Speaking of which, should the `channel` parameter in `FlatSamples::index` be first? So that it the general order of indices consistent with `extents` and `ImageBuffer::dimensions()`

HeroicKatora:
@bvssvni @fintelia @dobkeratops I'm pretty sure the current design is pretty stable, only missing a few converters and utilities. Can anyone do a code review? Or just generally comment on it.

src/dynimage.rs
```diff
@@ -286,6 +287,11 @@ impl DynamicImage {
         image_to_bytes(self)
     }
 
+    /// Return a view on the raw sample buffer.
+    pub fn as_flattened(&self) -> FlatSamples<&[u8]> {
```
fintelia: How about as_flattened -> as_flat_samples

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+/// of library functions will always guarantee this. To manually make this
+/// check use `check_index_validities` and maybe put that inside an assert.
+#[derive(Clone, Debug)]
+pub struct FlatSamples<Buffer> {
+    /// Underlying linear container holding sample values.
+    pub samples: Buffer,
+
+    /// A `repr(C)` description of the buffer format.
+    pub format: MatrixFormat,
+}
```
fintelia: How would you feel about also including and `Option<ColorType>` field for when the image format is known and is one of the supported ones?

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+
+    /// Shrink the image to the minimum of current and given extents.
+    ///
+    /// This does not modify the strides, so that the resulting sample buffer may have holes
+    /// created by the shrinking operation. Shrinking could also lead to an non-aliasing image when
+    /// samples had aliased each other before.
+    pub fn shrink_to(&mut self, width: u32, height: u32, channels: u8) {
+        self.format.shrink_to(width, height, channels)
+    }
+}
```
fintelia: Could FlatSamples also have a get() function implemented whenever Buffer: Index<usize>?

HeroicKatora:
> It is worth thinking through whether we can make sure that FlatSamples is the last one we need to add.

It is not. But I don't think there is a way around this when faced with the real world and efficiency in that setting. There is [Yuv420p](https://en.wikipedia.org/wiki/YUV#Y%E2%80%B2UV420p_(and_Y%E2%80%B2V12_or_YV12)_to_RGB888_conversion) and other graphics native buffer types that even violate the notion of having a consistent number of channels (think that you might want to compress channels on the edges of your screen more because you glance there less often, also a real thing but I can't find sources). 

However, I think it is the last type to add *for all formats viable as mutable images* until a later point. Specifically because mutating compressed information has glaringly bad edge cases for keeping changes revertible. And blend etc. becomes a pain to even define.

Now, with regards to `DynamicImage` in particular, this is the thing that would hold me back from a `1.0` release. It does not even try to support sample types other than `u8` right now. But as you said yourself, this is general enough and independent from that issue.

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+    /// care about the value of the alpha channel even though you need `Rgba`.
+    WrongColor(ColorType),
+}
+
+/// Different normal forms of buffers.
+///
+/// A normal form is an unaliased buffer with some additional constraints.  The `ÌmageBuffer` uses
+/// row major form with packed samples. 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum NormalForm {
```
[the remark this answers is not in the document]
HeroicKatora: I'd rather have offered this publically with and additional `normalize` function in the spirit of `slice::sort` or `slice::reverse`. That way, one could import a general buffer and normalize it at a convenient time before calling a never-failing `into_buffer`, instead of having to copy a `View` into a newly allocated `ImageBuffer` or having to handroll this.

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+    pub samples: Buffer,
+
+    /// A `repr(C)` description of the buffer format.
+    pub format: MatrixFormat,
+}
+
+/// A ffi compatible description of a sample buffer.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct MatrixFormat {
```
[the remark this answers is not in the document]
HeroicKatora: I like the `Layout` part, but as for `Buffer` I'd like to avoid the reference to `ImageBuffer`.

src/flat.rs
HeroicKatora: I'm not sure how that would work exactly. For using the index kinds of the underlying buffer, I generally though `samples()[..]` to be good enough. If this is about `Index<(u32, u32)>`, sure.

src/flat.rs
HeroicKatora: I'm torn on this. This feels like it deserved an additional consistency check when calling `as_view` but you don't generally gain help in the form of extra type checking. A `into_dynamic_image` would sure make this much more useful.

src/flat.rs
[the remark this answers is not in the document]
fintelia: Oh, good point

src/dynimage.rs
fintelia: I like `FlatSamples` slightly better, but I don't feel strongly either way

src/flat.rs
fintelia: I was imagining something like:
```rust
fn get<T>(x: u32, y: u32, channel: u8) -> T
where
    Self::Buffer: Index<usize, T>,
    T: Copy
{
    self.buffer[self.format.index(x, y, channel).unwrap()]
}

File: src/dynimage.rs (hunk 5 above)
Comment (HeroicKatora) [reply src/dynimage.rs]: It has the association to `image planes`, which are possible here in contrast to `ImageBuffer`, so then it'll stay.

File: src/flat.rs (hunk 7 above)
Comment (HeroicKatora) [reply src/flat.rs]: Ah, I see, I for some reason thought this was on `View`. `panic` makes sense due to behaviour of `ImageBuffer::get()` so yes, good idea.

Comment (HeroicKatora) [reply src/flat.rs]: One odd thing is that the arguments of returning `T` instead of `&T` like `Vec::get` is not as strong as on `GenericImage`. That is, neither is the api old nor are we aggregating data.

**Comment by fintelia** [comment]
Yeah, I think we probably need to punt compressed image representations and colorspace handling if we want to have a 1.0 release any time soon.

File: src/flat.rs [hunk 10]
```diff
@@ -0,0 +1,1139 @@
...
+            if self.height as usize*self.height_stride != self.width_stride {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /// Check that the pixel and the channel index are in bounds.
+    pub fn in_bounds(&self, x: u32, y: u32, channel: u8) -> bool {
```
HeroicKatora: One concern of myself, which I am not yet sure how to resolve, is how the order here and in `index` is inconsistent with `strides_hwc` and `extents`. We want to stay consistent with both `Image::get(x, y)` and the order of potentially row-major strides (`hwc` or `cwh`). This suggest switching to `fn in_bounds(&self, channel: u8, x: u32, y: u32)`.

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+        return x < self.width && y < self.height && channel < self.channels
+    }
+
+    /// Resolve the index of a particular sample.
+    ///
+    /// `None` if the index is outside the bounds or does not fit into a `usize`.
+    pub fn index(&self, x: u32, y: u32, channel: u8) -> Option<usize> {
```
HeroicKatora: Also affected

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+        }
+
+        self.index_ignoring_bounds(x as usize, y as usize, channel as usize)
+    }
+
+    /// Get the theoretical position of sample (x, y, channel).
+    ///
+    /// The 'check' is for overflow during index calculation, not that it is contained in the
+    /// image.
+    pub fn index_ignoring_bounds(&self, x: usize, y: usize, channel: usize) -> Option<usize> {
```
HeroicKatora: Also affected

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+        Some(0usize)
+            .and_then(|b| b.checked_add(idx_c))
+            .and_then(|b| b.checked_add(idx_x))
+            .and_then(|b| b.checked_add(idx_y))
+    }
+
+    /// Get an index provided it is inbouds.
+    ///
+    /// The computation can not overflow as we could represent the maximum coordinate.
+    pub fn in_bounds_index(&self, x: u32, y: u32, c: u8) -> usize {
```
HeroicKatora: Also affected

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+        (y as usize * y_stride) + (x as usize * x_stride) + (c as usize * c_stride)
+    }
+
+
+    /// Shrink the image to the minimum of current and given extents.
+    ///
+    /// This does not modify the strides, so that the resulting sample buffer may have holes
+    /// created by the shrinking operation. Shrinking could also lead to an non-aliasing image when
+    /// samples had aliased each other before.
+    pub fn shrink_to(&mut self, width: u32, height: u32, channels: u8) {
```
HeroicKatora: Also affected.

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+        self.0*self.1
+    }
+}
+
+impl<Buffer> FlatSamples<Buffer> {
+    /// Get the strides for indexing matrix-like [(h, w, c)].
+    ///
+    /// For a row-major layout with grouped samples, this tuple is strictly
+    /// decreasing.
+    pub fn strides_hwc(&self) -> (usize, usize, usize) {
```
HeroicKatora: Maybe switch to `strides_cwh`?

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+    /// captured (by design) by the type system. These are then checked before the conversion. Such
+    /// checks can all be done in constant time and will not inspect the buffer content. You can
+    /// perform these checks yourself when the conversion is not required at this moment but maybe
+    /// still performed later.
+    pub fn is_normal(&self, form: NormalForm) -> bool {
+        self.format.is_normal(form)
+    }
+
+    /// Check that the pixel and the channel index are in bounds.
+    pub fn in_bounds(&self, x: u32, y: u32, channel: u8) -> bool {
```
HeroicKatora: Don't forget these5

src/flat.rs
```diff
@@ -0,0 +1,1139 @@
...
+    /// conversion from `FlatSamples`–the resulting slice may still contain holes.
+    pub fn image_slice(&self) -> &[P::Subpixel] {
+        &self.inner.samples.as_ref()[..self.min_length()]
+    }
+
+    /// Shrink the inner image.
+    ///
+    /// The new dimensions will be the minimum of the previous dimensions. Since the set of
+    /// in-bounds pixels afterwards is a subset of the current ones, this is allowed on a `View`.
+    pub fn shrink_to(&mut self, width: u32, height: u32, channels: u8) {
```
HeroicKatora: And this.

HeroicKatora:
>  by allowing things like Buffer=BitVec<T> doesn't even assume that components take up a full 8 bits.

Well, interaction with other components is a bit harder in those cases. Since we only ever create `View<&[Subpixel]>` it does not matter terribly if, in the future, we make a change to it to allow buffer types that are not `AsRef<[T]>` but more generally `Index<usize, Output=T> + Index<Range<usize>>`. I don't want to make it as generic right now because I think there might be some traps in the exact trait bounds and once it is allowed, breaking the interface feels terrible.

src/flat.rs
HeroicKatora: Added: 

```
impl<Buffer> Index<(u8, u32, u32)> for FlatSamples<Buffer>
    where Buffer: Index<usize>
{
    …
}
```

src/flat.rs
HeroicKatora: I'll add a `color_hint: Option<ColorType>` with an explicit warning that it is not checked in any of the conversions. A later dynamic converter could then be `into_hinted_image`.

src/flat.rs
HeroicKatora: And added mutable variants and more as well

HeroicKatora:
With the addition of `get_mut_sample`, a `View` constructor that allows `get_mut_sample` and later upgrading to `ViewMut` I believe this is now a very well rounded and complete unit. After checks complete we should likely merge it as such if nothing comes up.

Ping @bvssvni so you can add final comments if you want.

Issue #823, opened 2018-10-14
Support loading gif frames one at a time.


ArturKovacs:
It really seems to me that the current implementation loads all the frames in the animated gif at once. I would really like to be able to read the first frame as soon as it is loaded and only then start loading the next frame.

In my opinion the API is fine, the only change that may be done is to load the frame within the `next` function of the Frames struct i.e. to make the elements of the Frames struct be evaluated in a lazy fashion.

If there is no opposition to this idea then I would be happy to implement it.

ArturKovacs:
After spending some time writing code I changed my view about the idea that it could be done within the `next` function of the Frames struct.

The aforementioned `next` function has the following signature.

```rust
fn next(&mut self) -> Option<Frame>
```

The problem is that this function returns an `Option<Frame>` which cannot give any information about an error that might occur while loading a frame. The return value therefore should be `Option<ImageResult<Frame>>`.

To solve this without breaking the API a new function alongisde `into_frames` could be introduced to the `ImageDecoder` trait. The signature of such a function might look like the following.

```rust
fn into_lazy_iter(self) -> Box<Iterator<Item = ImageResult<Frame>> + 'a>
```

The specific decoders that implement this trait could then return an iterator which takes ownership of any implementation specific structure. For example the gif decoder would return a struct that owns the underlying `gif::Reader`.

Note the `'a` lifetime attached to the boxed iterator. I'm not that knowledgeable about lifetimes but as far as I understand this has to be matched with the lifetime of the `Read` trait an implementer of which (eg File) the gif decoder uses and passes to the iterator within the `gif::Reader`.

I currently have an implementation that follows this description. I think this topic may require some discussion but otherwise I'll post a pull request in the near future.

fintelia:
Any change to the `ImageDecoder` trait would necessarily be a breaking change. However, I'm actually in the process of [proposing a bunch of other cleanup to the trait](#819) which would also constitute breaking changes. It might make sense to apply these at the same time so we have only a single version bump.

ArturKovacs:
The link is broken but I assume you are referring to #819.

I completely agree that the two should be done at the same time. In that case the best approach in my opinion would be to change the `Frames` object to read the frames in a lazy fashion. Whenever someone needs all the frames at once they can just call `collect` on it.

Shnatsel:
Fixed by #846

PR #831, opened 2018-11-09
Remove Frames, add FrameIterator that reads the frames in a lazy manner.

ArturKovacs:
Implements a solution for #823.

Suggestions are welcome.

### DIFF
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -1,20 +1,26 @@
+use std::iter::Iterator;
+
 use num_rational::Ratio;
 
 use buffer::RgbaImage;
+use image::ImageResult;
 
-/// Holds the frames of the animated image
-pub struct Frames {
-    frames: Vec<Frame>,
-    current_frame: usize,
+/// Struct that can be used to iterate through the frames of an animated image such as a gif.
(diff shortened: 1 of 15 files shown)


ArturKovacs:
It may be better to have `gif::Encoder::encode_frames` take an `Iterator<Item = animation::Frame>` or simply a `&Vec<animation::Frame>`  instead of an `Iterator<Item = ImageResult<animation::Frame>>`. Also `FrameIterator` could actually just be named `Frames` just like before.

What do you think?

bvssvni:
I think "Frames" sounds better than "FrameIterator".

ArturKovacs:
Yes, I actually agree. Alright I will wait for #798 to be merged and as soon that's done I will probably delete this PR and make a new one, based on those changes.

bvssvni:
Nice! I was wondering about that. I'll check in tomorrow to see if there any objections, but I expect that https://github.com/PistonDevelopers/image/pull/819 will be go smoothly (it has been open since October and nobody has objected yet).

ArturKovacs:
New pull request, implementing changes on the latest commit will be opened.

PR #833, opened 2018-11-10
Replace copy_memory with copy_from_slice + Fix ImageDecoder::load_rect

kngwyu:
I think it's make our code a bit simpler.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -267,9 +267,8 @@ pub trait ImageDecoder: Sized {
             {
                 let from = &tmp[x as usize * bpp..width as usize * bpp];
 
-                let to = &mut buf[i * width as usize * bpp..width as usize * bpp];
-
-                ::copy_memory(from, to);
+                let start = i * width as usize;
+                buf[start..start + from.len()].copy_from_slice(from);
             }
 
             let _ = try!(self.read_scanline(&mut tmp));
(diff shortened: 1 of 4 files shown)


src/image.rs
```diff
@@ -267,9 +267,8 @@ pub trait ImageDecoder: Sized {
             {
                 let from = &tmp[x as usize * bpp..width as usize * bpp];
 
-                let to = &mut buf[i * width as usize * bpp..width as usize * bpp];
-
-                ::copy_memory(from, to);
+                let start = i * width as usize;
+                buf[start..start + from.len()].copy_from_slice(from);
```
philipc: It looks like you're trying to fix a bug here at the same time, but it still looks wrong to me. I haven't tested, but I would expect it to be:
```rust
let from = &tmp[x as usize * bpp..][..width as usize * bpp];
let to = &mut buf[i * width as usize * bpp..][..width as usize * bpp];
to.copy_from_slice(from);
```

kngwyu: Your code looks correct.
Thanks.

kngwyu:
Since `load_rect` had additional bugs other than @philipc commented, I added a test for it and fixed it.
In addition, how about changing its args from `(&mut self, x: u32, y: u32, length: u32, width: u32)` to `(&mut self, x: u32, y: u32, width: u32, length: u32)`, which is the same order as `GenericImage::sub_image`?
This is a breaking change, but I think not so many people will be affected by it, since this method is currently completely broken.

fintelia:
I'm currently working on more extreme breaking changes to this functionality in #819, so I'd advocate for holding off on changing the argument order here. Otherwise, I'm in favor of making the method slightly less buggy.

kngwyu:
Closed due to #819 merged

PR #841, opened 2018-12-16
Correct to_bgra docstring for DynamicImage

sameer:
The docstring says RGBA when it should say BGRA, this PR corrects it.


fintelia:
Looks good to me

PR #843, opened 2018-12-17
bmp/encoder: encode RGBA using bitfields with alpha mask

### LINKED ISSUE #840: Support for encoding BMP transparency
Related to #539 
It's confusing that the encoder just silently ignores the request for RGBA format... we should either support it or return an error.
This problem negatively affects https://github.com/kvark/vange-rs/pull/54, where I had to resort to PNG (for some reason, TGA also refused to encode), and PNG is less than ideal there, since compression makes 2Kx16K images a bit too slow.

philipc:
Fixes #840 
cc @kvark

### DIFF
--- a/src/bmp/encoder.rs
+++ b/src/bmp/encoder.rs
@@ -3,6 +3,10 @@ use std::io::{self, Write};
 
 use color;
 
+const BITMAPFILEHEADER_SIZE: u32 = 14;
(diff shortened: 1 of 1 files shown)


bvssvni:
Looks good to me.

bvssvni:
Merging.

bvssvni:
Thanks!

PR #846, opened 2018-12-23
The `animation::Frames` struct now loads gif frames one at a time.

ArturKovacs:
This implements a solution to #823.

### DIFF
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -20,7 +20,7 @@ impl<'a> Frames<'a> {
     /// a `Vec`.
     /// If en error is encountered that error is returned instead.
     /// 
-    /// Note: This is equvalent to `Frames::collect::<ImageResult<Vec<Frame>>>()`
+    /// Note: This is equivalent to `Frames::collect::<ImageResult<Vec<Frame>>>()`
     pub fn collect_frames(self) -> ImageResult<Vec<Frame>> {
         self.collect()
     }
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -19,15 +19,10 @@ impl<'a> Frames<'a> {
     /// Steps through the iterator from the current frame until the end and pushes each frame into
     /// a `Vec`.
     /// If en error is encountered that error is returned instead.
-    pub fn collect_frames(mut self) -> ImageResult<Vec<Frame>> {
-        let mut result = Vec::new();
-        while let Some(item) = self.next() {
-            match item {
-                Ok(item) => result.push(item),
-                Err(err) => return Err(err),
-            }
-        }
-        Ok(result)
+    /// 
+    /// Note: This is equvalent to `Frames::collect::<ImageResult<Vec<Frame>>>()`
+    pub fn collect_frames(self) -> ImageResult<Vec<Frame>> {
+        self.collect()
     }
 }
 
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -1,20 +1,40 @@
+use std::iter::Iterator;
+
 use num_rational::Ratio;
 
 use buffer::RgbaImage;
+use image::ImageResult;
 
-/// Holds the frames of the animated image
-pub struct Frames {
-    frames: Vec<Frame>,
-    current_frame: usize,
+/// An implementation dependent iterator, reading the frames as requested
+pub struct Frames<'a> {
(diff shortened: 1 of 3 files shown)


ArturKovacs:
@bvssvni this is the new PR as discussed in #831.

src/animation.rs
```diff
@@ -1,20 +1,40 @@
...
+    /// Creates a new `Frames` from an implementation specific iterator.
+    pub fn new(iterator: Box<Iterator<Item = ImageResult<Frame>> + 'a>) -> Self {
+        Frames { iterator }
+    }
+
+    /// Steps through the iterator from the current frame until the end and pushes each frame into
+    /// a `Vec`.
+    /// If en error is encountered that error is returned instead.
+    pub fn collect_frames(mut self) -> ImageResult<Vec<Frame>> {
+        let mut result = Vec::new();
```
HeroicKatora: This whole procedure should be equivalent to `self.collect::<ImageResult<Vec<Frame>>()` if I'm not mistaken. The `FromIterator` impls for `Result` is very broad.

src/gif.rs
```diff
@@ -224,7 +269,10 @@ impl<W: Write> Encoder<W> {
         result
     }
     /// Encodes Frames.
-    pub fn encode_frames(&mut self, frames: animation::Frames) -> ImageResult<()> {
+    pub fn encode_frames<Frames>(&mut self, frames: Frames) -> ImageResult<()>
+    where 
+        Frames: Iterator<Item = animation::Frame>
```
HeroicKatora: I'd recommend `IntoIterator<animation::Frame>`, that makes this api more strictly more broad.

It strikes me as odd that I can not use `animation::Frames` as an argument to `encode_frames`. Generic argument should probably be named different, and maybe a `try_encode_frames` which takes `IntoIterator<ImageResult<animation::Frame>>`?

Minor stylistic choice: Rewriting the `for`-body in terms of `map` and `collect`, avoiding the `if let Err(_)` control flow at the end and `Ok(())`.

ArturKovacs: First of all, thank you for the review!

Yes, using `IntoIterator` instead of `Iterator` is a great idea, I will certainly do that. Also `try_encode_frames` seems like it would have some nice use cases so I'll make that as well, having it return the error in case of an `Err`.

On the other hand I don't really see how `if let Err(_)` could be alleviated. Could you elaborate on that?

HeroicKatora: Sure, the body could be written as:

```
frames.into_iter().map(|frame| {
    // get the delay before coverting img_frame
    let frame_delay = img_frame.delay().to_integer();
    // convert img_frame into RgbaImage
    let rbga_frame = img_frame.into_buffer();

    // Create the gif::Frame from the animation::Frame
    let mut frame = Frame::from_rgba(rbga_frame.width() as u16, rbga_frame.height() as u16, &mut rbga_frame.into_raw());
    frame.delay = frame_delay;

    // encode the gif::Frame
    self.encode(&frame)
}).collect()
```

It's mostly a matter of style and personal taste though.

ArturKovacs: Hmm but in this case `collect` would resolve to `collect<ImageResult<()>>`. As long as I'm not missing something I think this would go against the general expectation of what `collect` means and what `ImageResult` means so I prefer the original form.

HeroicKatora: Depends on what is general expectation of `collect`. From its own documentation:

```
This is one of the more powerful methods in the standard library, used in a variety of contexts.
[…]
One of the keys to collect()'s power is that many things you might not think of as 'collections' actually are. […] And a collection of Result<T, E> can be thought of as single Result<Collection<T>, E>. See the examples below for more.
```

ArturKovacs:
I ended up using the `?` operator as I didn't find a nice way to avoid it in the `try_encode_frames` function so I just used that in both function to remain consistent.

Also I kept the `collect_frames` function but I changed the implementation. However I'm not sure if it's good to have that function.

ArturKovacs:
The CI builds seem they failed at installing the nightly toolchain. I will close and reopen this pull request in an attempt to trigger a travis rebuild.

bvssvni:
Looks good. I think we can keep `collect_frames`. Any more changes before merging?

bvssvni:
Merging.

bvssvni:
Thanks!

PR #848, opened 2018-12-25
fix imageops::overlay() range_height

martinlindhe:
Fixes range_height calculation under certain conditions.

Without the fix, this caused a crash with input
// bottom_height 298
// top_height 32
// y 301

since 298 - 301 overflows u32

### DIFF
--- a/src/imageops/mod.rs
+++ b/src/imageops/mod.rs
@@ -57,7 +57,11 @@ pub fn overlay<I: GenericImage>(bottom: &mut I, top: &I, x: u32, y: u32) {
     };
 
     let range_height = if y + top_height > bottom_height {
-        bottom_height - y
+        if y < bottom_height {
+            bottom_height - y
+        } else {
+            0
+        }
     } else {
         top_height
     };


HeroicKatora:
Good point, those range behaviours should be documented or changed.

1. The issue seems to be the same for `width` as well, would be nice to have that fix in there.
2. I'd replace all the comparison math by `min` and `max`-based rectangle intersection. If we are talking overflow, consider that `x + top_width` could overflow as well!

Suggestion how the function could also look:

```
range_width = x.saturating_add(top_width) // Non-overflow max coordinate calculation
    .min(bottom_width) // Limit to full width
    .saturating_sub(x); // Actual width
range_height = y.saturating_add(top_height)
    .min(bottom_height)
    .saturating_sub(y);
```

Note that, all non-augmented operations not in integer behaviour but mathematically speaking, due to the saturating operations we have:
* `x + range_width <= x.saturating_add(top_width).min(bottom_width) <= bottom_width`
* `x + range_width <= x.saturating_add(top_width).min(bottom_width) <= x.saturating_add(top_width)  <= x + top_width` and therefore `range_width <= top_width`
* similar for the height coordinates

This means that all our coordinates are provably in bounds.

HeroicKatora:
Thanks for the report and initial fix. Closed in favour of the more rigorous other solution.

PR #853, opened 2019-01-14
Unifies, fixes, checks and benchmarks `ImageBuffer` index calculation

### LINKED ISSUE #836: ImageBuffer has wrong index calculation
All buffer index calculations for accessing pixel samples in the internal buffer are done in `u32` for the most part:

https://github.com/PistonDevelopers/image/blob/50e29cda9470ddb579948c590c8b4b09e90e2863/src/buffer.rs#L375

The cast to `usize` is late in that case, it should happen before the multiplication. This leads to overflow–silent in release mode–when using a large image and large indices.

That means that for large images not all pixels are accessible, and the function `get_pixel_mut` is not injective, even for images with `width > 0 && height > 0 && P::channel_count() > 0`. See https://github.com/PistonDevelopers/imageproc/issues/281 for a scenario where that can turn out to facilitate undefined behaviour.

Also, contrary to the documentation, the function will not always panic when the indices were out of bounds of the image. Rather, it will only panic when the indices are out of bounds of the underlying buffer that might have been bigger (`ImageBuffer::from_raw`).

HeroicKatora:
Fixes #836 (and by extension addresses one of the issues in https://github.com/PistonDevelopers/imageproc/issues/281). I'll leave this open for a few more days or hours in case an additional idea for a performance improvement comes up (thanks @fintelia for the current calculation). Right now:

```
Before: test … bench:   1,541,207 ns/iter (+/- 224,258) = 1946 MB/s
After: test … bench:   1,759,194 ns/iter (+/- 62,646) = 1705 MB/s
```

That seems acceptable but maybe not yet optimal.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -371,9 +371,30 @@ where
     ///
     /// Panics if `(x, y)` is out of the bounds `(width, height)`.
     pub fn get_pixel(&self, x: u32, y: u32) -> &P {
+        let pixel_indices = self.pixel_indices(x, y)
+            .unwrap_or_else(|| panic!(
+                "Image index {:?} out of bounds {:?}", (x, y), (self.width, self.height),
+            ));
+        <P as Pixel>::from_slice(&self.data[pixel_indices])
+    }
+
+    #[inline(always)]
+    fn pixel_indices(&self, x: u32, y: u32) -> Option<std::ops::Range<usize>> {
         let no_channels = <P as Pixel>::channel_count() as usize;
-        let index = no_channels * (y * self.width + x) as usize;
-        <P as Pixel>::from_slice(&self.data[index..index + no_channels])
+        let max_index = Some(self.width as usize)
+            .and_then(|w| w.checked_mul(y as usize))
+            .and_then(|y| y.checked_add(x as usize))
+            .and_then(|x| x.checked_mul(no_channels))
+            .and_then(|i| i.checked_add(no_channels))?;
+        // Can't wrap.
+        let min_index = max_index.checked_sub(no_channels).unwrap();
+        Some(min_index..max_index)
+    }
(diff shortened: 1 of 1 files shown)


src/buffer.rs
```diff
@@ -786,4 +828,34 @@ mod test {
...
+            rgb[0] = 255;
+            rgb[1] = 23;
+            rgb[2] = 42;
+        }
+
+        b.iter(move || {
+            let image: &RgbImage = test::black_box(&a);
+            let mut sum: usize = 0;
+            for x in 0..1000 {
+                for y in 0..1000 {
```
fintelia: Swapping these two loops makes the access more cache friendly (and thus speeds up the benchmark) but I'm not sure if this way was intentional.

HeroicKatora: Me neither. Maybe we just benchmark both.

HeroicKatora: Without respect to your very likely different cache sizes, roughly double the speed. But the speed difference between previously and now does not seem to matter for row-by-row cache friendly patterns.

Previous:
```
test … bench:   1,560,805 ns/iter (+/- 117,600) = 1922 MB/s
test … bench:     971,162 ns/iter (+/- 42,709) = 3089 MB/s
```

Now:
```
test … bench:   1,770,711 ns/iter (+/- 63,445) = 1694 MB/s
test … bench:     982,404 ns/iter (+/- 45,258) = 3053 MB/s
```

HeroicKatora:
Ehm, this fails some tests? Note that tests are only executed for nightly on Travis.

src/buffer.rs
HeroicKatora: That means the performance regression only happens when you iterate the wrong way. Very nice!

PR #859, opened 2019-02-01
README minor: adding type for the compiler and deref for pixel

Isan-Rivkin:
Hi :), seems like the compiler cannot infer the type so I added a concrete type for Pixel in the ImageBuffer and * to deref as the function signature states. `pub fn put_pixel(&mut self, x: u32, y: u32, pixel: P)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,8 @@ extern crate image;
 
 use image::{GenericImage, ImageBuffer};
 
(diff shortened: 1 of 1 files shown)


bvssvni:
Merging.

bvssvni:
Thanks!

Issue #858, opened 2019-01-31
Incremental TIFF loading


birktj:
Currently `TIFFDecoder::Reader` is `Cursor<Vec<u8>>`. My use case is processing of huge images and would like too have incremental loading of images (one strip at a time). This is something I would love to contribute, but I just want to check that this is something would be appreciated. At first I would like to start out with tiff, but if successful then I would like to look into other formats.

fintelia:
Yes! This is definitely a feature we'd like to have but haven't yet had anyone with the expertise/time to implement it.

PR #860, opened 2019-02-01
Implemented incremental png loading

birktj:
This adds an incremental reader to the png type. The reader uses `png::Reader::next_row` to try to decode the png one row at a time.

This is a small breaking api change as `PNGDecoder::Reader` now becomes `PNGReader`.

There should probably be some new tests to check the reader with different buffers ect... but I am not entirely sure where they should go?

Related to #858

### DIFF
--- a/src/png.rs
+++ b/src/png.rs
@@ -10,10 +10,85 @@ extern crate png;
 
 use self::png::HasParameters;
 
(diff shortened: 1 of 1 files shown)


birktj:
@fintelia or @HeroicKatora could you have a look at this?

PR #863, opened 2019-02-10
Upgrade versions of some dependencies

### LINKED ISSUE #816: Panic in gif::Decoder::into_frames on invalid background color
This small program, along with this image: `invalid-bg-color.gif` - invalid-bg-color

```rust

extern crate image;

use image::ImageDecoder;

fn main() {
   
    let file = std::fs::File::open("invalid-bg-color.gif").unwrap();
    let decoder = image::gif::Decoder::new(std::io::BufReader::new(file));
    let _frames = decoder.into_frames().unwrap();
}
```
Produces the following crash:
```
   8: core::slice::slice_index_len_fail
             at libcore/slice/mod.rs:1971
   9: <core::ops::range::Range<usize> as core::slice::SliceIndex<[T]>>::index
             at /checkout/src/libcore/slice/mod.rs:2136
  10: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at /checkout/src/libcore/slice/mod.rs:1953
  11: <image::gif::Decoder<R> as image::image::ImageDecoder>::into_frames
             at /home/XXX/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.20.0/./src/gif.rs:139
  12: gif_test::main
             at src/main.rs:10
```

But opening the image with `image::load("invalid-bg-color.gif")` works correctly.

-----
The input image is a copy of [this image](https://github.com/PistonDevelopers/image-gif/blob/master/tests/samples/sample_1.gif), but with the background color set to the invalid value `0xFF`, which causes an out-of-bound access in the Global Color Table.

Some GIFs in the wild use such an invalid value when the background color is never used, so these files should be parsed correctly (even if this *technically* is an invalid GIF).

fintelia:
Fixes #816 and possibly others

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,38 +29,38 @@ name = "image"
 path = "./src/lib.rs"
 
 [dependencies]
-byteorder = "1.2.1"
-num-iter = "0.1.32"
+byteorder = "1.3.1"
+num-iter = "0.1.37"
 num-rational = { version = "0.2.1", default-features = false }
-num-traits = "0.2.0"
+num-traits = "0.2.6"
 lzw = "0.10.0"
 safe-transmute = "0.10.1"
 
 [dependencies.gif]
-version = "0.10.0"
+version = "0.10.1"
 optional = true
 
 [dependencies.jpeg-decoder]
-version = "0.1"
+version = "0.1.15"
 default-features = false
 optional = true
 
 [dependencies.png]
-version = "0.14"
+version = "0.14.0"
 optional = true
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
Cargo uses the latest compatible version automatically. If you compile `image` or a crate depending on it and inspect `Cargo.lock`, you will see that version 0.10.1 is what's actually used. Therefore, #816 us already fixed and this PR is not needed.

bvssvni:
Closing.

PR #864, opened 2019-02-10
Make Pixel structs more idiomatic

fintelia:
Right now we have both structs and functions named Rgb, Rgba, etc. There is no reason for this duplication, and it clutters up the crate documentation page.  This is a breaking change, though I'm not sure how much code will be broken in practice.

### DIFF
--- a/examples/fractal.rs
+++ b/examples/fractal.rs
@@ -35,7 +35,7 @@ fn main() {
             }
 
             let pixel = imgbuf.get_pixel_mut(x, y);
-            let data = (*pixel as image::Rgb<u8>).data;
+            let data = (*pixel as image::Rgb<u8>).0;
             *pixel = image::Rgb([data[0], i as u8, data[2]]);
         }
     }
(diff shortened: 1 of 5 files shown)


bvssvni:
I agree with this change.

bvssvni:
It might break a lot of code though, so I would like to hear what other people think.

fintelia:
Yeah, I'm not actually sure how much code will break. Though fixing any code should hopefully be quite manageable: I think it would mostly come down to changing .data to .0 or swapping braces for parenthesis. 

It should be fine if you have code like:
```rust
let p = Rgb(1, 2, 3); // Works fine
```
or even:
```rust
let make_pixel = Rgb;
let p = make_pixel(1, 2, 3); // Works fine
```

However it does break code that worked with the struct directly:
```rust
let p = Rgb { data: [1, 2, 3] }; // BROKEN!
```
or
```rust
let red = rgb.data[0]; // BROKEN!
```

fintelia:
@HeroicKatora or anyone else, any objections to merging this?

HeroicKatora:
No, but can you manually rerun tests? I remember some line involving `Rgb` in the `unsafe` patches, so maybe something changed from that.

HeroicKatora:
@fintelia Can you rebase this onto current master? There is a new instance that was introduced into `src/hdr/hdr_decoder.rs:57:27`. After that, I plan to merge this.

PR #865, opened 2019-02-10
Refactor the ColorType enum

### LINKED ISSUE #855: Finalize ColorType enum
The [ColorType](https://docs.rs/image/0.21.0/image/enum.ColorType.html) enum is used to represent the meaning of subcomponents of Pixels and Decoders. However it has a few issues:
* Many values of it are completely ridiculous and almost surely never used in practice. Why would anyone have an image with 157 bits per pixel BGR?
* Parts of the library are unimplemented for certain color types. At very least it should be possible to convert between any pair of color types (filling any missing values with some reasonable default).
* Adding enum variants would be a breaking change to the library, yet some relatively useful/common formats (like floating point RGBA) are missing. This needs to be resolved before a 1.0 release.
 
One option is to replace the enum with the something similar to the following (based on the supported modes in the [Pillow library](https://pillow.readthedocs.io/en/latest/handbook/concepts.html#modes)):
```rust
enum ColorType {
    // 8-bits per channel
    Palette,
    Luminance,
    LuminanceAlpha,
    RGB,
    RGBA,

    // Reversed channel order
    BGR,
    BGRA,

    // Exotic single channel bit depths
    U1,
    I32,
    F32,

    // High bit depth color images
    RGBA16,
    RGBA32F,
}
```

### Questions
[ ] Do we want to include BGR / BGRA formats?
[ ] Should we include U1 (packed 1 bit per pixel, 8 pixels per byte)?
[ ] Should we define a smaller set of "core" types, say Luminance + RGB + RGBA + RGBA32F, and then focus on conversions to and between them?

fintelia:
Fixes #855

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -37,6 +37,9 @@ pub enum ColorType {
     /// are associated with an external palette. In that case, the pixel value is an index into the
     /// palette.
     Unknown(u8),
+
+    #[doc(hidden)]
+    __Nonexhaustive,
 }
 
 /// Returns the number of bits contained in a pixel of `ColorType` ```c```
@@ -51,6 +54,7 @@ pub fn bits_per_pixel(c: ColorType) -> usize {
         ColorType::RGB16 => 48,
         ColorType::RGBA16 => 64,
         ColorType::Unknown(n) => n as usize,
+        ColorType::__Nonexhaustive => unreachable!(),
     }
 }
 
@@ -61,6 +65,7 @@ pub fn num_components(c: ColorType) -> usize {
         ColorType::LA | ColorType::LA16 => 2,
         ColorType::RGB | ColorType::RGB16| ColorType::BGR => 3,
         ColorType::RGBA | ColorType::RGBA16 | ColorType::BGRA => 4,
+        ColorType::__Nonexhaustive => unreachable!(),
     }
 }
 
--- a/src/color.rs
+++ b/src/color.rs
@@ -7,9 +7,6 @@ use traits::Primitive;
 /// An enumeration over supported color types and bit depths
 #[derive(Copy, PartialEq, Eq, Debug, Clone, Hash)]
 pub enum ColorType {
-    /// Pixel is an index into a color palette
-    Palette(u8),
-
     /// Pixel is 1-bit luminance
     L1,
 
@@ -35,6 +32,11 @@ pub enum ColorType {
     BGR,
     /// Pixel is 8-bit BGR with an alpha channel
     BGRA,
+
+    /// Pixel is of unknown color type with the specified bit depth. This can apply to pixels which
+    /// are associated with an external palette. In that case, the pixel value is an index into the
+    /// palette.
+    Unknown(u8),
 }
 
 /// Returns the number of bits contained in a pixel of `ColorType` ```c```
@@ -43,21 +45,21 @@ pub fn bits_per_pixel(c: ColorType) -> usize {
         ColorType::L1 => 1,
         ColorType::L8 => 8,
         ColorType::L16 => 16,
-        ColorType::Palette(n) => 3 * n as usize,
         ColorType::LA => 16,
         ColorType::RGB | ColorType::BGR => 24,
         ColorType::RGBA | ColorType::BGRA | ColorType::LA16 => 32,
         ColorType::RGB16 => 48,
         ColorType::RGBA16 => 64,
+        ColorType::Unknown(n) => n as usize,
     }
 }
 
 /// Returns the number of color channels that make up this pixel
 pub fn num_components(c: ColorType) -> usize {
     match c {
-        ColorType::L1 | ColorType::L8 | ColorType::L16 => 1,
+        ColorType::Unknown(_) | ColorType::L1 | ColorType::L8 | ColorType::L16 => 1,
         ColorType::LA | ColorType::LA16 => 2,
-        ColorType::RGB | ColorType::RGB16 | ColorType::Palette(_) | ColorType::BGR => 3,
+        ColorType::RGB | ColorType::RGB16| ColorType::BGR => 3,
         ColorType::RGBA | ColorType::RGBA16 | ColorType::BGRA => 4,
     }
 }
--- a/src/color.rs
+++ b/src/color.rs
@@ -33,9 +33,9 @@ pub enum ColorType {
     /// Pixel is 8-bit BGR with an alpha channel
     BGRA,
 
-    /// Pixel is of unknown color type with the specified bit depth. This can apply to pixels which
-    /// are associated with an external palette. In that case, the pixel value is an index into the
-    /// palette.
+    /// Pixel is of unknown color type with the specified bits per pixel. This can apply to pixels
+    /// which are associated with an external palette. In that case, the pixel value is an index
+    /// into the palette.
     Unknown(u8),
 
     #[doc(hidden)]
--- a/src/color.rs
+++ b/src/color.rs
@@ -1,54 +1,64 @@
 use num_traits::{NumCast, Zero};
-use std::mem;
 use std::ops::{Index, IndexMut};
 
 use buffer::Pixel;
 use traits::Primitive;
 
-/// An enumeration over supported color types and their bit depths
+/// An enumeration over supported color types and bit depths
 #[derive(Copy, PartialEq, Eq, Debug, Clone, Hash)]
 pub enum ColorType {
-    /// Pixel is grayscale
-    Gray(u8),
-
-    /// Pixel contains R, G and B channels
-    RGB(u8),
-
     /// Pixel is an index into a color palette
     Palette(u8),
 
-    /// Pixel is grayscale with an alpha channel
-    GrayA(u8),
+    /// Pixel is 1-bit luminance
+    L1,
 
-    /// Pixel is RGB with an alpha channel
-    RGBA(u8),
+    /// Pixel is 8-bit luminance
+    L8,
+    /// Pixel is 8-bit luminance with an alpha channel
+    LA,
+    /// Pixel contains 8-bit R, G and B channels
+    RGB,
+    /// Pixel is 8-bit RGB with an alpha channel
+    RGBA,
 
-    /// Pixel contains B, G and R channels
-    BGR(u8),
+    /// Pixel is 16-bit luminance
+    L16,
+    /// Pixel is 16-bit luminance with an alpha channel
+    LA16,
+    /// Pixel is 16-bit RGB
+    RGB16,
+    /// Pixel is 16-bit RGBA
+    RGBA16,
 
-    /// Pixel is BGR with an alpha channel
-    BGRA(u8),
-
+    /// Pixel contains 8-bit B, G and R channels
+    BGR,
+    /// Pixel is 8-bit BGR with an alpha channel
+    BGRA,
 }
 
 /// Returns the number of bits contained in a pixel of `ColorType` ```c```
 pub fn bits_per_pixel(c: ColorType) -> usize {
     match c {
-        ColorType::Gray(n) => n as usize,
-        ColorType::GrayA(n) => 2 * n as usize,
-        ColorType::RGB(n) | ColorType::Palette(n)| ColorType::BGR(n) => 3 * n as usize,
-        ColorType::RGBA(n) | ColorType::BGRA(n) => 4 * n as usize,
+        ColorType::L1 => 1,
+        ColorType::L8 => 8,
+        ColorType::L16 => 16,
+        ColorType::Palette(n) => 3 * n as usize,
+        ColorType::LA => 16,
+        ColorType::RGB | ColorType::BGR => 24,
+        ColorType::RGBA | ColorType::BGRA | ColorType::LA16 => 32,
+        ColorType::RGB16 => 48,
+        ColorType::RGBA16 => 64,
     }
 }
 
 /// Returns the number of color channels that make up this pixel
 pub fn num_components(c: ColorType) -> usize {
     match c {
-        ColorType::Gray(_) => 1,
-        ColorType::GrayA(_) => 2,
-        ColorType::RGB(_) | ColorType::Palette(_) | ColorType::BGR(_)=> 3,
-        ColorType::RGBA(_) | ColorType::BGRA(_) => 4,
-
+        ColorType::L1 | ColorType::L8 | ColorType::L16 => 1,
+        ColorType::LA | ColorType::LA16 => 2,
+        ColorType::RGB | ColorType::RGB16 | ColorType::Palette(_) | ColorType::BGR => 3,
+        ColorType::RGBA | ColorType::RGBA16 | ColorType::BGRA => 4,
     }
 }
 
@@ -58,7 +68,7 @@ macro_rules! define_colors {
         $channels: expr,
         $alphas: expr,
         $interpretation: expr,
-        $color_type: ident,
+        $color_type: expr,
         #[$doc:meta];
     )*} => {
 
@@ -71,14 +81,13 @@ $( // START Structure definitions
 pub struct $ident<T: Primitive> (pub [T; $channels]);
 
 impl<T: Primitive + 'static> Pixel for $ident<T> {
-
     type Subpixel = T;
 
     const CHANNEL_COUNT: u8 = $channels;
 
     const COLOR_MODEL: &'static str = $interpretation;
 
-    const COLOR_TYPE: ColorType = ColorType::$color_type(mem::size_of::<T>() as u8 * 8); 
+    const COLOR_TYPE: ColorType = $color_type;
 
     #[inline(always)]
     fn channels(&self) -> &[T] {
@@ -216,12 +225,12 @@ impl<T: Primitive> IndexMut<usize> for $ident<T> {
 }
 
 define_colors! {
-    Rgb, 3, 0, "RGB", RGB, #[doc = "RGB colors"];
-    Bgr, 3, 0, "BGR", BGR, #[doc = "BGR colors"];
-    Luma, 1, 0, "Y", Gray, #[doc = "Grayscale colors"];
-    Rgba, 4, 1, "RGBA", RGBA, #[doc = "RGB colors + alpha channel"];
-    Bgra, 4, 1, "BGRA", BGRA, #[doc = "BGR colors + alpha channel"];
-    LumaA, 2, 1, "YA", GrayA, #[doc = "Grayscale colors + alpha channel"];
+    Rgb, 3, 0, "RGB", ColorType::RGB, #[doc = "RGB colors"];
+    Bgr, 3, 0, "BGR", ColorType::BGR, #[doc = "BGR colors"];
+    Luma, 1, 0, "Y", ColorType::L8, #[doc = "Grayscale colors"];
+    Rgba, 4, 1, "RGBA", ColorType::RGBA, #[doc = "RGB colors + alpha channel"];
+    Bgra, 4, 1, "BGRA", ColorType::BGRA, #[doc = "BGR colors + alpha channel"];
+    LumaA, 2, 1, "YA", ColorType::LA, #[doc = "Grayscale colors + alpha channel"];
 }
 
 /// Provides color conversions for the different pixel types.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,7 +25,7 @@ extern crate quickcheck;
 
 use std::io::Write;
 
-pub use color::ColorType::{self, Gray, GrayA, Palette, RGB, RGBA, BGR, BGRA};
+pub use color::ColorType;
 
 pub use color::{Luma, LumaA, Rgb, Rgba, Bgr, Bgra};
 
--- a/src/png.rs
+++ b/src/png.rs
@@ -166,13 +166,18 @@ impl<W: Write> PNGEncoder<W> {
 impl From<(png::ColorType, png::BitDepth)> for ColorType {
     fn from((ct, bits): (png::ColorType, png::BitDepth)) -> ColorType {
         use self::png::ColorType::*;
-        let bits = bits as u8;
-        match ct {
-            Grayscale => ColorType::Gray(bits),
-            RGB => ColorType::RGB(bits),
-            Indexed => ColorType::Palette(bits),
-            GrayscaleAlpha => ColorType::GrayA(bits),
-            RGBA => ColorType::RGBA(bits),
+        match (ct, bits as u8) {
+            (Indexed, bits) => ColorType::Palette(bits),
+            (Grayscale, 1) => ColorType::L1,
+            (Grayscale, 8) => ColorType::L8,
+            (Grayscale, 16) => ColorType::L16,
+            (GrayscaleAlpha, 8) => ColorType::LA,
+            (GrayscaleAlpha, 16) => ColorType::LA16,
+            (RGB, 8) => ColorType::RGB,
+            (RGB, 16) => ColorType::RGB16,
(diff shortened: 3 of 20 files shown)


HeroicKatora:
I wouldn't remove Palette completely. It's basically the archetype of a color type that has no inherent color space information, which is definitely useful exactly for the case where the palette is not supplied by the library but the user. And `png::Indexed` is definitely **reachable** in the language sense.

I'm in favour of removing the `u8` argument from RGB's but it seems rash to not introduce 16-bit colors that are already used, e.g. in `pnm`.

src/lib.rs
```diff
@@ -24,7 +24,7 @@ extern crate quickcheck;
 
 use std::io::Write;
 
-pub use color::ColorType::{self, Gray, GrayA, Palette, RGB, RGBA, BGR, BGRA};
```
fintelia: Not sure if this is the right move. It isn't really idiomatic to `pub use` a bunch of enum variants, but removing them could unnecessarily break people's code.

src/color.rs
```diff
@@ -1,54 +1,64 @@
...
 pub fn bits_per_pixel(c: ColorType) -> usize {
     match c {
-        ColorType::Gray(n) => n as usize,
-        ColorType::GrayA(n) => 2 * n as usize,
-        ColorType::RGB(n) | ColorType::Palette(n)| ColorType::BGR(n) => 3 * n as usize,
-        ColorType::RGBA(n) | ColorType::BGRA(n) => 4 * n as usize,
+        ColorType::L1 => 1,
+        ColorType::L8 => 8,
+        ColorType::L16 => 16,
+        ColorType::Palette(n) => 3 * n as usize,
```
fintelia: This seems really suspect to me, but it is how the code currently works.

src/color.rs
```diff
@@ -1,54 +1,64 @@
...
+        ColorType::RGBA16 => 64,
     }
 }
 
 /// Returns the number of color channels that make up this pixel
 pub fn num_components(c: ColorType) -> usize {
     match c {
-        ColorType::Gray(_) => 1,
-        ColorType::GrayA(_) => 2,
-        ColorType::RGB(_) | ColorType::Palette(_) | ColorType::BGR(_)=> 3,
```
fintelia: Seems like Palette should have only a single component, but didn't want to break this without asking

HeroicKatora: Mhm, yes, seems mostly reasonable if it is opaque anyways. But I'm not quite sure myself.

src/lib.rs
HeroicKatora: `libstd` does the same by using `Option::*` so that `Some()` and `None` works. I'd say it's idiomatic enough where useful and unambiguous.

fintelia:
@HeroicKatora I've added back Palette (now called Unknown) and included 16 bit colors. However, I wasn't quite sure which luminance color types to include. L2 and L4 seemed too obscure so I left them out, but I could be persuaded otherwise. Conversely, I'm not so sure how common L1 is and if we removed it then all of the well defined color types would be an integer number of bytes. I'm also open to bikeshedding variant names if you (or anyone else) has suggestions.

src/lib.rs
fintelia: The prelude is a bit different though isn't it? The Some and None variants are automatically in scope whereas bringing bring in these variants would typically involve either `use image::{RGB, RGBA, ...}` or `use image::ColorType::*` which are basically the same (people generally don't pull in entire crates with a wildcard, with the exception of special "prelude" submodules).

HeroicKatora: Fair points, and it does clobber the documentation a little.

birktj:
Would it be an idea to have `Colortype::Unknown(u8, u8)`, where the first argument is number of components, and the other is bitdepth per component? It could then be used for cmyk and other potentially exotic colortypes.

HeroicKatora:
The problem with arbitrary bit depths and components counts is that they only appear supported by the interfaces consuming `ColorType`, but in reality are not and will almost always just panic. The idea is to restrict `ColorTypes` to those which can be support in and implementation of `Pixel`, e.g. to utilize it in `ImageBuffer`. That said, cmyk for example would fit this perfectly fine with 4 channels.

Decoders potentially have the option of decoding the raw bytes, and signaling another color type in their own api.

birktj:
Would it be better to not use `From`and `Into` for conversions to/from colortypes of decoders given that many of them can produce colortypes that per now fall under `unimplemented!` clauses?

Instead there could be simple functions like `fn convert_from_png_colortype(colortype: png::Colortype) -> ImageResult<Colortype>` and have decoders fail  with a `UnsupportedColortype` on open if you try to open a file with a colortype not under `ColorType`.

fintelia:
I'd really like to split out a `image-core` crate that could be used by all the decoder crates and thus eliminate png::ColorType entirely. Right now that code branch is actually unreachable unless you pass an (undocumented) ColorType::__Nonexaustive which you really shouldn't create in the first place...

birktj:
I agree with the `image-core` idea, but i know that at least `image-tiff` has a few colortypes that the decoder can return that currently goes into an `unimplemented!` clause when converting to `ColorType`. You could thus force a panic simply by giving a normal tiff file to any application using this crate.

aschampion:
Particular formats will almost always have niche color types that will not map to a sensible common set, so it's fine for them to have their own enums for users directly using those format crates interested in those niche types.

I am also in favor of an `image-core` so that format crates can define how and which of their internal color types map to the common set, but I don't think that necessarily means removing use of the format-specific color type in format crates. I think this is also what @birktj is suggesting.

Re: `L1` -- it's not uncommon in image processing and computer vision. I occasionally run into `L2`/`L4`, but am already in the habit of pre-converting them to `L8` because of rare support in most software and libraries. So :+1: including `L1` but omitting `L2`/`L4`.

32bpc (float) RGB (96bpp) and RGBA (128bpp) are around, too, but until we have JPEG XR or TIFF implementations that support them they can be left out.

fintelia:
@HeroicKatora @birktj how would you feel about just merging this? This PR doesn't have any story around color type conversion or format capability detection, but those can be added later, and I do think that this is a clear improvement from the current state of the ColorType enum.

I've also just addressed @birktj's concern about panicking when converting foreign crate's color types to image::ColorType. Now we just map any types to ColorType::Unknown with the proper bit depth.

fintelia:
If possible, I'd like to include this in the `0.22` release so that we don't have to do it in a separate breaking release

birktj:
I think we can merge this, I don't think all the questions around `DynamicImage`, `ColorType`, `ImageDecoder` and etc. are answered yet, but I think this PR is definitely an improvement over the current situation and a step in the right direction. We can have more discussions at a later stage.

src/pnm/encoder.rs
```diff
@@ -326,43 +329,38 @@ impl<'a> CheckedDimensions<'a> {
...
                 decoded: HeaderRecord::Bitmap(_),
                 ..
             } => match color {
-                ColorType::Gray(_) => (),
+                ColorType::L1 | ColorType::L8 | ColorType::L16 => (),
                 _ => {
                     return Err(io::Error::new(
                         io::ErrorKind::InvalidInput,
-                        "PBM format only support ColorType::Gray",
+                        "PBM format only support ColorType::L",
```
HeroicKatora: That's a rather odd way to put it. `::{L1, L8, L16}` would be clearer.

src/pnm/encoder.rs
```diff
@@ -326,43 +329,38 @@ impl<'a> CheckedDimensions<'a> {
...
                 decoded: HeaderRecord::Graymap(_),
                 ..
             } => match color {
-                ColorType::Gray(_) => (),
+                ColorType::L1 | ColorType::L8 | ColorType::L16 => (),
                 _ => {
                     return Err(io::Error::new(
                         io::ErrorKind::InvalidInput,
-                        "PGM format only support ColorType::Gray",
+                        "PGM format only support ColorType::L",
```
HeroicKatora: Same here.

src/tiff.rs
```diff
@@ -56,12 +56,21 @@ impl From<tiff::TiffError> for ImageError {
 impl From<tiff::ColorType> for ColorType {
     fn from(ct: tiff::ColorType) -> ColorType {
         match ct {
-            tiff::ColorType::Gray(depth) => ColorType::Gray(depth),
-            tiff::ColorType::RGB(depth) => ColorType::RGB(depth),
-            tiff::ColorType::Palette(depth) => ColorType::Palette(depth),
-            tiff::ColorType::GrayA(depth) => ColorType::GrayA(depth),
-            tiff::ColorType::RGBA(depth) => ColorType::RGBA(depth),
-            tiff::ColorType::CMYK(_) => unimplemented!()
```
HeroicKatora: Ok yeah, this line is *the* primary reason to get it through. Even we seemed confused about what to support and how.

src/pnm/encoder.rs
```diff
@@ -636,9 +637,9 @@ mod tests {
         let data: [u8; 12] = [0, 0, 0, 1, 1, 1, 255, 255, 255, 0, 0, 0];
 
         let header = ArbitraryHeader {
-            width: 2,
-            height: 2,
-            depth: 3,
+            width: 3,
+            height: 4,
+            depth: 1,
```
HeroicKatora: Does that mean `depth: 3` no longer passes? I would have expected `ColorType::Unknown(24)` or something but still working. Or with an extra count for the number of color planes, note that `depth` in the context of `pam` refers actually to the number of color planes.. Weird but that could keep it working.

src/color.rs
```diff
@@ -1,54 +1,71 @@
...
+
+    /// Pixel contains 8-bit B, G and R channels
+    BGR,
+    /// Pixel is 8-bit BGR with an alpha channel
+    BGRA,
+
+    /// Pixel is of unknown color type with the specified bit depth. This can apply to pixels which
+    /// are associated with an external palette. In that case, the pixel value is an index into the
+    /// palette.
+    Unknown(u8),
```
HeroicKatora: I'm not sure if this is a sufficient 'catch all' type for what it looks like it is trying to achieve. On the one hand it is still too open to have somewhat guaranteed support (no encoder in its right mind has 255-bit depth support) but too restricted to abstract color planes. I don't think we should restrict the representation here, `Error::Unsupported` can be returned where necessary. So I'd rather embrace it as (or similar):

```
Unknown {
    planes: u8,
    bits_per_plane: u8,
},
```

src/color.rs
```diff
@@ -1,54 +1,71 @@
...
 
 /// Returns the number of color channels that make up this pixel
 pub fn num_components(c: ColorType) -> usize {
     match c {
-        ColorType::Gray(_) => 1,
-        ColorType::GrayA(_) => 2,
-        ColorType::RGB(_) | ColorType::Palette(_) | ColorType::BGR(_)=> 3,
-        ColorType::RGBA(_) | ColorType::BGRA(_) => 4,
-
+        ColorType::Unknown(_) | ColorType::L1 | ColorType::L8 | ColorType::L16 => 1,
```
HeroicKatora: Note how `Palette` had previously returned `3` but now whatever maps to `Unknown` will return `1`.

src/png.rs
```diff
@@ -181,13 +189,19 @@ impl From<ColorType> for (png::ColorType, png::BitDepth) {
...
         use self::png::ColorType::*;
         let (ct, bits) = match ct {
-            ColorType::Gray(bits) => (Grayscale, bits),
-            ColorType::RGB(bits) => (RGB, bits),
-            ColorType::Palette(bits) => (Indexed, bits),
-            ColorType::GrayA(bits) => (GrayscaleAlpha, bits),
-            ColorType::RGBA(bits) => (RGBA, bits),
-            ColorType::BGRA(bits) => (RGBA, bits),
-            ColorType::BGR(bits) => (RGB, bits),
+            ColorType::Unknown(bits) => (Indexed, bits),
```
HeroicKatora: This is not the reverse of the `From` conversion above. Maybe we should avoid a direction one-to-one conversion for now.

src/pnm/encoder.rs
birktj: Just saying "luma colortypes" is maybe simpler?

birktj: I mean since I guess errors like this can end up being user-facing in some applications, and thus referencing the specific enum  could look sort of strange in some cases.

src/pnm/encoder.rs
fintelia: This is a consequence of the `num_channels(ColorType::Unknown(..)) = 1` change from above. I'm really not sure why we support non-standard encodings like this, but if we want to get the old behavior back the best way would just be to tell the encoder that the input data has 3 8-bit channels by passing `ColorType::RGB`. The ArbitraryTuplType::Custom argument already causes the output header from including the true input color type.

src/png.rs
HeroicKatora: `From` should *never* panic. But we might want to just not give a `From` implementation for that reverse case anyhow.

src/pnm/encoder.rs
HeroicKatora: That's because `pnm` completely separate 'color tuple type' and 'channel count/depth' information. The `ArbitraryTuplType` only sets the former and the encoder only ensures that that information is consistent with the `ColorType` for non-`Custom` variants.

HeroicKatora: But that also means that including the true input color type is only possible if setting arbitrary color channel count and bit-depth information is also available. I had previously used `Palette` to at least allow a count of `3`. It's not strictly necessary.

src/color.rs
HeroicKatora: But as a pure escape hatch with out-of-band information it's kind of weird. The decoder needs to return byte data. The maximum flexibility would then be it returning simply how many bytes-per-pixel are required since it can not expect the further processing steps to have any idea about the correct transformation for the color information. Thus `bit depth` is almost meaningless to the environment.

What happens if we just drop the variant entirely? Decoders still have the option of returning an `Unsupported` error for some sort.

birktj: I am also in favor of dropping `Unknown` and replacing it with some sort of `Unsupported` error, and simply remove all the `From` impls that can fail. We have the argument that `Unknown` really cannot be used to do anything which is pretty strong. 

A simple way to do this could probably be with a `#[doc(hidden)]` enum variant `__Unknown`. Since the enum already is unexhaustive any place where it is used needs some sort of wildcards which should return some `Err(Unsupported(...))`.

birktj: Maybe that was not such a good idea, it would make the enum quite a bit more painful to use.

src/png.rs
fintelia: I've just removed this From implementation and now return an error if an unsupported ColorType is passed to encode. Really that function should return an ImageResult instead of an io::Result but that is a much deeper rabbit hole than I wanted to get into for this PR.

src/pnm/encoder.rs
fintelia: The current code checks that `depth == num_components(color)` even for Custom variants, which just happened to work because Palette previously was declared to have 3 color channels. That might be fixable but I'd like to hold off making more intrusive changes to the pnm decoder for now

src/color.rs
fintelia: Would it be clearer if the description was changed from `bit depth` to `bits per pixel`? This is the meaning I've been intending. If we removed the case entirely, we'd have a worse regression until proper decoding of RGBA2 and similar formats are supported.

HeroicKatora: Maybe `bytes per pixel` would be a feasible middle ground. All internal loading post-processing expands pixels to byte boundaries, i.e. `L1` is turned to `u8`. But for arbitrary `bits` there is no clear process to do this.

fintelia: I'm not sure all decoders do the post processing you are describing. From my understanding, the PNG decoder doesn't do any expansion for BitDepth::{One, Two, Four} and so as a result RGB images will have bits per pixels that aren't multiples of 8.

fintelia: If that conversion is actually happening that is a bigger bug: we're claiming the returned image data will be in one format and then providing it in a different one

HeroicKatora: Afaik, it does: https://github.com/image-rs/image-png/blob/4f2e8eb44586442133c0222e47e1eb762aaef5a5/src/decoder/mod.rs#L304-L311 and https://github.com/image-rs/image-png/blob/4f2e8eb44586442133c0222e47e1eb762aaef5a5/src/decoder/mod.rs#L349-L355 and `Transformations::EXPAND` is set by default. 

I'm not sure if I consider that a bug by the way. The format is that of the *returned buffer* which is `L8` since we don't support others. It doesn't seem too weird to me that you don't get the genuine format of the *image* itself through that same interface, that seems by definition like it exceeds the possibilities of the limited number of enum variants.

fintelia: Does this mean that L1 is also impossible to get from decoders? If we can guarantee that pixels will always be an integer number of bytes then I think it does make sense to have Unknown bytes_per_pixel and to remove L1.

Regarding 2-bit RGB, 4-bit palette, and such the PNG crate actually has enums for those so I (wrongly?) assumed they were supported not just for input but also for the format of returned buffers. If we can just match over the results of png::Reader::output_color_type() and panic unimplemented!() on any sub 8-bit formats that would make the code much nicer.

HeroicKatora: This really boils down to what `ColorType` is supposed to represent, as according to the discussion in #855 after is RFC was opened. Either

* Collection of known color formats in the wild (basically what it was previously but less wildly unstructured).
* The color formats from which we can always convert to one of the formats of images. Then `Unknown` doesn't quite fit except if we introduce an `Opaque` pixel data representation. But `L4` should then be surely included as well.
* The color formats that the image buffers can *actually* have, in particular those on which generic algorithms can operate. I think that's too restrictive.

One design goal could be: Be liberal in what to expect and restrictive in what to emit. That would point to create *two* color types namely one for the each of the last two options. Sounds like a hassle but that might actually solve a few problems. There is no buffer representation utilizing `L4` but a roundtrip of `png` encoded with it should definitely be possible. Two different color types would make it possible to store the original encoding independent of the color type of the buffer and offer an interface that accepts the former while providing a default if only the latter is provided (as in `DynamicImage::write_to`).

fintelia: The `ColorType` is currently used to mean (and I'd argue should continue to mean):

* The set of color formats that decoders can chose for their output data or that users can given to encoders as input. Thus `L4` and similar aren't needed since the decoder can always "decode" that to `L8` (and that is easier for the user because they don't have to deal with weird formats)

In #873 I previously tried to create a separate enum for (3), but discovered that it would be used almost nowhere since both encoders and decoders want a less restricted set of formats and the Pixel trait / DynamicImage control which formats generic algorithms operate over.

I don't have strong feelings on `Unknown` vs `Opaque`, they mean roughly the same thing to me.

HeroicKatora: Your definition seems phrases somewhere between my `(2)` and `(3)` but I fail to see the distinction between requiring the decoder to already convert to `L8` and requiring it to convert to the set of colors that `DynamicImage` is supposed to represent. I.e it sounds exactly like (3) but the paragraph below says that it isn't supposed to.

fintelia: `DynamicImage` doesn't support ColorType::{L, LA, RGB, RGBA}16, though I suppose that may also change at some point. For now I see two main directions forward:

1. We can merge things as they are now, perhaps changing Unknown -> Opaque first.

2. We can remove L1, make Unknown/Opaque take *bytes* per pixel, and require at all future colortypes have pixels which are an integer number of bytes. Any decoders which produce output less than 8 bpp would be converted to return an error (and then a future PR could do an expand pass instead).

HeroicKatora: One thing to consider with regards to `Unknown`: due to the reference semantics we will have to require that any `impl Pixel` type is an integer number of bytes in any case (meaning `L1` will *never* be represented as an `image::Pixel` anytime soon). Since unknown can not be converted (including 'scaling') it *can never* be stored in a `GenericImage`-like buffer if a pixel is not composed of a whole number of bytes. I don't think it would be prospective to suggest otherwise.

Why I think it should be two types: The primary concern of `DynamicImage` is to ensure that the data is represented in some `Pixel` implementation (to allow it to implement `GenericImage`). However `L1` and `L4` and several other formats can not be a pixel as it conflates two samples in a single byte and that won't work with references. That means, in particular, that `ColorType` does not seem to be intended as a discriminant of a `Pixel`-type enum. On an image with such a source color type there has to, necessarily, be some post-processing.

I agree it would be nice however *not* to lose that information. And I also think that for some color types the transformation should be done by this library when such post-processing has a clearly defined meaning. That avoids redundancy and makes decoders easier to write. *Be lenient in what to accept*. But then why stop at `L1` and not allow `L4`? The conversion is reasonably clear in both directions. What about `Yuv420`? Wouldn't it be nice to have the `image` library automatically duplicate the `cr` channel into each pixel's sample?

As laid out above, it would also be inconvenient to have `Pixel::color_type` defined in terms of that same enum due to the integer-byte-count etc. In fact, the mapping of a `ColorType` to representations in pixel buffers would be non-injective (and thus can not be reversed). And due to L16 as observed by you it might error. This question would be avoided when having two enums. One for the original output of the decoder and one for the resulting representation as a `GenericImage`. Perhaps name one `ColorType` and one `PixelType`.

It would also (possibly) solve the problem for encoders. Since it is encouraged not to do the preprocessing on its own the input color format information is preserved – only the image buffer into which it is loaded doesn't have the information on its own but it can be queried from the decoder beforehand in standardized types. Then, the encoder can be supplied with both during encoding or the data be pre-processed to that 

So, about the resolution of this. If we split this into `ColorType` and `PixelType` we have no more problems. The `DynamicImage` code would try to find a `PixelType` that the `ColorType` returned by the decoder can be converted into, then do that conversion if possible. In reverse, the `write_to` implementation would find a default `ColorType` for the `PixelType` (possibly being wasteful compared to the original encoding) but another, new api could offer supplying a `ColorType` directly and convert the data back before feeding it into the encoder. And with regards to `ColorType::Unknown(bytes)`; the `DynamicImage` loading would fail with an error of `Unsupported` but reading directly would still work. This seems to be the most flexible option where required. We could then also expand on the variants of `ColorType` to offer more decoders with zero-loss of information while maintaining the `PixelType` of their output images (e.g. introduce `Rgb555` but the output image would still be `RGB8`).

HeroicKatora: Tl;dr: Basically option 1. but expanded right away to not lose functionality?

fintelia: That approach seems reasonable. Any opposition to merging this now and then adding `PixelType` in a follow up?

HeroicKatora:
@fintelia Also want to add that it's quite draining to rework such a fundamental aspect of the library, having to avoid loss of functionality without expanding the scope of the rework right and left. I think the biggest problem is that merging as-is would likely require *another* breaking change if expanded afterwards. If we are to do that major release soon we'd knowingly break functionality and commit to a forseeable future breaking change as we would likely not have time to replace everything that's gone missing until that release. Thanks for the hard work :heart:

fintelia:
Yeah, this sort of large change is very hard to pull off, especially considering there are so many alternatives all with their own tradeoffs. Repeated breaking changes are unfortunate but I think unavoidable if we want to get a clean result. There's just a huge (and growing) API surface much of it written before idiomatic Rust was even established. My hope is that a couple of iterations of these big breaking changes can get the library in good enough shape to finally do a 1.0 release.

Thanks for all your time reviewing.

HeroicKatora:
We could also do the following, since I'm convinced that the changes *are* necessary regardless of their final form and that this is already a good start.

1. Release one version without these changes as `0.22`
2. Additionally, also release a version `0.23.0-preview.0` based off a new branch where this is merged.
3. Incrementally work on feature-parity in the new branch while maintaining the other changes on `0.22`.

Issue #854, opened 2019-01-15
Fixing `MutPixels`


HeroicKatora:
# Motivation

As we well know, `PixelsMut` is currently not strictly well-defined. It works out just okay for all internal types but not for external ones. I propose improving this by improving its interface. Note that the standard library offers [`std::slice::ChunksMut`](https://doc.rust-lang.org/std/slice/struct.ChunksMut.html) or alternative the stricter `std::slice::ExactChunksMut` from `1.31.0` onwards. We already make use of it in `ImageBuffer::pixels_mut`. So, instead of just giving up on safely implementing this, why not move this burden on implementors of `GenericImage` through boxing but offer a fast-path for the most generic cases of matrix strides?

So, instead of making `MutPixels` a magic type without any interface, offer constructors for the special, blessed cases and an abstract `Box<Iterator<Item=(u32, u32, &mut Self::Pixel)> + 'a>` as an escape hatch. Once we discover additional requirements, we can enlarge the set of special cases if we keep this internal. Note that by `fn get_pixel_mut(&mut self, …) -> &mut Self::Pixel` we *already* require that pixels are actually backed by memory within the `GenericImage` impl. So in the worst case a really weird implementation could box a `&mut self` to make iteration work?

We can also provide an `unsafe` constructor exactly what the current default implementation does. This way, all unsafety is explicitely opted-in by creators without much overhead over the current situation if appropriate.

# Implementation

Remove the default impl and somehow make this interface work, optimizing for all necessary impls in this library and the Piston project:

```
pub struct MutPixels<'a, I: GenericImage + ?Sized + 'a> {
    /// Make pixels from row-major subpixels.
    ///
    /// # Panics
    ///
    /// When the slice is too short.
    pub fn row_major_chunks(slice: &'a mut [I::Pixel::Subpixel], width: u32, height: u32) -> Self;

    /// Maybe we want this?
    pub fn col_major_chunks(slice: &'a mut [I::Pixel::Subpixel], width: u32, height: u32) -> Self;

    /// Fallback if there is no blessed implementation.
    pub fn from_box(fallback: Box<…>) -> Self;

    /// Reverse compatible replacement, unsafe is now the burden of the implementor
    ///
    /// Move all the doc and requirements that usually apply to the reasoning why 
    /// `slice::iter_mut` is safe to here, to explain this method.
    unsafe pub fn unsafe_wrap_image(image: &mut I) -> Self;

    /// Some special case for the implementation in `SubImage`?
}
```

[...]

PR #866, opened 2019-02-11
[WIP] Fix mutable iteration over pixels

HeroicKatora:
Addresses points made in #854

The `unsafe` burden of constructing the iterator type will now lie on
the implementor of `GenericImage` instead of the library. This should
ensure that this decision is made more conciously. Some additional
blessed internal variants are provided to ensure that performant
iterators are still possible for the library types.


HeroicKatora:
Closing this now that `PixelsMut` has been removed. Still may be the basis for future work to bring it back in another shape.

PR #880, opened 2019-03-08
allow different image types in overlay and replace

dgulotta:
This commit allows the `bottom` and `top` parameters of `imageops::overlay` and `imageops::replace` to have different types.


HeroicKatora:
Thank you for the contribution!

PR #887, opened 2019-03-24
Remove nearly all instances of unsafe

fintelia:
This sketches out how we can remove nearly all instances of unsafe code in this crate. A majority of these cases are performance related (removing bounds checks or avoiding zeroing memory before use) but I think the safe versions should optimize reasonably well. The only places left are in the implementation of `Pixel` for the various pixel structs so if we could find a way to eliminate them we could mark the entire crate `#![deny(unsafe_code)]`:
```rust
fn from_slice(slice: &[T]) -> &$ident<T> {
    assert_eq!(slice.len(), $channels);
    unsafe { &*(slice.as_ptr() as *const $ident<T>) }
}

fn from_slice_mut(slice: &mut [T]) -> &mut $ident<T> {
    assert_eq!(slice.len(), $channels);
    unsafe { &mut *(slice.as_ptr() as *mut $ident<T>) }
}
```

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -852,13 +852,6 @@ where
         *self.get_pixel(x, y)
     }
 
-    /// Returns the pixel located at (x, y), ignoring bounds checking.
-    #[inline(always)]
-    unsafe fn unsafe_get_pixel(&self, x: u32, y: u32) -> P {
-        let indices = self.pixel_indices_unchecked(x, y);
-        *<P as Pixel>::from_slice(self.data.get_unchecked(indices))
-    }
-
     fn inner(&self) -> &Self::InnerImageView {
         self
     }
@@ -878,14 +871,6 @@ where
 
     fn put_pixel(&mut self, x: u32, y: u32, pixel: P) {
         *self.get_pixel_mut(x, y) = pixel
-    }
-
-    /// Puts a pixel at location (x, y), ignoring bounds checking.
-    #[inline(always)]
-    unsafe fn unsafe_put_pixel(&mut self, x: u32, y: u32, pixel: P) {
-        let indices = self.pixel_indices_unchecked(x, y);
-        let p = <P as Pixel>::from_slice_mut(self.data.get_unchecked_mut(indices));
-        *p = pixel
     }
 
     /// Put a pixel at location (x, y), taking into account alpha channels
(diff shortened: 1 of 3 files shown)


HeroicKatora:
Some of them are more convincing as individual changes (e.g. removing `PixelsMut` is largely independent from the unecessary `unsafe set_len` in hdr) but I like this, a lot. For `PixelsMut` specifically there's also #866 in the making but I've neglected it for a while and it would replace it with a new, safe implementation instead of completely removing the functionality.

HeroicKatora:
> The only places left are in the implementation of Pixel for the various pixel structs so if we could find a way to eliminate them we could mark the entire crate #![deny(unsafe_code)]:

There are new additions to the standard library that could enable this: `TryFrom` for casting between slices and arrays: [see here, though I can't seem to link them directly](https://doc.rust-lang.org/std/convert/trait.TryFrom.html). However, this still requires casting between the array and the wrapper type.

fintelia:
Wouldn't the cast from `&[T; 3]` to `Rgb<T>` still require unsafe?

HeroicKatora:
> However, this still requires casting between the array and the wrapper type

> Wouldn't the cast from say &[T; 3] to &Rgb<T> still require unsafe?

Oh yes, that cast is of course still unsafe. We could change `#[repr(C)]` to `#[repr(transparent)]` on the primitve pixel structs to make that cast much less controversial (and possibly fully safe with `std` or library support maybe in the future).

fintelia:
Allowing casts between any two `#[repr(transparent)]` types would let you cast from a u64 to a NonZeroU64 or otherwise break type level invariants, so there isn't an obvious (to me) way a library could safely allow that

HeroicKatora:
> Allowing casts between any two #[repr(transparent)] types would let you cast from a u64 to a NonZeroU64 or otherwise break type level invariants, so there isn't an obvious (to me) way a library could safely allow that.

I don't think that would break `U64`, and I would not imagine allowing it just arbitrarily. But the wrapping type itself could be allowed to do that cast *on references*, right? References require valid bit-pattern (✓), correct alignment (✓) and inbounds (✓).

HeroicKatora:
I'd really like to get rid of unsafety but contrary to the other changes this affects a lot of interfaces. I'm fine with removing the one that was already marked as `#[deprecated]` (`pixels_mut`) but would rather get the other deprecated for one major release before removing them. Is that possible?

fintelia:
Sounds reasonable. I've created #960 960

HeroicKatora:
All changes were merged incrementally or kept as performance critical.

PR #889, opened 2019-04-02
Add function to retrieve image dimensions without loading full image

vbarrielle:
This is way faster than loading a full image, and is useful if one needs to, for instance, check that all images in a folder have the same resolution, or if one needs to know the dimensions of all the images to compute some layout.

I've been using a variant of this function in [a photo album generator](https://github.com/vbarrielle/phototex/blob/master/src/main.rs#L81-L119) but I thought this might be useful for other users of the image crate.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -760,6 +760,59 @@ fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
     load(fin, format)
 }
 
+/// Read the dimensions of the image located at the specified path.
+/// This is faster than fully loading the image and then getting its dimensions.
+pub fn image_dimensions<P>(path: P) -> ImageResult<(u32, u32)>
+where
+    P: AsRef<Path>
+{
+    // thin wrapper function to strip generics before calling open_impl
+    image_dimensions_impl(path.as_ref())
+}
+
+fn image_dimensions_impl(path: &Path) -> ImageResult<(u32, u32)> {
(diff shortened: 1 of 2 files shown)


src/dynimage.rs
```diff
@@ -760,6 +760,59 @@ fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
...
+            Ok(pnm::PNMDecoder::new(fin)?.dimensions())
+        }
+        format => Err(image::ImageError::UnsupportedError(format!(
+            "Image format image/{:?} is not supported.",
+            format
+        ))),
+    }
+     // TODO possible overflow here, though not probable since it would
+     // require an extremely large image.
+    .map(|(w, h)| (w as u32, h as u32))
```
HeroicKatora: While not probable, the correct way to handle them is to return an error of either `ImageError::DimensionError` or `ImageError::UnsupportedError` with explanation in these cases. Otherwise, an allocation may be larger or smaller than expected, or a malicious image could appear small but then consume a large amount of memory when actually loaded.

vbarrielle: Thanks for the review, that makes the function clearer to read and the overflowing behaviour is now more explicit.

vbarrielle:
Latest commit should addres all issues pointed out by @HeroicKatora. Thanks for the review!

HeroicKatora:
Thank you!

PR #895, opened 2019-04-09
remove power of two check

SirWaddles:
Removing the Power of two check for DXT Decoder.

This is only necessary for DDS files which support mipmaps, but not for individual textures.


HeroicKatora:
Thanks you. Since this is your first contribution, you can choose to join the collaborators of `image-rs`, an invitation has been sent.

PR #898, opened 2019-04-13
Adds the readme as a doctest

HeroicKatora:
Uses a clever hack found during implementation of the standard library and popularized in this crate: https://github.com/GuillaumeGomez/doc-comment

Since this actually compiles the `README` now, we also needed to fix some of the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ All pixels are parameterised by their component type.
 A trait that provides functions for manipulating images, parameterised over the image's pixel type.
 
 ```rust
+# use image::{Pixel, Pixels};
 pub trait GenericImage {
     /// The pixel type.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,6 +69,7 @@ All pixels are parameterised by their component type.
 A trait that provides functions for manipulating images, parameterised over the image's pixel type.
 
 ```rust
+# use image::{Pixel, Pixels};
```
birktj: A bit late, but should the `#` be here?

HeroicKatora: Not necessarily. In the documentation this would of course make it invisible but for the readme this fails its purpose. I guess we can remove it if we want.

PR #903, opened 2019-04-15
Add lifetime bound on ImageDecoder trait

fintelia:
This was overlooked in the initial overhaul of the trait, but is necessary for writing certain sorts of generic code over decoders.

### DIFF
--- a/src/pnm/decoder.rs
+++ b/src/pnm/decoder.rs
@@ -406,7 +406,7 @@ trait HeaderReader: BufRead {
 
 impl<R: Read> HeaderReader for BufReader<R> {}
 
-impl<'a, R: Read> ImageDecoder<'a> for PNMDecoder<R> {
+impl<'a, R: 'a + Read> ImageDecoder<'a> for PNMDecoder<R> {
     type Reader = Cursor<Vec<u8>>;
 
     fn dimensions(&self) -> (u64, u64) {
--- a/src/pnm/decoder.rs
+++ b/src/pnm/decoder.rs
@@ -406,7 +406,7 @@ trait HeaderReader: BufRead {
 
 impl<R: Read> HeaderReader for BufReader<R> {}
 
-impl<R: Read> ImageDecoder for PNMDecoder<R> {
+impl<'a, R: Read> ImageDecoder<'a> for PNMDecoder<R> {
     type Reader = Cursor<Vec<u8>>;
 
     fn dimensions(&self) -> (u64, u64) {
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1255,7 +1255,7 @@ impl<R: Read + Seek> BMPDecoder<R> {
     }
 }
 
-impl<'a, R: Read + Seek> ImageDecoder<'a> for BMPDecoder<R> {
+impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for BMPDecoder<R> {
     type Reader = Cursor<Vec<u8>>;
 
     fn dimensions(&self) -> (u64, u64) {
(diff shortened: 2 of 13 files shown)


fintelia:
If there are no objections to this (and no new CI conflicts) I'll merge this in a bit

src/pnm/decoder.rs
```diff
@@ -406,7 +406,7 @@ trait HeaderReader: BufRead {
 
 impl<R: Read> HeaderReader for BufReader<R> {}
 
-impl<R: Read> ImageDecoder for PNMDecoder<R> {
+impl<'a, R: Read> ImageDecoder<'a> for PNMDecoder<R> {
```
HeroicKatora: Why doesn't this require an explicit `'a + Read` bound?

birktj: I was wondering the same, but I think maybe(?) that `rustc` is smart enough to understand this?

fintelia: I think this should actually be changed to a `'a + Read` bound. The reason that rustc doesn't complain is that the lifetime of a `Cursor<Vec<u8>>` doesn't depend on the lifetime of of the reader we use to create the data for it

HeroicKatora: Having less bouds seems more useful. And any change of the `Reader` type will be SemVer break in any case, the only advantage I can spot directly is an easier transition in case we decide to change the `Reader` anyways.

birktj: Since that bound isn't used it wouldn't have any practical implications would it?

fintelia: If changing the Reader type is a breaking change, then I think we need to move to an opaque wrapper type per implementation. And we would also want to have the constrained lifetime bound so that we don't rule out incremental decoding for any image types. For now, the more restrictive lifetime bounds should prevent users from relying on the previous behavior.

HeroicKatora: Not quite, the bound restricts the usability of the following:

```
into_boxed_reader<'a, D>(decoder: D) -> Result<Box<Read>, ImageError>
    where D: ImageDecoder<'a>, D::Reader: 'static
{
    Box::new(decoder.into_reader()?)
}
```

which can not be instantiated with some `PNMDecoder<&'a File>`. Since: `&'a File` is not `'static` thus the `impl` block does not apply.

birktj: It is probably not to important but since `PNMDecoder::Reader = Cursor<Vec<u8>>` and thus `'static` I would think that it should be usable with this function?

HeroicKatora: Sorry, I meant `D: ImageDecoder<'static>`, `D: ImageDecoder<'a>, D::Reader: 'static` works.

HeroicKatora: Testable example:

```

fn into_boxed_reader<'a, D>(decoder: D) -> Result<Box<Read>, ImageError>
    where D: ImageDecoder<'static>
{
    Ok(Box::new(decoder.into_reader()?))
}

fn test() -> Box<Read> {
    use std::fs;
    let file = fs::File::open("/tmp/something")
        .unwrap();
    into_boxed_reader(
        ::pnm::PNMDecoder::new(&file).unwrap()
    ).unwrap()
}
```

works fine without `R: 'a + Read` but doesn't work with that bound.

fintelia: No, that couldn't work regardless since ImageDecoder has the bound that `type Reader: Read + 'a`. (In fact, that's the only place the lifetime bound on the decoder applies). 

What the change is preventing is a function like:

```rust
fn make_decoder(filename: &str) -> impl ImageDecoder<'static> {
    let mut file = File::open(filename).unwrap();
    PnmDecoder::new(&mut file).unwrap().into_reader()
    // file is closed here, but the interface pretends that the user can still do 
    // incremental decoding from it. 
}
```

This would be fine for PNMDecoder because it doesn't actually support incremental decoding, but would break if we switched to say PNGDecoder which does decode incrementally.

HeroicKatora: It works, since the bound from `ImageDecoder` on affects the associated type `Reader` but the bound on the impls affects the internal type, but is not necessarily contained in the associated one.

birktj: @fintelia But that doesn't make sense, your function is of type `fn make_decoder(filname: &str) -> impl Read` which should compile.

HeroicKatora: PNMDecoder not being incremental is a big part of it. Ok, I would be fine with actually having the bound to make sure that transitioning isn't too much of a change when it gets incremental support.

PR #904, opened 2019-04-15
Replace the ImageFormat enum with mime types

fintelia:
This PR replaces all uses of the ImageFormat enum with with the Mime struct (from the crate of the same name). This allows us to tap into the broader Rust ecosystem for file type handling, including the mime_guess crate which provides conversions from file extension to file type.

Depends on #903 and abonander/mime_guess#45

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ num-traits = "0.2.0"
 lzw = "0.10.0"
 mime_guess = "2.0.0-alpha.6"
 mime = "0.3.13"
+lazy_static = "1.3.0"
 
 [dependencies.gif]
 version = "0.10.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,8 @@ num-iter = "0.1.32"
 num-rational = { version = "0.2.1", default-features = false }
 num-traits = "0.2.0"
 lzw = "0.10.0"
+mime_guess = "2.0.0-alpha.6"
+mime = "0.3.13"
 
 [dependencies.gif]
 version = "0.10.0"
--- a/examples/opening.rs
+++ b/examples/opening.rs
@@ -27,5 +27,5 @@ fn main() {
     let fout = &mut File::create(&Path::new(&format!("{}.png", file))).unwrap();
 
     // Write the contents of this image to the Writer in PNG format.
-    im.write_to(fout, image::PNG).unwrap();
+    im.write_to(fout, image::ImageOutputFormat::PNG).unwrap();
 }
--- a/examples/scaledown/main.rs
+++ b/examples/scaledown/main.rs
@@ -1,6 +1,6 @@
 extern crate image;
 
-use image::{FilterType, PNG};
+use image::{FilterType, ImageOutputFormat::PNG};
 use std::fmt;
 use std::fs::File;
 use std::time::{Duration, Instant};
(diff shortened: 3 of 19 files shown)


HeroicKatora:
I don't really like it. Adding converters to choose the image type from `mime` and conversely seems good. But I think just replacing the type looses a lot of expressivness. The need to have the format magic be replaced by a `lazy_static` alone is indicative of this.

What is the reason to have `DynamicDecoder` as an enum, but not the `Reader`? I think this could do without the dependency on #903 if it would be coded like that.

fintelia:
I agree on the expressivity/lazy_static front. Do you think it would be enough if we were able to provide constants for the various mime types? 

The motivation for using an enum for `DynamicDecoder` but not Reader is actually reversed. Using `Box<dyn TraitObject>` is simpler and adds negligible overhead (hence why I use it for Reader), but `ImageDecoder` can't be object safe, so an enum is required along with all the code boilerplate that entails. The need for #903 is somewhat orthogonal: the trait always should have had an explicit way to say "the reader produced by this decoder can't outlive the one used to create it", it just so happened that I didn't realize that was missing until trying to implement this PR.

HeroicKatora:
> Using Box<dyn TraitObject> is simpler and adds negligible overhead (hence why I use it for Reader), but ImageDecoder can't be object safe, so an enum is required along with all the code boilerplate that entails.

Could you still use `Box<dyn ImageDecoder<Reader=Box<dyn Read>>>` as the dynamic decoder type (and implement `ImageDecoder` on that)? And while the impact of the box could be neglible in some or many cases, I'd be surprised if it doesn't thwart some constant propagation for the methods where the output type is not the decoder but the result of calling the method on it. And it's at least two allocations there. A proper performance test could for example be querying the `image_dimensions` on a medium or large collection of images.

HeroicKatora:
> I agree on the expressivity/lazy_static front. Do you think it would be enough if we were able to provide constants for the various mime types?

That, or either `TryFrom` or `From` or `ImageFormat::from_mime(Mime) -> Option<Self>`. Open to argument for the positive and negative aspect of each.

fintelia:
> Could you still use Box<dyn ImageDecoder<Reader=Box<dyn Read>>> as the dynamic decoder type (and implement ImageDecoder on that)?

ImageDecoder has a Self: Sized bound because some methods consume self so it isn't object safe. Consequently, we aren't able to create a `Box<dyn ImageDecoder>`

> That, or either TryFrom or From or ImageFormat::from_mime(Mime) -> Option<Self>. Open to argument for the positive and negative aspect of each.

The main benefit of having constants would be that we wouldn't have to keep our own `ImageType` enum, which would reduce our API surface. Though I'm not sure whether it is possible to create Mime objects statically, so that might not actually be an option.

HeroicKatora:
Ah, `Self: Sized` is the reason. I somehow missed that.

>  The main benefit of having constants would be that we wouldn't have to keep our own ImageType enum, which would reduce our API surface. Though I'm not sure whether it is possible to create Mime objects statically, so that might not actually be an option.

They can be constants with `lazy_static` just like the table would be now but I hope that we don't need to do more than re-export some existing constants? When we get reasonable support for `const fn` and want to upgrade the compiler requirements then `mime` and ourselves can upgrade to full statics I hope. In the meantime, I figure that `ImageFormat::PNG.mime()` would not be too bad and even fully inlined by the compiler (I hope), where there is no constant in the `mime` crate. And our own enum variants can be used where an actual `static`/`const` is necessary.

fintelia:
It doesn't seem that the mime crate is going to get support for creating constant Mimes in the near future, so I'm going to close this for now.

PR #905, opened 2019-04-15
[API Change] Make copy_from return an ImageResult

### LINKED ISSUE #901: [API design] copy_from returns bool
Hello, and thank you for this very useful crate!
  I stumbled upon [`GenericImage::copy_from`](https://docs.rs/image/0.21.1/image/trait.GenericImage.html#method.copy_from), that returns `false` when the operation fails. This is not very idiomatic in rust, and makes error handling more difficult. It would be nice if the method could return a `Result<()>`. This would prevent crate users from inadvertently ignoring an error.

lovasoa:
Fixes #901

Returning a Result instead of a boolean should simplify error handling, and prevent accidentally ignoring errors.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -566,6 +566,7 @@ pub trait GenericImageView {
     fn inner(&self) -> &Self::InnerImageView;
 
     /// Returns an subimage that is an immutable view into this image.
+    /// You can use [`GenericImage::sub_image`] if you need a mutable view instead.
     fn view(&self, x: u32, y: u32, width: u32, height: u32) -> SubImage<&Self::InnerImageView> {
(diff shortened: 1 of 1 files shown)


lovasoa:
Hurrah ! :confetti_ball:

lovasoa:
Oh geez ! I just realized I force pushed to the wrong branch ! I forgot about this old PR while working on #1003.  What was merged was not the fix for #901

HeroicKatora:
There really should be a special warning for this, it's just odd.

Feel free to reopen against the `next` branch which aggregates breaking changes for the next major release, for issues that don't require immediate attention.

lovasoa:
I opened a new PR (#1004) sorry for the inconvenience.

Issue #774, opened 2018-05-18
webp decoder panic


quininer:
I try to decode [this webp file](https://www.gstatic.com/webp/gallery/1.webp).

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/test-image-webp`
thread 'main' panicked at 'index out of bounds: the len is 202400 but the index is 202400', /checkout/src/libcore/slice/mod.rs:2342:14
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:463
   5: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:350
   6: rust_begin_unwind
             at libstd/panicking.rs:328
   7: core::panicking::panic_fmt
             at libcore/panicking.rs:71
   8: core::panicking::panic_bounds_check
             at libcore/panicking.rs:58
   9: <usize as core::slice::SliceIndex<[T]>>::index_mut
             at /checkout/src/libcore/slice/mod.rs:2342
  10: core::slice::<impl core::ops::index::IndexMut<I> for [T]>::index_mut
             at /checkout/src/libcore/slice/mod.rs:2249
  11: <alloc::vec::Vec<T> as core::ops::index::IndexMut<I>>::index_mut
             at /checkout/src/liballoc/vec.rs:1720
  12: <image::webp::vp8::VP8Decoder<R>>::intra_predict
             at /home/quininer/.cargo/git/checkouts/image-ac7d41938885a19a/6f46cf3/./src/webp/vp8.rs:1264
  13: <image::webp::vp8::VP8Decoder<R>>::decode_frame
             at /home/quininer/.cargo/git/checkouts/image-ac7d41938885a19a/6f46cf3/./src/webp/vp8.rs:1450
  14: <image::webp::decoder::WebpDecoder<R>>::read_frame
             at /home/quininer/.cargo/git/checkouts/image-ac7d41938885a19a/6f46cf3/./src/webp/decoder.rs:80
[...]

PR #918, opened 2019-05-03
Fix vp8 border block indexing

HeroicKatora:
Closes: #774


HeroicKatora:
I completely remove the `if` based implementation. It should be *much* more obvious what the code is trying to achieve.

martinlindhe:
Looks great, the new calculation is much more readable

HeroicKatora:
Forgot that `From<u16> for usize` was stabilized only recently.

PR #922, opened 2019-05-03
Implement support for tiff encoding

birktj:
Updated `tiff` version to 0.3.0, which supports
tiff encoding and implemented a basic tiff encoder.

### DIFF
--- a/src/tiff.rs
+++ b/src/tiff.rs
@@ -8,7 +8,7 @@
 
 extern crate tiff;
 
-use std::io::{self, Cursor, Read, Seek};
+use std::io::{self, Cursor, Read, Write, Seek};
 use std::marker::PhantomData;
 use std::mem;
 
@@ -48,6 +48,7 @@ impl From<tiff::TiffError> for ImageError {
             tiff::TiffError::IoError(err) => ImageError::IoError(err),
             tiff::TiffError::FormatError(desc) => ImageError::FormatError(desc.to_string()),
             tiff::TiffError::UnsupportedError(desc) => ImageError::UnsupportedError(desc.to_string()),
+            tiff::TiffError::LimitsExceeded => ImageError::InsufficientMemory,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/tiff.rs
```diff
@@ -85,3 +86,33 @@ impl<R: Read + Seek> ImageDecoder for TIFFDecoder<R> {
...
     }
 }
+
+/// Encoder for tiff images
+pub struct TiffEncoder<W> {
+    w: W,
+}
+
+impl<W: Write + Seek> TiffEncoder<W> {
+    /// Create a new encoder that writes its output to ```w```
```
HeroicKatora: Is there a semantic difference in Commonmark Markdown between triple quoted (```these```) and single quoted (`these`) inline blocks?

birktj: I would suppose not, however I think single quotes look better so I will change it.

HeroicKatora: I always wondered that since I saw it in several code bases. Might be a historic reason or an older spec that had a difference.

birktj:
It seems like some changes introduced in `tiff` 0.3.0 is not supported by rust 1.24.1, notably the use of `crate` in paths and `std` without `use std`. What would be a good solution to this problem? Publish a new version 0.3.1 of the tiff crate where we fix these?

HeroicKatora:
Oh, I didn't really consider that. How good is it to stay dependent on Rust `1.24.1` is a question I've been asking myself a while now.

HeroicKatora:
But if it is possible, and not too awkward, then lowering the required `rustc` version is helpful, go for it.

birktj:
You mean lowering the required version of `rustc` in the tiff crate right?

HeroicKatora:
Yep. My default criterion for choosing when to bump required version is [the next Debian stable release](https://wiki.debian.org/DebianReleases#Production_Releases) likely this year as that is the 'laggiest' distribution I know of with a large share of users. That may be a bit too conservative, we may want to do it sooner and some dependencies for `image` may have already done so as far as I am aware (`edition = 2018` being not compatible with `1.24.1` as you saw).

HeroicKatora:
So, by all means tell us if it turns out backporting is a bit too involved. We could pre-emptively choose to adopt a newer version of Rust, and do the next major release sometime in summer. I'd like to get the `ImageDecoder` stuff sorted out in that interface as well.

HeroicKatora:
@birktj As far as I can tell, `tiff` is not a public dependency of `image`–none of its items are accessible through image and there are no traits implemented for types from `tiff`. We should be free to change the version we depend on however we like. That means, I think it could be fine to simply set the version requirement to exactly one that contains the 1.24.1 compatible stuff, like `v0.3.1` with https://github.com/image-rs/image-tiff/pull/44

HeroicKatora:
Forgive me the bunch of force pushes to the branch, I rebase onto current master with an additional commit to update the dependency as just published. Then I got confused about the exact interpretation of the requirements syntax in `Cargo.toml`.

Also, did you even know that maintainers can force push into PR branches of forks? :joy:

birktj:
It's fine, great that we could do this while keeping the compiler support. I would guess the maintainer write access to the PR branches is a full write access and github dosn't care if it just a commit or a force-push?

HeroicKatora:
Woooo! :tada:

Issue #919, opened 2019-05-03
VP8 decoder will panic on unsupport input images


HeroicKatora:
Some instances of `panic` depend on input data but only signal an umimplemented feature. Those should be replaced with `Err(ImageError::NotSupported(..))` instead. Futhermore, at least one enumeration (`{DC, V, H, TM}_PRED`) taken from the reference implementation may be replaced with an actual `enum` in Rust to avoid having `unreachable!()` branches on `match`.

PR #923, opened 2019-05-03
Fix multiple cases of panic on Webp images

HeroicKatora:
After the initial 'obvious' results, I've simply turned to fuzzing. That yielded almost 2 separate bugs per hour, huge productivity boost. Resource exhaustion is not yet within the scope of consideration but the number of timeouts is still relatively large. Also, no functional changes for existing images, so sadly no other bugfixes. The final minimized corpus of samples will be added when the rest is ready, and some known-good reference images will also be added to the tests for CI purposes.

So far, it's been a huge boost to the stability.

Now, I only find memory exhaustion failures. While still annoying, they aren't nearly as critical right now.

Closes: #919


HeroicKatora:
The last of these commits has me slightly worried. The reference implementation works on `i16` data in these methods, I can't find my way through all the macro insanity they used for it, and there's something weird with multiplication of two 16-bit fixed point numbers overflowing an `i32` (one of the tests cases from the corpus exhibits this, so you can revert the commit locally to inspect the crash).

HeroicKatora:
What? How does `1.24.1` fail on this line:

https://github.com/image-rs/image/blob/f416e5d4af13953bbb8021401e359882cf8d8f06/src/webp/vp8.rs#L1305

That doesn't make sense, I haven't touched that at all. But:

```text
error[E0277]: the trait bound `i16: std::convert::From<bool>` is not satisfied
    --> src/webp/vp8.rs:1375:49
     |
1375 |                         extra = extra + extra + i16::from(self.partitions[p].read_bool(t[j]));
     |                                                 ^^^^^^^^^ the trait `std::convert::From<bool>` is not implemented for `i16`
     |
     = help: the following implementations were found:
               <i16 as std::convert::From<u8>>
               <i16 as std::convert::From<i8>>
     = note: required by `std::convert::From::from`
```

src/webp/vp8.rs
```diff
@@ -1148,67 +1211,75 @@ impl<R: Read> VP8Decoder<R> {
...
             }
 
-            mb.chroma_mode =
-                self.b
-                    .read_with_tree(&KEYFRAME_UV_MODE_TREE, &KEYFRAME_UV_MODE_PROBS, 0);
+            let chroma = self.b
+                .read_with_tree(&KEYFRAME_UV_MODE_TREE, &KEYFRAME_UV_MODE_PROBS, 0);
+            mb.chroma_mode = ChromaMode::from_i8(chroma)
+                .ok_or_else(|| ImageError::FormatError(
+                    format!("Invalid chroma prediction mode {}", luma))
```
birktj: Shouln't `luma` here be `chroma`

HeroicKatora: Yes. Thank you.

src/webp/vp8.rs
```diff
@@ -884,33 +932,38 @@ impl<R: Read> VP8Decoder<R> {
...
 
-                self.partitions[i].init(buf);
+                self.partitions[i].init(buf)?;
             }
         }
 
         let mut buf = Vec::new();
-        try!(self.r.read_to_end(&mut buf));
-        self.partitions[n - 1].init(buf);
+        self.r.read_to_end(&mut buf)?;
```
birktj: This looks sort of weird. Shouldn't the decoder know how large the last partition is, since more frames can follow? I would guess this easily could get some sort of memory exhaustion error?

HeroicKatora: No, the last partition is actually implicitely sized by the length of the upper partition in the `RIFF` file. In this particular instance, the `Read` instance should be instantiated with a `file.take(riff.len)` in the `webp` decoder. The reference implementation keeps track of all remaining bytes by propagating a `unsigned int sz` into every function call instead, which is effectively the same.

HeroicKatora: Hm, it seems that this is however not the case in the upper layer: https://github.com/image-rs/image/blob/445b60bcd6e5578f28d1c81fffa37187877a2554/src/webp/decoder.rs#L70

HeroicKatora: Will fix that, and include it as a test case.

birktj: But is still don't completely understand. I thought `read_frame_header` would run once for each of the possibly very many frames in a vp8 stream, and then `read_to_end` could be called many times?

birktj: From the reference:
> At the highest level, every compressed frame has three or more
   pieces.  It begins with an uncompressed data chunk comprising
   10 bytes in the case of key frames and 3 bytes for interframes.  This
   is followed by two or more blocks of compressed data (called
   partitions).  These compressed data partitions begin and end on byte
   boundaries.

To me this seems like the partitions only last until the next frame header?

HeroicKatora: From [section-9.5](https://tools.ietf.org/html/rfc6386#section-9.5):
> Offsets are embedded in the bitstream to provide the decoder direct
   access to token partitions.  If the number of data partitions is
   greater than 1, the size of each partition (except the last) is
   written in 3 bytes (24 bits).  The size of the last partition is the
   remainder of the data not used by any of the previous partitions.

I was under the impression that there was one `VP8` [chunk](https://developers.google.com/speed/webp/docs/riff_container#simple_file_format_lossy) in the `RIFF` for every 'compressed frame'. Otherwise, those two statements don't exactly make sense to me. For image (single-frame vp8) this doesn't make a difference, the only contained frame in the chunk is the image.

HeroicKatora: But I didn't really get it completely, and am not really sure. However, it seems to work.

birktj: Yeah, that makes sense. So it is up to the container format so define frame boundaries, and the vp8 decoder just decodes whatever data is given to it as a single frame?

HeroicKatora: It seems so? At least then it makes sense that there is seemingly one length number too few in the frame header.

HeroicKatora:
I'll rebase and force-push on top of current master, where we also have some reference image test cases thanks to dd74b23955896b17ca3dcb2cad02e7a4ab871b28

birktj:
I think we also should split `vp8.rs` into some smaller modules, 1.7k lines is quite exhausting to go through. I also think that when chroma decoding is implemented (I am looking into it, but not promising anything) we should create a separate `vp8` crate.

PR #924, opened 2019-05-03
Reduce builds on windows and osx to full check

HeroicKatora:
We don't gain any interesting insight with feature flag combinations
other than 'all features' (the default). Still builds on stable and
nightly, but not beta, and 1.24.1 so that we are (somewhat) warned when
a dependency requires updating the minimum compiler version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ notifications:
     irc: "irc.mozilla.org#piston-internals"
 os:
     - linux
-    - osx
-    - windows
 rust:
     - 1.24.1
     - stable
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Reference times from recent `master`:
```
 Ran for 1 hr 43 min 37 sec
Total time 6 hrs 39 min 44 sec
```

HeroicKatora:
And with the new `.travis.yml`:

```text
Ran for 31 min 33 sec
Total time 1 hr 47 min 25 sec
```

One huge influence in this improvement:

* Linux, stable: `2min 53s`
* Windows, stable: `7min 56s`

Basically removing most Windows builds alone cuts our build time by 40% or more.

martinlindhe:
I think this is the way to go! CI running for hours is not exactly optimal.

PR #928, opened 2019-05-04
Error clean up

### LINKED ISSUE #921: Encoders use `io::Error` instead of `ImageError`
I just realized that encoders use `io::Error` as errors instead of `ImageError`, wouldn't it be better to use `ImageError` in order to signal more specific encoding errors like unsupported color formats, etc?

birktj:
This is a start towards cleaning up error handling in this library, there is still a lot to do and it can be we want to rethink error handling somewhat, having error variants for specific decoders like a `ImageError::PngError` vs general variants like `ImageError::FormatError`. See some discussion in #921.

This should fix #921

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,98 @@
+use std::error::Error;
+use std::fmt;
+use std::io;
+use color::ColorType;
+
+/// An enumeration of Image errors
+#[derive(Debug)]
+pub enum ImageError {
+    /// The Image is not formatted properly
+    FormatError(String),
+
+    /// The Image's dimensions are either too small or too large
+    DimensionError,
+
+    /// The Decoder does not support this image format
+    UnsupportedError(String),
+
+    /// The Decoder does not support this color type
+    UnsupportedColor(ColorType),
+
+    /// Not enough data was provided to the Decoder
+    /// to decode the image
+    NotEnoughData,
+
+    /// An I/O Error occurred while decoding the image
+    IoError(io::Error),
+
+    /// The end of the image has been reached
+    ImageEnd,
+
+    /// There is not enough memory to complete the given operation
+    InsufficientMemory,
+}
+
+impl fmt::Display for ImageError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        match *self {
+            ImageError::FormatError(ref e) => write!(fmt, "Format error: {}", e),
+            ImageError::DimensionError => write!(
+                fmt,
+                "The Image's dimensions are either too \
+                 small or too large"
+            ),
+            ImageError::UnsupportedError(ref f) => write!(
+                fmt,
+                "The Decoder does not support the \
+                 image format `{}`",
+                f
+            ),
+            ImageError::UnsupportedColor(ref c) => write!(
+                fmt,
+                "The decoder does not support \
+                 the color type `{:?}`",
+                c
+            ),
+            ImageError::NotEnoughData => write!(
+                fmt,
+                "Not enough data was provided to the \
+                 Decoder to decode the image"
+            ),
+            ImageError::IoError(ref e) => e.fmt(fmt),
+            ImageError::ImageEnd => write!(fmt, "The end of the image has been reached"),
+            ImageError::InsufficientMemory => write!(fmt, "Insufficient memory"),
+        }
+    }
+}
+
+impl Error for ImageError {
+    fn description(&self) -> &str {
+        match *self {
+            ImageError::FormatError(..) => "Format error",
+            ImageError::DimensionError => "Dimension error",
+            ImageError::UnsupportedError(..) => "Unsupported error",
+            ImageError::UnsupportedColor(..) => "Unsupported color",
+            ImageError::NotEnoughData => "Not enough data",
+            ImageError::IoError(..) => "IO error",
+            ImageError::ImageEnd => "Image end",
+            ImageError::InsufficientMemory => "Insufficient memory",
+        }
+    }
+
+    // TODO: use `Error::source` when minimal rust version is updated
+    fn cause(&self) -> Option<&Error> {
+        match *self {
+            ImageError::IoError(ref e) => Some(e),
+            _ => None,
+        }
+    }
+}
+
+impl From<io::Error> for ImageError {
+    fn from(err: io::Error) -> ImageError {
+        ImageError::IoError(err)
+    }
+}
+
+/// Result of an image decoding/encoding process
+pub type ImageResult<T> = Result<T, ImageError>;
--- a/src/error.rs
+++ b/src/error.rs
@@ -16,7 +16,7 @@ pub enum ImageError {
     /// This image format is not supported
     UnsupportedFormat(String),
 
-    /// The Decoder does not support this color type
+    /// This color type is not supported
     UnsupportedColor(ColorType),
 
     /// Unsupported feature in a image format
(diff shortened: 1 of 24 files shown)


src/error.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::fmt;
+use std::io;
+use color::ColorType;
+use image::ImageFormat;
+
+/// An enumeration of Image errors
+#[derive(Debug)]
+pub enum ImageError {
+    /// The Image is not formatted properly
+    FormatError(String),
```
fintelia: Perhaps `FileCorrupt(String)`?

fintelia:
I like the direction this is going. Slightly longer term, I'd like to include the error type in a future `image-core` crate. That way the various encoding/decoding libraries could directly return `ImageError`s instead of having to make their own type

src/error.rs
birktj: Yeah, that would probably be a bit more specific. Only problem is that I use this variant for encoders as well, but with some sort of `InvalidData` it could be a nice combination.

src/error.rs
```diff
@@ -0,0 +1,107 @@
...
+    DimensionError,
+
+    /// This image format is not supported
+    UnsupportedFormat(String),
+
+    /// This color type is not supported
+    UnsupportedColor(ColorType),
+
+    /// Unsupported feature in a image format
+    UnsupportedFeature(ImageFormat, String),
```
fintelia: I'm not sure if we should include `ImageFormat` here. The problem is that if encoding/decoding crates were to start returning `ImageError`s instead of their own error types, then we'd have to add a variant to ImageFormat before we could start implementing the decoding library (so that it could produce correctly typed UnsupportedFeature errors)

HeroicKatora: Maybe this could be a (struct) variant similar to the following. Preferring struct to give the `Option` a better meaning.
```
UnsupportedFeature {
    format: Option<ImageFormat>,
    details: String,
}
```

src/error.rs
HeroicKatora: One more thing we could be clearer on here: If the format contains a `length` field but the file ends prematurely. Is it

* `FormatError`
* `NotEnoughData`
* `io::ErrorKind::UnexpectedEof`, the return when reading that data with `r.read_exact(&mut slice)?`

HeroicKatora: This doesn't need a definitive answer, but definitely examples in the doc to make choosing a variant easier in all implementations.

src/error.rs
```diff
@@ -0,0 +1,107 @@
...
+use image::ImageFormat;
+
+/// An enumeration of Image errors
+#[derive(Debug)]
+pub enum ImageError {
+    /// The Image is not formatted properly
+    FormatError(String),
+
+    /// The Image's dimensions are either too small or too large
+    DimensionError,
```
HeroicKatora: I would find it good to remove the ambiguity of `too small or too large` in the error message here. Maybe an `expected: Option<(u32, u32)>` and an `actual: (u32, u32)` field which could give much better error formatting.

src/error.rs
fintelia: Hm, I'm imagining `ImaegError` becoming part of image-core with `ImageFormat` remaining part of the main image crate. That would preclude mentioning ImageFormat at all, even inside an option. Also, wouldn't we expect the user to have an idea of which image type their decoding anyway? After all, most of the other variants don't list which image format...

src/error.rs
fintelia: Thinking about this more, it isn't really clear to me what this error is even supposed to mean. How could an image be too small? We support 1x1 images and you can't get smaller than that! I think this variant was actually created for DXT images which error out if the width and height aren't multiples of 4 or dimensions that overflow a u32

src/error.rs
HeroicKatora: True, maybe we should avoid mentioning it altogether.

Is `String` the best representation we have for unsupported features? `Cow<'static, str>` seems like a better representation even if we stringly type this variant.

src/error.rs
birktj: It seems like it is mostly used in dynamicimage and generally when the size of image data `!= width * height`. A `NotEnoughData` or something like that is maybe better?

src/error.rs
birktj: This is probably better. I think we have to stringly type this variant as it can be so many different things from any of the decoders, however I don't know the answer to the `Cow` vs `String` question, I see both appear here and there. I would say that the small amount of allocations that will happen using string is unproblematic and the small complexity of using `Cow` isn't really worth it.

src/error.rs
HeroicKatora: I don't quite undersand the usage in `dynimage.rs`. It seems to protect against a decoder reporting a size that doesn't match the pixel data returned. But in that case, shouldn't the decoder already return an error for reading? Wouldn't a mismatching success be an error in the implementation of the decoder rather than the input? I feel like an assertion could be fine here.

Opposing this is the assertion that the size fits within `u32`! That is not the problem of a buggy decoder but simply not supported for actual images. *Here* we could return `DimensionError` with good reason.

In `DXT`, the decoder will also check that the size is a multiple of 4 in all directions. It seem this limit is imposed by the format. This one would also fit `DimensionError` but neither `too small` no `too large` is correct.

Finally, in `flat` the check if a buffer is well-formed for being used as an image includes ensuring all indices are smaller than the length. That reports its own error type first as `TooLarge` which then gets converted into `ImageError::DimensionError` in the `From<_>` impl.

fintelia:
What is the status of this? It would be good if we could fix up the lingering issues and merge this

birktj:
Yeah, I too would like to finish this. However there are some problems, the biggest of which is that I am not entirely sure how we should think about error messages. One way is to create all error messages on error construction and only use the enums to match error type, this is the most flexible, but requires a bit of thought when constructing the errors so that they make sense and is a slippery slope towards stringly typed errors. The other way is to use the enums to create some sort of standardized error message where we simply fill in the blanks. I tried to do this a few places in this PR, but I am not convinced that it is the best solution, in many cases I think the error messages ends up not really making too much sense.

I would like some discussion about how we should solve these problems before continuing with this PR, the reason I haven't got around to asking before is that I have been slightly busy the last couple of weeks, but if we figure out a nice way to solve error messages I should be able to finish this up.

birktj:
Case in point, I think `UnsupportedFormat` should work pretty well the way it is defined in this PR, but I think `UnsupportedFeature` is too general to work this way.

HeroicKatora:
While getting it into `0.22` would be nice, we should neither block the next version on this cleanup nor rush it because of the release. From a quick look there are several open questions, some of them more fundamental than others. Don't get me wrong, it's very good that the process is started here but the problem as I see it is mostly an ergonomical one.

With regards to `UnsupportedFeature`, maybe one possibility:

```
trait Feature: Any {
    …
}

enum Error {
    …
    UnsupportedFeature(&'static Feature),
}
```

fintelia:
At this point I'm not sure how actionable any of the error variants really are. How is programmatically determining which feature wasn't supported actually going to help? Is the program going to do anything other than print an error to the screen and then exit/ask for a different image file? From that perspective, building a whole collection of types to encode different features we don't have might not provide much value. Also thinking about it, if the types are public wouldn't removing them (because said feature is no-longer unimplemented) be a breaking change?

To me, the two most useful things for an error to carry would be a string (to display to the user) and a backtrace so that the developer could figure out exactly why the library decided to bail.

HeroicKatora:
@fintelia That's actually a very good point. In retrospect, it was some part of the motivation for the design of providing just the abstract trait and a static reference to an instance in the variant. The zero-sized type that implements it can stay hidden and doesn't need to be `pub` qualified. Then it can also be removed without a semver break.

But that doesn't address the point of what opportunity it provides, but only if the format is part of the error. A user of `image` could then possibly try to decode and convert the image with either another library or the underlying library directly. This is may make sense as a fallback since it may involve menial or less efficient code.

HeroicKatora:
Was closed by #1119.

PR #929, opened 2019-05-04
document how to run benchmarks with a benches/README.md

martinlindhe:


HeroicKatora:
Can we move this `Makefile` and possibly also a `Readme.md` to the same effect into the `benches` folder? A top-level `Makefile` could be confusing people into running it instead of `cargo` to build the crate itself.

martinlindhe:
I dropped the makefile in favor for a benches/README.md

benches/README.md
```diff
@@ -0,0 +1,6 @@
+# Getting started with benchmarking
+
+To run the benchmarks you need a nightly rust toolchain.
+Then you launch it with
+
+    rustup run nightly cargo bench --features=benchmarks
```
fintelia: Couldn't you just do `cargo +nightly bench --features/benchmarks`?

martinlindhe: Wasn't aware of this syntax. I'll update PR

martinlindhe: Sorted

PR #930, opened 2019-05-04
Flif decoding support implemented

### LINKED ISSUE #469: Support flif
http://flif.info/

birktj:
This PR adds flif decoding support using the [`flif.rs` crate](https://github.com/dgriffen/flif.rs). Only small problem is that this crate is rust 2018, and thus we can't merge this PR before we update the minimal supported `rustc` version.

This fixes #469


birktj:
A possible solution to the `rustc` version problem could be to not make the `flif` feature a default feature? And say that `rustc` 1.24.1 is not supported when using this feature.

HeroicKatora:
Seems like a possible solution until the next major version that upgrades compiler requirements. It also makes it clear that the older compiler support is deprecated

est31:
MSRV has been increased to 1.34 in 087b9a6ca66613272651d242eadfc836c68a1ac1 so merging this shouldn't be a problem (after a rebase).

Shnatsel:
I'm not sure this is worth the trouble right now. flif.rs is still very limited and will likely have API breaks as it evolves. Besides, the last activity on flif.rs repository was almost a year ago.

fintelia:
The flif.rs crate hasn't been update since 2019. Going to go ahead and close this

Shnatsel:
FLIF is effectively superseded by the lossless mode of JPEG XL.

PR #933, opened 2019-05-04
rework open() to use magic bytes with a fallback on file extension, fixes #577

### LINKED ISSUE #577: Support guessing format when loading from file
`image::open` only uses the filename extension.  `image::load_from_memory` can guess the format based on magic numbers, but requires loading the image into memory first in a separate step.  Please consider adding a function to read an image from a file but guess the format based on magic numbers.

martinlindhe:
Not fully satisfied with the function name but I'm out of ideas.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -927,8 +927,9 @@ pub fn load<R: Read + Seek>(r: R, format: ImageFormat) -> ImageResult<DynamicIma
 
 fn load_guess<R: Read + Seek>(mut r: R) -> ImageResult<DynamicImage> {
     let mut buffer: [u8; 16] = [0u8; 16];
+    let cur = r.seek(SeekFrom::Current(0))?;
     r.read_exact(&mut buffer[..])?;
-    r.seek(SeekFrom::Start(0))?;
+    r.seek(SeekFrom::Start(cur))?;
     load(r, guess_format(&buffer)?)
 }
 
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -1,7 +1,7 @@
 use num_iter;
 use std::fs::File;
 use std::io;
-use std::io::{BufRead, BufReader, BufWriter, Read, Seek, SeekFrom, Write};
+use std::io::{BufRead, BufReader, BufWriter, Seek, SeekFrom, Write};
 use std::path::Path;
 use std::u32;
 
@@ -776,13 +776,7 @@ fn open_magic_impl(path: &Path) -> ImageResult<DynamicImage> {
         Ok(f) => f,
         Err(err) => return Err(image::ImageError::IoError(err)),
     };
-    let mut fin = BufReader::new(fin);
-
-    let mut buffer: [u8; 16] = [0u8; 16];
-    fin.read(&mut buffer[..])?;
-    fin.seek(SeekFrom::Start(0))?;
-
-    load(fin, guess_format(&buffer)?)
+    load_guess(BufReader::new(fin))
 }
 
 /// Read the dimensions of the image located at the specified path.
@@ -937,6 +931,15 @@ pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<Dynamic
     }
 }
 
+fn load_guess<R: BufRead + Seek>(mut r: R) -> ImageResult<DynamicImage> {
+    let mut buffer: [u8; 16] = [0u8; 16];
+    r.read(&mut buffer[..])?;
+    r.seek(SeekFrom::Start(0))?;
+
+    let format = guess_format(&buffer)?;
+    load(r, format)
+}
+
 static MAGIC_BYTES: [(&'static [u8], ImageFormat); 17] = [
     (b"\x89PNG\r\n\x1a\n", ImageFormat::PNG),
     (&[0xff, 0xd8, 0xff], ImageFormat::JPEG),
@@ -962,7 +965,8 @@ static MAGIC_BYTES: [(&'static [u8], ImageFormat); 17] = [
 /// Makes an educated guess about the image format.
 /// TGA is not supported by this function.
 pub fn load_from_memory(buffer: &[u8]) -> ImageResult<DynamicImage> {
-    load_from_memory_with_format(buffer, try!(guess_format(buffer)))
+    let b = io::Cursor::new(buffer);
+    load_guess(b)
 }
 
 /// Create a new image from a byte slice
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -767,19 +767,6 @@ fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
     load(BufReader::new(&fin), format)
 }
 
-/// Open the image located at the path specified.
-/// The image's format is determined from the file header magic bytes.
-pub fn open_magic<P>(path: P) -> ImageResult<DynamicImage>
-where
-    P: AsRef<Path>,
-{
-    let fin = match File::open(path) {
-        Ok(f) => f,
-        Err(err) => return Err(image::ImageError::IoError(err)),
-    };
-    load_guess(BufReader::new(fin))
-}
-
 /// Read the dimensions of the image located at the specified path.
 /// This is faster than fully loading the image and then getting its dimensions.
 pub fn image_dimensions<P>(path: P) -> ImageResult<(u32, u32)>
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -736,8 +736,14 @@ fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
 
     match load_guess(BufReader::new(&fin)) {
         Ok(im) => return Ok(im),
-        Err(_e) => {
-            // println!("load_guess failed loading {}: {} - failling back to extension format matching", path.display(), e);
+        Err(e) => {
+            match e {
+                image::ImageError::UnsupportedError(_) => {
+                    // fall-thru to extension based detection (should only happen for TGA images)
+                    // println!("load_guess failed loading {}: {}", path.display(), e);
+                }
+                _ => return Err(e),
+            };
         }
     };
 
@@ -1112,13 +1118,5 @@ mod test {
         let im_path = "./tests/images/jpg/progressive/cat.jpg";
         let dims = super::image_dimensions(im_path).unwrap();
         assert_eq!(dims, (320, 240));
-    }
-
-    #[cfg(feature = "jpeg")]
-    #[test]
-    fn open_magic() {
-        let im_path = "./tests/images/jpg/progressive/cat.jpg";
-        let im = super::open_magic(im_path).unwrap();
-        assert_eq!(im.raw_pixels().len(), 230400);
     }
 }
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -767,11 +767,6 @@ pub fn open_magic<P>(path: P) -> ImageResult<DynamicImage>
 where
     P: AsRef<Path>,
 {
-    // thin wrapper function to strip generics before calling open_magic_impl
-    open_magic_impl(path.as_ref())
-}
-
-fn open_magic_impl(path: &Path) -> ImageResult<DynamicImage> {
     let fin = match File::open(path) {
         Ok(f) => f,
         Err(err) => return Err(image::ImageError::IoError(err)),
@@ -933,11 +928,9 @@ pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<Dynamic
 
 fn load_guess<R: BufRead + Seek>(mut r: R) -> ImageResult<DynamicImage> {
     let mut buffer: [u8; 16] = [0u8; 16];
-    r.read(&mut buffer[..])?;
+    r.read_exact(&mut buffer[..])?;
     r.seek(SeekFrom::Start(0))?;
-
-    let format = guess_format(&buffer)?;
-    load(r, format)
+    load(r, guess_format(&buffer)?)
 }
 
 static MAGIC_BYTES: [(&'static [u8], ImageFormat); 17] = [
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -1,7 +1,7 @@
 use num_iter;
 use std::fs::File;
 use std::io;
-use std::io::{BufRead, BufReader, BufWriter, Seek, SeekFrom, Write};
+use std::io::{BufReader, BufWriter, Read, Seek, SeekFrom, Write};
 use std::path::Path;
 use std::u32;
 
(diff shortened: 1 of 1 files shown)


birktj:
I think it would be better to be generic in `Read + Write`. We can then have two functions like:
```rust
fn load<R: Read + Seek>(reader: R, format: ImageFormat) -> ImageResult<DynamicImage> {...}
fn load_guess<R: Read + Seek>(reader: R) -> ImageResult<DynamicImage> {...}
```
We could then base the other loading functions, `open` and `load_from_memory` on these two.

birktj:
In fact I think it would be best with `load`, and `load_with_format` as names, but I guess that would be too much of a breaking change.

martinlindhe:
Read or BufRead?

I'm afraid changing `pub fn load<R: BufRead + Seek>` to Read might break stuff elsewhere?

martinlindhe:
Reworked into a generic load_guess() hopefully as you imagined

birktj:
@martinlindhe looks good! Just some small things

src/dynimage.rs
```diff
@@ -761,13 +761,31 @@ fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
...
+/// The image's format is determined from the file header magic bytes.
+pub fn open_magic<P>(path: P) -> ImageResult<DynamicImage>
+where
+    P: AsRef<Path>,
+{
+    // thin wrapper function to strip generics before calling open_magic_impl
+    open_magic_impl(path.as_ref())
+}
+
+fn open_magic_impl(path: &Path) -> ImageResult<DynamicImage> {
```
birktj: What purpose does`open_magic_impl` serve, can't the body of this function be used directly in `open_magic`?

I also wonder if we should use magic bytes as primary way of detecting file type in `open`, and use a file-extension fall-back if magic byte detection failed. This should be more versatile than using file-extensions.

src/dynimage.rs
```diff
@@ -910,6 +928,15 @@ pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<Dynamic
     }
 }
 
+fn load_guess<R: BufRead + Seek>(mut r: R) -> ImageResult<DynamicImage> {
+    let mut buffer: [u8; 16] = [0u8; 16];
+    r.read(&mut buffer[..])?;
```
birktj: This should probably use `read_exact` so that the loading doesn't fail if `read` only reads a few bytes.

src/dynimage.rs
martinlindhe: open_magic_impl() is just a pattern I copied from how [open() was implemented](https://github.com/image-rs/image/blob/master/src/dynimage.rs#L723).
Indeed code can be used directly in open_magic(), thanks for noticing!

I think magic bytes detection would be a good default, but should I try address that in this PR or a later one?

src/dynimage.rs
martinlindhe: When I wrote that I was considering what would happen if input file is < 16 bytes, then .read() should fill buffer with available bytes.
Since buffer is 16 bytes big, I think read_exact() will error then.
But I'm just speculating about the behavior really, didn't test any of it.

src/dynimage.rs
birktj: I don't see why you shouldn't do it in this PR, we could simply wait a few days before merging to see if anyone has any objections.

src/dynimage.rs
martinlindhe: Pushed change to use read_exact

birktj: I am pretty sure that `read_exact` will error, but you would probably get a similar error from all of the decoders. I guess no image file can be less than 16 bytes long?

martinlindhe: Yea I don't think I a valid image file could be <16 bytes.

src/dynimage.rs
martinlindhe: Okay, sure

martinlindhe:
Reworked open() to default to magic bytes detection and fall back to extension loading.

The magic detection currently errors on the following from the test files, looking at some of them now...

```
load_guess failed loading ./tests/images/bmp/images/Bad_clrsUsed.bad_bmp: Format error: Palette size 4278190086 exceeds maximum size for BMP with bit count of 8 - failling back to extension format matching
load_guess failed loading ./tests/images/bmp/images/Bad_width.bad_bmp: Format error: Image too large - failling back to extension format matching
load_guess failed loading ./tests/images/bmp/images/Bad_height.bad_bmp: Format error: Image too large - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/cbw8.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/ccm8.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/ctc24.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/ubw8.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/ucm8.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/utc16.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/utc24.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tga/testsuite/utc32.tga: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tiff/testsuite/hpredict.tiff: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tiff/testsuite/hpredict_packbits.tiff: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
load_guess failed loading tests/images/tiff/testsuite/lenna.tiff: The Decoder does not support the image format `Unsupported image format` - failling back to extension format matching
```

birktj:
The `.bad_bmp` files have that extension because they are not supposed to be loaded I think. `tga` cannot be detected by magic bytes and the magic byte constants for tiff are wrong.

martinlindhe:
Yea, looking at the TIFF ones some more

birktj:
The `.` should be `\x00` I think

martinlindhe:
You are correct, I just discovered the same

martinlindhe:
Correct tiff headers from our samples:

```
tests/images/tiff/testsuite/hpredict.tiff:
00000000: 4949 2a00 5207 0000 8000 2050 3824 160d  II*.R..... P8$..


tests/images/tiff/testsuite/hpredict_packbits.tiff:
00000000: 4949 2a00 c00e 0000 8100 fd00 ff10 7500  II*...........u.


tests/images/tiff/testsuite/lenna.tiff:
00000000: 4d4d 002a 000c 0008 e289 7de2 897d df89  MM.*......}..}..
```

src/dynimage.rs
```diff
@@ -733,7 +733,13 @@ fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
         Ok(f) => f,
         Err(err) => return Err(image::ImageError::IoError(err)),
     };
-    let fin = BufReader::new(fin);
+
+    match load_guess(BufReader::new(&fin)) {
+        Ok(im) => return Ok(im),
+        Err(_e) => {
```
birktj: I am not completely sure this is correct. What happens if there is a decoding error later in the image, wouldn't it try to read it again using the extension, and starting where-ever the first decoder stopped? You probably need to duplicate some functionality from `load_guess`?

src/dynimage.rs
```diff
@@ -758,7 +764,20 @@ fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
         }
     };
 
-    load(fin, format)
+    load(BufReader::new(&fin), format)
+}
+
+/// Open the image located at the path specified.
+/// The image's format is determined from the file header magic bytes.
+pub fn open_magic<P>(path: P) -> ImageResult<DynamicImage>
```
birktj: With magic bytes detection in `open`, do we need this function?

martinlindhe: Nope, haha!

src/dynimage.rs
martinlindhe: Not sure I follow what you mean here.
Like, if load_guess() guesses its a PNG and we return Ok(im) here, gut the png data stream is corrupt ?

martinlindhe: Do you mean like if load_guess() detects PNG,  but fails to decode it and load_guess() then returns an Err here,
and further down we will again try to decode this png but this time because of the file extension?

birktj: Exactly!

martinlindhe: Alright, yea in the error path it would do 2x work.
I'll see if i can figure something better out.

martinlindhe: Ok resolved this too with a new commit.

Verified the only files who fallthru are the .tga files as expected since they're not detected by magic bytes.

birktj: With #928 this should work, but unfortunately `UnsupportedError` is currently used for more than just unsupported formats.

martinlindhe: Lets pause this until #928 lands then.

martinlindhe:
This PR should not be merged until #928 has landed.

src/dynimage.rs
```diff
@@ -910,14 +922,21 @@ pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<Dynamic
     }
 }
 
+fn load_guess<R: Read + Seek>(mut r: R) -> ImageResult<DynamicImage> {
+    let mut buffer: [u8; 16] = [0u8; 16];
+    r.read_exact(&mut buffer[..])?;
+    r.seek(SeekFrom::Start(0))?;
```
HeroicKatora: This doesn't seem like the proper behaviour for arbitrary `Seek` impls. The file could be intended to start at a position other than `0`. It should rather use the sequence:

```
let cur = r.seek(SeekFrom::Current(0))?;
r.read_exact(&mut buffer[..])?;
r.seek(SeekFrom::Start(cur))?;
```

For a fixed-length sequence however, it could also be an option to not require `Seek` at all and utilize [`std::io::Chain`](https://doc.rust-lang.org/std/io/struct.Chain.html) on the inspected data.

martinlindhe: Good point, I didn't consider this. I went ahead with your suggested code.

HeroicKatora:
Not that it concerns this PR directly but wait a minute, why do we have `load_from_memory_with_format` at all? It is covered by `load(io::Cursor::new(buffer))` as evident from the implementation thus being just a very, very lengthy name for something that should be a documentation item.

HeroicKatora:
With these changes I think there should be exactly two based on `Read + Seek`: one which guesses the format and one which doesn't but takes an argument instead. And then one which falls back to filepath extensions for some extra fanciness.

And I'd similarly expect one for `save` to also have a variant that is *not* based on filepaths but takes an arbitrary `impl Write`. For the moment this only exists on an `impl` of DynamicImage as far as I can tell. That has now lead to at least two confused users in the last two months, most recently #967 .

martinlindhe:
@HeroicKatora agreed. I could add a commit dropping load_from_memory_with_format() to this PR if desired.

Also, branch rebased to master.

HeroicKatora:
Something went wrong during rebase, the array has now 18 elements instead of 17.

src/dynimage.rs
```diff
@@ -913,14 +925,23 @@ pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<Dynamic
...
+
 static MAGIC_BYTES: [(&'static [u8], ImageFormat); 17] = [
     (b"\x89PNG\r\n\x1a\n", ImageFormat::PNG),
     (&[0xff, 0xd8, 0xff], ImageFormat::JPEG),
     (b"GIF89a", ImageFormat::GIF),
     (b"GIF87a", ImageFormat::GIF),
     (b"RIFF", ImageFormat::WEBP), // TODO: better magic byte detection, see https://github.com/image-rs/image/issues/660
-    (b"MM.*", ImageFormat::TIFF),
-    (b"II*.", ImageFormat::TIFF),
+    (b"WEBP", ImageFormat::WEBP),
```
birktj: I think this line shouldn't be here?

HeroicKatora:
@martinlindhe I'm fully for removing the outdated version, just make sure to add ~10 lines of explicit mention and example in the more generic ones (maybe just the current implementation). Don't know about other's opinions on this though, let's consider it for a while.

src/dynimage.rs
martinlindhe: rebase went wrong, fixed.

thanks for noticing!

martinlindhe:
@HeroicKatora fixed the borked rebase, sorry about that!

Okay, let's do that one in a later PR then, hopefully we can land this eventually

src/dynimage.rs
HeroicKatora: You can comfortably fit a pnm in less than that space even with an unecessary comment on its content: (15 bytes)

```
P1\n1 1#white\n\n1
```

martinlindhe: @HeroicKatora What do you think, should we revert this behaviour then in order to handle a 15 byte PBM ?

birktj: You could simply shorten some of the magic byte constants so that none of them are more than 5 bytes. They should still be uniquely identifiable I think? If this turns out to be a problem later, we should probably consider a slightly smarter system.

birktj: An other solution is to use something like and simply fill as much as possible under 16 bytes:
```rust
let mut buf = [0; 16];
let mut len = 0;
while !buf.is_empty() {
    match self.read(&mut buf[len..]) {
        Ok(0) => break,
        Ok(n) => len += n,
        Err(ref e) if e.kind() == ErrorKind::Interrupted => (),
        Err(e) => return Err(e),
    }
}
```

birktj: Probably even better would be to use:
```rust
match r.read_exact(&mut buf) {
    Ok(()) => (),
    Err(e) if e.kind() == ErrorKind::UnexpectedEof => (),
    Err(e) => return Err(e.into()),
}
```

HeroicKatora: No, don't revert it. There are more options than those mentioned by @birktj . You could wrap the buffer into a `io::Cursor` to read potentially less bytes and use `take` to limit the maximum number:

```
    let mut buffer: [u8; 16] = [0u8; 16];
    let mut out = io::Cursor::new(&mut buffer[..]);
    let cur = r.seek(SeekFrom::Current(0))?;
    let len = io::copy(&mut cur.take(16), &mut out)?;
    r.seek(SeekFrom::Start(cur))?;
    load(r, guess_format(&buffer[..len as usize])?)
```

HeroicKatora: No live updates on review conversations so now you have ample alternatives @martinlindhe

martinlindhe: I'll hopefully get back to this shortly- if you are reading this from the future and I didn't, please ping me!

HeroicKatora:
Closed by: #1014

PR #937, opened 2019-05-05
Fix webp magic bytes

### LINKED ISSUE #660: Issue with guessing format of WEBP images
Presently, using `image::load_from_memory` will make an attempt to guess the image's format using magic numbers at the beginning of the image's bytes. WebP's magic bytes are defined here:

```
static MAGIC_BYTES: [(&'static [u8], ImageFormat); 11] = [
    //omitted other definitions
    (b"WEBP", ImageFormat::WEBP),
    //omitted other definitions

];
```
This would cause any byte sequence starting with the bytes for "WEBP" to be recognized as WebP images; however, WebP images do not actually begin with the ASCII bytes "WEBP", but instead, their file header looks like this:
```
 0                   1                   2                   3
 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|      'R'      |      'I'      |      'F'      |      'F'      |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                           File Size                           |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|      'W'      |      'E'      |      'B'      |      'P'      |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
(taken from the [WebP Container Specification](https://developers.google.com/speed/webp/docs/riff_container#riff_file_format))
Because of this, WEBP images' bytes are not properly recognized by `load_from_memory`.

birktj:
Magic bytes for webp are wrong.
This fixes #660


martinlindhe:
If you have a small valid webp image handy, could you possibly also add this to this PR into tests/images/webp or something? Seems like we currently lack any samples.

martinlindhe:
Also, I think in the long run this may not be an optimal solution as RIFF is a container format. However I much prefer header based format detection than full blown format parsers, but again we may have to.

For example TGA cannot simply be detected with magic bytes, one would need to parse more of the file to conclude its valid TGA.

birktj:
I agree, but I think that would be for a later PR. For now webp is the only image format supported by this library where the file can start with `RIFF` so I assume this should be safe for now.

martinlindhe:
Yea, for sure. The only other format using RIFF container I'm aware of that's even remotely resembling our use case is Windows 95 animated cursors (.ANI).

PR #941, opened 2019-05-05
Add templates for issues and pull request

HeroicKatora:
Github documentation makes it harder than necessary to create this
outside the webinterface but hopefully this is correct.

### DIFF
--- a/docs/ISSUE_TEMPLATE/bug-template.md
+++ b/docs/ISSUE_TEMPLATE/bug-template.md
@@ -1,0 +1,19 @@
+----
+labels: bug
+name: Bug report
+about: Reporting a bug
+----
+
+This happens in ,...
+
+## Expected
+
+What behaviour should have happened
+
+## Actual behaviour
+
+What did happen
+
+## Reproduction steps
+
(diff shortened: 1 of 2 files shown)


martinlindhe:
Trying to follow (possible outdated) shattered documents about this, I feel with you.

At [one place](https://github.blog/2016-02-17-issue-and-pull-request-templates/) they said

> Pull Request templates follows the same pattern: add a file called PULL_REQUEST_TEMPLATE to the root directory of your repository.
>
> If you’re worried about the added clutter in the root directory of your project, we also added support for a .github/ folder. You can put CONTRIBUTING.md, ISSUE_TEMPLATE.md, and PULL_REQUEST_TEMPLATE.md files in .github/ and everything will work as expected.

martinlindhe:
It is somewhat ironic how they insist we should use their web tools rather than just adding the proper files to the repository directly.

HeroicKatora:
I see no direct way to test these beforehand, but I have confirmed the general format by comparison with another repo of mine where this already more or less works.

PR #943, opened 2019-05-05
Make hdr module more consistent with others

fintelia:
There is no need to have "hdr" multiple times in the filenames...


src/lib.rs
```diff
@@ -14,6 +14,7 @@ extern crate lzw;
 extern crate num_iter;
 extern crate num_rational;
 extern crate num_traits;
+extern crate scoped_threadpool;
```
HeroicKatora: This dependency is included only for `hdr` and wouldn't work in `src/lib.rs`. Probably CI will flag this as well.

PR #944, opened 2019-05-05
Fix a regression in the DXT encoder.

CensoredUsername:
I think someone made an accidental typo while doing cargo fmt and clippy fixing. This error would occasionally cause 2 pixels in a DXT block to be encoded wrong.

This commit simply restores the line of code to how it was before.

If necessary: I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.

### DIFF
--- a/src/dxt.rs
+++ b/src/dxt.rs
@@ -618,7 +618,7 @@ fn encode_dxt_colors(source: &[u8], dest: &mut [u8]) {
             swap(&mut color0, &mut color1);
             // Indexes are packed 2 bits wide, swap index 0/1 but preserve 2/3.
             let filter = (chosen_indices & 0xAAAA_AAAA) >> 1;
-            chosen_indices ^= filter ^ 0x555_5555;
+            chosen_indices ^= filter ^ 0x5555_5555;
         }
     } else if !chosen_use_0 {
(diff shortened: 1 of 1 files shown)


HeroicKatora:
A way to prevent a regression like this in the future would be to include some images in `./tests/{images,reference}`. That would be nice.

HeroicKatora:
See https://github.com/image-rs/image/pull/937 for a PR that added such tests for `webp`.

HeroicKatora:
Oh, nvm. DXT is not autodetected via file extension nor via magic bytes.

CensoredUsername:
The DXT backend does not support any actual file format, just the encoding/decoding scheme, which makes integration in the current test suite complicated. I previously planned to add a proper .dds implementation but I ended up not having the time for it unfortunately.

HeroicKatora:
@CensoredUsername I didn't realize you hadn't joined the new organization yet, so I forgot to ask: You've been invited to the `image-rs` collaborators team, which allows you access to this and the other repos if you want. See [the contribution policy](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/image-rs/organization/blob/master/CODE_OF_CONDUCT.md).

PR #945, opened 2019-05-05
Always use custom struct for ImageDecoder::Reader

fintelia:
As @HeroicKatora pointed out, changing the Reader type of a Decoder is a breaking change. Thus have all Decoders use an opaque struct as their Reader type instead of `Cursor<Vec<u8>>`. This way we can trigger all the breakage now and be able to add incremental decoding without breaking downstream users.

Side note: In this PR I made all newly added types follow the current Rust naming convention of capitalizing only the first letter of an acronym (i.e. GifReader rather than GIFReader). I can change this if desired.

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1,6 +1,5 @@
-use std::cmp;
-use std::io;
-use std::io::{Cursor, Read, Seek, SeekFrom};
+use std::{cmp, mem};
+use std::io::{self, Cursor, Read, Seek, SeekFrom};
 use std::iter::{repeat, Iterator, Rev};
 use std::slice::ChunksMut;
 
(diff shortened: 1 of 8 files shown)


src/bmp/decoder.rs
```diff
@@ -1255,8 +1254,24 @@ impl<R: Read + Seek> BMPDecoder<R> {
     }
 }
 
+/// Wrapper struct around a `Cursor<Vec<u8>>`
+pub struct BmpReader(Cursor<Vec<u8>>);
```
birktj: What about doing something like:
```rust
pub struct BmpReader<R> {
    inner: Cursor<Vec<u8>>,
    _phantom: PhantomData<R>,
}
```
Because right now when we update these reader types to take the inner reader as an argument it will still be a breaking change.

HeroicKatora:
@fintelia Since this change something in `hdr_decoder.rs`, it collided with the rename of that file. I think this is the right way.

PR #949, opened 2019-05-06
README: add crates.io badge and drop the usage section, fixes #942

### LINKED ISSUE #942: Update usage section in README
The recommended way of adding dependency from the README is currently 
```toml
[dependencies]
image = "*"
```
at https://github.com/image-rs/image/blame/master/README.md#L18-L19

From [The Cargo Book:](https://doc.rust-lang.org/cargo/faq.html#can-libraries-use--as-a-version-for-their-dependencies)

> Can libraries use * as a version for their dependencies?
>
> As of January 22nd, 2016, crates.io rejects all packages (not just libraries) with wildcard dependency constraints.
>
> While libraries can, strictly speaking, they should not. A version requirement of * says “This will work with every version ever,” which is never going to be true. Libraries should always specify the range that they do work with, even if it’s something as general as “every 1.x.y version.”

martinlindhe:


README.md
```diff
@@ -1,4 +1,4 @@
-# Image [![Build Status](https://travis-ci.org/image-rs/image.svg?branch=master)](https://travis-ci.org/image-rs/image) [![Gitter](https://badges.gitter.im/image-rs/image.svg)](https://gitter.im/image-rs/image?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
+# Image [![crates.io](https://img.shields.io/crates/v/image.svg)](https://crates.io/crates/image) [![Build Status](https://travis-ci.org/image-rs/image.svg?branch=master)](https://travis-ci.org/image-rs/image) [![Gitter](https://badges.gitter.im/image-rs/image.svg)](https://gitter.im/image-rs/image?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
 Maintainers: @nwin, @ccgn
```
birktj: I guess this is kinda out of scope for this PR, but we should probably change these as both nwin an ccgn have been inactive for some time.

martinlindhe: Agree. Maybe just drop the Maintainers line, or add a link to https://github.com/orgs/image-rs/people or something?

HeroicKatora: Link to the team sounds like a possible solution?

martinlindhe: I opened #950 for the maintainer part. Could we merge this PR ?

HeroicKatora: Sure

PR #953, opened 2019-05-07
Associated constants

### LINKED ISSUE #835: Associated constants would make the API more idiomatic
Now that [this feature](https://rust-lang-nursery.github.io/edition-guide/rust-2018/trait-system/associated-constants.html) is available, it is a better option.

I'm looking specifically at the `Pixel` trait, but this probably applies in other places too.

Currently, the trait is like this:

```rust
pub trait Pixel: Copy + Clone {
    type Subpixel: Primitive;
    fn channel_count() -> u8;
    fn color_model() -> &'static str;
    fn color_type() -> ColorType;
    // etc.
}
```

With associated consts:

```rust
pub trait Pixel: Copy + Clone {
    type Subpixel: Primitive;
    const CHANNEL_COUNT: u8;
    const COLOR_MODEL: &'static str;
    const COLOR_TYPE: ColorType;
    // etc.
}
```

This would be a breaking API change. Keeping the methods, deprecated, as part of the trait definition (with an implementation returning the constant) would prevent breaking code using the trait, but not implementing it (I don't know how common it is in practice to implement traits like this outside the crate).

Practically speaking, this has the advantage of allowing the value to be used in constant contexts, which could be useful. Making the methods `const fn`s should do that too, but apparently `const fn`s can't be part of a trait definition, and are still an unstable feature.

I don't know if this is worth the potential API breakage, but it's at least something to keep in mind for any future API redesign.

lazypassion:
Closes #835


src/buffer.rs
```diff
@@ -20,7 +20,10 @@ pub trait Pixel: Copy + Clone {
     type Subpixel: Primitive;
 
     /// Returns the number of channels of this pixel type.
+    #[deprecated(note="please use CHANNEL_COUNT associated constant")]
     fn channel_count() -> u8;
```
birktj: This should probably be a default impl like:
```rust
fn channel_count() -> u8 {
    return Self::CHANNEL_COUNT;
}
```

lazypassion: Should i do the same for the other two constants?

src/color.rs
```diff
@@ -77,12 +77,18 @@ impl<T: Primitive + 'static> Pixel for $ident<T> {
     fn channel_count() -> u8 {
```
birktj: With the default impls, we probably don't need to implement these manually.

birktj:
Looks good, just one more small thing before it is ready to be merged.

birktj:
What do you say @HeroicKatora? I am in favor of this change.

HeroicKatora:
Seems fine :+1:

Issue #956, opened 2019-05-09
Add alpha count associated constant to Pixel trait


lazypassion:
I would like to be able to get the number of alpha values of the Pixel type found in an image buffer. 

My specific use case for this functionality is: 
This is to write generic code over gray scale and color images in an imperative style as opposed to using `apply_with_alpha`

```rust
pub fn invert_mut<I, P, S>(image: &mut I) 
where
    I: GenericImage<Pixel = P>,
    P: Pixel<Subpixel = S> + 'static,
    S: Primitive + 'static,
{
    let channel_count = <P>::CHANNEL_COUNT;
    // example of alpha_count 
    let alpha_count = <P>::ALPHA_COUNT;
    let slice = channel_count - alpha_count;

    for pixel in image.pixels_mut() {
        for value in p.data[..slice].iter_mut() {
            *value = 255 - *value;
        }
    }
}
```
The alternative using `apply_with_alpha` looks like this:
```rust
pub fn invert_mut<I, P, S>(image: &mut I) 
where
    I: GenericImage<Pixel = P>,
    P: Pixel<Subpixel = S> + 'static,
    S: Primitive + 'static,
{
    let map_color = |x| MAX_VALUE - x;
    let map_alpha = |x| x; //identity transform
    let (width, height) = image.dimensions();
    for x in 0..width {
        for y in 0..height {
            image.get_pixel_mut(x, y)
                .apply_with_alpha(map_color, map_alpha);
        }
    }
}
```
This might also be useful if someone needs to get the alpha count in the case there is an alpha value for each color channel.

fintelia:
Do you think that the Pixel trait interface is otherwise relatively reasonable? Looking at it, I'm somewhat skeptical that there is really all that much generic code that can be written against it without making a bunch more assumptions. This proposal is a (small) step in the right direction, but much more drastic API changes might be necessary. Given that we're not yet at 1.0, now is the time to get the interface right.

lazypassion:
I think its overall pretty fine. I think that some of the methods shouldn't be mandatory. For example if it were possible to get the `alpha_count` then methods like `apply_with_alpha` wouldn't be necessary, but nice to have. I don't think every implementor for that trait would necessarily want to use that method. I think this could extend to almost all of the methods, but I'm not sure what kind of person would implement a very bare bones Pixel type. I hope this makes sense lol.

HeroicKatora:
[...]

PR #957, opened 2019-05-09
Added apply_without_alpha and map_without_alpha to Pixel trait with their implementations and tests

lazypassion:


src/color.rs
```diff
@@ -174,6 +174,22 @@ impl<T: Primitive + 'static> Pixel for $ident<T> {
         }
     }
 
+    fn map_without_alpha<F>(&mut self, f: F) -> Self
```
HeroicKatora: After the conclusion in #956 these should be default implementations in the trait in `src/buffer.rs` instead, right?

HeroicKatora: Also `(*self).clone()` can be simply `*self`, since `trait Pixel: Copy + Clone`.

HeroicKatora: And `map_without_alpha` doesn't require `&mut self` but `&self` suffices.

lazypassion: Why does making `*self` only work instead of `(*self).clone()`?

HeroicKatora: Because of the marker trait `Copy` which allows the compiler to make bytewise copies without invoking any additional code. It will then always copy the value instead of moving it from the place. This allows simply reading the value through a reference (because it will leave the original value unmoved), instead of requiring cloning it.

lazypassion: Ohhhh, haha I forgot what Copy meant. Thanks

HeroicKatora:
Thank you!

PR #960, opened 2019-05-25
Remove pixels_mut and deprecate unsafe_get_pixel/unsafe_set_pixel

fintelia:
This is a continuation of the effort to purge unsafe code from the image crate.

### DIFF
--- a/src/flat.rs
+++ b/src/flat.rs
@@ -1418,7 +1418,7 @@ impl PartialOrd for NormalForm {
 mod tests {
     use super::*;
     use buffer::GrayAlphaImage;
-    use color::Rgb;
+    use color::{LumaA, Rgb};
 
     #[test]
     fn aliasing_view() {
@@ -1441,6 +1441,35 @@ mod tests {
            .inspect(|pixel| assert!(pixel.2 == Rgb([42, 42, 42])))
            .count();
        assert_eq!(pixel_count, 100*100);
+    }
+
+    #[test]
+    fn mutable_view() {
+        let mut buffer = FlatSamples {
+            samples: [0; 18],
+            layout: SampleLayout {
+                channels: 2,
+                channel_stride: 1,
+                width: 3,
+                width_stride: 2,
+                height: 3,
+                height_stride: 6,
+            },
+            color_hint: None,
+        };
+
+        {
+            let mut view = buffer.as_view_mut::<LumaA<usize>>()
+                .expect("This should be a valid mutable buffer");
+            assert_eq!(view.dimensions(), (3, 3));
+            for i in 0..9 {
+                *view.get_pixel_mut(i % 3, i / 3) = LumaA([2 * i as usize, 2 * i as usize + 1]);
+            }
+        }
+
+        buffer.samples.iter()
+            .enumerate()
+            .for_each(|(idx, sample)| assert_eq!(idx, *sample));
     }
 
     #[test]
--- a/src/flat.rs
+++ b/src/flat.rs
@@ -1444,37 +1444,6 @@ mod tests {
     }
 
     #[test]
-    fn mutable_view() {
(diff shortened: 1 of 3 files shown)


src/flat.rs
```diff
@@ -1443,37 +1443,6 @@ mod tests {
        assert_eq!(pixel_count, 100*100);
     }
 
-    #[test]
-    fn mutable_view() {
```
HeroicKatora: That test had did some assertions other than constructing the mutable iterator. Can you replace

```
            #[allow(deprecated)]
            let pixel_count = view.pixels_mut()
                .enumerate()
                .map(|(idx, (_, _, pixel))| *pixel = LumaA([2*idx, 2*idx + 1]))
                .count();
```

with:

```
assert_eq!(view.dimensions(), (3, 3));
for idx in 0..9 {
    *view.get_pixel_mut(idx % 3, idx / 3) = LumaA([2*idx, 2*idx + 1]);
}
```

src/image.rs
```diff
@@ -585,6 +542,7 @@ pub trait GenericImageView {
     /// Returns the pixel located at (x, y)
     ///
     /// This function can be implemented in a way that ignores bounds checking.
```
birktj: Should we maybe add some comments about the deprecation here?

fintelia: Probably, though I'm not quite sure what to say

HeroicKatora: I would imagine something along the lines of 'safety preconditions are hard to uphold in generic code'. But the function is actually quite clear about it: Caller must ensure that the coordinates are in bound while callee must not rely on other preconditions. 

Is there another reason remove these functions other than working towards marking the crate as `#[forbid(unsafe_code)]`? If not then I have to remark that the performance impact of that seems still unclear to me while performance regressions are clearly not in our interest and should, imho, only occur when we fix *actual* undefined behaviour arising from wrong use of `unsafe`.

fintelia: Yes, the point of these deprecations is to eventually have `#[forbid(unsafe_code)]`. It is definitely possible that some user code will experience performance degradations moving from the unsafe versions of these functions to the safe ones, but worst case users could always implement the functions themselves. For ImageBuffer (the only type that has an implementation of these functions) a custom implementation would be a single line that deref's the object and then calls get_unchecked on the underlying slice.

fintelia: I should probably also add that IMO this library doesn't necessarily need the highest possible performance because anyone who is doing super performance critical image manipulation in Rust is either going to use bindings some highly optimized SIMD/GPGPU library or is going to want to write such vector code themselves

HeroicKatora: I disagree on the performance side: it should be possible to utilize the abstractions provided by the library within high performance code. Nevertheless, there are good reasons why not *all* parts need to be optimized specifically for such a use case if that distinction is reasonably clear. In conclusion I'm not opposed to deprecating the `unsafe` methods on the trait for as long as `ImageBuffer` still offers somewhat of an opportunity for fast code. Which it does because it offers an interface to generic array elements that can be used in SIMD/SSE/... and an `ffi` interface with `flat`.

So yes, deprecating these functions is fine and the deprecation note should likely point towards utilizing the additional guarantees provided by and utilized in `ImageBuffer`/`flat::*` directly.

HeroicKatora: Can we get this as a note in then? I'd really prefer this PR to be in the next release, more or less  get this merged and then see if any updates on dependencies are outstanding and then release it. Proposed deprecation note:

```
Offers little advantage for most types. If you must, prefer dedicated methods on the specific image type instead.
```

If someone finds a case where there is more than little advantage they'll likely tell us.

HeroicKatora:
Took a while to get all the `unsafe` removals merged in waves but very nice :+1:

PR #970, opened 2019-06-05
Backport of fixes to release 0.21.2

HeroicKatora:
Backports the fixes that happened on `master` to the `0.21` version 
series. Due to the pressure from memory-safety issues surrounding
the release of `0.21.1` the master branch had already significant
interface changes applied. However, [`crates.io`](https://crates.io/crates/image)
indicates that a significant amount of users lags behind in major versions 
of the crate. 

Reducing the ecosystem burden by trying to maintain at least this 
version for a longer amount should give more users the chance to 
update to a version that fixed the soundness issues. It should be 
stressed that this does **NOT** remove the `PixelMut` type despite 
its issues but 'only' rectifies internal issues It was already deprecated
and such a change absolutely requires a major version. And there are
no promises attached to how long this branch is maintained.

Release of the newer interface with `0.22`, which removes that 
outstanding unsafe interface, will follow shortly.


src/webp/vp8.rs
```diff
@@ -1421,19 +1483,19 @@ impl<R: Read> VP8Decoder<R> {
     }
 
     /// Decodes the current frame and returns a reference to it
-    pub fn decode_frame(&mut self) -> io::Result<&Frame> {
-        try!(self.read_frame_header());
+    pub fn decode_frame(&mut self) -> ImageResult<&Frame> {
```
HeroicKatora: This is [part of the public interface](https://docs.rs/image/0.21.1/image/webp/vp8/struct.VP8Decoder.html). Need to minimally adjust that change to not change the return type here for now.

fintelia:
I think this plan makes a lot of sense. We have a bunch of breaking changes planned/merged on master so it is good to have one final release without them and a lingering branch in case there are any further patches we want to backport.

birktj:
Seems like a good idea. This makes me wonder if we should have stricter commit policies in order to make this kind of work easier going forward? @HeroicKatora you should be the most qualified to to answer this, are unstructured/unclear commits a pain when doing this kind of work or is it fine? A quick look through the commit history seems like things are pretty clear, but we have a quite liberal policy with a lot of contributors with different styles. 

I guess some marker that differentiates commits with and without breaking changes could be useful? And a policy to keep bug-fixes separate from breaking changes if possible?

I am not trying to say that I think people here write bad commits, just that we maybe with some kind of policy could make stuff like this easier to do.

HeroicKatora:
@birktj It was still alright. There are obviously things that could be improved but the commit style is manageable since almost all PRs were either bugfixes or feature additions. Maybe I missed something but a few additional commits should not be too painful to rebase into a `0.21.3` if required.

I think the biggest lesson to learn here is that the nature of the repository, partially an aggregate of multiple independent image related projects, obviously promotes features and fixes arriving in mixed succession. But rather than purely require better commit messages (though guiding contributors to a style guide wouldn't be wrong) maybe we could have a `next` branch for breaking changes while developing the incremental fixes and non-breaking changes in `master`. I see two major advantages that this could have:
* Less confusion due to `master` having diverged from the published crate on `docs.rs`
* Quicker bugfix releases due to less pressure from rapid breaking releases

Something that might be helpful in communicating this: contributors theoretically have the power to amend and force push to the PR branch. I don't think it's a good idea to normalize it but at the same time am not sure if contributors are typically aware of it or would be comfortable with a little help on wording and style?

PR #972, opened 2019-06-09
chore: Update glob to 0.3

ignatenkobrain:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ optional = true
 
 [dev-dependencies]
 num-complex = "0.2.0"
-glob = "0.2.10"
+glob = "0.3"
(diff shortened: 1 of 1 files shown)


HeroicKatora:
The major version upgrade made this seem like a breaking change. But it's only within `dev-dependencies` so it doesn't matter greatly that it was not included in the release.

Is there any bug or other critical reason for upgrading this?

ignatenkobrain:
Just trying to get rid out of downstream patches in Fedora :)

HeroicKatora:
Ah I see. Not sure if this warrants its own release (what's your preference and opinion?) but it will definitely be in the next one of this version then.

ignatenkobrain:
As long as new version would come out within few months -- I'm okay :)

PR #973, opened 2019-06-11
Release version 0.22

HeroicKatora:
As discussed in #865 we're deferring the changes to `ColorType` onto another 
branch and/or prerelease versions of the next api for now due to several other 
API changes that await release. The same seems sensible for the error cleanup.
This is a huge merge commit to ensure we don't miss any changes that 
happened in the version-0.21 branch.

This is also a call to suggest outstanding breaking changes that should be part 
of 0.22, such as #901.


HeroicKatora:
@fintelia I'm not sure what to do, I see no easy way to make this work. Together with `png` having been converted to edition 2018 already and getting some major fixes for depth < 8 right now, I think we might have reached a tipping point where the old compiler version support is no longer maintainable. I had previously justified it as not having too big consequences but that might no longer hold true.

Any ideas?

fintelia:
I'd be in favor of just bumping the minimum supported version whenever our dependencies do. Without ecosystem wide guidance, deciding when to bump is mostly guesswork anyways. The past comment about what version we'd have to advance to is gone (lost in the force push?) but if I recall correctly it was 1.26.1 which is just ever so slightly more than a year old. Supporting a year of past versions seems like a very reasonable policy

HeroicKatora:
The dependency of `rayon` is a year old now but it seems that rayon itself had not bumped version for that time. But the Rust version is even much older, true. I'm also more inclined to just bump version now instead of doing another 'major' release that bumps version later. Let's also not forget that we have the existing `0.21` release with all bugfix contained here (although we might want to update the `.travis.yml` on its branch to reflect the supported status) and this is mainly an interface change and adding new encoders.

HeroicKatora:
Ok, then how to select the version to use and which version? At least the current requirement of all our fully updated dependencies but possibly more.

fintelia:
I really don't have any ideas on how to pick a version, and I'm fine with pretty much anything at least a few releases ago. The first release with `edition = 2018` seems like a totally defensible option though.

HeroicKatora:
>  The first release with edition = 2018 seems like a totally defensible option though.

Which would be [`1.31`](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1310-2018-12-06).

lovasoa:
Will the API change suggested in #901 be part of version 0.22 ?

HeroicKatora:
@lovasoa There are a few open points of discussion that have yet to be addressed. I'd personally be happy if they can be answered before the release (and included) but don't want to delay for too long.

HeroicKatora:
Debian Buster is scheduled to release as stable today and includes Rust `1.34.2`. I suppose the alleviates some compatibility concerns. Is there any reason not to bump to that version instead of `1.31` now that bumping is necessary?

martinlindhe:
I support using 1.34.2 since its in debian stable starting today

fintelia:
That sounds good to me, although I don't think we should commit to always supporting the Rust version in Debian stable since that can potentially get to be several years old.

HeroicKatora:
Good, I've provided the bump in a seperate PR (#984) so that remaining tentative changes for the release (#980 and potentially resolution of #901) can be tested.

PR #979, opened 2019-06-22
chore: Update tiff to 0.3

ignatenkobrain:


HeroicKatora:
Part of `0.22`. For `0.21` this is unfortunately a breaking change

PR #984, opened 2019-07-07
Bump compiler requirements to 1.34.2

HeroicKatora:
The version supports edition=2018 and several `const` improvements. It
is also the one shipped with current Debian stable (Buster, released one
day ago).

Since the oldest version supporting the new edition is not 18 weeks
older than the chosen one. But the stabilization of TryFrom, several
integer const functions, integer atomics and NonZero_ types are not
unlikely to get used eventually.

This PR is provided independent from the release PR to get last-minute
fixes the opportunity of good CI support. Does not affect the `0.21` branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ notifications:
 os:
     - linux
 rust:
-    - 1.24.1
+    - 1.34.2
     - stable
     - beta
     - nightly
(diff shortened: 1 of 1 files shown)


HeroicKatora:
I messed up the commit description so I've reworded it slightly.

Issue #980, opened 2019-06-30
Unnecessary unsafety in HDR decoder


64:
https://github.com/image-rs/image/blob/984e092827a6ec83fc7673047d5ca6cacf77b3ea/src/hdr/decoder.rs#L346

`Vec::set_len` leads to the creation of uninitialized memory which can be dropped if an error occurs during decoding. Additionally the other functions like `decode_component` have to be very careful to overwrite every single byte in the buffer lest uninitialized memory leak to the outside.

Is there any reason why `vec![0; n]` can't be used here? I doubt there would be a significant performance impact, if at all.

HeroicKatora:
It's indeed not ok, maybe worse than you'd think since the type of `ret` is not a `Vec<u8>` but a generic `Vec<impl Send>`. @fintelia I believe the necessary changes were part of the original #887 but dropped for the `0.21` branch since they change the interface. Now would be a good time for a rebase of those since `0.22` is delayed either way.

HeroicKatora:
Indeed: this PR removes it https://github.com/image-rs/image/pull/887/files#diff-9e9af5d18e72ec657d6b3c02cf35c490

PR #985, opened 2019-07-08
Remove actual unsafety in hdr

HeroicKatora:
Third part of the effort of removing `unsafe` from the library, the hdr
decoder actually required an interface change. Several other instances
had been removed in earlier commits. The method would use `Vec::set_len`
even though it did not have full control over the type parameter, as it
was a public interface.

Now simply requires the memory into which to emplace the result to be
provided and allocated externally. This makes the callers responsible
for correct optimization.

Removes the optimization for the library internal structures; it is
expected that the compiler picks up on the pattern well enough for it to
not be the main bottleneck.

Leaves the instances that are deprecated methods, they should enjoy some
deprecation period afterall and are not inherently bad as they are
correctly marked with `unsafe` themselves.

Closes: #980

### DIFF
--- a/src/hdr/decoder.rs
+++ b/src/hdr/decoder.rs
@@ -330,57 +330,48 @@ impl<R: BufRead> HDRDecoder<R> {
     pub fn read_image_transform<T: Send, F: Send + Sync + Fn(RGBE8Pixel) -> T>(
         mut self,
         f: F,
-    ) -> ImageResult<Vec<T>> {
+        output_slice: &mut [T],
+    ) -> ImageResult<()> {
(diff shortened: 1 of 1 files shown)


Shnatsel:
Do I understand correctly that this code could potentially drop uninitialized memory of arbitrary type?

HeroicKatora:
I fear so, but I didn't realize until a few days ago that it was actually that severe in the old interface, was always under the impression that the own type would be dropped, no the type parameter. Hence the backport.

@Shnatsel Should probably get an advisory for this.

Shnatsel:
Yeah, please open a RustSec advisory: https://github.com/RustSec/advisory-db 
Let me know if you need any help.

Issue #982, opened 2019-07-05
implement GenericImageView::save


lovasoa:
Hello, and thank you for this great library! 

### The feature

This is a feature request for something that is already doable, but could be easier: **saving a `GenericImageView`**, that is adding a method :

```rust
pub fn save<Q: AsRef<Path>>(&self, path: Q) -> Result<()>
```

to [GenericImageView](https://docs.rs/image/0.21.2/image/trait.GenericImageView.html)

### My specific use case for this functionality is ...
I am implementing the `GenericImageView` trait on structs, and I currently have to either implement a `save` method myself, or instruct the users of the type that they have to manually create a `DynamicImage` if they want to save the image.

### This is more generally applicable to ...

It would notably allow using `save()` on [SubImage](https://docs.rs/image/0.21.2/image/struct.SubImage.html)

fintelia:
I think that any implementation of save will ultimately have to go through [image::save_buffer](https://docs.rs/image/0.21.2/image/fn.save_buffer.html) since that is roughly the interface that the encoders themselves take. Since GenericImageView doesn't provide a way to directly get a byte slice of the data backing the image, this would mean that any generic implementation of save would first require pixel by pixel copying of the image data into another buffer. 

It is a kind of unfortunate consequence of the interface for GenericImageView being so generic that even most of the rest of this library insists on taking inputs with more well defined layouts.

fintelia:
Also, are you able to share the code you're working on? I'd be curious if a trait like this would be better for most uses:

```rust
trait ImageView<P: Pixel> {
    fn dimensions(&self) -> (u32, u32);
    fn image_data(&mut self) -> &mut [u8];

    fn get_pixel(&self, x: u32, y: u32) -> P {...}
    fn set_pixel(&mut self, x: u32, y: u32, p: P) {...}
    ... 
}
```

HeroicKatora:
@fintelia Pretty sure that `image_data(&mut self)` should not be `mut` in `ImageView`. But overall, if saving is first copying to a dedicated byte buffer then so be it. If we add `save` to the trait then it can be specialized in `impl` to avoid that copy.

HeroicKatora:
Related, I wouldn't add 

> `pub fn save<Q: AsRef<Path>>(&self, path: Q) -> Result<()>`

but

> `pub fn write(&self, path: impl std::io::Write) -> Result<()>`

[...]

Issue #834, opened 2018-11-10
Support for saving to memory needs better documention or consistent api


Timmmm:
Currently the only way to write a PNG image is to save it to a file. It would be nice if there was an API that took a `std::io::Write` (or `std::io::Write + std::io::Seek`) for cases when you want an in-memory image.

My use case is writing an SVG with an embedded image. I need to encode to PNG, then base64-encode the result and *then* write to disk.

HeroicKatora:
This is possible with a [`DynamicImage`](https://docs.rs/image/0.20.1/image/enum.DynamicImage.html#method.write_to) or by using the [`png` encoder interface](https://docs.rs/image/0.20.1/image/png/struct.PNGEncoder.html) directly. `DynamicImage` is a public enum of different `ImageBuffer` variants. But the api could be more consistent.

Timmmm:
Ah yeah that works - probably sufficient to mention those in the `ImageBuffer::save()` API docs for now.

SaiintBrisson:
@HeroicKatora I think you may close this issue now.

PR #988, opened 2019-07-11
Save image with specified format

Bruflot:
Ran into the use case where I wanted to save an image without specifying the file extension in the filename. This PR lets you do just that:
```rust
img.save_with_format(&"path/to/img_file", ImageFormat::PNG);
```
I'm not sure if this is a wanted addition or not, so I'm mostly looking for feedback.

I used the `image::ImageFormat` enum for accepting formats. This enum doesn't include `pbm`, `pgm`, `ppm`, and `pam`, so I left those out as I'm not familiar enough with the library to make changes to the enum. Alternatively we can just make another enum.

------

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -775,6 +775,9 @@ where
 {
     /// Saves the buffer to a file at the specified path in
     /// the specified format.
+    ///
+    /// See [`save_buffer_with_format`](fn.save_buffer_with_format.html) for
+    /// supported types.
     pub fn save_with_format<Q>(&self, path: Q, format: ImageFormat) -> io::Result<()>
     where
         Q: AsRef<Path>,
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -773,7 +773,7 @@ where
     P: Pixel<Subpixel = u8> + 'static,
     Container: Deref<Target = [u8]>,
 {
-    /// Saves the buffer to a file at the path specified in
+    /// Saves the buffer to a file at the specified path in
     /// the specified format.
     pub fn save_with_format<Q>(&self, path: Q, format: ImageFormat) -> io::Result<()>
     where
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -7,8 +7,8 @@ use std::slice::{Chunks, ChunksMut};
 
 use color::{ColorType, FromColor, Luma, LumaA, Rgb, Rgba, Bgr, Bgra};
 use flat::{FlatSamples, SampleLayout};
-use dynimage::save_buffer;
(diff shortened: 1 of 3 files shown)


src/buffer.rs
```diff
@@ -768,6 +768,29 @@ where
...
 }
 
+impl<P, Container> ImageBuffer<P, Container>
+where
+    P: Pixel<Subpixel = u8> + 'static,
+    Container: Deref<Target = [u8]>,
+{
+    /// Saves the buffer to a file at the specified path in
+    /// the specified format.
+    pub fn save_with_format<Q>(&self, path: Q, format: ImageFormat) -> io::Result<()>
```
HeroicKatora: Can this doc get one sentence linking to `save_buffer_with_format` which contains the extensive documentation about supported images?

src/dynimage.rs
```diff
@@ -564,6 +564,17 @@ impl DynamicImage {
             p.save(path)
         })
     }
+
+    /// Saves the buffer to a file at the specified path in
+    /// the specified format.
+    pub fn save_with_format<Q>(&self, path: Q, format: ImageFormat) -> io::Result<()>
```
HeroicKatora: Here as well.

HeroicKatora:
> This enum doesn't include pbm, pgm, ppm, and pam, so I left those out as I'm not familiar enough with the library to make changes to the enum. 

The `PNM` variant uses `pam` for all encoders, since the format is flexible enough to be basically a superset of the other netbpm formats. Nothing to worry about. It's not the most used format (without having any stats for it but the encoder didn't exist for 5 years) and can be added afterwards if required.

HeroicKatora:
> I'm not sure if this is a wanted addition or not, so I'm mostly looking for feedback.

Just to be clear. YES! Definitely very much appreciated. This is also the first and a necessary step towards #982 . And it addresses #834 partially. Just making a note here, so that we don't forget to mark it.

Bruflot:
Thanks for the feedback! 😄 I made the changes to the docs, and exposed the `save_buffer_with_format` function in `lib.rs` (I forgot to do that earlier).

HeroicKatora:
Great work, thank you. Want to join `image-rs`? You can be invited to the `image-rs` collaborators team, which allows you access to this and the other repos if you want. See [the contribution policy](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/image-rs/organization/blob/master/CODE_OF_CONDUCT.md).

Bruflot:
Sure, I'd like an invite! I don't have too much free time nowadays, but I'll try to help out now and then.

PR #989, opened 2019-07-18
fix example code of Generating Fractals

kmkzt:

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ fn main() {
             }
 
             let pixel = imgbuf.get_pixel_mut(x, y);
-            let image::Rgb(data) = *pixel;
+            let data = *pixel;
             *pixel = image::Rgb([data[0], i as u8, data[2]]);
         }
     }


HeroicKatora:
Thanks for your concerns. [We are running CI on the example](https://github.com/image-rs/image/blob/7c5873c082bea22d68513e62be9953ae3e83c9b4/src/lib.rs#L128) so I can assure you it is not broken with the code of master. However, your fix incidentally also works due to an implementation of `Index<usize>` on `image::Rgb` which makes the color struct and the raw array interchangable in the code.

So this is a matter of style, mostly. Due to auto-deref you could even omit the whole dereferncing assignment and work directly with `pixel`. For the sake of example, I would tend to want more verbose code.

kmkzt:
@HeroicKatora 
Thank you for the polite commentary and code review.
Sorry for not understanding that the existing code works well.

How about rewriting to redundant code like below?

```rust
let data = pixel.clone()
```

I am sorry if it is my mistake, but it seems that the example code is not executed by this commit.
https://github.com/image-rs/image/commit/40269b92943eaaf17b444d76dceaedbfc72c4922

PR #993, opened 2019-07-23
[WIP] Grayscale png 16bits

SnoozeTime:
I'm working on a project where I need to manipulate grayscale 16-bits image. This is a WIP pull request that adds loading and saving from/to 16-bit png and tiff images. It is not using `DynamicImage` yet as it would require a lot of refactoring and impact on the existing codebase. I added an example for reference.

Main changes from existing code is:
- macro to generate the implementation of `ImageBuffer::save`. Adding a specialized implementation for u16 resulted in compiler error for duplicated implementation.
- another factory for PNGDecoder. The default allows loading 16-bits image but I added a `with_strip16` factory to keep existing behavior in `DynamicImage` (which is converting 16-bits to 8-bits)

The rest is additional code. I need to add more test and image samples to this branch before merge.

Please kindly review ;)

new contributor license consent:


examples/grayscale_16bits.rs
```diff
@@ -6,12 +6,17 @@ use std::env;
...
 use std::io::BufReader;
 
-use image::png::PNGDecoder;
 use image::ImageBuffer;
 use image::Gray16Image;
 use image::imageops::thumbnail;
 
-fn main() {
+
+#[cfg(feature = "png")]
```
HeroicKatora: The proper way to do this is to add the example with a `required-features` entry in Cargo.toml. [More information here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-required-features-field-optional). This required naming the file directly but gives a very neat error message in cargo when trying to execute the example without the feature enabled.

```
[[example]]
name = "grayscale_16bits"
path = "examples/grayscale_16bits.rs"
required-features = ["png"]
```

SnoozeTime: Neat, I'll change that. I had the same issue with unit testing where I had to add some some `#[cfg(feature = "png")]`

SnoozeTime: I just tested. Adding an explicit example configuration to the cargo file will disable example discovery (this is pre-2018 behavior. In 2018 edition, the examples will be auto-discovered).


What I can see:
- Specify all the examples in cargo file
- Keep the feature specific flags in the example source code
- Use 2018 (I supposed this has larger consequences not in scope here...)

Do you have any other suggestion or preference?

```
warning: An explicit [[example]] section is specified in Cargo.toml which currently
disables Cargo from automatically inferring other example targets.
This inference behavior will change in the Rust 2018 edition and the following
files will be included as a example target:

* /home/benoit/PROJECT/image-rs/examples/fractal.rs
* /home/benoit/PROJECT/image-rs/examples/opening.rs
* /home/benoit/PROJECT/image-rs/examples/scaleup/main.rs
* /home/benoit/PROJECT/image-rs/examples/scaledown/main.rs
```

fintelia:
This change works very hard to avoid breaking changes to the API, but in the process adds a whole bunch of complexity. Since this crate is not yet 1.0 and we already want to change a bunch of the APIs involved, I think this is the wrong trade-off. In particular, I think that DynamicImage should be changed to support 16-bit images and decoding shouldn't strip them to 8-bits

SnoozeTime:
Yes I fully agree with DynamicImage. There is some code in this PR that could be removed when DynamicImage supports 16 bits.
 Do you plan to change the `ImageBuffer` save implementation? This is where I found the most complexity with the need to add a macro.

fintelia:
It can also be changed as necessary, though it isn't obvious to me what would actually go wrong with the current code. It looks like ImageBuffer just tries to save with the ColorType associated with its own Pixel format, right?

SnoozeTime:
I tried to just add an implementation for `save` for ImageBuffer that has Pixel with Subpixel u16 and Container Vec<u16>.
For example: https://gist.github.com/SnoozeTime/730e697816bc3ea9ea5edf41b0e6b0c4
This is the error message: https://gist.github.com/SnoozeTime/31295c6a7f67b24574afe9d2a4d59d73

If I specialize more the implementation (by specifying pixel instead of subpixel), then it goes well but I need to add implementation for all existing Pixels, thus the macro.

fintelia:
I see what you are saying. The `save` function currently has a bound requiring that P::Subpixel == u8, which prevents use for 16 bpp images. That should be fixable by instead casting the &[P::Subpixel] to a &[u8] inside the function

examples/grayscale_16bits.rs
HeroicKatora: I hadn't realized that this was edition specific. Since it is only 4 examples it shouldn't be too complicated to have them entered into `Cargo.toml`.

fintelia:
@minybot I think encoding 16-bit PNG files should work at this point, so perhaps file a bug? See for instance this line:

https://github.com/image-rs/image/blob/54355ed17f8f71965a8a7c99bf2b9e25bfe3a183/src/codecs/png.rs#L544

fintelia:
I'm going to go ahead and close this PR since it is very stale and a lot (all?) of the functionality has since been added.

PR #995, opened 2019-07-23
Continue refactoring ColorType

fintelia:
This introduces a bunch of (hopefully) improvements to the color type handling in the crate. The idea is that the main `ColorType` will be the set of formats supported by `DynamicImage`s and produced by decoders. The `ExtendedColorType` is used to track the original encodings from image files and in the future should be accepted by encoders to save files back in their original format. (The exact details of what that interface should look like are still TBD.)


HeroicKatora:
I'm asking myself if the `ExtendedColorType` should also present proper [color space](https://en.wikipedia.org/wiki/Color_space) information [^1]. The current `Rgb8` only refers to the *color model* and all other functionality assumes `sRGB` I think. I wouldn't necessarily change this *right now* since it is a useful default but a consumer that does know better than assume `sRGB` for everything should be able to retrieve proper information about at least some set of color spaces.


[^1]: Or introduce another type `ColorSpace` with

```
struct ColorSpace {
    // opaque
}

enum ColorSpaceStandard {
  // some set of standards we have integrated
  // non-exhaustive
}

impl ColorSpace {
    fn color_type(&self) -> ExtendedColorType { … }
    // badly typed and highly non-exhaustive
    // Easy to add even non-standard ones, e.g. for weird image formats
    fn name(&self) -> &str { … }
    // well typed but not always present
    fn standard(&self) -> ColorSpaceStandard { … }
}

impl From<ExtendedColorType> for ColorSpace {
    // into some internal 'unknown' implementation
}

trait<'a> ImageDecoder<'a> {
    // instead of `original_colortype`
    fn color_space(&self) -> ColorSpace { … }
}
```

fintelia:
Aren't color spaces somewhat orthogonal to color models? Rgb8, Rgb16 and Bgr8 could all store data in the same color space, right? Would it be enough to later on just add a single (non-exhaustive, extended every time we/someone else wants to add a new image format new color spaces) enum and an optional method to image decoders:

```rust
pub enum ColorSpace {
    sRGB,
    ...
}
trait ImageDecoder {
    /// Return the color space of this image, or None if unknown.
    fn color_space(&self) -> Option<ColorSpace> { None }
}
```

HeroicKatora:
> Aren't color spaces somewhat orthogonal to color models?

Not quite, no. `sRGB` can be `Rgb8` or `Rgba8` but surely not `La8`.

fintelia:
Interesting. We'll definitely have to consider than when exposing color space information then. 

Do these changes otherwise look reasonable? If so I'll go ahead and merge them into the `next` branch.

PR #996, opened 2019-07-24
Style fixes

HeroicKatora:
Replace `try!` with `?`-operator (finally!) and preemptively applies some future deprecations for trait dynamically sized trait-objects without `dyn`, which is forbidden in Rust edition 2018 and thus helps transition if wanted.


HeroicKatora:
Note: eventually do the same again with clippy, I'm just too lazy to do this manually now that there is an automatic integration on the way: https://github.com/rust-lang-nursery/rustfix/issues/130

PR #997, opened 2019-07-25
Update fractal exemple.

redwarp:
The destructuring did not work, as Rgb is a struct and not a tuple.


HeroicKatora:
It is, as of version `0.22`.

PR #998, opened 2019-07-31
Fixed broken build with no default features

bvssvni:
- Bumped to 0.22.1


bvssvni:
`cargo test --no-default-features` fails because the type of the tuple `(u64, u64)` can not be inferred from the returning error case.

bvssvni:
I bumped the patch version because this is not a breaking change. Looking at the logs, I could not find breaking changes since last published version (0.22.0). Best to double check this, though.

HeroicKatora:
Oh no.. :smile: Can you add a CI test for that one as well? And it should be fine as a minor patch.

bvssvni:
Done.

HeroicKatora:
Published

bvssvni:
Thanks!

PR #1002, opened 2019-08-04
Provide bits_per_pixel and num_components as ColorType methods

dobrakmato:
It is possible to get `ColorType` by calling `.color()` on `DynamicImage` but `bits_per_pixel()` nor `num_components()` function are available because they are not exported.

This PR exports them so they are available to use with `ColorType` structs returned from `DynamicImage`.

@nwin, @ccgn @HeroicKatora 

--------

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -31,24 +31,35 @@ pub enum ColorType {
 
 }
 
+impl ColorType {
+    /// Returns the number of bits contained in a pixel of this `ColorType`
+    pub fn bits_per_pixel(&self) -> u16 {
+        bits_per_pixel(*self)
+    }
+
+    /// Returns the number of color channels that make up this pixel
+    pub fn channel_count(&self) -> u8 {
+        channel_count(*self)
+    }
+}
+
 /// Returns the number of bits contained in a pixel of `ColorType` ```c```
-pub fn bits_per_pixel(c: ColorType) -> usize {
+pub fn bits_per_pixel(c: ColorType) -> u16 {
     match c {
-        ColorType::Gray(n) => n as usize,
-        ColorType::GrayA(n) => 2 * n as usize,
-        ColorType::RGB(n) | ColorType::Palette(n)| ColorType::BGR(n) => 3 * n as usize,
-        ColorType::RGBA(n) | ColorType::BGRA(n) => 4 * n as usize,
+        ColorType::Gray(n) => n as u16,
+        ColorType::GrayA(n) => 2 * n as u16,
+        ColorType::RGB(n) | ColorType::Palette(n)| ColorType::BGR(n) => 3 * n as u16,
+        ColorType::RGBA(n) | ColorType::BGRA(n) => 4 * n as u16,
     }
 }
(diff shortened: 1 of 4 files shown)


HeroicKatora:
I agree that these would make sense. There exposed interface should simply be inherent methods on `ColorType` instead though, so maybe create inherent ones that forward to the current free functions instead of exporting the free functions.

HeroicKatora:
Sorry for the wait for such a small change.

src/color.rs
```diff
@@ -31,6 +31,18 @@ pub enum ColorType {
 
 }
 
+impl ColorType {
+    /// Returns the number of bits contained in a pixel of this `ColorType`
+    pub fn bits_per_pixel(&self) -> usize {
```
HeroicKatora: Return type `usize` makes a limited amount of sense. It's the one of the internal method, true, but its internal use is also restricted to: 

* casts to `u32`
* comparison to constants `0`, `1`, `2`, `3`
* as a buffer index (which may be the reason for the internal choice)

As a public method however we may prefer a platform independent type, for which `u8` may be too small for the future (think four 64-bit components) but `u16` should be plenty. Also, `u16` can be converted into `usize` using `Into` since `1.26.1` so this is would not be a loss of functionality for internal use.

@fintelia Any comments, also with regards to the in-progress `ColorType` rework?

src/color.rs
```diff
@@ -31,6 +31,18 @@ pub enum ColorType {
...
+        bits_per_pixel(*self)
+    }
+
+    /// Returns the number of color channels that make up this pixel
+    pub fn num_components(&self) -> usize {
```
HeroicKatora: Same as the above but `u8` is maybe large enough already.

HeroicKatora: Also, consider renaming to `channel_count` for naming alignment with the `Pixel` trait.

src/color.rs
fintelia: I agree that there should be a public method exposing this information. The current plan with the `ColorType` rework is to have a `ColorType::bytes_per_pixel` function and make all pixels an integer number of bytes -- which is good for everyone's sanity.

HeroicKatora: For `bytes_per_pixel`, which data type would provide enough range?

fintelia: A `u8` should be fine, right? I don't know of any pixel format with more than 4 channels, but this could fit up to 31 x `f64` channels.

dobrakmato: Should I change the internal free functions too, or only change the public interface?

HeroicKatora: It would be cleaner to change the internal functions as well. All current use should be able to be replaced with proper `_.into()` or `u32::from(_)` conversion I think.

HeroicKatora:
@dobrakmato Is there something we can help you with for progressing this PR?

dobrakmato:
Sorry I am currently on a vacation but I will get to the PR as soon as I
home.

HeroicKatora:
Thank you!

PR #1003, opened 2019-08-09
Fix #558

### LINKED ISSUE #558: Add more documentation for InnerImageView (eg. SubImage) and InnerImage (eg. SubImageMut)
There's currently only one SubImage type which requires the parent image reference to be mutable. Creating a subimage of an immutable image isn't possible, despite this having obvious use cases. The standard Rust pattern of parallel mutable and immutable accessor types (eg. IterMut and Iter) could be used here. Rename current SubImage with the mutable access to the parent image to SubImageMut and add a new SubImage type that takes an immutable parent reference and only provides read access.

lovasoa:
This should help avoiding confusion between `sub_image` and `view`

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -566,6 +566,7 @@ pub trait GenericImageView {
     fn inner(&self) -> &Self::InnerImageView;
 
     /// Returns an subimage that is an immutable view into this image.
+    /// You can use [`GenericImage::sub_image`] if you need a mutable view instead.
     fn view(&self, x: u32, y: u32, width: u32, height: u32) -> SubImage<&Self::InnerImageView> {
         SubImage::new(self.inner(), x, y, width, height)
     }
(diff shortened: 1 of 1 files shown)


src/image.rs
```diff
@@ -638,6 +638,7 @@ pub trait GenericImage: GenericImageView {
     fn inner_mut(&mut self) -> &mut Self::InnerImage;
 
     /// Returns a subimage that is a view into this image.
+    /// If you want an immutable subimage instead, use `GenericImageView::view`
```
HeroicKatora: This addresses only parts of the missing documentation. The other way around may also be helpful together with a short paragraph on these methods in `SubImage`. Potentially similar to [`std::iter::Map`](https://doc.rust-lang.org/std/iter/struct.Map.html) but with a clear indication that these are helpers and `SubImage::new` is also available.

HeroicKatora: Also, these should be proper links, not only code markup.

lovasoa:
@HeroicKatora I implemented the changes you wanted.

HeroicKatora:
Thank you :smile:

PR #1004, opened 2019-08-09
Make copy_from return an ImageResult instead of a boolean

### LINKED ISSUE #901: [API design] copy_from returns bool

lovasoa:
Fixes #901

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -618,8 +618,7 @@ pub trait GenericImage: GenericImageView {
     /// In order to copy only a piece of the other image, use `sub_image`.
     ///
     /// # Returns
-    /// `true` if the copy was successful, `false` if the image could not
-    /// be copied due to size constraints.
+    /// Returns an error if the image is too large to be copied at the given position
     fn copy_from<O>(&mut self, other: &O, x: u32, y: u32) -> bool
     where
         O: GenericImageView<Pixel = Self::Pixel>,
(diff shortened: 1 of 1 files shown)


src/image.rs
```diff
@@ -618,16 +618,15 @@ pub trait GenericImage: GenericImageView {
     /// In order to copy only a piece of the other image, use `sub_image`.
     ///
     /// # Returns
-    /// `true` if the copy was successful, `false` if the image could not
-    /// be copied due to size constraints.
+    /// Returns an error if the image is too large to be copied at the given position
     fn copy_from<O>(&mut self, other: &O, x: u32, y: u32) -> bool
```
fintelia: The `bool` here should actually be `ImageResult<()>`...

lovasoa: Wow ! I messed this up when reopening the old PR.

fintelia:
Thanks!

lovasoa:
@fintelia Thanks to you for merging this

PR #1005, opened 2019-08-10
Improve GIF support

yvt:
This pull request improves the support for GIF images by addressing the following issues:

- An image containing a frame smaller than the image size caused an `UnsupportedError("Unknown error occured while reading gif frame")` error when loaded via `<Decoder<R> as AnimationDecoder<'_>>`.
- The same image from the previous item caused `ImageError::DimensionError` when loaded via `image::open`.
- The background color field of the image global header was interpreted. According to web browsers' source code, the "web compatible" way to handle it seems to be to just ignore it and set the background color to transparency.
- `GifFrameIterator` had extra fields which were effectively used as local variables. They were moved to *real* local variables.
- The lack of test suites for `<Decoder<R> as AnimationDecoder<'_>>`.


HeroicKatora:
[Slightly extended insight into for the web compatibility claim found here](https://github.com/image-rs/image-gif/issues/52).

src/gif.rs
```diff
@@ -93,13 +94,27 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for Decoder<R> {
...
+            f_height = u32::from(frame.height);
         } else {
-            Err(ImageError::ImageEnd)
+            return Err(ImageError::ImageEnd);
         }
+
+        let mut buf = vec![0; self.reader.buffer_size()];
+        self.reader.read_into_buffer(&mut buf)?;
+
+        let image_buffer_raw = ImageBuffer::from_raw(f_width, f_height, buf).unwrap();
```
HeroicKatora: `buffer_size` is input derived from the file. This `unwrap` should instead be a `match` or similar to check if the assumption was indeed correct.

Note that within `gif` the method `buffer_size()` involves converting the image dimensions (`u16`) to the native pointer size `usize` and multiplying several times. At 32-bit and 16-bit these are not necessarily lossless and also may not fit as asserted by `ImageBuffer::from` and thus crash here.

src/gif.rs
```diff
@@ -201,12 +182,17 @@ impl<R: Read> Iterator for GifFrameIterator<R> {
         }
 
         // create the image buffer from the raw frame
-        let mut image_buffer = match ImageBuffer::from_raw(self.width, self.height, vec) {
-            Some(buffer) => buffer,
-            None => return Some(Err(ImageError::UnsupportedError(
-                "Unknown error occured while reading gif frame".into()
-            ))),
-        };
+        let image_buffer_raw = ImageBuffer::from_raw(f_width, f_height, vec).unwrap();
```
HeroicKatora: Similar to the other comment, the previous code already tested for the error case even though its wording was very imprecise.

src/gif.rs
yvt: I somehow assumed that `gif ` would reject image sizes that do not fit in `usize`. Will fix.

yvt:
@HeroicKatora Implemented the requested changes. Can you review again?

tests/reference_images.rs
```diff
@@ -95,49 +97,135 @@ fn check_references() {
...
+            };
+        }
+
+        let ref_img = coalesce_transparent_pixels(ref_img);
+        let test_img = coalesce_transparent_pixels(test_img);
+
+        //ref_img.save("/tmp/test-ref.png").unwrap();
+        //test_img.save("/tmp/test-in.png").unwrap();
+
+        if *ref_img != *test_img {
```
HeroicKatora: These three lines above relaxes the tests on all images: No longer compares exact checksums and doesn't compare visually transparent pixels. If for some reason, the strict tests can not be applied to animated `gif` then it would make sense to have a seperate set of assertions for them instead of modifying the assertions for all other images.

tests/reference_images.rs
```diff
@@ -95,49 +97,135 @@ fn check_references() {
...
+            .unwrap()
+            .split('.')
+            .collect::<Vec<_>>();
+
+        // Ignore the file extension
+        filename_parts.pop().unwrap();
+
+        // The penultimate part of `filename_parts` represents the metadata.
+        let meta_str = filename_parts.pop().unwrap();
+        let mut meta = meta_str.split('_').collect::<Vec<_>>();
```
HeroicKatora: Can this, together with the `if` conditions below, be a seperate method that returns a `Option<TestKind>` where `TestKind` is an enum with two variants for the two cases.

The whole parsing of test metadata creates a lot of clutter before the actual test and code flow it isn't really obvious which assertions are part of the test itself.

tests/reference_images.rs
```diff
@@ -95,49 +97,135 @@ fn check_references() {
...
+        let ref_crc;
+        let anim_frame: Option<usize>;
```
HeroicKatora: Move these in front of the block which assigns them, without spacing. I'm not sure the type ascription is necessary but it really doesn't hurt. Together with the enum, I'd expect something like:

```
let ref_crc;
let anim_frame: Option<usize>;
match meta {
    TestKind::Crc => {
        ref_crc = meta.pop().unwrap();
        anim_frame = None;
    }
…
}
```

tests/reference_images.rs
```diff
@@ -95,49 +97,135 @@ fn check_references() {
...
+            // `CRC`
+            ref_crc = meta.pop().unwrap();
+            anim_frame = None;
+        } else if meta.len() == 3 && meta[0] == "anim" {
+            // `anim_FRAME_CRC`
+            ref_crc = meta.pop().unwrap();
+            let frame: usize = meta.pop().unwrap().parse().unwrap();
+            anim_frame = Some(frame.checked_sub(1).expect("frame number cannot be 0"));
+        } else {
+            panic!(
```
HeroicKatora: Rather use `unreachable!` here. `panic` makes it seem like a failing test could trigger this assertion but it should be the structure of the test suite itself. Maybe also include a comment.

tests/reference_images.rs
```diff
@@ -95,49 +97,135 @@ fn check_references() {
...
+        let ref_crc = u32::from_str_radix(ref_crc, 16).unwrap();
+
+        let ref_crc_actual = {
+            let mut hasher = Crc32::new();
+            hasher.update(&*ref_img);
+            hasher.finalize()
+        };
+
+        if ref_crc_actual != ref_crc {
+            panic!(
```
HeroicKatora: This test, comparing the crc of the test suite with the actual crc of the test suite image is new, right? It doesn't hurt and will properly trigger for faulty new test suite images (good) but should also be an `unreachable!`.

But it seems to have *replaced* the test for crc of `test_img`. Please restore the crc test for the opened `test_img`.

tests/reference_images.rs
```diff
@@ -95,49 +97,135 @@ fn check_references() {
...
+                Err(image::ImageError::UnsupportedError(_)) => return,
+                Err(err) => panic!(format!("decoding of {:?} failed with: {}", img_path, err)),
+            };
+        }
+
+        let ref_img = coalesce_transparent_pixels(ref_img);
+        let test_img = coalesce_transparent_pixels(test_img);
+
+        //ref_img.save("/tmp/test-ref.png").unwrap();
+        //test_img.save("/tmp/test-in.png").unwrap();
```
HeroicKatora: Delete these comments.

HeroicKatora:
@yvt Personal request: Are the reviews in 'milestones' (first implementation, then testing) more annoying than if I had reviewed everything in its entirety? Because it would possibly be more work to do everything in a single review than separately but I don't really want the contribution experience to suffer as a result.

yvt:
@HeroicKatora Kinda? Probably you should indicate in advance that the review doesn't cover all of the changes, because otherwise people would assume that the other part is okay, and get surprised when it turned out to be not.

yvt:
Reverted the modifications to test conditions, meanwhile moving the whole parsing of test metadata to a separate place.

HeroicKatora:
Great work, thank you. If you have continued interest in image processing in Rust, want to join image-rs? You can be invited to the image-rs collaborators team, which allows you access to this and the other repos if you want. See [the contribution policy](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/image-rs/organization/blob/master/CODE_OF_CONDUCT.md).

yvt:
Sorry for the delay! Yes, I'd like to join.

PR #1006, opened 2019-08-12
Fix some clippy warnings

JohnTitor:
Fix some clippy warnings:

* use `from` for lossless cast
* remove redundant code
* replace deprecated attributes

etc.

### DIFF
--- a/src/dxt.rs
+++ b/src/dxt.rs
@@ -114,7 +114,7 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for DXTDecoder<R> {
     type Reader = DXTReader<R>;
 
     fn dimensions(&self) -> (u64, u64) {
-        (self.width_blocks as u64 * 4, self.height_blocks as u64 * 4)
+        (u64::from(self.width_blocks) * 4, u64::from(self.height_blocks) * 4)
     }
 
     fn colortype(&self) -> ColorType {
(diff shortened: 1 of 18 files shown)


src/dxt.rs
```diff
@@ -181,7 +181,7 @@ pub struct DXTReader<R: Read> {
 }
 impl<R: Read> Read for DXTReader<R> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let ref mut decoder = &mut self.decoder;
+        let decoder = &mut (&mut self.decoder);
```
HeroicKatora: The double mutable borrow should not be necessary.

src/hdr/encoder.rs
```diff
@@ -32,15 +32,15 @@ impl<W: Write> HDREncoder<W> {
             // new RLE marker contains scanline width
             let marker = rgbe8(2, 2, (width / 256) as u8, (width % 256) as u8);
             // buffers for encoded pixels
-            let mut bufr = Vec::with_capacity(width);
+            let mut bufr = vec![0; width];
             bufr.resize(width, 0);
```
HeroicKatora: Since the buffer is directly allocated, `resize` should be removed. (3 more instances below).

src/tga/decoder.rs
```diff
@@ -535,7 +535,7 @@ pub struct TGAReader<R> {
 }
 impl<R: Read + Seek> Read for TGAReader<R> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let ref mut decoder = &mut self.decoder;
+        let decoder = &mut (&mut self.decoder);
```
HeroicKatora: Another instance of a puzzling double `&mut`.

JohnTitor:
Thanks! Applied review comments.

HeroicKatora:
Thank you!

Issue #1007, opened 2019-08-13
Missing image_dimensions_from_memory_and_format or similar that allows getting the dimensions from memory


FSMaxB-dooshop:
I would like to be able to read the dimensions of an image in memory without needing to decode all of it. Essentially `image_dimensions` but without requiring a file.

I need this because I want to estimate the size of a decoded image before actually decoding it. I'm currently implementing an image conversion pipeline and need to prevent large (in terms of dimensions) images from being decoded because they can sometimes easily surpass the available RAM.

## Draft

This could either be:

```rust
fn image_dimensions_from_memory_and_format(buffer: &[u8], format: ImageFormat) -> ImageResult<(u32, u32)>
```
or something that takes `impl std::io::Reader`.

FSMaxB-dooshop:
Just to be clear: This functionality is already possible with the current API and I will do that for now, but this essentially requires the duplication of `image_dimension_impl` in my code.

fintelia:
Personally I'd like there to be a function like:
```rust
fn decoder_from_format<R: Read>(reader: Read, format: ImageFormat) -> ImageResult<impl ImageDecoder>;
```
You could then easily get the dimensions by calling `ImageDecoder::dimensions`, but would also have access to all the other image metadata and could decode without having to repeat any work.

HeroicKatora:
I don't like creating even more free functions floating around in the API, there may be a better way to solve this somewhat comprehensively:

```
/// Name bikeshedding.
struct ImageReader<R> {
    reader: R,
    format: Option<ImageFormat>,
}

impl<R: Read> ImageReader<R> {
    /// Construct a reader that will guess the format.
    pub fn new(reader: R) -> Self { … }

    /// Construct a reader with specified format.
    pub fn with_format(reader: R, format: ImageFormat) -> Self { … }

    /// Open a file to read. Format will be guessed.
    /// Note: Failure to guess is not an error. The `Err` is only for `File::open`.
    pub fn with_format(reader: R, format: ImageFormat) -> io::Result<Self> { … }

    /// Get the current format.
    fn format(&self) -> Option<ImageFormat> { … }
}

impl<R: Read + Seek> ImageReader<R> {
    /// Replace the format with a guess based on the content.
    ///
    /// If the guess fails, the format is unchanged. Error is for seek, hence io::Error 
    /// and not ImageError.
    pub fn guess_format(&mut self) -> io::Result<()> { … }
    /// Read the image dimensions.
[...]

PR #1010, opened 2019-08-16
Add a function to derive a ImageFormat from a path

rockwotj:
This was originally inlined as part of `open`'s implementation, however, I
have a usecase where I want to open a file handle and extract metadata
about the image (exif) then turn the file into an image for doing things
such as creating thumbnails, etc.

This seems like the best way to achieve this, is to expose a function
like this (I'm open to a different API FWIW). My code would then be something like:

```rust
let mut file = File::open(path);

let metadata = extract_image_metadata(&file);
let image = image::load(BufReader::new(file), image::ImageFormat::from_path(path)?)?;
image.thumbnail(256, 256).save(thumbnail_path)?;
```

Is there a better way to do something like this using this library?
Otherwise I'm inclined to believe this patch is useful for my usecase
and (hopefully) in general to other users of this library as well.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -4,6 +4,7 @@ use std::error::Error;
 use std::fmt;
 use std::io;
 use std::io::Read;
+use std::path::Path;
 use std::ops::{Deref, DerefMut};
 
 use buffer::{ImageBuffer, Pixel};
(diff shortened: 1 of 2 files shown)


fintelia:
In #1007 @HeroicKatora proposed a more comprehensive API change that exposes this capability (and a bunch of others). It introduces a single struct to replace a whole bunch of one-off top level functions, which I think is a cleaner design.

HeroicKatora:
This seems somewhat orthogonal, adding `ImageFormat::from_path` was not part of the planned changes. I think it would fit well together although and that it's possible to reuse the extract `Path -> ImageFormat` conversion in the more generic reader api as well.

rockwotj:
Hrm, ok, reading over #1007, I think this fits in nicely there (as @HeroicKatora mentioned).

@fintelia note it's not a free (what you mean by top-level?) function, but a factory method on the ImageFormat enum.

src/image.rs
```diff
@@ -143,6 +144,41 @@ pub enum ImageFormat {
     HDR,
 }
 
+impl ImageFormat {
+    /// Return the image format specified by the path's file extension.
+    pub fn from_path<P>(path: P) -> ImageResult<Self> where P : AsRef<Path> {
```
HeroicKatora: In my other rework I had, internally, something similar as `from_path<P>(path: P) -> Option<ImageFormat>`. It's not truely an error here, though I'm not sure which version fits better and `ImageResult<Self>` is the most generic option. On the one hand, there are file extensions which do not specify an image but on the other hand we do not support all image formats.

rockwotj: Usually I consider Result to be a better API than Option, and it's easy to convert to an option with ok().

Here a hypothetical case for result, although it's a little much and at the end of the day you're the maintainer and I will make the API to be what you're willing to support.

The nice thing is Result gives context over *why* the result is not ok and be able to special case certain failures, for example, a user may in the future want to special case when the extension isn't specified and fall back to guessing the format from the magic bytes. Currently the error message for the missing extension case isn't great.

HeroicKatora: Wanting to differentiate various causes and having a better error message is a good point, so let's keep it as a `ImageResult<_>` with a note to #928/#921 which should eventually provide better distinction of the cases.

HeroicKatora:
Thank you!

PR #1012, opened 2019-08-18
A bunch more ColorType improvements for image:next

fintelia:
With this PR, the ColorType handling is hopefully approaching its final form (at least until color space support is added in the mid/long term):
* This change makes all variants of ColorType have pixels with an integer number of bytes. Baking in this assumption simplifies a bunch of code in this crate and in downstream uses.
* There is no longer a "Unknown" fallback variant with unspecified meaning, decoders are expected to convert to one of the supported color types if at all possible. Individual decoders are of course free to provide extra functions for reading raw data in strange pixel types if they want to
* It also takes the chance to rename `colortype()` methods to `color_type()` to match Rust naming conventions, and fixes a few comments.


src/color.rs
```diff
@@ -30,38 +30,28 @@ pub enum ColorType {
...
-        ColorType::La8 => 16,
-        ColorType::Rgb8 | ColorType::Bgr8 => 24,
-        ColorType::Rgba8 | ColorType::Bgra8 | ColorType::La16 => 32,
-        ColorType::Rgb16 => 48,
-        ColorType::Rgba16 => 64,
-        ColorType::Unknown(n) => n as usize,
-        ColorType::__Nonexhaustive => unreachable!(),
+impl ColorType {
+    /// Returns the number of bits contained in a pixel of `ColorType` ```c```
+    pub fn bytes_per_pixel(self) -> usize {
```
HeroicKatora: Make it `u8`, as we agreed would be the a useful range.

src/color.rs
```diff
@@ -30,38 +30,28 @@ pub enum ColorType {
...
-    /// which are associated with an external palette. In that case, the pixel value is an index
-    /// into the palette.
-    Unknown(u8),
-
     #[doc(hidden)]
     __Nonexhaustive,
 }
 
-/// Returns the number of bits contained in a pixel of `ColorType` ```c```
-pub fn bits_per_pixel(c: ColorType) -> usize {
```
HeroicKatora: Or you could leave it as:

```rust
fn bits_per_pixel(c: ColorType) -> u16 {
    u16::from(self.bytes_per_pixel()) * 8
}
```

It doesn't really hurt to have it, the `bytes_per_pixel` already commits to having a whole number of bytes without needing the removal of this method.

src/color.rs
HeroicKatora: Also, run clippy after that change because I'm sure it will turn up many instances of conversion that could be turned into `_::from` if the type is not platform dependent.

src/pnm/encoder.rs
```diff
@@ -626,28 +626,3 @@ impl<'a> TupleEncoding<'a> {
         }
     }
 }
-
-#[cfg(test)]
-mod tests {
```
HeroicKatora: This reminds me: We should think about splitting `pnm` into a seperate crate where we can support the more arcance details, similar to `png` etc. Please remind me if I forget ;)

src/tga/decoder.rs
```diff
@@ -505,12 +505,12 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for TGADecoder<R> {
...
     }
 
-    fn colortype(&self) -> ColorType {
+    fn color_type(&self) -> ColorType {
         self.color_type
     }
 
     fn scanline_bytes(&self) -> u64 {
-        self.row_bytes()
+        self.color_type.bytes_per_pixel() as u64 * self.width as u64
```
HeroicKatora: Should have some justification of cast/overflow behaviour. I believe it can not if `bytes_per_pixel` is `u8`.

src/image.rs
```diff
@@ -366,27 +362,21 @@ pub trait ImageDecoder<'a>: Sized {
...
-    /// to a byte boundary.
-    fn row_bytes(&self) -> u64 {
-        (self.dimensions().0 * color::bits_per_pixel(self.colortype()) as u64 + 7) / 8
-    }
-
-    /// Returns the total number of bytes in the image.
+    /// Returns the total number of bytes in the decoded image.
     fn total_bytes(&self) -> u64 {
-        self.dimensions().1 * self.row_bytes()
+        self.dimensions().0 * self.dimensions().1 * self.color_type().bytes_per_pixel() as u64
```
HeroicKatora: Should have some justification of overflow behaviour. I believe it is that it should not be possible for a valid image? But I'm not sure, this has been a weak spot for a long time but it may be okay.

birktj: Is there a library-wide restriction on width and height?

HeroicKatora:
Overall, I too think it's a good change to have integer assumption for all image pixel representations *in memory*. Note that, in the long term, it doesn't make `rgb555` or other variants impossible, so long as they treat pixels as an unseperable single sample. (And since they represent non-linear color channels and are quantized too coarsely, most arithmetic must be applied with extreme care anyways.) Particularly with a look at SIMD and parallelization, byte boundaries alignment is surely the better abstraction.

src/image.rs
fintelia: This can theoretically overflow, but only if the decoder's reported dimensions + color type indicated a decoded size of greater than 16 EB. That doesn't remotely come close to fitting in memory any time in the foreseeable future, so I'm really not too concerned about it. If you'd prefer I can change it to using checked multiplication instead though.

fintelia:
I've made a bunch of small tweaks as suggested by reviews/clippy. To get clippy to give reasonable output I had to merge in the changes from master.

Issue #716, opened 2018-01-29
Reading metadata without reading the entire file


elbaro:
I have large image files (>5MB). I want only the dimensions of them.
This code is very slow, taking >1s for each open.

```
let img = image::open(path).unwrap();
let (height, width) = img.dimensions();
```

How can I only retrieve the metadata?
I tried `image-utils` crate but it also uses `image` crate.

elbaro:
Just found `immeta` crate works great for this. Please close this if lazy loading is not one of your goals.

nwin:
Rust heavily relies on optimizations. Using `--release` when building your problem should drastically reduce the time it takes to open the image.

With regards to lazy loading, this is already supported just not for the convenience function `image::open`. The individual `ImageDecoder` implementation usually only read the meta data when calling `ImageDecoder::dimensions`. I agree that there should be a convenient way to do that.

jramapuram:
Would love to see this feature. Trying to use `image` to crop a section of a 100k x 100k medical image, but this becomes intractable when using `image::open`.

HeroicKatora:
@jramapuram It seems your use case is not restricted to metadata but rather as a streaming decoder. I'm not completely sure if the current [`Reader`] api already covers your use case since there is no generic equivalent to `open`, only format specific implementations.

Please open that as a new issue if it is not yet possible.

[`Reader`]: https://docs.rs/image/0.22.2/image/trait.ImageDecoder.html#associatedtype.Reader

PR #1014, opened 2019-08-21
Image reader

HeroicKatora:
Replace free reader functions with `Reader` structure as sketched [in this comment of #1007](https://github.com/image-rs/image/issues/1007#issuecomment-521040748). This provides an api for reading image metadata and pixels without coupling to format guessing instead of adding all possible combinations of methods.

Also makes the interface for directly turning an `ImageDecoder` into a `DynamicImage` public.

Closes: #1007 
Closes: #716
Closes: #577

Unresolved questions:
- [x] ~~Should opening a file automatically try to guess the format from the first bytes? Or only from the path? From neither?~~ ~~Offer two methods.~~ Offer a method and a builder style continuation.
- [x] ~~Should the path be contained in the data stored in the `Reader`? This would allow other methods to emulate having opened a file, without going through the actual `fs` api.~~ `ImageFormat::from_path` can be used.
- [x] ~~Should there be a special `load` for formats which do not require `Seek`?~~ Later, if the need arises.
- [x] ~~What about formats not requiring `BufRead` but only `Read`?~~ Easy to wrap in `BufReader`.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -1,7 +1,6 @@
 use num_iter;
-use std::fs::File;
 use std::io;
-use std::io::{BufRead, BufReader, BufWriter, Seek, Write};
+use std::io::Write;
 use std::path::Path;
 use std::u32;
 
@@ -9,8 +8,6 @@ use std::u32;
 use bmp;
 #[cfg(feature = "gif_codec")]
 use gif;
-#[cfg(feature = "hdr")]
-use hdr;
 #[cfg(feature = "ico")]
 use ico;
 #[cfg(feature = "jpeg")]
@@ -19,12 +16,6 @@ use jpeg;
 use png;
 #[cfg(feature = "pnm")]
 use pnm;
-#[cfg(feature = "tga")]
-use tga;
-#[cfg(feature = "tiff")]
-use tiff;
-#[cfg(feature = "webp")]
-use webp;
 
 use buffer::{
     BgrImage, BgraImage, ConvertBuffer, GrayAlphaImage, GrayImage, ImageBuffer, Pixel, RgbImage,
@@ -36,6 +27,7 @@ use image;
 use image::{
     GenericImage, GenericImageView, ImageDecoder, ImageFormat, ImageOutputFormat, ImageResult,
 };
+use io::free_functions;
 use imageops;
 
 /// A Dynamic Image
@@ -735,17 +727,7 @@ where
     P: AsRef<Path>,
 {
     // thin wrapper function to strip generics before calling open_impl
-    open_impl(path.as_ref())
-}
-
-fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
-    let fin = match File::open(path) {
-        Ok(f) => f,
-        Err(err) => return Err(image::ImageError::IoError(err)),
-    };
-    let fin = BufReader::new(fin);
-
-    load(fin, ImageFormat::from_path(path)?)
+    free_functions::open_impl(path.as_ref())
 }
 
 /// Read the dimensions of the image located at the specified path.
@@ -755,47 +737,7 @@ where
     P: AsRef<Path>,
 {
     // thin wrapper function to strip generics before calling open_impl
-    image_dimensions_impl(path.as_ref())
-}
-
-fn image_dimensions_impl(path: &Path) -> ImageResult<(u32, u32)> {
-    let fin = File::open(path)?;
-    let fin = BufReader::new(fin);
-
-    #[allow(unreachable_patterns)]
-    // Default is unreachable if all features are supported.
-    let (w, h): (u64, u64) = match image::ImageFormat::from_path(path)? {
-        #[cfg(feature = "jpeg")]
-        image::ImageFormat::JPEG => jpeg::JPEGDecoder::new(fin)?.dimensions(),
-        #[cfg(feature = "png_codec")]
-        image::ImageFormat::PNG => png::PNGDecoder::new(fin)?.dimensions(),
-        #[cfg(feature = "gif_codec")]
-        image::ImageFormat::GIF => gif::Decoder::new(fin)?.dimensions(),
-        #[cfg(feature = "webp")]
-        image::ImageFormat::WEBP => webp::WebpDecoder::new(fin)?.dimensions(),
-        #[cfg(feature = "tiff")]
-        image::ImageFormat::TIFF => tiff::TIFFDecoder::new(fin)?.dimensions(),
-        #[cfg(feature = "tga")]
-        image::ImageFormat::TGA => tga::TGADecoder::new(fin)?.dimensions(),
-        #[cfg(feature = "bmp")]
-        image::ImageFormat::BMP => bmp::BMPDecoder::new(fin)?.dimensions(),
-        #[cfg(feature = "ico")]
-        image::ImageFormat::ICO => ico::ICODecoder::new(fin)?.dimensions(),
-        #[cfg(feature = "hdr")]
-        image::ImageFormat::HDR => hdr::HDRAdapter::new(fin)?.dimensions(),
-        #[cfg(feature = "pnm")]
-        image::ImageFormat::PNM => pnm::PNMDecoder::new(fin)?.dimensions(),
-        format => {
-            return Err(image::ImageError::UnsupportedError(format!(
-                "Image format image/{:?} is not supported.",
-                format
-            )));
-        }
-    };
-    if w >= u64::from(u32::MAX) || h >= u64::from(u32::MAX) {
-        return Err(image::ImageError::DimensionError);
-    }
-    Ok((w as u32, h as u32))
+    free_functions::image_dimensions_impl(path.as_ref())
 }
 
 /// Saves the supplied buffer to a file at the path specified.
@@ -816,54 +758,7 @@ where
     P: AsRef<Path>,
 {
     // thin wrapper function to strip generics before calling save_buffer_impl
-    save_buffer_impl(path.as_ref(), buf, width, height, color)
-}
-
-fn save_buffer_impl(
-    path: &Path,
-    buf: &[u8],
-    width: u32,
-    height: u32,
-    color: color::ColorType,
-) -> io::Result<()> {
-    let fout = &mut BufWriter::new(File::create(path)?);
-    let ext = path
-        .extension()
-        .and_then(|s| s.to_str())
-        .map_or("".to_string(), |s| s.to_ascii_lowercase());
-
-    match &*ext {
-        #[cfg(feature = "ico")]
-        "ico" => ico::ICOEncoder::new(fout).encode(buf, width, height, color),
-        #[cfg(feature = "jpeg")]
-        "jpg" | "jpeg" => jpeg::JPEGEncoder::new(fout).encode(buf, width, height, color),
-        #[cfg(feature = "png_codec")]
-        "png" => png::PNGEncoder::new(fout).encode(buf, width, height, color),
-        #[cfg(feature = "pnm")]
-        "pbm" => pnm::PNMEncoder::new(fout)
-            .with_subtype(pnm::PNMSubtype::Bitmap(pnm::SampleEncoding::Binary))
-            .encode(buf, width, height, color),
-        #[cfg(feature = "pnm")]
-        "pgm" => pnm::PNMEncoder::new(fout)
-            .with_subtype(pnm::PNMSubtype::Graymap(pnm::SampleEncoding::Binary))
-            .encode(buf, width, height, color),
-        #[cfg(feature = "pnm")]
-        "ppm" => pnm::PNMEncoder::new(fout)
-            .with_subtype(pnm::PNMSubtype::Pixmap(pnm::SampleEncoding::Binary))
-            .encode(buf, width, height, color),
-        #[cfg(feature = "pnm")]
-        "pam" => pnm::PNMEncoder::new(fout).encode(buf, width, height, color),
-        #[cfg(feature = "bmp")]
-        "bmp" => bmp::BMPEncoder::new(fout).encode(buf, width, height, color),
-        #[cfg(feature = "tiff")]
-        "tif" | "tiff" => tiff::TiffEncoder::new(fout)
-            .encode(buf, width, height, color)
-            .map_err(|e| io::Error::new(io::ErrorKind::Other, Box::new(e))), // FIXME: see https://github.com/image-rs/image/issues/921
-        format => Err(io::Error::new(
-            io::ErrorKind::InvalidInput,
-            &format!("Unsupported image format image/{:?}", format)[..],
-        )),
-    }
+    free_functions::save_buffer_impl(path.as_ref(), buf, width, height, color)
 }
 
 /// Saves the supplied buffer to a file at the path specified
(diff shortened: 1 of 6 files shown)


src/dynimage.rs
```diff
@@ -465,6 +456,13 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::rotate270(p))
     }
 
+    /// Decodes an encoded image into a dynamic image.
+    pub fn decode<'a>(decoder: impl ImageDecoder<'a>)
```
fintelia: Perhaps `from_decoder`?

src/io/free_functions.rs
```diff
@@ -0,0 +1,293 @@
...
+
+/// Internal error type for guessing format from path.
+pub(crate) enum PathError {
+    /// The extension did not fit a supported format.
+    UnknownExtension(OsString),
+    /// Extension could not be converted to `str`.
+    NoExtension,
+}
+
+pub fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
```
fintelia: This can be `pub(crate)`, right?

src/io/reader.rs
```diff
@@ -0,0 +1,108 @@
...
+use image::ImageFormat;
+use {ImageError, ImageResult};
+
+use super::free_functions;
+
+/// A multi-format image reader.
+///
+/// Wraps an input reader to facilitate automatic detection of an image's format, appropriate
+/// decoding method, and dispatches into the set of supported `ImageDecoder` implementations.
+pub struct Reader<R: Read> {
```
fintelia: I prefer something like `ImageReader` just to give users a bit more context.

src/lib.rs
```diff
@@ -61,7 +61,8 @@ pub use flat::{FlatSamples};
 pub use traits::Primitive;
 
 // Opening and loading images
-pub use dynimage::{guess_format, load, load_from_memory, load_from_memory_with_format, open,
+pub use io::free_functions::{guess_format, load};
+pub use dynimage::{load_from_memory, load_from_memory_with_format, open,
```
fintelia: How would you feel about deprecating some of these functions? We wouldn't have to remove them any time soon, but it would push people towards using the new (cleaner) API.

src/io/reader.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Unwrap the reader.
+    pub fn into_inner(self) -> R {
+        self.inner
+    }
+}
+
+impl Reader<BufReader<File>> {
+    /// Open a file to read. Format will be guessed.
+    /// Note: Failure to guess is not an error. The `Err` is only for `File::open`.
+    pub fn open<P>(path: P) -> io::Result<Self> where P: AsRef<Path> {
```
fintelia: Perhaps call this `from_path`?

src/io/reader.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Uses the current format to construct the correct reader.
+    pub fn load(self) -> ImageResult<DynamicImage> {
+        unimplemented!()
+    }
+}
+
+/// Potential apis, not part of the core proposal:
+impl<R: Read> Reader<R> {
+    /// Load for image formats that don't require seeking.
+    pub fn load_forward(self) -> ImageResult<DynamicImage> {
```
fintelia: I don't think this is required. In almost all cases users should either have a reader that's already seek, or can use Cursor + read_to_end to get one. The rest of the time they can use the ImageDecoder for their specific format. (Which they have to know because none of the format guessing code would work for them...)

src/io/reader.rs
```diff
@@ -0,0 +1,108 @@
...
+            &mut Cursor::new(&mut start[..]))?;
+        self.inner.seek(SeekFrom::Start(cur))?;
+
+        self.format = free_functions::guess_format(&start[..len as usize]).ok();
+        Ok(())
+    }
+
+    /// Read the image dimensions.
+    /// Uses the current format to construct the correct reader.
+    pub fn image_dimensions(self) -> ImageResult<(u32, u32)> {
```
fintelia: I'd rather there be a `into_decoder(self) -> ImageResult<impl ImageDecoder>` function instead. That way after getting the dimensions you don't have to start over to get other pieces of metadata or to decode the image.

src/io/reader.rs
```diff
@@ -0,0 +1,108 @@
...
+        let format = self.format
+            .map_or_else(|| self.guess_format(), Ok)
+            .map_err(ImageError::IoError)?;
+
+        free_functions::image_dimensions_with_format_impl(self.inner, format)
+    }
+
+    /// Read the image (replaces `load`).
+    /// Uses the current format to construct the correct reader.
+    pub fn load(self) -> ImageResult<DynamicImage> {
```
fintelia: Naming here can be bikeshedded. I don't love "load" (the image data might already be in memory) but I can't think of anything much clearer than that.

fintelia:
> What about formats not requiring BufRead but only Read?

It seems easy enough to just dynamically create a `BufReader` for them. They aren't that expensive to construct

src/io/free_functions.rs
HeroicKatora: `free_functions` is `pub(crate)`, hence these are public so we can choose more freely what to export under which name.

src/io/reader.rs
HeroicKatora: `into_decoder` needs to dynamically allocate, this one doesn't. That was the rationale at least. 

But yeah, there should be an `into_decoder` in addition.

src/lib.rs
HeroicKatora: Depends on how sure we are that we want the new set of methods to behave the same, or at least very similar, in terms of format deduction. Also, consider that the `Writer` equivalents do not yet exist. I'd keep them around undeprecated until at least the next major release to evaluate if any problem arises.

We should at minimum add a note to the documentation, however.

src/io/reader.rs
HeroicKatora: Well, `decode`?

src/io/reader.rs
HeroicKatora: The most important part of this operation is, in my eyes, `File::open`. I suppose the answer depends on whether the file content should be inspect for magic bytes. The name `from_path` suggests to me that this is not the case, we could also provide both.

src/io/free_functions.rs
fintelia: My personal preference would be to mark only the things that are actually public as `pub` and make everything else `pub(crate)`.

src/io/reader.rs
fintelia: Why is dynamic allocation required? I'm just imagining that function would return an enum over the different decoder types (a Box<dyn ImageDecoder> wouldn't work because the trait isn't object safe).

HeroicKatora: You're right, an enum may be possible.

But even if it is private it will be hard to maintain, if it's a public enum it will be even harder. Each feature flags would add or remove a variant which would mess with a lot of `match` statements. In the end, I fear that it would turn into a similar state as `DynamicImage` currently. Still might be one way forward but I don't see myself adding it to this PR right now, that's something I would like to consider and sketch separately.

fintelia: I actually wrote out the code for a DynamicDecoder enum in #904. It worked out to be about 40 lines total. With some refactoring, that type plus the one case in `into_decoder` could replace nearly all of the other instances where we currently mach over image types.

HeroicKatora: With `type Reader = Box<dyn Read + 'a>` which is not optimal since it could have a similar `enum`. All together, the interface for `DynamicDecoder` is somewhat removed from the interface for `Reader` and we can always replace the internals when it is ready.

src/io/free_functions.rs
HeroicKatora: Sounds reasonable and less error prone.

src/io/reader.rs
HeroicKatora: The type is available from a module with full path `image::io::Reader`. That should be reasonable context, it can be imported `as ImageReader` if the user requires even more. Overall, it seems to be similar to [RFC356](https://github.com/aturon/rfcs/blob/no-module-prefixes/active/0000-no-module-prefixes.md) on naming conventions.

src/io/reader.rs
HeroicKatora: Maybe `Reader::open` should not already read parts of the file, so that it alludes to `File::open` more closely. Instead, we may want the file magic method to appear under another name if we can find one that is sufficiently distinct from `open`. This would clear up the confusion as well and seems slightly cleaner to me.

-----------

I could have sworn that I submitted that comment earlier but it doesn't appear so.

src/io/free_functions.rs
```diff
@@ -0,0 +1,293 @@
...
+            io::ErrorKind::InvalidInput,
+            &format!("Unsupported image format image/{:?}", format)[..],
+        )),
+    }
+}
+
+/// Guess format from a path.
+/// If the path has no extension or it can not be convert to a `str` for comparison then this
+/// method will simply return `None`.
+pub(crate) fn guess_format_from_path_impl(path: &Path) -> Result<ImageFormat, PathError> {
```
theotherphil: This comment is no longer true

src/io/reader.rs
HeroicKatora: A very good argument for `open` only inspecting the filename would the the current `image::open` function only doing that.

This means we need a better name for the function that *does* inspect the file. Not sure if `read` (similar to `fs::read`) fits that functionality correctly but it is at least a candidate.

HeroicKatora: Not withstanding the already open PR to change that behaviour, #933 of course.

src/io/free_functions.rs
HeroicKatora: This comment is no longer true :grinning:

src/io/reader.rs
fintelia: There is already a `guess_format` function, so I think it is fine to have `open only consider the path.

HeroicKatora: It's not quite the same, but that may be fixable on its own.

* The current `open` guess the format, and if that doesn't return anything falls back to path.

* `Reader::from_path(..)?.guess_format()?` would always overwrite the guess from the path.

Maybe it should simply not overwrite the current format in `guess_format` when the guess comes back empty.

HeroicKatora: It's what the documentation says already, just my implementation being wrong.

fintelia: Can `guess_format` ever be wrong? Specifically, if it returns some format (rather than "unknown"), it is ever possible that the image could be decoded as something else? If not, I'd be in favor of something like `Reader::from_path(..)?.guess_format()?` being the default method.

HeroicKatora: Yes, if that behaviour is unwanted. Specifically there are a variety of programs who *only* rely on paths for formats (particularly on Windows but also `graphicx` in Latex) and where it would be surprising if a broken image gets parsed with some other format and leads to unexpected pixel types.

It's maybe a too cautious but the combination of operations is really straightforward.

HeroicKatora: Does that even make sense? One could also `clear_format` afterwards if it is unwanted.

fintelia: Guessing the format could certainly break user expectations, but I was more wondering if there are any "[polyglot](https://en.wikipedia.org/wiki/Polyglot_(computing))" images that could be say both a valid PNG and a valid JPEG? That particular pair isn't possible because any PNG starts with `89 50 4E 47` and any valid JPEG starts with `FF D8`, but perhaps there are some pairs where that could happen?

The reason I ask is that it leads to a clear priority between filenames and magic bytes: if the user asks us to consider both, then we always want to use what the magic bytes say first and only fall back to file extensions if there aren't any recognizable magic bytes.

HeroicKatora: I don't think there are any based on the existing table of magic bytes. But some (`ico`: `\x00\x00\x01\x00`) are certainly not very unique overall files.

src/io/free_functions.rs
```diff
@@ -0,0 +1,298 @@
...
+}
+
+static MAGIC_BYTES: [(&'static [u8], ImageFormat); 17] = [
+    (b"\x89PNG\r\n\x1a\n", ImageFormat::PNG),
+    (&[0xff, 0xd8, 0xff], ImageFormat::JPEG),
+    (b"GIF89a", ImageFormat::GIF),
+    (b"GIF87a", ImageFormat::GIF),
+    (b"RIFF", ImageFormat::WEBP), // TODO: better magic byte detection, see https://github.com/image-rs/image/issues/660
+    (b"MM\x00*", ImageFormat::TIFF),
+    (b"II*\x00", ImageFormat::TIFF),
```
HeroicKatora: Since this likely replaces #933, should remember to cherry pick the fix to magic bytes here.
@martinlindhe

HeroicKatora: Would you say this PR is a fitting replacement?

martinlindhe: @HeroicKatora yea, sure lets drop #933 for this PR

src/io/reader.rs
```diff
@@ -0,0 +1,209 @@
...
+        self.inner.seek(SeekFrom::Start(cur))?;
+
+        self.format = free_functions::guess_format(&start[..len as usize]).ok();
+        Ok(())
+    }
+
+    /// Read the image dimensions.
+    ///
+    /// Uses the current format to construct the correct reader for the format.
+    pub fn dimensions(mut self) -> ImageResult<(u32, u32)> {
```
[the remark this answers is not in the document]
fintelia: My personal preference is an `into_decoder()` method. All general metadata about an image is available directly on the decoder, and they can also be used to decode the image afterwards. Having other metadata related methods feels redundant, especially if we need to have two copies of each due to worrying about Seek.

src/io/reader.rs
```diff
@@ -0,0 +1,209 @@
...
+        })
+    }
+}
+
+impl<R: BufRead + Seek> Reader<R> {
+    /// Replace the format with a guess based on the content.
+    ///
+    /// If the guess fails, the format is unchanged. Error is for seek, hence io::Error
+    /// and not ImageError.
+    pub fn guess_format(&mut self) -> io::Result<()> {
```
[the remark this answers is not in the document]
fintelia: My personal preference would be to have a single function:
* `with_guessed_formatk(&mut self) -> io::Result<Self>`

In the normal case users could one of:
* `let r = image::Reader::from_path(...)?;` or
* `let r = image::Reader::from_path(...)?.with_guessed_format()?;`

And if they had some obscure requirements there is still the option to have complete flexibility via something fancy like:
```rust
let mut r = image::Reader::from_path(...)?;
let ext = r.format();
r.clear_format();
r.with_guessed_format()?;
let guessed = r.format();
r.set_format(/* some decision based on `ext` and `guessed` */)
```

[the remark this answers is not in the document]
HeroicKatora: Methods of the style `fn(self) -> Result<Self>` have two disadvantages that make awkward to use as inherent methods of builders with io, that are not `Copy`. This applies to all methods that take `self` by value.

* They can not be applied behind a (mutable) reference. If some component should configure the reader but not consume the stream then it should not require being passed the potentially large value. It should be possible to operate on a reference here.
* The error case with return type `io::Result<Self>` always drops the inner reader. This makes it more complicated and expensive than necessary to use a reader with cleanup such as `BufReader`.

I'd prefer having methods that can operate on `self` by reference which would be simpler than an artifical interface that allows this despite the value semantics on everything else.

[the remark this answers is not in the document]
fintelia: Those cases don't seem super common so instead of duplicating every method, perhaps we could just make a method `as_mut(&mut self) -> &mut R`? This would allow a lot of the same flexibility: For any `Reader<R>` you can make a `Reader<&mut R>` that wouldn't drop/consume the original, and you do also do anything else you wanted to the underlying reader prior to decoding.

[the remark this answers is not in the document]
HeroicKatora: I don't see the point exposing a mutable reference to reader if there is no utility around seeking to the prior position (which needs to exist anyways since `guess_format` requires this). I can live with dropping the method that only deduces the format but doesn't set it but a method taking `fn set_guessed_format(&mut self)` instead of `with_guess_format(self)` would be nearly as easy to use:

```
let mut r = image::Reader::from_path(...)?;
r.set_guessed_format()?;
r
```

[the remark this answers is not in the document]
HeroicKatora: Well, except maybe `by_ref(&mut self) -> Reader<&mut R>` which also preserves the deduced metadata.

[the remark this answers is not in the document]
fintelia: If the underlying reader supported seek then you'd be able to use it directly. If not, then guess_format wasn't going to be available anyway.

Regardless, I'd like to see this land more than I care about the exact interface. Is there anything else blocking merging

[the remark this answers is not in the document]
HeroicKatora: The decision of whether to have `into_decoder` right now or simply have `decode` and provide a `DynamicDecoder` (w.o.r.t. bikeshedding) later. Since the current free functions provide a way to query the image dimension metadata without having to construct a decoder I'd rather have an `into_dimensions` similar to that as well (and maybe for the other methods of decoder as well). We can always soft deprecate them later.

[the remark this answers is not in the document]
HeroicKatora: @fintelia Alright, I've been swayed. The main purpose is to expose a reader and not as an uncommon way of getting `guess_format` to work. So let's do `with_guessed_format(self) -> io::Result<Self>` and see where it takes us. Can always add the other methods later, as above there is enough room in the naming scheme.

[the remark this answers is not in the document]
fintelia: Lets hold off on `into_decoder` for now, I think that could use a dedicated PR for discussion. It seems fine to have some into metadata functions now and deprecate them once we get proper dynamic decoder support

HeroicKatora:
That's a weird failure, seems spurious and within the `cargo` tool itself:
https://travis-ci.org/image-rs/image/jobs/589672851

HeroicKatora:
No, it's not spurious. But I'm not sure what to make of it:
https://travis-ci.org/image-rs/image/jobs/589672851

HeroicKatora:
Going to ignore it: https://github.com/rust-lang/cargo/pull/7429

HeroicKatora:
I'm glad to finally have this landed. Thanks @fintelia for the review discussions.

PR #1015, opened 2019-08-21
Backport fix for unsafety in hdr decoder

HeroicKatora:
Must initialize all memory of Vec<T> (according to valid T) prior to
calling set_len, even for zeroed memory. Furthermore, we can leak some
type instances-since they should not have a Drop implementation as per
documentation-which is safe. This allows us to complete all read
operations prior to setting the vector length, ensuring initialization
of all elements without default values.

This is a backport that avoids the interface changes made in the change
that fixed in in `0.22`, #985. It may look scary since it involves even *more* 
unsafe blocks but each has reasoning attached, that's how `unsafe` is 
supposed to be used and the reason why it is available. Please review the
reasoning carefully.

Backport feels somewhat necessary since ~50% of usage is still on `0.21` 
(as predicted from adoption speed of previous major versions). This will
get a release `0.21.3` once landed.

### DIFF
--- a/src/hdr/decoder.rs
+++ b/src/hdr/decoder.rs
@@ -309,11 +309,15 @@ impl<R: BufRead> HDRDecoder<R> {
         Ok(ret)
     }
 
-    /// Consumes decoder and returns a vector of transformed pixels
+    /// Consumes decoder and returns a vector of transformed pixels.
+    ///
+    /// **Note**: This method should not be called with a parameter `T` that has a `Drop` impl. In case of
+    /// error or panic, it will never call this `Drop` implementation even for pixels that have
+    /// already been successfully decoded.
     #[deprecated(
         since = "0.21.3",
         note = "For trivial types `T` this interface is less safe and less efficient than one taking\
-                an output slice. Consider upgrading to v0.22.")]
+                an output slice. Semantics may be confusing for non-trivial types. Consider upgrading to v0.22.")]
     pub fn read_image_transform<T: Send, F: Send + Sync + Fn(RGBE8Pixel) -> T>(
         mut self,
         f: F,
--- a/src/hdr/decoder.rs
+++ b/src/hdr/decoder.rs
@@ -1,7 +1,7 @@
 use scoped_threadpool::Pool;
 use num_traits::cast::NumCast;
 use num_traits::identities::Zero;
-use std::mem;
+use std::{mem, ptr};
 #[cfg(test)]
 use std::borrow::Cow;
 use std::error::Error;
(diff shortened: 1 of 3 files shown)


src/hdr/decoder.rs
```diff
@@ -323,46 +327,80 @@ impl<R: BufRead> HDRDecoder<R> {
...
             try!(pool.scoped(|scope| {
-                for chunk in chunks_iter {
+                for chunk_index in 0..self.height as usize {
                     let mut buf = Vec::<RGBE8Pixel>::with_capacity(uszwidth);
                     unsafe {
+                        // SAFETY: allocation has exactly that size.
+                        ptr::write_bytes(buf.as_mut_ptr(), 0u8, uszwidth);
+                        // SAFETY: zeroed is valid for RGBE8Pixel and we initialized just above.
                         buf.set_len(uszwidth);
                     }
```
fintelia: Couldn't this unsafe block be replaced by:

```rust
let mut buf = vec![rgbe8(0,0,0,0); uszwidth];
```

src/hdr/decoder.rs
```diff
@@ -323,46 +327,80 @@ impl<R: BufRead> HDRDecoder<R> {
...
                     try!(read_scanline(&mut self.r, &mut buf[..]));
+
                     let f = &f;
+                    // Each chunk is a unique slice of exactly `uszwidth` uninitialized elements.
+                    let chunk = unsafe {
+                        // SAFETY: within the allocation as per `pixel_count` assertions.This is
+                        // in-bounds since the `Vec` has enough capacity and is never re-allocated.
+                        let start: *mut T = chunks_base.add(chunk_index * uszwidth);
```
fintelia: If you change `add` to `wrapping_offset` then this line doesn't need to be in the unsafe block.

src/hdr/decoder.rs
```diff
@@ -323,46 +327,80 @@ impl<R: BufRead> HDRDecoder<R> {
...
+                        // INVARIANT: unique since all slices are spaced `uszwidth` elements and
+                        // also exactly as large.
+                        UninitSlice(ptr::NonNull::<T>::new_unchecked(start), uszwidth)
```
fintelia: With `NonNull::new(start).unwrap()` this wouldn't need to be in an unsafe block.

HeroicKatora: Since it doesn't come with any more proof obligations than the `add` in the line above, I didn't think it would make any difference but simply make the intent even clearer. Note that the null pointer would not be inbounds and thus can not be the result of the computation of `start`.

src/hdr/decoder.rs
HeroicKatora: Right. That's similar to the current solution. I may have been a bit too literal with transferring the performance aspects of the old code.

src/hdr/decoder.rs
HeroicKatora: This is easier to analyze for llvm and should be (maybe only slightly) faster. But it doesn't add significant assertions that the following blocks do not also add. It must be aligned and writable for the whole chunk until index `pixel_count` to allow `ptr::write`. And the chunk must not be aliased by any of the other chunks, else the writes would be data races.

HeroicKatora:
> I've pointed out a couple things that don't need to be done unsafely, though I don't think changing them would really impact correctness much: any bugs in the safe versions would also cause undefined behavior in the subsequent unsafe blocks.

They are good test for confidence in the safety anyways. At least they made it reasonably obvious to myself if I understood the my own reasonings.

fintelia:
:+1: Feel free to merge when you're ready

PR #1021, opened 2019-08-25
Replace try! macros with ? operators

JohnTitor:
Nightly build says `try!` macro is now deprecated and use `?` operators.
This PR replaces `try!` macros with `?` operators.
Note that, this may contain formatting issue but I leave it because the diff will be large when I run `cargo fmt`. I'm happy to send a PR to run `cargo fmt` later.


fintelia:
Thanks!

PR #1022, opened 2019-08-26
Format code by rustfmt

JohnTitor:
Format code that I recently touched by rustfmt.
(Should we run rustfmt to all the code?)

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -664,17 +664,13 @@ impl<R: Read + Seek> BMPDecoder<R> {
                 1 | 2 | 4 | 8 => ImageType::Palette,
                 16 => ImageType::RGB16,
                 24 => ImageType::RGB24,
-                32 => {
-                    if self.add_alpha_channel {
-                        ImageType::RGBA32
-                    } else {
-                        ImageType::RGB32
-                    }
-                }
+                32 if self.add_alpha_channel => ImageType::RGBA32,
+                32 => ImageType::RGB32,
                 _ => {
-                    return Err(ImageError::FormatError(
-                        format!("Invalid RGB bit count {}", self.bit_count).to_string(),
-                    ))
+                    return Err(ImageError::FormatError(format!(
+                        "Invalid RGB bit count {}",
+                        self.bit_count
+                    )))
                 }
             },
             1 => match self.bit_count {
@@ -1340,11 +1336,9 @@ impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for BMPDecoder<R> {
 
         let data = data?;
 
+        #[rustfmt::skip]
         image::load_rect(
-            x,
-            y,
-            width,
-            height,
+            x, y, width, height,
             buf,
             progress_callback,
             self,
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1,8 +1,8 @@
-use std::{cmp, mem};
 use std::io::{self, Cursor, Read, Seek, SeekFrom};
 use std::iter::{repeat, Iterator, Rev};
 use std::marker::PhantomData;
 use std::slice::ChunksMut;
+use std::{cmp, mem};
 
 use byteorder::{LittleEndian, ReadBytesExt};
 
@@ -293,6 +293,7 @@ fn set_4bit_pixel_run<'a, T: Iterator<Item = &'a u8>>(
     true
 }
 
+#[rustfmt::skip]
 fn set_2bit_pixel_run<'a, T: Iterator<Item = &'a u8>>(
     pixel_iter: &mut ChunksMut<u8>,
     palette: &[(u8, u8, u8)],
@@ -598,17 +599,11 @@ impl<R: Read + Seek> BMPDecoder<R> {
         self.width = i32::from(self.r.read_u16::<LittleEndian>()?);
         self.height = i32::from(self.r.read_u16::<LittleEndian>()?);
 
-        check_for_overflow(
-            self.width,
-            self.height,
-            self.num_channels()
-        )?;
+        check_for_overflow(self.width, self.height, self.num_channels())?;
 
         // Number of planes (format specifies that this should be 1).
         if self.r.read_u16::<LittleEndian>()? != 1 {
-            return Err(ImageError::FormatError(
-                "More than one plane".to_string(),
-            ));
+            return Err(ImageError::FormatError("More than one plane".to_string()));
         }
 
         self.bit_count = self.r.read_u16::<LittleEndian>()?;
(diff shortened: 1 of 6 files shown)


src/bmp/decoder.rs
```diff
@@ -1344,8 +1340,20 @@ impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for BMPDecoder<R> {
 
         let data = data?;
 
-        image::load_rect(x, y, width, height, buf, progress_callback, self, |_, _| unreachable!(),
-                         |_, buf| { buf.copy_from_slice(&data); Ok(buf.len()) })?;
+        image::load_rect(
+            x,
+            y,
+            width,
+            height,
```
HeroicKatora: Can we teach it to keep `x, y, width, height` on one or two lines? This just takes a ton of space for virtually nothing.

src/bmp/encoder.rs
```diff
@@ -71,23 +71,17 @@ impl<'a, W: Write + 'a> BMPEncoder<'a, W> {
             self.writer.write_u32::<LittleEndian>(0xff << 0)?; // blue mask
             self.writer.write_u32::<LittleEndian>(0xff << 24)?; // alpha mask
             self.writer.write_u32::<LittleEndian>(0x73524742)?; // colorspace - sRGB
-            // endpoints (3x3) and gamma (3)
+                                                                // endpoints (3x3) and gamma (3)
```
HeroicKatora: This is probably because there is no newline separating the above block from the loop.

src/dynimage.rs
```diff
@@ -788,21 +778,20 @@ fn image_dimensions_impl(path: &Path) -> ImageResult<(u32, u32)> {
...
         image::ImageFormat::HDR => hdr::HDRAdapter::new(fin)?.dimensions(),
         #[cfg(feature = "pnm")]
-        image::ImageFormat::PNM => {
-            pnm::PNMDecoder::new(fin)?.dimensions()
+        image::ImageFormat::PNM => pnm::PNMDecoder::new(fin)?.dimensions(),
+        format => {
+            return Err(image::ImageError::UnsupportedError(format!(
+                "Image format image/{:?} is not supported.",
+                format
+            )))
```
HeroicKatora: I'd follow-up the lint by adding a `;` at the end here. Somewhat surprised that it doesn't recommend that.

src/dynimage.rs
```diff
@@ -1109,28 +1100,37 @@ mod test {
...
         check(
-            1, 10, 2,
+            1,
+            10,
+            2,
             &[0b11110000, 0b11000000, 0b00001111, 0b11000000],
-            vec![255, 255, 255, 255, 0, 0, 0, 0, 255, 255, 0, 0, 0, 0, 255, 255, 255, 255, 255, 255]);
+            vec![
+                255, 255, 255, 255, 0, 0, 0, 0, 255, 255, 0, 0, 0, 0, 255, 255, 255, 255, 255, 255,
+            ],
```
HeroicKatora: Ugh. Somewhat.

src/tga/decoder.rs
```diff
@@ -261,7 +261,8 @@ impl<R: Read + Seek> TGADecoder<R> {
                     format!(
                         "Color format not supported. Bit depth: {}, Alpha bits: {}",
                         other_channel_bits, num_alpha_bits
-                    ).to_string(),
+                    )
+                    .to_string(),
```
HeroicKatora: Wait, `format!().to_string()`?

JohnTitor:
Yeah, rustfmt is a bit aggressive and we can have `#[rustfmt::skip]` to avoid formatting. If we go, we should add the attribute carefully.

src/tga/decoder.rs
JohnTitor: Hehe, it's odd. `format!()` should return `String`.

src/dynimage.rs
JohnTitor: Probably the line is too long for rustfmt. We can add config `max_width`.

src/dynimage.rs
JohnTitor: Agreed. I'm going to add `;` here.

src/bmp/decoder.rs
JohnTitor: We can add `#[rustfmt::skip]` here.

src/bmp/decoder.rs
```diff
@@ -874,7 +869,8 @@ impl<R: Read + Seek> BMPDecoder<R> {
             buf.resize(max_length, 0);
         } else if length > max_length {
             // Ignore any excess palette colors.
-            self.r.seek(SeekFrom::Current((length - max_length) as i64))?;
+            self.r
+                .seek(SeekFrom::Current((length - max_length) as i64))?;
```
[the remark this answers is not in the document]
HeroicKatora: @JohnTitor Maybe this is also related to the line length settings.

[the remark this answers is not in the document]
JohnTitor: The default value for `max_width` is 100. How far should we extend it?

src/dynimage.rs
HeroicKatora: Yep, exactly 99 characters right now. It's even an understandable decision by rustfmt to try and wrap here but it is annoying that it does not allow any arguments to be grouped. Is there an option to allow that, as long as the line width is not exceeded? Then this would best be formatted in groups of four to show the rgb pixels.

src/bmp/decoder.rs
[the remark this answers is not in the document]
HeroicKatora: I can live with this one. It's the only occurance and partially caused by `r` not having a descriptive name.

src/dynimage.rs
JohnTitor: In imageproc, we added `#[rustfmt::skip]` and made it grid style. Should we do the same?

src/bmp/decoder.rs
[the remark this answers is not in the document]
JohnTitor: Should we rename `r` of `BMPDecoder`?

src/dynimage.rs
HeroicKatora: Probably. I'll see if there is an upstream issue for `rustfmt` since I now encountered that problem in at least 5 projects, in different forms.

src/bmp/decoder.rs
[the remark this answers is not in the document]
HeroicKatora: I wouldn't say no to that change :)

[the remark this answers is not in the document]
HeroicKatora: Either `inner` or `reader` seem fine.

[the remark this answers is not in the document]
JohnTitor: Renamed to `reader`

PR #1028, opened 2019-08-31
Remove unused lzw dependency

est31:


HeroicKatora:
Was no longer used directly since `tiff` is its own crate (`v0.20`).

Issue #877, opened 2019-03-02
Panic on multiplication overflow in image::gif::GifFrameIterator


Shnatsel:
The following GIF files trigger a panic in image::gif::GifFrameIterator on attempting to decode them: 
[gif-overflow.zip](https://github.com/PistonDevelopers/image/files/2921688/gif-overflow.zip)

This happens in release mode, not just in debug mode. Code to reproduce the issue is the same as in #876. Found using [AFL.rs](https://github.com/rust-fuzz/afl.rs).

Shnatsel:
Fixed in 0.23, closing.

PR #1036, opened 2019-09-16
Fix potential overflow in GIF decoding

fintelia:
Fixes: #877 

That issue was caused by an overflow in the "frame delay" field between individual frames of the GIF frames when expressed in milliseconds. This translates to highly unrealistic frame rates of greater than 65 seconds per frame (i.e. over 1000x slower than a normal monitor refresh rate), but ideally should still not result in panics. This PR switches to using u32's to prevent the possibility of overflow while still allowing all delay values expressible by GIF images.

At the same time I noticed that there was some confusion about the actual units used to store frame delays in various places. I've renamed a bunch of places to "delay_ms" to be clear when milliseconds are in use.

**Open questions**: Is num_rational::Ratio<u32> really the right type to express the delay in milliseconds between frames of an animation?

### DIFF
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -30,6 +30,7 @@ extern crate gif;
 extern crate num_rational;
 
 use std::clone::Clone;
+use std::convert::TryInto;
 use std::cmp::min;
 use std::convert::TryFrom;
 use std::io::{self, Cursor, Read, Write};
@@ -190,7 +191,7 @@ impl<R: Read> Iterator for GifFrameIterator<R> {
                     f_height = u32::from(frame.height);
 
                     // frame.delay is in units of 10ms so frame.delay*10 is in ms
-                    delay = Ratio::new(frame.delay * 10, 1);
+                    delay = Ratio::new(u32::from(frame.delay) * 10, 1);
                     dispose = frame.dispose;
                 } else {
                     // no more frames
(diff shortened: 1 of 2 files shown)


src/gif.rs
```diff
@@ -170,7 +171,7 @@ impl<R: Read> Iterator for GifFrameIterator<R> {
                     f_height = u32::from(frame.height);
 
                     // frame.delay is in units of 10ms so frame.delay*10 is in ms
-                    delay = Ratio::new(frame.delay * 10, 1);
+                    delay = Ratio::new(frame.delay as u32 * 10, 1);
```
HeroicKatora: This should be `u32::from(frame.delay)` to make it more obvious this is lossless.

HeroicKatora:
> Open questions: Is num_rational::Ratio really the right type to express the delay in milliseconds between frames of an animation?

Is it even used for any delay with numerator `!= 1`?

fintelia:
> Is it even used for any delay with numerator != 1?

No, but it would be necessary to exactly encode something like 60 Hz. Though an integer number of nanoseconds for frame delay might also have low enough error to matter?

HeroicKatora:
Apparently it doesn't really matter for `gif`, the format itself doesn't allow for better timings, but there is no obvious disadvantage despite compile times either. Address one issue at a time I'd say, so let's simply keep that for now.

fintelia:
Sounds good. Also, this is a breaking change since I've renamed some things/changed the type. Should I be pushing this to the *next* branch?

HeroicKatora:
Right. That poses the question: Are the type changes necessary? We will have *some* arbitrary maximum supported frame delay in either case. While `u16` is (obviously?) a bit too short in the long term other formats may/will have other limits and we shouldn't have to adjust the type every time. Maybe the pragmatic choice is to instead do:

> `Ratio::new(frame.delay.saturating_mul(10), 1)`

and clearly document this saturation. We surely don't want to error on decoding this and it's the most accurate replacement value.

HeroicKatora:
Add: I just noted that `Frame` is completely private so we are of course free to change the internals to `Ration<u32>` without breaking the external interface. If `delay_u32` were an *addition* instead of a replacement we could have the best of both worlds.

HeroicKatora:
@fintelia Do you think we can/should land this for `0.22.3`?

fintelia:
This changes a couple methods on `animation::Frame` so it is technically a breaking change. Merging onto the next branch might be better?

HeroicKatora:
Ah right, yes, change the target to `next` in that case. *OR* we could maybe get away with a saturating multiplication and no type changes instead for an alternative `0.22` compatible change.

HeroicKatora:
@fintelia Reminder that you wanted to rebase this on `next` and it's a partial blocker for comprehensive fuzzing :smile:

fintelia:
If everything looks good, lets merge once CI passes

PR #1037, opened 2019-09-20
Fix the WebP spelling in README

XVilka:


JohnTitor:
Thanks!

PR #1040, opened 2019-09-23
FilterType: Add examples and timings to docs

dbrgn:
Preview:

img

Unfortunately Rustdoc does not provide a simple method to include images in the docs (see https://github.com/rust-lang/rust/issues/32104). I quickly considered including them as inline base64, but that would add a *huge* amount of text to the enum documentation. Instead, I linked to them from GitHub.

This means that when doing local changes, these need to be previewed by manually replacing the github URL with a relative link (`../../../docs/`). However, it should work on docs.rs.


HeroicKatora:
If I understand the changes then these images would be included in the distributed crate. This will bloat it quite a bit, from `~150 kB` to roughly 5 times that amount. Do you think that is worth the documentation clarity? Are there alternatives that could avoid that downside?

dbrgn:
Ah, I think you're right that the `docs` directory is being included in the crate. I wasn't aware of that. We could move the pictures into the `examples` directory, which is excluded in `Cargo.toml`.

HeroicKatora:
Triage: It seems clear how to address the requested changes. Is there some hold-up?

dbrgn:
@HeroicKatora fixed!

HeroicKatora:
Thank you :)

PR #1044, opened 2019-09-27
Opinionated changes to ImageDecoder

### LINKED ISSUE #1029: Should ImageDecoder::read_image take a &mut [u8] instead of returning a Vec<u8>
I can see a couple of tradeoffs arguing both ways.

**Pros:**
* Consistent interface with ImageDecoderExt::read_rect
* Avoids alignment concerns (EDIT: *for decoder implementations*)
* Sidesteps memory limit considerations by making users do memory allocation

**Cons:**
* Breaking change
* More inconvenient for users since they have to allocate a buffer themselves

fintelia:
This PR makes breaking changes to the ImageDecoder trait. Now is a good time to do these because the next branch already has a bunch of breaking changes queued and the longer we wait the more code will be written against the current interfaces.

#### _read_image()_ and _read_image_ with_progress()_ now take `&mut [u8]`'s 
Alignment issues make returning a Vec awkward, and this seems like the simplest way of avoiding that. If the interface proves annoying to use we can always add helper functions or higher level abstractions. One possible future path would be to have image-canvas be an optional dependency of image-core so that there could be a function like `ImageDecoder::read_to_canvas(self) -> ImageResult<Canvas>` or similar.

#### _dimensions()_ now returns `u32`'s instead of `u64`'s
I initially picked u64's for consistency with the *__bytes_ functions + forward compatibility. However I've since realized that image formats generally don't support dimensions over 2^32 and there really aren't images in the wild within several orders of magnitude of hitting that limit. Not to mention almost the entire rest of the library / Rust image ecosystem uses u32's for sizes.

For now, I've only the trait definition has been updated but if other people are on board I'll make a pass updating the rest of the code.

Fixes #1029

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1254,8 +1254,8 @@ impl<R: Read + Seek> BmpDecoder<R> {
 
     /// Read the actual data of the image. This function is deliberately not public because it
     /// cannot be called multiple times without seeking back the underlying reader in between.
-    pub(crate) fn read_image_data(&mut self) -> ImageResult<Vec<u8>> {
-        match self.image_type {
+    pub(crate) fn read_image_data(&mut self, buf: &mut [u8]) -> ImageResult<()> {
+        let data = match self.image_type {
             ImageType::Palette => self.read_palettized_pixel_data(),
             ImageType::RGB16 => self.read_16_bit_pixel_data(Some(&R5_G5_B5_COLOR_MASK)),
             ImageType::RGB24 => self.read_full_byte_pixel_data(&FormatFullBytes::RGB24),
@@ -1278,7 +1278,10 @@ impl<R: Read + Seek> BmpDecoder<R> {
                     "Missing 32-bit bitfield masks".to_string(),
                 )),
             },
-        }
+        }?;
+
+        buf.copy_from_slice(&data);
+        Ok(())
     }
 }
 
@@ -1301,8 +1304,8 @@ impl<R> Read for BmpReader<R> {
 impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for BmpDecoder<R> {
     type Reader = BmpReader<R>;
 
-    fn dimensions(&self) -> (u64, u64) {
-        (self.width as u64, self.height as u64)
+    fn dimensions(&self) -> (u32, u32) {
+        (self.width as u32, self.height as u32)
     }
 
     fn color_type(&self) -> ColorType {
@@ -1314,11 +1317,19 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for BmpDecoder<R> {
     }
 
     fn into_reader(self) -> ImageResult<Self::Reader> {
-        Ok(BmpReader(Cursor::new(self.read_image()?), PhantomData))
+        if self.total_bytes() > usize::max_value() as u64 {
+            return Err(ImageError::InsufficientMemory);
+        }
+
+        let mut buf = vec![0; self.total_bytes() as usize];
+        self.read_image(&mut buf)?;
+
+        Ok(BmpReader(Cursor::new(buf), PhantomData))
     }
 
-    fn read_image(mut self) -> ImageResult<Vec<u8>> {
-        self.read_image_data()
+    fn read_image(mut self, buf: &mut [u8]) -> ImageResult<()> {
+        assert!(buf.len() as u64 == self.total_bytes());
+        self.read_image_data(buf)
     }
 }
 
@@ -1333,23 +1344,9 @@ impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for BmpDecoder<R> {
         progress_callback: F,
     ) -> ImageResult<()> {
         let start = self.reader.seek(SeekFrom::Current(0))?;
-        let data = self.read_image_data();
+        image::load_rect(x, y, width, height, buf, progress_callback, self, |_, _| unreachable!(),
+                         |s, buf| { s.read_image_data(buf).map(|_| buf.len()) })?;
         self.reader.seek(SeekFrom::Start(start))?;
-
-        let data = data?;
-
-        #[rustfmt::skip]
-        image::load_rect(
-            x, y, width, height,
-            buf,
-            progress_callback,
-            self,
-            |_, _| unreachable!(),
-            |_, buf| {
-                buf.copy_from_slice(&data);
-                Ok(buf.len())
-            },
-        )?;
         Ok(())
     }
 }
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1317,14 +1317,7 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for BmpDecoder<R> {
     }
 
     fn into_reader(self) -> ImageResult<Self::Reader> {
-        if self.total_bytes() > usize::max_value() as u64 {
-            return Err(ImageError::InsufficientMemory);
-        }
-
-        let mut buf = vec![0; self.total_bytes() as usize];
-        self.read_image(&mut buf)?;
-
-        Ok(BmpReader(Cursor::new(buf), PhantomData))
+        Ok(BmpReader(Cursor::new(image::decoder_to_vec(self)?), PhantomData))
     }
 
     fn read_image(mut self, buf: &mut [u8]) -> ImageResult<()> {
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1,3 +1,4 @@
+use std::convert::TryFrom;
 use std::io::{self, Cursor, Read, Seek, SeekFrom};
 use std::iter::{repeat, Iterator, Rev};
 use std::marker::PhantomData;
@@ -1321,7 +1322,7 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for BmpDecoder<R> {
     }
 
     fn read_image(mut self, buf: &mut [u8]) -> ImageResult<()> {
-        assert!(buf.len() as u64 == self.total_bytes());
+        assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
         self.read_image_data(buf)
     }
 }
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -24,7 +24,7 @@ use color;
 use flat::FlatSamples;
 use image;
 use image::{
-    GenericImage, GenericImageView, ImageDecoder, ImageFormat, ImageOutputFormat, ImageResult,
+    GenericImage, GenericImageView, ImageDecoder, ImageError, ImageFormat, ImageOutputFormat, ImageResult,
 };
 use io::free_functions;
 use imageops;
@@ -639,17 +639,19 @@ impl GenericImage for DynamicImage {
 }
 
 /// Decodes an image and stores it into a dynamic image
-fn decoder_to_image<'a, I: ImageDecoder<'a>>(codec: I) -> ImageResult<DynamicImage> {
-    let color = codec.color_type();
-    let (w, h) = codec.dimensions();
-    let buf = codec.read_image()?;
+fn decoder_to_image<'a, I: ImageDecoder<'a>>(decoder: I) -> ImageResult<DynamicImage> {
+    let color_type = decoder.color_type();
+    let (w, h) = decoder.dimensions();
+    let total_bytes = decoder.total_bytes();
 
-    // TODO: Avoid this cast by having ImageBuffer use u64's
-    assert!(w <= u64::from(u32::max_value()));
-    assert!(h <= u64::from(u32::max_value()));
-    let (w, h) = (w as u32, h as u32);
+    if total_bytes > usize::max_value() as u64 {
+        return Err(ImageError::InsufficientMemory);
+    }
 
-    let image = match color {
+    let mut buf = vec![0; total_bytes as usize];
+    decoder.read_image(&mut buf)?;
+
+    let image = match color_type {
         color::ColorType::Rgb8 => {
             ImageBuffer::from_raw(w, h, buf).map(DynamicImage::ImageRgb8)
         }
@@ -673,11 +675,11 @@ fn decoder_to_image<'a, I: ImageDecoder<'a>>(codec: I) -> ImageResult<DynamicIma
         color::ColorType::La8 => {
             ImageBuffer::from_raw(w, h, buf).map(DynamicImage::ImageLumaA8)
         }
-        _ => return Err(image::ImageError::UnsupportedColor(color.into())),
+        _ => return Err(ImageError::UnsupportedColor(color_type.into())),
     };
     match image {
         Some(image) => Ok(image),
-        None => Err(image::ImageError::DimensionError),
+        None => Err(ImageError::DimensionError),
     }
 }
 
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -640,16 +640,9 @@ impl GenericImage for DynamicImage {
 
 /// Decodes an image and stores it into a dynamic image
 fn decoder_to_image<'a, I: ImageDecoder<'a>>(decoder: I) -> ImageResult<DynamicImage> {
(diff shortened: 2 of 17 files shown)


src/image.rs
```diff
@@ -423,16 +428,36 @@ pub trait ImageDecoder<'a>: Sized {
...
+    /// such casts, the returned contents will always be in native endian.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if buf.len() != self.total_bytes().
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use zerocopy::{AsBytes, FromBytes};
```
HeroicKatora: This idea to add `zerocopy` as a new dependency suggests itself basically. If so, that seems interesting to consider. It could make the implementation itself safe although I would want to avoid adding it to the public interface if possible.

src/image.rs
```diff
@@ -423,16 +428,36 @@ pub trait ImageDecoder<'a>: Sized {
     }
 
     /// Returns all the bytes in the image.
-    fn read_image(self) -> ImageResult<Vec<u8>> {
+    ///
+    /// This function takes a slices of bytes and writes the pixel data of the image into it.
+    /// Although not required, for certain color types callers may want to pass buffers which are
```
HeroicKatora: Maybe it should be required. This remains to be evaluated but I would suppose the actual decoder implementation would be easier and potentially faster (avoiding an intermediate buffer) if it could cast to the actual type slice internally as well while, as seen below, it's not incredibly complicated to align the buffer.

src/image.rs
```diff
@@ -412,6 +412,11 @@ pub trait ImageDecoder<'a>: Sized {
     fn into_reader(self) -> ImageResult<Self::Reader>;
 
     /// Returns the total number of bytes in the decoded image.
+    ///
+    /// This is the size of the buffer that must be passed to `read_image` or
```
HeroicKatora: Hm, that seems backwards. Note that partial reads need to be supported in some form anyways since `Self::Reader` requires it from being an `io::Read`. Allowing reading a part of the image without reading it completely sounds incredibly useful.

src/image.rs
fintelia: The decoders have to have `into_reader() -> impl Read` so decoding into misaligned buffers already has to be supported (the Read trait doesn't say anything about alignment).  But in practice I don't expect this to be much of a burden. Many formats only have 8-bit channels or decoders only support 8-bit images so they completely dodge this. For the rest, they are essentially implementing a function from [u8] -> [u8]. If they need to they can use the relevant read_X + write_X functions from the byteorder crate to convert integers to the right endian and write them back into the output buffer, which should optimize into an efficient bswap loop / call to memcpy.

As a side note, I expect that buffers will almost always be aligned because allocators generally only return chucks aligned to 4 - 16 byte boundaries. This means we probably won't end up paying penalties for doing misaligned accesses even if we don't require alignment

src/image.rs
fintelia: The _read_rect_ / _read_rect_with_progress_ functions on ImageDecoderExt allow partial reads. If decoders can't provide efficient random access but can still incrementally decode then they can provide a non-trivial reader implementation (i.e. one that doesn't just read from a Cursor<Vec<u8>>). If they can do neither then _read_image_ is intended as their primary interface and everything else will be implemented in terms of it.

src/image.rs
HeroicKatora: > But in practice I don't expect this to be much of a burden. Many formats only have 8-bit channels or decoders only support 8-bit images so they completely dodge this.

I think this is very much a chicken-and-egg problem. I can personally attest that I had avoided 16-bit support when expanding pnm *only* because it wasn't clear how to actually implement this nicely (and partically because `to_be_bytes()` wasn't in `std` back then).

It's true that many buffers will be naturally aligned enough. The question then becomes, can we exploit this with a hot path or similar. It's certain that using `byteorder::read_X` in a loop instead of iterating `&mut X` is more error prone, nowhere near as nice to read (even if the optimizer picks it up) due to `unwrap` from the length etc. nor does it compose as well. These details however don't need to be addressed right now.

But it's also true that leaving the possibility to pass unaligned buffers open is the better interface. I don't think the `Read` trait is a good argument *for* this, since `ImageDecoder` should be a stronger trait and so it wouldn't be unexpected to have stronger some preconditions traded for the higher level interface. Neither is it an argument *against* it. With *some* help to allocate proper buffers this may be okay to use. In particular, the exact `Layout` needs to be calculated by the caller even though it depends only on the `ColorType` and `total_bytes`.

src/image.rs
HeroicKatora: > The read_rect / read_rect_with_progress functions on ImageDecoderExt allow partial reads.

I think neither `ImageDecoderExt` nor `Self::Reader: io::Read` are very good candidates for an optimal partial read interface. Let me explain:

The `Reader` implementation can cover *some* of this but not all. With `io::Read` it offers a very standard but also inflexible interface. Part of that is the error type not being under our control, no conversion back and hence losing the access to the metadata, the implictness of the single possible order of pixels within the stream, ...

Offering a standard `io::Read` interface, the `Reader` can't optimize for discarding parts since every pixel *will* be loaded into the buffer at some point. It can't speed up with a seek on the underlying reader, and this means every single pixel will be loaded, decoded, inspected. Adding `Seek` to such an implementation sounds possible but requiring it in the trait on the other hand would be a complication especially since then `Seek` will also be required on the underlying stream where it currently is not for some decoders. This together with the fact that we don't even want a generic `Seek`, e.g. don't want to go backwards, greatly outweighs the potential benefits for me.

In their most prominent implementation, `ImageDecoderExt` also doesn't provide a big benefit or none at all. The only actual implementation available reads scanlines at a time (iirc) which still requires a *huge* buffer and this also requires, as explained above, *the exact same throughput* as reading the complete image.

And also importantly, `ImageDecoderExt` is not always available and especially not when one wants a generic interface over all decoders. The typical `*Ext` naming is for traits that purely extend an interface and are implementation for all `T: BaseTrait`, so this is also a confusing naming.

fintelia: I don't think there are any perfect solutions to incremental image decoding. I think the current options we have represent good points in the trade-off space, but I'd be open to suggestions on alternatives. Really, a lot of the issue right now is that the current slate of decoders don't actually have much in the way of incremental decoding support. There's no reason why not, other than no one taking the time to implement them. Just about ever decoder's Reader<R> could easily also implement Seek when R did, and almost every format can use Seek to make a rather good read_rect implementation.

> In their most prominent implementation, ImageDecoderExt also doesn't provide a big benefit or none at all. The only actual implementation available reads scanlines at a time (iirc) which still requires a huge buffer and this also requires, as explained above, the exact same throughput as reading the complete image.

The only "real" implementation of ImageDecoderExt is for DXT. Here it actually does seek to the first row of the target rect, but then reads/decodes 4 entire rows at a time (a "scanline") and splices the needed ranges into the target buffer. Depending on shape of the rect, it may or may not be faster to use seeking to only read the parts of rows that are needed.

> The typical *Ext naming is for traits that purely extend an interface and are implementation for all T: BaseTrait, so this is also a confusing naming.

This is a very good point. Any ideas for alternate names? Perhaps `ImageDecoderSeekable` or something?

HeroicKatora: >  The only "real" implementation of ImageDecoderExt is for DXT. Here it actually does seek to the first row of the target rect, but then reads/decodes 4 entire rows at a time (a "scanline") and splices the needed ranges into the target buffer. Depending on shape of the rect, it may or may not be faster to use seeking to only read the parts of rows that are needed.

That seems like an important point in itself: Image formats are explicitely not built for a linear stream of values, and thus the locality of values will not be linear at all but sometimes/often block based. Having the complete target buffer available is sometimes a real advantage that an `io::Read`-based implementation can never enjoy!

src/image.rs
```diff
@@ -423,16 +428,36 @@ pub trait ImageDecoder<'a>: Sized {
     }
 
     /// Returns all the bytes in the image.
-    fn read_image(self) -> ImageResult<Vec<u8>> {
+    ///
+    /// This function takes a slices of bytes and writes the pixel data of the image into it.
```
est31:
```suggestion
    /// This function takes a slice of bytes and writes the pixel data of the image into it.
```

src/image.rs
```diff
@@ -423,16 +428,36 @@ pub trait ImageDecoder<'a>: Sized {
...
+    /// Although not required, for certain color types callers may want to pass buffers which are
+    /// aligned to 2 or 4 byte boundries to the slice can be cast to a [u16] or [u32]. To accommodate
```
est31:
```suggestion
    /// aligned to 2 or 4 byte boundaries to the slice can be cast to a [u16] or [u32]. To accommodate
```

src/image.rs
HeroicKatora: Just because it doesn't seem as clear on a second read: This means that I think the requirement of passing in the whole decoding buffer is the *correct* choice since it is a very minimal interface that guarantees that an output region is available no matter the ordering of pixels within the encoded format.

HeroicKatora:
Lifting this from the review since the concern itself has been resolved:

> > The typical *Ext naming is for traits that purely extend an interface and are implementation for all T: BaseTrait, so this is also a confusing naming.
>
> This is a very good point. Any ideas for alternate names? Perhaps ImageDecoderSeekable or something?

`ImageSeek` may be a bit terse but not inaccurate.

src/image.rs
```diff
@@ -412,6 +412,11 @@ pub trait ImageDecoder<'a>: Sized {
     fn into_reader(self) -> ImageResult<Self::Reader>;
 
     /// Returns the total number of bytes in the decoded image.
+    ///
+    /// This is the size of the buffer that must be passed to `read_image` or
+    /// `read_image_with_progress`. It is possible that the returned value exceeds usize::MAX, in
```
JohnTitor: It may verbose a bit.
```suggestion
    /// `read_image_with_progress`. The returned value may exceed usize::MAX, in
```

fintelia:
I've now implemented these changes for all the decoders and tests. In a bunch of cases I wasn't able to avoid doing an extra copy without a bunch more refactoring, so I left that for a followup PR. This sadly means that in changing the API to support zero copies, we've actually for some cases made it copy everything twice but once the interface has fully stabilized we'll be in a better position to improve decoders across the board.

src/bmp/decoder.rs
```diff
@@ -1314,11 +1317,19 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for BMPDecoder<R> {
...
+        let mut buf = vec![0; self.total_bytes() as usize];
+        self.read_image(&mut buf)?;
+
+        Ok(BmpReader(Cursor::new(buf), PhantomData))
     }
 
-    fn read_image(mut self) -> ImageResult<Vec<u8>> {
-        self.read_image_data()
+    fn read_image(mut self, buf: &mut [u8]) -> ImageResult<()> {
+        assert!(buf.len() as u64 == self.total_bytes());
```
HeroicKatora: This comparison should consider the case that `usize > u64`. One possible alternative assertion:

>  `assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes());`

src/dynimage.rs
```diff
@@ -639,17 +639,19 @@ impl GenericImage for DynamicImage {
...
 
-    // TODO: Avoid this cast by having ImageBuffer use u64's
-    assert!(w <= u64::from(u32::max_value()));
-    assert!(h <= u64::from(u32::max_value()));
-    let (w, h) = (w as u32, h as u32);
+    if total_bytes > usize::max_value() as u64 {
+        return Err(ImageError::InsufficientMemory);
+    }
+
+    let mut buf = vec![0; total_bytes as usize];
```
HeroicKatora: Maybe add a note that both zero-extension and cutoff lead to a useful comparison and make the reverse conversion correct. A better *limit* here might also be `isize::max_value()` which is the maximum number of bytes any `Vec` will try to allocate on 32-bit platforms. But maybe this should not act like a limit and use fallible conversion instead:

```
let total_bytes = usize::try_from(total_bytes)
    .map_err(|_| ImageError::InsufficientMemory)?;
```

src/bmp/decoder.rs
HeroicKatora: Instances of similar code appear in other decoders.

src/gif.rs
```diff
@@ -81,19 +81,28 @@ impl<R> Read for GifReader<R> {
...
     }
 
     fn into_reader(self) -> ImageResult<Self::Reader> {
-        Ok(GifReader(Cursor::new(self.read_image()?), PhantomData))
+        if self.total_bytes() > usize::max_value() as u64 {
+            return Err(ImageError::InsufficientMemory);
+        }
+
+        let mut buf = vec![0; self.total_bytes() as usize];
+        self.read_image(&mut buf)?;
```
HeroicKatora: This seems like a duplicate of the above. Maybe it should be a utility or a default implemented extension method.

src/gif.rs
```diff
@@ -105,20 +114,31 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for Decoder<R> {
...
-        )?;
+        self.reader.read_into_buffer(buf)?;
 
-        // Recover the full image
         let (width, height) = (u32::from(self.reader.width()), u32::from(self.reader.height()));
-        let image_buffer = full_image_from_frame(width, height, image_buffer_raw, left, top);
-
-        Ok(image_buffer.into_raw())
+        if (left, top) != (0, 0) || (width, height) != (f_width, f_height) {
+            // TODO: Avoid unnecessary allocation here.
```
HeroicKatora: It would be neat to have more context here. Why is this control flow necessary, what it tries to perform and why it currently allocates.

src/image.rs
```diff
@@ -274,19 +274,20 @@ impl ImageReadBuffer {
 
 /// Decodes a specific region of the image, represented by the rectangle
 /// starting from ```x``` and ```y``` and having ```length``` and ```width```
-pub(crate) fn load_rect<'a, D, F, F1, F2>(x: u64, y: u64, width: u64, height: u64, buf: &mut [u8],
+pub(crate) fn load_rect<'a, D, F, F1, F2, E>(x: u64, y: u64, width: u64, height: u64, buf: &mut [u8],
```
HeroicKatora: You may want to open a new issue for changing these `u64` types as well.

HeroicKatora:
> I've now implemented these changes for all the decoders and tests. In a bunch of cases I wasn't able to avoid doing an extra copy without a bunch more refactoring, so I left that for a followup PR. This sadly means that in changing the API to support zero copies, we've actually for some cases made it copy everything twice but once the interface has fully stabilized we'll be in a better position to improve decoders across the board.

Yep. For example I can already guarantee that `pnm` could be reworked to read directly into the output image buffer.

src/image.rs
fintelia: Oh good point

src/bmp/decoder.rs
fintelia: I think this case is fine because to trigger the overflow the user would have be running on a >64bit machine _and_ somehow actually produce a 16+ exabyte buffer. But your proposed assertion seems clean enough that I'd be fine switching it.

src/gif.rs
fintelia: I added a comment explaining things better. None of the code in that block is new, it is just moved from elsewhere in the file.

src/dynimage.rs
fintelia: I thought about this a bit more, and I think the right thing to do in these cases is actually to panic. If total_bytes / scanline_bytes is under isize::max_value() but still more than the size of RAM, then the user is going to get a OOM panic. It seems silly to try to protect them if it is even larger than that.

src/image.rs
fintelia: Created #1063

src/image.rs
birktj: You could always have two code paths: one for aligned, and one for unaligned buffers. Then you would check the alignment before choosing.

HeroicKatora:
This seems more or less ready right now? It fails CI in one function where some borrows should end sooner than lexically aparent. Edition 2015 does not have NLL enabled by default on `1.34.2` but this should be fixed rather easily with a new scope.

PR #1047, opened 2019-10-01
Use `match` instead of `if-else` chain

JohnTitor:
Use more common code style


HeroicKatora:
Oh great, nightly breakage. Feels like that's happening more frequently recently. It feels wrong to ignore these outright but shouldn't really block PR merges as well. I'm on it.

HeroicKatora:
Overruling Travis here, failure was related to #1049 . Only nightly failed.

Issue #1043, opened 2019-09-25
Unecessary test runs


HeroicKatora:
Every current CI job runs the test for `--no-default-features`.

https://github.com/image-rs/image/blob/4a7bdb861d8681f144ea4b8bb0c1e99ff3c7b7c6/.travis.yml#L52-L62

This is since the case of an empty `FEATURES` environment variable has been used as a run with default features in the past. This however, unecessarily duplicates test compilations and executions. It also leads to puzzling and hard-to-debug failures where a feature that works with `png` enabled can fail the `png` test if it does not work without default features.

A better way to do this would instead to have two environment variables and two separate scripts:

* `FEATURES`: disabled default features and only run *specific* feature combinations.
* `DEFAULT_FEATURSE`: run a simple `cargo test` when this variable exists at all.

This would also make the script somewhat easier to parse mentally.

PR #1048, opened 2019-10-01
Align travis build to actions

HeroicKatora:
Failure that occurs only on nightly seems to be more often due to Rust
breakage than a bug. Take deprecation warnings more serious by actively
denying them which warns us of incompatible future versions as nightly
would.
  Note: May remove this again in the future if it doesn't help. We've
  already had the case where the rustc version commitment left us
  without a proper replacement for a while since that had only been
  introduce shortly before deprecation.

Also ensures that tests are ran on all relevant builds instead of only
on nightly.

Closes: #1043

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,11 @@ env:
     - FEATURES='tiff'
     - FEATURES='webp'
     - FEATURES='hdr'
+    - DEFAULT_FEATURES='yes'
 matrix:
   allow_failures:
     - name: "Clippy"
+    - rust: nightly
   include:
     - os: osx
       rust: 1.34.2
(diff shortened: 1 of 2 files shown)


JohnTitor:
Does it resolve #1043?

HeroicKatora:
Yes it should. Thank you for noting.

.travis.yml
```diff
@@ -50,13 +52,11 @@ matrix:
         - rustup component add clippy
         - cargo clippy
 script:
-  - if [ -z "$FEATURES" ]; then
-      cargo build -v &&
-      if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then cargo test -v; fi &&
-      cargo doc -v;
-    else
+  - if [ -v FEATURES ]; then
```
HeroicKatora: `-v` seems to be a bash addition and doesn't appear to work correctly on `osx`.

HeroicKatora: Found a POSIX compliant possibility on the internet: `-n ${FEATURE+word}`.

PR #1049, opened 2019-10-01
Fix strict const_err

HeroicKatora:
An array is indexed with constants that exceed its size. This is
inopportune even in a branch that never gets executed. This changes it
so that the indexing happens on a slice. This is a potential forward 
compatibility hazard that isn't too hard to fix.

Also removes a bunch of annotations by avoiding primitive casts with
explicit type annotations in constants.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -100,15 +100,16 @@ impl<T: Primitive + 'static> Pixel for $ident<T> {
         &mut self.0
     }
 
-    #[allow(trivial_casts)]
     fn channels4(&self) -> (T, T, T, T) {
+        const CHANNELS: usize = $channels;
         let mut channels = [T::max_value(); 4];
-        channels[0..$channels].copy_from_slice(&self.0);
+        channels[0..CHANNELS].copy_from_slice(&self.0);
         (channels[0], channels[1], channels[2], channels[3])
     }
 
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Waiting on some feedback from `rust`: https://github.com/rust-lang/rust/issues/65005

HeroicKatora:
So looks like `const_err` is a deny-by-default *lint* and lints are allowed to be arbitrarily broadened. Since the 'bug' in question exists since [`0.19`](https://github.com/image-rs/image/commit/42ccbf08b3724bfc7622869f080621cbe6752a69) and versions before only work because they are buggy :laughing: . I'll probably backport to `0.21` like before, it's still about equal to `0.22` in terms of users, slowly declining.

The better fix probably involves adding `allow(const_err)` to the macro in question since I'm not sure if the allow changes may not even potentially allow evaluating `self.0[..][4]` as if a `const` expression, despite the slice type in the middel.

HeroicKatora:
So, the difference between `deny(lint)` and an error is that the former still gets silenced when `image` is used as a non-path dependency. Phew. So no risk of downstream crates breaking (until it gets turned into a hard error which is likely not soon).

HeroicKatora:
And I found a cute way to avoid the issue regardless. Since old versions as dependencies won't break there is no reason to backport except if someone complains because they use it by path.

PR #1054, opened 2019-10-03
Update next with latest changes to master

HeroicKatora:
Keeps the amount of final merges lower.  This updates a few dependencies and Travis-CI scripts as well as better compatibility with `nightly`.

Also syncs up the introduced methods on `ColorType`:
* `channel_count` instead of `num_components` as it was determined to align to `Pixel` trait

Adds the `io::Reader` structure for experimentation.


fintelia:
Seems reasonable to me.

Issue #974, opened 2019-06-13
[API/Minor/Uncertain] Change acronym enum variants to CamelCase


AregevDev:
I noticed there that acronym enums has a SCREAM naming, which should be changed according to the naming conventions.
This includes all the `ImageFormat` variants and some of the `ColorType` variants.

This is very minor and could be changed easily, however I wouldn't rush to do it as many projects use this crate and this change can break things...

First, I want to know if it is known and get a response.

Thanks!

birktj:
See also https://github.com/image-rs/image/issues/741. https://stackoverflow.com/questions/15526107/acronyms-in-camelcase is also an interesting discussion.

birktj:
@HeroicKatora what are your thoughts on this, I have thought about it before, but I have sort of mixed feelings. On one hand, more consistency with the api guidelines and the library itself would be nice, ~but the breaking changes are quite large and the question is if the (relatively small) improvements are worth it.~ Forget that, as I was writing it I realized that it is stupid, we have many more breaking changes to go before 1.0, and we should definitely do what we can to make the library as good as possible before then.

HeroicKatora:
It's probably a good idea to change it for the color types for sure. For the formats I don't see such a great benefit but for consistency it should be done as well.

AregevDev:
Should I change these and send a PR?

HeroicKatora:
@AregevDev Feel free to do remaining changes on `next`, e.g. the decoder still have this problem. It's been noted before that `WebP` is the more precise capitailization and we've also been in favor of this form in #1056. Remember that change for decoders, there we have [`WebpDecoder`](https://docs.rs/image/0.22.3/image/webp/struct.WebpDecoder.html) instead of `WebPDecoder`.

AregevDev:
Ok, which changes do I need to submit then, in addition to the formats?

HeroicKatora:
Most of the format specific wrappers, encoders. We could also *remove* the format prefix from these. The Rust api guidelines correct version is `gif::Encoder`, not `dxt::DXTEncoder` for example.

* `DXTVariant` and variants
* `DXTDecoder`
* `*Decoder` and `*Encoder` are mostly in capitals (see above decisions)
* `PNMSubtype` (prefix also slated for removal imho)

fintelia:
[...]

PR #1056, opened 2019-10-10
Convert ImageFormat variants to CamelCase

HeroicKatora:
Relates to: #974 (but doesn't completely fix it)

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -122,7 +122,7 @@ pub enum ImageFormat {
     Gif,
 
     /// An Image in WEBP Format
-    Webp,
+    WebP,
 
     /// An Image in general PNM Format
     Pnm,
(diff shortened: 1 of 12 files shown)


src/image.rs
```diff
@@ -113,34 +113,34 @@ pub type ImageResult<T> = Result<T, ImageError>;
...
-    JPEG,
+    Jpeg,
 
     /// An Image in GIF Format
-    GIF,
+    Gif,
 
     /// An Image in WEBP Format
-    WEBP,
+    Webp,
```
HeroicKatora: An alternate spelling would be `WebP`, the official capitalization.

JohnTitor: I have no strong opinion, but `Webp` is fine.

Issue #927, opened 2019-05-04
Not all encoders are supported by `save_buffer`


birktj:
It seems we have both a `gif` encoder and a `dxt` encoder, but none of them are implemented in `save_buffer`. I assume this is just an oversight, but it would be nice to fix.

See also #676

Josh015:
Anyone working on this one? If not, I can try to knock it out really quick.

Josh015:
Actually, wait. It looks like ```gif``` is already implemented in ```save_buffer_impl```. As for ```dxt```, I'm a little confused about that one. Is it supposed to mean ```dds```? If so, that'll require implementing an encoder for that file format first.

fintelia:
No, we have a separate `dds` module for that format. `dxt` refers to raw block compressed image data with no metadata, and is used by the dds code. Honestly, it isn't really an image format itself and should probably not be treated as one in this crate...

Josh015:
Yeah, that's why I was confused. Also, I'm now realizing the issue that was merged into this one seems to have a more detailed discussion. I think I'll leave this one alone for now.

HeroicKatora:
This was fixed in #1391 I believe.

PR #1057, opened 2019-10-11
Adding  gif format to `save_buffer`

deg4uss3r:

WIP for issue #927 

I have the basics down for `gif` and `dxt` in `save_buffer`, and `gif` for `save_buffer_with_format_impl`. I just would to check a few things:

- [x] Will a `gif` type have all `frame`s in a single `&[u8]` for the `buf` parameter? 
- [x] How should we map `dxt::DXTVariant` to `color::ColorType`? 
- [x] Should we add a `image::ImageFormat::DXT` for a `dxt` `save_buffer_with_format_impl`?

### DIFF
--- a/src/io/free_functions.rs
+++ b/src/io/free_functions.rs
@@ -25,6 +25,8 @@ use tga;
 use tiff;
 #[cfg(feature = "webp")]
 use webp;
+#[cfg(feature = "dxt")]
+use dxt;
 
 use color;
 use image;
@@ -147,6 +149,12 @@ pub(crate) fn save_buffer_impl(
         .map_or("".to_string(), |s| s.to_ascii_lowercase());
 
     match &*ext {
+        #[cfg(feature = "dxt")]
+        "dxt" => dxt::DXTEncoder::new(fout).encode(&buf, width, height, dxt::DXTVariant::DXT1)
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, Box::new(e))), // FIXME: see https://github.com/image-rs/image/issues/921
+        #[cfg(feature = "gif_codec")]
+        "gif" => gif::Encoder::new(fout).encode(&gif::Frame::from_rgb(width as u16, height as u16, &buf))
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, Box::new(e))), // FIXME: see https://github.com/image-rs/image/issues/921
         #[cfg(feature = "ico")]
         "ico" => ico::ICOEncoder::new(fout).encode(buf, width, height, color),
         #[cfg(feature = "jpeg")]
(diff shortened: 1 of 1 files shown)


HeroicKatora:
> Will a gif type have all frames in a single &[u8] for the buf parameter?

The `save_*` functions are not really versatile enough for animations, maybe it should only accept a single frame and a different interface for multiple frames.

> How should we map dxt::DXTVariant to color::ColorType?

There is [`DXTVariant::colortype`](https://docs.rs/image/0.22.3/image/dxt/enum.DXTVariant.html#method.colortype) if that's what you are asking.

> Should we add a image::ImageFormat::DXT for a dxt save_buffer_with_format_impl?

Regarding the file extension mapping, we shouldn't forget that DXT is not a file format and it seems inappropriate to guess it based on an file path extension. Would be fine to add an `ImageOutputFormat` and `ImageFormat` and permit it via an explicit format only?

deg4uss3r:
@HeroicKatora I agree with all the above so how should we move forward. 

I can:
- [ ] Check the save function I currently have for specifically a `&[u8]` to make sure it's one frame
- [ ] Add a save_all_frames function, in a very similar manner, but only allowing `gif` types. 
- [ ] Remove the `dxt` from `save_buffer` and work through using the `DXTVariant::colortype` to properly output an image in that format within `save_buffer_with_format_impl` only. 

Thoughts, I should have some time to work on this in the near future so looking to get some suggestions to complete this PR.

HeroicKatora:
> I can:
>
> - [ ] Check the save function I currently have for specifically a &[u8] to make sure it's one frame
> - [ ] Add a save_all_frames function, in a very similar manner, but only allowing gif types.
> - [ ] Remove the dxt from save_buffer and work through using the DXTVariant::colortype to properly output an image in that format within save_buffer_with_format_impl only.

That sounds good. It would be nice to have `save_all_frames` in a separate PR since the interface will likely need some feedback cycles until it is ready. For example, it probably should have a `delay` argument that specifies the timing between frames but there are other possible options as well.

fintelia:
I'm in favor of adding GIF output support, but it is less clear to me that including DXT is the right move. My understanding is that DXT compressed images are generally stored in the DDS format because it also carries the usual metadata regarding image size, etc. At very least  this patch document somewhere that "ImageFormat::DXT" is a non-standard image format and requires out of band metadata to decode.

Relatedly, at some point we should probably add DDS support to the library...

HeroicKatora:
> I'm in favor of adding GIF output support, but it is less clear to me that including DXT is the right move. 

I don't think every format in `ImageFormat` is necessarily a *file* format although `from_path` somewhat suggest that, true. The inverse `to_extension` does not yet exist but for its design it would be good to make a decision here.

fintelia:
I've been imagining `ImageFormat` as being a file format though I suppose it doesn't have to be. DXT is probably best described as an in-memory representation which makes it kind of the odd one out among our other image formats. 

We shouldn't get rid of the functionality all together, but perhaps it makes sense to spin the DXT code into a separate crate and have _image_ only expose a DDS encoder/decoder? There is already [_ddsfile_](https://github.com/PistonDevelopers/ddsfile) which looks like it can understand DDS headers.  Combined with our existing compression/decompression code I don't think it would be too hard to get something working. (The _ddsfile_ crate was passed over when spinning image-rs out of Piston but I don't think it was deliberate?).

deg4uss3r:
I'm happy to just add the `gif` functionality for this PR, and create an issue for the `dxt` discussion as it seems we need to discuss this more.

PR #1062, opened 2019-10-18
Less in root

fintelia:
There are a bunch of types are structs that are exported from the crate root but don't need to be. Not having them makes the documentation less cluttered.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,9 +41,7 @@ pub use image::{AnimationDecoder,
                 Pixels,
                 SubImage};
 
-pub use imageops::FilterType::{self, CatmullRom, Gaussian, Lanczos3, Nearest, Triangle};
-
-pub use image::ImageFormat::{self, Bmp, Gif, Ico, Jpeg, Png, Pnm, WebP};
+pub use image::ImageFormat;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -28,7 +28,10 @@ use std::io::Write;
 
 pub use color::ColorType;
 
-pub use color::{Luma, LumaA, Rgb, Rgba, Bgr, Bgra};
+/// Structs that represent a single pixel of some color type.
+pub mod pixel {
+    pub use color::{Luma, LumaA, Rgb, Rgba, Bgr, Bgra};
+}
```
HeroicKatora: These are not re-exports currently and also don't add clutter (structs are towards the bottom anyways). They are also likely among the most-used items. Moving them into a sepearte module only appears to complicate imports, I don't know what we gain from it.

fintelia: The fact that the listing of structs feels so far down in the docs is part of the clutter issue I'm trying to address. Either way, we don't need to decide on this now

fintelia:
Unless anyone has any last minute objections, I'm going to go ahead and merge this

PR #1066, opened 2019-10-21
Fixes #1058. Migrated to ImageError from std::io::Error

### LINKED ISSUE #1058: TIFF encoder has unique return type
This happens with the encoder for TIFF: image::tiff::TiffEncoder;

## Expected

`encode()` on any other encoder (bmp, ico, jpg, png) returns a `Result<()>`.

## Actual behaviour

`encode()` on the tiff encoder returns so-called `ImageResult<()>`

## Reproduction steps

https://docs.rs/image/0.22.3/image/tiff/struct.TiffEncoder.html
https://docs.rs/image/0.22.3/image/png/struct.PNGEncoder.html

Mike-Neto:
Fixes #1058.

A ton of changes are from cargo fmt, which I'm not sure if I ran wrong or something, otherwise we should merge a fmt commit into master so the this diff is way easier to read.

### DIFF
--- a/src/pnm/encoder.rs
+++ b/src/pnm/encoder.rs
@@ -176,10 +176,7 @@ impl<W: Write> PNMEncoder<W> {
             ColorType::RGB(n @ 1..=16) => ((1 << n) - 1, ArbitraryTuplType::RGB),
             ColorType::RGBA(n @ 1..=16) => ((1 << n) - 1, ArbitraryTuplType::RGBAlpha),
             _ => {
-                return Err(ImageError::IoError(io::Error::new(
-                    io::ErrorKind::InvalidInput,
-                    &format!("Encoding colour type {:?} is not supported", color)[..],
-                )))
+                return Err(ImageError::UnsupportedColor(color))
             }
         };
 
@@ -238,6 +235,7 @@ impl<W: Write> PNMEncoder<W> {
                 encoded: None,
             },
             (_, _) => {
+                // FIXME https://github.com/image-rs/image/issues/921
                 return Err(ImageError::IoError(io::Error::new(
                     io::ErrorKind::InvalidInput,
                     "Color type can not be represented in the chosen format",
@@ -284,13 +282,7 @@ impl<'a> CheckedImageBuffer<'a> {
             .and_then(|v| v.checked_mul(uwidth))
             .and_then(|v| v.checked_mul(uheight))
         {
-            None => Err(ImageError::IoError(io::Error::new(
-                io::ErrorKind::InvalidInput,
-                &format!(
-                    "Image dimensions invalid: {}×{}×{} (w×h×d)",
-                    width, height, components
-                )[..],
-            ))),
+            None => Err(ImageError::DimensionError),
             Some(v) if v == image.len() => Ok(CheckedImageBuffer {
                 _image: image,
                 _width: width,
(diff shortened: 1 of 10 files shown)


fintelia:
Could you propose a version without the cargo fmt? We currently have two branches running in parallel (master and next) so doing an auto-format over the whole repository will make merging between them much harder

Mike-Neto:
I did the formatting revert to the PR. Tough this kind of makes it a pain to contribute to this crate, it might be the right call to do a fmt pass on both branches you mention.

HeroicKatora:
> I did the formatting revert to the PR. Tough this kind of makes it a pain to contribute to this crate, it might be the right call to do a fmt pass on both branches you mention.

You could disable autoformatting? I don't see how the *remaining* formatting fixes influence readability greatly. There has been an fmt pass a short while ago that fixed some actual formatting issue but we also noticed that cargo fmt doesn't play well with matrices etc. So the tool is in fact not an obviously best solution to this problem. Squashing the reverts would also be appreciated.

Mike-Neto:
Ahh I see what you mean, well I could look into cargo fmt config's and see if I can get something that preserves matrixes and other specific type formatting in a follow up PR.

Yeah the remaining formatting changes are not bad at all, I was referring to trailing white spaces that most editors will remove by default even if we don't run cargo fmt.

Either way I will work on squashing the commits ;)

fintelia:
I should also add that the pause on running rustfmt is intended to be temporary. Once the next branch is merged into master, we'll run a formatter over the whole thing so that this isn't an issue in the future

Mike-Neto:
Fixed CI as well, I think this is done 👍

src/pnm/encoder.rs
```diff
@@ -237,10 +238,10 @@ impl<W: Write> PNMEncoder<W> {
                 encoded: None,
             },
             (_, _) => {
-                return Err(io::Error::new(
+                return Err(ImageError::IoError(io::Error::new(
```
HeroicKatora: Here a FIXME note would be appreciated if the line is touched anyways. Maybe `io::Error` is not the worst here since the request was invalid, not unsupported.

src/pnm/encoder.rs
```diff
@@ -275,42 +276,46 @@ impl<'a> CheckedImageBuffer<'a> {
...
+    ) -> ImageResult<CheckedImageBuffer<'a>> {
         let components = channel_count(color) as usize;
         let uwidth = width as usize;
         let uheight = height as usize;
         match Some(components)
             .and_then(|v| v.checked_mul(uwidth))
             .and_then(|v| v.checked_mul(uheight))
         {
-            None => Err(io::Error::new(
+            None => Err(ImageError::IoError(io::Error::new(
```
HeroicKatora: There is `DimensionError` but that looses some information on the error. Probably discuss in #928 for some details?

Mike-Neto:
Thanks for the review, I will look into changing the errors instead of just wrapping them 👍

Mike-Neto:
I'm not sure I understood you first 2 comments, can you elaborate a bit please :)

Updated code as suggested in the other comments.

HeroicKatora:
> I'm not sure I understood you first 2 comments, can you elaborate a bit please :)
>
> Updated code as suggested in the other comments.

Sure. An implementation of a trait is part of the public and SemVer interface of a crate. That is, removing an `impl` requires a major version bump. In case such an impl references a dependency this extends to changing the version of the dependency itself. This is the case here.

```rust
impl From<png::EncodingError> for ImageError {
```

This creates a trait impl for `png::EncodingError` with the specific crate version chosen in dependency resolution. This means that updating the `png` dependency is not internal change but a publicly visible change that requires a major version bump.

Mike-Neto:
I see your point now, It makes a lot of sense especially for a lib which is meant to be depended upon, I removed the impl you mentioned as to avoid the issues you mentioned and mapped them into ImageError::IoError.

fintelia:
@Mike-Neto is there anything left on this patch other than a rebase and merge?

Mike-Neto:
No, I don't think so. I will update the PR soon.

Mike-Neto:
Should be good to go! 👍

PR #1068, opened 2019-10-27
Convert error with private methods, not From

HeroicKatora:
The latter is part of the a publicly visible crate interface. This makes
it impossible to upgrade the image format crates used for these specific
format over major versions. Using this pattern is problematic in several
aspects:

* less mature formats are harder to adopt as they are a huge commitment
* bugs and issues that require breaking changes require a coordinate
  effort across the whole dependency chain and ecosystem
* dependent crates we control do not evolve as independently as I would
  wish

It should arguably be avoided to have an unecessary public interface
dependency *unless a format reaches a stable 1.0 interface itself* where
it is no large liability. For consistency this should be uniform across
formats without regards to whether they are internally implemented, a
crate in image-rs, or otherwise.

### DIFF
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -57,7 +57,7 @@ impl<R: Read> GifDecoder<R> {
         decoder.set(ColorOutput::RGBA);
 
         Ok(GifDecoder {
-            reader: decoder.read_info()?,
+            reader: decoder.read_info().map_err(ImageError::from_gif)?,
         })
     }
 }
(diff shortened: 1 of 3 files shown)


HeroicKatora:
Note: converting error types is one instance of a breaking change. The different format have inherited different versions of the flawed (#928) enum definition and may want to expand in several ways on top of a 'standard' one (e.g. [here in `tiff`](https://github.com/image-rs/image-tiff/pull/49/commits/6f9f2adfbfbabd1fbe93667155baea303a982713#diff-9a269209bed069ffb36d5565a91c9840R38)). I expect these to not be the last instances of such changes, and *not* to happen at once across all formats. Arguably, it's easier to experiment with a better error style in specific format before unifying the errors in `image`.

PR #1070, opened 2019-10-29
Added making shapes example

chitalian:
Added some an example to help people who want to draw on an image buffer.

I have my examples repo on gitlab here: https://gitlab.com/justintorre75/image-examples

### DIFF
--- a/examples/basic_shapes.rs
+++ b/examples/basic_shapes.rs
@@ -1,0 +1,84 @@
+//! An example of generating basic shapes
+extern crate image;
+extern crate num_complex;
+extern crate line_drawing;
+extern crate palette;
+use palette::{LinSrgb, rgb::{Rgb}, encoding::{Srgb, Linear}, Gradient};
+
+fn circle_bres(imgbuf: image::RgbaImage, xc: i32, yc: i32, r: i32) -> image::RgbaImage {
+	let mut imgbuf_clone = imgbuf.clone();
+	for (x, y) in line_drawing::BresenhamCircle::new(xc, yc, r) {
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Good job on the example, and `palette` is a very good crate for colors so showing its interaction is definitely useful. However, this adds quite a lot of (dev-)dependencies and the implementation is dominated not by *utilizing* `image` itself but integrating it. Since `image` is already a dev-dependency of both [`line_drawing`](https://github.com/expenses/line_drawing/) and [`palette`](https://github.com/Ogeon/palette) -- in different outdated versions -- maybe you could contribute the relevant parts of the example to those crates? For example, `line_drawing` does not actually produce output images for its `XiaolinWu` implementation that you used here.

We might want to consider adding a 'built with `image`' section to link to more extended examples, guides, users which repositories as this would fit better. I'm assuming (part of) the motivation for this PR is that such material was hard to find.

chitalian:
@HeroicKatora Thanks a lot for your feedback! I took your advice.
I like your idea for a 'built with `image`' section. Do you know any other repos that use image?

Right now:
line_drawing/examples
palette/palette/examples

fintelia:
https://crates.io/crates/image/reverse_dependencies lists about 400 crates that depend on `image`

HeroicKatora:
Some screenshots from a few users on the [webpage](https://github.com/image-rs/image-rs.github.io) wouldn't hurt? In terms of code teaching portion though it's a bit harder to estimate. Many of them are dev-dependencies, or old versions and the list could use a bit of hand selection to find the most illustrative users.

fintelia:
I'm going to go ahead and close this for now

PR #1071, opened 2019-11-01
PNG encoding should return error on unsupported color type

### LINKED ISSUE #826: Png encoder treats BGRA as RGBA
The current encoder will encode BGRA without returning an `Err` but will wrongfully treat it was RGBA instead. Relevant line of code:

https://github.com/PistonDevelopers/image/blob/3a61c016572e8642dffb06e65affd3216ffd8796/src/png.rs#L134

Expected behaviour: Trying to encode an image with that color type should return an error instead.

Behaviour is more confusing for `DynamicImage` as a special code path will ensure that the subpixels will be swapped into `rgba` order but the `ColorType` will be indicated as `ColorType::Bgra` to the encoder.

fintelia:
Simple change to avoid silently writing PNG images in the wrong channel order. It would be slightly nicer to change the return type to ImageResult, but that would be a breaking change so let's hold off on that for now.

Fixes #826

### DIFF
--- a/src/png.rs
+++ b/src/png.rs
@@ -161,6 +161,7 @@ impl<W: Write> PNGEncoder<W> {
             _ => None,
         }
         .and_then(|(ct, bits)| Some((ct, png::BitDepth::from_u8(bits)?)))
+        // FIXME: After #1066 lands, return the ImageError directly.
         .ok_or(io::Error::new(io::ErrorKind::InvalidInput, ImageError::UnsupportedColor(color)))?;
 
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Can you include a `FIXME` note reminding that the return type might/should change?

fintelia:
Just as a doc comment above the function?

HeroicKatora:
Before the error construction itself, the function return is fixed with #1066 .

Issue #1076, opened 2019-11-23
Design question: endianness during encoding


aschampion:
I'm close to PRing 16 bit per channel support for colors, image buffers, and dynamic images. The last remaining feature for completeness is endian ordering during image encoding. For decoding, I implemented the byte reordering in formats' `ImageDecoder::read_image` so that it is always returned native endian (as is currently specified in the doc comment but was not implemented). (Reinterpreting the buffer from `u8` bytes to the `u16` type happens later in `decoder_to_vec`.)

However, for encoding, there is no equivalent interface trait. I could change the behavior of each encoder, but this would be a breaking change for people who were already using 16bpc `ColorType`s with the encoders by managing byte order manually.

The alternative would be introducing an `ImageEncoder` trait similar to `ImageDecoder`, whose only meaningful method would be `encode`. For most formats this would be a noop wrapper around the format encoder's `encode`, but for 16b capable formats this would do reordering when appropriate. In this way the format encoders would remain format-specific endianness interfaces, while `ImageEncoder` would provide a native endian input guarantee.

Before I implement this last piece, are there any objections to this design?

I'm only dealing with 16bpc I/O support in PNG and TIFF for this PR.

cc @fintelia since you've done most of the recent color refactoring.

fintelia:
That all sounds reasonable; I like the idea of having an `ImageEncoder` trait

PR #1078, opened 2019-11-23
Allow setting a custom pixel density in the JPEG encoder

### LINKED ISSUE #1077: The jpeg encoder hardcodes the pixel density of the image
I have a rust application that deals mainly with painting reproductions, for which some people find the density information embedded in JPEG files important.

The JPEG encoder of this crate [contains](https://github.com/image-rs/image/blob/master/src/jpeg/encoder.rs#L582-L584) a hardcoded [JFIF header](https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format#JFIF_APP0_marker_segment) for all the files it produces.

It would be nice if it were possible to customize the information in this header.

I am interested in the pixel density information, but the header can also contain an uncompressed thumbnail. It would be nice if it were possible to set it too.

lovasoa:
This pull request allows setting a custom DPI value for JPEG images (the encoder used to hardcode a pixel aspect ratio of 1x1 for all images).

Fixes #1077

### DIFF
--- a/src/jpeg/encoder.rs
+++ b/src/jpeg/encoder.rs
@@ -274,6 +274,59 @@ impl<'a, W: Write + 'a> BitWriter<'a, W> {
     }
 }
 
+/// Represents a unit in which the density of an image is measured
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum PixelDensityUnit {
+    /// Represents the absence of a unit, the values indicate only a
+    /// [pixel aspect ratio](https://en.wikipedia.org/wiki/Pixel_aspect_ratio)
+    PixelAspectRatio,
+
+    /// Pixels per inch (2.54 cm)
+    Inches,
+
+    /// Pixels per centimeter
+    Centimeters,
+}
+
+/// Represents the pixel density of an image
+///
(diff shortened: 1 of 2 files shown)


fintelia:
It might be good to have the pixel density setting be part of the ImageEncoder trait that @aschampion is looking into adding for #1076

lovasoa:
@fintelia : do you want to wait until ImageEncoder is merged, and then re-do this PR ?

lovasoa:
Also: Do you think it would be a good idea to reuse the types defined in https://github.com/image-rs/jpeg-decoder/pull/115 once it is merged ?

aschampion:
> It might be good to have the pixel density setting be part of the ImageEncoder trait that @aschampion is looking into adding for #1076

Many of the format encoders' constructors and parameters are similar, so it might eventually make sense to make some common subset of these or other settings part of `ImageEncoder`'s constructor/builder/configuration, but for now I was going to abstract only the `encode` method as a minimal initial change. So one still manually constructs the format encoder they want (as in `save_buffer_impl`), which would not conflict with these changes. No need to wait on my PR.

lovasoa:
Thank you !

PR #1085, opened 2019-11-27
16 bit per channel image buffers and dynamic images

aschampion:
Notes:

- Tests will fail until the `image-png` dependency can be updated to include image-rs/image-png#174, so this is not yet mergeable. Tests pass with current `image-png` master. I'm going ahead and PRing since there may be revisions needed.
- Nomenclature: the PR refers to *channels* and *bits per channel* and not *subpixels* (as is used in `Pixel`), because in the scientific and biomedical image processing communities subpixels are samples at higher spatial resolution than pixels, different from the channel components of a pixel.
- To make this as noncontroversial as possible to merge, minimal changes are made to the `DynamicImage` interface, even if this results in extra boilerplate or obvious missing conversions. This can be fleshed out later.
- One notable deletion is `DynamicImage::as_flat_samples`, because this cannot have a generic return type. While there are alternatives, a code search of GitHub turned up no cases where `as_flat_samples` was used on `DynamicImage` directly, only where it is used on `ImageBuffer`, which still works.
- Only `u16` 16 bit colors are supported, because having more general support for other bit depth types isn't really feasible without `specialization` or some new type magic. Since, practically speaking, only `u8` is fully supported for most existing images I assume this is fine.
- This is a breaking change for anyone using the `save_buffer` free functions to save 16 bit per channel PNGs, because those functions now take native endian buffers rather than format-specific endian (big endian for PNG). Anyone using `PNGEncoder` directly is unaffected.


src/buffer.rs
```diff
@@ -42,13 +44,10 @@ pub trait Pixel: Copy + Clone {
         Self::COLOR_MODEL
     }
 
-    /// ColorType for this pixel format
-    const COLOR_TYPE: ColorType;
+    /// Channels for this pixel format
+    const CHANNELS_TYPE: ChannelsType;
```
HeroicKatora: `ColorType` has a method mapping to the `ChannelsType`. I would have assumed to instead gain a new:

```
fn channels_type() -> ChannelsType {
    ExtendedColorType::from(Self::COLOR_TYPE).channels_type()
}
```

src/color.rs
```diff
@@ -48,6 +78,11 @@ impl ColorType {
         }
     }
 
+    /// Returns the number of bytes contained in a channel per pixel of `ColorType` ```c```
+    pub fn bytes_per_channel(self) -> u8 {
```
HeroicKatora: This will essentially lock the `ColorType` to use the same amount of bytes for all channels, which will probably holds true for a good while. Still, I wonder if it wouldn't be simpler to have a local method in `png` that matches the color types themselves or does this arithemtic instead.

src/color.rs
```diff
@@ -329,65 +389,142 @@ impl<A: Copy> FromColor<A> for A {
     }
 }
 
-/// `FromColor` for Luma
+/// Copy-based conversions to target pixel types using `FromColor`.
+pub(crate) trait IntoColor<Other> {
```
HeroicKatora: Can you add a huge `FIXME: remove this trait`? Reinforcing the belief of `RGB == sRGB` is potentially harmful to the correctness of color processing and needs to be fixed eventually. This PR is not particularly bad, it doesn't add any new concepts in that regard, the problem should just be more prominently visible imho.

src/image.rs
```diff
@@ -512,6 +516,27 @@ pub trait AnimationDecoder<'a> {
     fn into_frames(self) -> Frames<'a>;
 }
 
+/// The trait all encoders implement
+pub trait ImageEncoder {
```
HeroicKatora: :rocket:

fintelia:
I don't have time for a full review right at the moment, but I do have a couple of comments from skimming:
- The channel_bytes function might break if future color types have channels that aren't an integer number of bytes or if different channels had different bit depths (we guarantee that the overall pixel takes an integer number of bytes, but that is a weaker assumption).
- ChannelsType should probably be non-exhaustive and ExtendedColorType::channels_type() should return an option.
- This PR emphasizes how awkward it is that our pixel structs are generic and thus we can't fully constrain them to supported inner types. If they weren't generic, we could also have them derive zerocopy::{FromBytes, AsBytes} which would simplify a couple things/eliminate some unsafe code. At some point we probably need to just replace them with separate structs for each variant of ColorType.

src/buffer.rs
```diff
@@ -745,8 +744,9 @@ where
 
 impl<P, Container> ImageBuffer<P, Container>
 where
-    P: Pixel<Subpixel = u8> + 'static,
-    Container: Deref<Target = [u8]>,
+    P: Pixel + 'static,
+    P::Subpixel: AsBytes,
```
HeroicKatora: I'm very sympathetic towards `zerocopy` but maybe we should hold off from making it part of the public interface for a short while at least. [There is some motion in internals](https://internals.rust-lang.org/t/pre-rfc-safe-transmute/11347) and if the current interface is indeed very stable, it is missing any regards towards `SIMD` and thus the crate could still be in need for a breaking change. Since we faired with only `u8` for a long while, it might be enough to simply define a very minimal (sealed) trait providing impls only for selected primitive types. These `impls` could use `zerocopy` for a safe internals implementation regardless.

src/dynimage.rs
```diff
@@ -279,14 +328,73 @@ impl DynamicImage {
...
+    /// Return a mutable reference to an 16bit Grayscale image with an alpha channel
+    pub fn as_mut_luma_alpha16(&mut self) -> Option<&mut GrayAlpha16Image> {
+        match *self {
+            DynamicImage::ImageLumaA16(ref mut p) => Some(p),
+            _ => None,
+        }
     }
 
-    /// Return a view on the raw sample buffer.
-    pub fn as_flat_samples(&self) -> FlatSamples<&[u8]> {
```
HeroicKatora: Could it instead be changed to return an `Option` here? I always saw it as a more generic version of the `as_rgb8(&self)` bunch of methods and that return type would fit in nicely.

HeroicKatora:
> ChannelsType should probably be non-exhaustive and ExtendedColorType::channels_type() should return an option.

`ExtendedColorType` is currentl also missing the marking as non-exhaustive, right? But if both are marked non-exhaustive we could also require new variants of `ExtendedColorType` to also create a new variant in `ChannelsType` if required.

src/buffer.rs
aschampion: The problem is each color/pixel struct doesn't have an unequivocal `ColorType` that can be an associated const, but it does have an unequivocal `ChannelsType`, e.g., both `ColorType::Rgb8` and `ColorType::Rgb16` are color types of the `Rgb` struct. @fintelia's comment mentions this, although I like the genericness of the current struct design and went down the "pixel type per color type" path first before finding it a mess and retreating to this approach.

Generic associated items could make this easier, but those still seem a ways off.

src/color.rs
aschampion: I thought it better put this arithmetic in a central location so if we are able to support inhomogeneous channel depths later we could pick most of them up by looking for usages. But I agree looking at it now it might have the opposite effect of making homogeneity look like a safe assumption. I'll move the logic to `png`.

aschampion:
Thanks for the comments. I should have time to make changes/respond this weekend.

src/buffer.rs
HeroicKatora: You're both right, the impl is too generic over `<T: Primitive>`.

src/buffer.rs
aschampion: I left the wrapper trait, `traits::AsBytes`, unsealed because `Primitive` was already unsealed, so one could presumably already use `ImageBuffer` for other types, such as on newtypes over primitives.

src/dynimage.rs
aschampion: There are now separate methods with optional returns for `u8` and `u16`.

aschampion:
> ChannelsType should probably be non-exhaustive and ExtendedColorType::channels_type() should return an option.

> If both ExtendedColorType and ChannelsType are marked non-exhaustive we could also require new variants of ExtendedColorType to also create a new variant in ChannelsType if required.

`ChannelsType` is now non-exhaustive, but per @HeroicKatora's comment I assumed that it covers the set of channel types in `ExtendedColorType`. The maintenance burden for adding to `ChannelsType` is low because it's used infrequently, unlike `ColorType`, so it seems ok to require it to cover extended cases.

> At some point we probably need to just replace them with separate structs for each variant of ColorType.

I went down this road with the PR first, but found it a mess of macros as so much logic is and should be shared generically based upon the channel type of the pixel. There's also an n^2 number of conversions between different pixel structs, and the current ability to use `u64`/`f32`/`f64` buffers is sometimes useful even if not ergonomic.

src/buffer.rs
HeroicKatora: This leaves `zerocopy` in the public interface, as it is appears as a bound for the provided impls of a public trait. 

I don't quite follow the reasoning with regards to `Primitive`. The two impl blocks guarded by `[P::Subpixel: AsBytes]` only implement `save` and `save_with_format`. Making their requirements strict does not decrease the usability of `ImageBuffer` as a whole but makes it clear which image types *can* be supported for image encoders. A custom pixel type is very unlikely to be encoded correctly and if it truly is byte-level compatible with one of the standard types than you can always create an `ImageBuffer` on the borrowed slice instead.

.travis.yml
```diff
@@ -10,7 +10,7 @@ notifications:
 os:
   - linux
 rust:
-  - 1.34.2
+  - 1.35.0
```
HeroicKatora: Oh darn. Given that `zerocopy` only requires the feature of `1.35` for **two** lines of code that we do not make use of.
See: https://fuchsia.googlesource.com/fuchsia/+/master/garnet/public/rust/zerocopy/src/lib.rs#1065

aschampion: Yeah I'd prefer to go to a lower version of zerocopy instead if possible, but haven't had time to go through their history since there's no changelog. The compiler bump here is just for CI testing the rest of the PR, unless doing a one version bump wouldn't be a dealbreaker.

HeroicKatora: There is none, I think. It was introduced somewhere around `0.2.2` for sure, I've waited on that feature being stabilized [in `image-canvas`](https://github.com/image-rs/canvas/issues/1). And note that all version prior to `0.2.4` have an unsound interface.

HeroicKatora: Also, this would entail restricting the version with an upper bound since `cargo` does not select crates based on compiler requirements (yet). So in essence, we would require an unsound version.. Not the route we want to go probably.

HeroicKatora: So, this leaves us with two choices:
* Bump the compiler version. This is somewhat contrary to making `this as noncontroversial as possible to merge` but honestly, it is a very large and useful feature and the PR is against the major version branch so it is a real possibility :+1: 
* Use something other than `zerocopy`. In the long term it is maybe the most sound solution available but this PR uses only a very small amount of its features so another crate (or porting the code in question, note that this route would *require* a sealed trait imo) could be a better fit.

src/buffer.rs
aschampion: This makes sense. The trait is now sealed as `EncodableLayout`. I'll squash and change the commit message once the rest of the issues are sorted out (if we stick with zerocopy).

.travis.yml
aschampion: Replacing zerocopy with code in the crate would be easy, < 4 lines of unsafe that's clearly safe as it's casting between integer slice types, but given how close image is to being unsafe-free it would be nice to use zerocopy or another similarly-proven alternative until when/if safe transmutes arrive.

If zerocopy would be useful elsewhere in the crate that could be an argument in its favor. @fintelia mentioned this for pixel structs, which could derive `AsBytes` if `Primitive` had an `AsBytes` bound. This might even get at removing the last two `unsafe`s.

For Rust version bumping, usually my opinion would be that it's a one week bump in the version number five months since image's last MSRV bump, so seems fine, but I also see the motive for staying with 1.34.2 in particular because its approximate to a LTS version being in Debian buster.

src/color.rs
```diff
@@ -104,38 +138,47 @@ pub enum ExtendedColorType {
...
-    pub fn channel_count(self) -> u8 {
+    pub fn channels_type(self) -> ChannelsType {
         match self {
             ExtendedColorType::L1 |
             ExtendedColorType::L2 |
             ExtendedColorType::L4 |
             ExtendedColorType::L8 |
             ExtendedColorType::L16 |
-            ExtendedColorType::Unknown(_) => 1,
+            ExtendedColorType::Unknown(_) => ChannelsType::L,
```
fintelia: This sets `ExtendedColorType::Unknown` to have `ChannnelsType::L`, I don't think that is quite right?

src/color.rs
```diff
@@ -314,18 +373,24 @@ impl<T: Primitive> IndexMut<usize> for $ident<T> {
...
-    Luma, 1, 0, "Y", ColorType::L8, #[doc = "Grayscale colors"];
-    Rgba, 4, 1, "RGBA", ColorType::Rgba8, #[doc = "RGB colors + alpha channel"];
-    Bgra, 4, 1, "BGRA", ColorType::Bgra8, #[doc = "BGR colors + alpha channel"];
-    LumaA, 2, 1, "YA", ColorType::La8, #[doc = "Grayscale colors + alpha channel"];
+    Rgb, 3, 0, "RGB", ChannelsType::Rgb, #[doc = "RGB colors"];
+    Bgr, 3, 0, "BGR", ChannelsType::Bgr, #[doc = "BGR colors"];
+    Luma, 1, 0, "Y", ChannelsType::L, #[doc = "Grayscale colors"];
+    Rgba, 4, 1, "RGBA", ChannelsType::Rgba, #[doc = "RGB colors + alpha channel"];
+    Bgra, 4, 1, "BGRA", ChannelsType::Bgra, #[doc = "BGR colors + alpha channel"];
+    LumaA, 2, 1, "YA", ChannelsType::La, #[doc = "Grayscale colors + alpha channel"];
```
fintelia: This and Pixel::CHANNELS_TYPE seems to be the only place that ChannelsType is actually used. Rather than adding an entire extra type to the public API for these, could you instead just pass the 8 and 16 bit versions of each ColorType as two macro arguments? Then color_type() could select between them based on the type argument

src/color.rs
```diff
@@ -180,7 +223,23 @@ impl<T: Primitive + 'static> Pixel for $ident<T> {
...
+            (ChannelsType::L, _) => ColorType::L16,
+            (ChannelsType::La, 1) => ColorType::La8,
+            (ChannelsType::La, _) => ColorType::La16,
+            (ChannelsType::Rgb, 1) => ColorType::Rgb8,
+            (ChannelsType::Rgb, _) => ColorType::Rgb16,
+            (ChannelsType::Rgba, 1) => ColorType::Rgba8,
+            (ChannelsType::Rgba, _) => ColorType::Rgba16,
+            (ChannelsType::Bgr, _) => ColorType::Bgr8,
+            (ChannelsType::Bgra, _) => ColorType::Bgra8,
+            (_, _) => unreachable!(),
```
fintelia: I don't love matching on the size of the subpixel type, but I'm not sure of a better option for now either. Though maybe we should panic any time the size isn't 1 or 2 bytes?

src/buffer.rs
```diff
@@ -1076,6 +1076,14 @@ pub type GrayAlphaImage = ImageBuffer<LumaA<u8>, Vec<u8>>;
...
 /// Sendable Bgr + alpha channel image buffer
 pub(crate) type BgraImage = ImageBuffer<Bgra<u8>, Vec<u8>>;
+/// Sendable 16-bit Rgb image buffer
+pub type Rgb16Image = ImageBuffer<Rgb<u16>, Vec<u16>>;
+/// Sendable 16-bit Rgb + alpha channel image buffer
+pub type Rgba16Image = ImageBuffer<Rgba<u16>, Vec<u16>>;
+/// Sendable 16-bit grayscale image buffer
+pub type Gray16Image = ImageBuffer<Luma<u16>, Vec<u16>>;
+/// Sendable 16-bit grayscale + alpha channel image buffer
+pub type GrayAlpha16Image = ImageBuffer<LumaA<u16>, Vec<u16>>;
```
fintelia: Could we hold off exposing these type aliases for now? We can add them later in any minor release, but shrinking the public interface is much disruptive

src/color.rs
aschampion: Prior to this PR if one was using another primitive type like `f32`, `color_type` wouldn't panic, so that's the behavior I preserved. This does change the behavior because one would now get `ColorType::Rgb16` for an `Rgb<f32>` type instead of `ColorType::Rgb8`, but that seems appropriate or at least less wrong. If `ChannelsType` remains, I could see an argument for having this panic or behave differently for any non-`u8`/`u16` type. However, if `ChannelsType` is removed and this panics, there's no way to do channel-based logic indifferent to the subpixel type for non-`u8`/`u16` subpixel types (e.g., give me the red channel index for this `Pixel`).

src/color.rs
aschampion: According to the comment the only reasonable interpretation for an unknown color type absent other context is as a single channel. It's not quite as clear that the *only* reasonable interpretation of an unknown single channel image absent other context is luminance, but that's the assumption that's made here. However I do think it is *a* reasonable default interpretation. An alternative (if `ChannelsType` remains) is to make this return an `Option`, as you suggested.

HeroicKatora: Would it also be possible to have a `ChannelsType` opaque?

src/color.rs
aschampion: This could be removed, and if that's the preference for now I will so to merge, but I think there needs to be some decision about the ability to abstract logic over channels without having to abstract it over the partial cartesian product of channels and bit depths that `ColorType` represents, as discussed in the `color_type` review comment.

src/buffer.rs
aschampion: Done.

src/color.rs
HeroicKatora: It should match explicitely on `ChannelsType::__NonExhaustive` though, to make it clear that this is actually unreachable.

src/color.rs
fintelia: Codifying that ColorType represents a Cartesian product of channels and bit depths is what concerns me. Currently the only promise we make is that pixels of any color type are an integer number of bytes. ExtendedColorType make the even weaker promise that all pixels are the same number of bits. 

To give examples of where this would break down, some pixel formats might use different numbers of bits for the RGB channels as they do for the alpha channel (i.e. R10G10B10A2). Or they might even have the number of bits per channel differ between different color channels (R5G6B5).
 
Perhaps we want to rule out these formats for ColorType, but I'd rather make that decision separately from the rest of this PR. (We probably do want to support weird formats like these for ExtendedColorType, the entire point of that enum is to encode weird formats we want to be able to name without claiming support for operating on them throughout the library)

aschampion: I agree with this, but see it as more argument in support having separate types to express what channels are in a pixel/image's color space abstracted from their depth, color space and encoding. However, I can see that this is something that can be dealt with independently from this PR, so have removed `ChannelsType`.

aschampion:
With `ChannelsType` removed, I think all comments have been addressed.

The remaining decision, then, is this:

> - Bump the compiler version. This is somewhat contrary to making this as noncontroversial as possible to merge but honestly, it is a very large and useful feature and the PR is against the major version branch so it is a real possibility 👍
> - Use something other than zerocopy. In the long term it is maybe the most sound solution available but this PR uses only a very small amount of its features so another crate (or porting the code in question, note that this route would require a sealed trait imo) could be a better fit.

If we are fine with going forward with zerocopy, I'll squash the cleanup commits and rebase on `next` to fix the merge conflict. If we're not, let me know and I'll start looking at alternatives.

src/buffer.rs
```diff
@@ -8,9 +8,10 @@ use color::{ColorType, FromColor, Luma, LumaA, Rgb, Rgba, Bgr, Bgra};
 use flat::{FlatSamples, SampleLayout};
 use dynimage::{save_buffer, save_buffer_with_format};
 use image::{GenericImage, GenericImageView, ImageFormat, ImageResult};
-use traits::Primitive;
+use traits::{EncodableLayout, Primitive};
 use utils::expand_packed;
 
+
```
fintelia: nit: extra newline here

src/color.rs
```diff
@@ -134,6 +134,7 @@ impl ExtendedColorType {
             ExtendedColorType::Rgba16 |
             ExtendedColorType::Bgra8 => 4,
             ExtendedColorType::__Nonexhaustive => unreachable!(),
+
```
fintelia: nit: extra newline

src/color.rs
```diff
@@ -180,7 +182,12 @@ impl<T: Primitive + 'static> Pixel for $ident<T> {
 
     const COLOR_MODEL: &'static str = $interpretation;
 
-    const COLOR_TYPE: ColorType = $color_type;
+    fn color_type() -> ColorType {
+        match std::mem::size_of::<T>() {
+            1 => $color_type_u8,
+            _ => $color_type_u16,
+        }
+    }
```
HeroicKatora: There might be a way to keep this an associated `const` despite the added logic. A very similar workaround is used within `std` and `core` in various places (and has a long C legacy).

```
const COLOR_TYPE = [$color_type_u8, $color_type_u16][(std::mem::size_of::<T>() > 1) as usize];
```

src/color.rs
HeroicKatora: Separate PR sounds good, let's not mix feature and bugfix changes if we can avoid them.

aschampion:
Nits done, rebased, and fix commits squashed. I could squash out the `ChannelsType` and `Pixel::COLOR_TYPE` noise entirely if that's preferred.

.travis.yml
HeroicKatora: This already matches my own thoughts closely, the PR makes use of `&[u16] -> &[u8]` conversion (trivial `unsafe`) and `&[u8] -> Option<&[u16]>` (not quite trivial but also not incredibly complex). There are no preconditions for the first conversion and the alignment checks that need to be done at runtime for the second case are clear.

With regards to safety, I don't think it matters particularly if `zerocopy` provides these or we include them here. First of all it is a small finite set, *none depend on type parameters* and the publicly exposed interface of it is rather minimal. In a way, the (total) unsafety introduced by depending on `zerocopy` is greater since these are not true for that crate!

The *potential* introduction into `std` (although similar thoughts had been voiced in previous years and still nothing concrete has happened) also points me in the direction of not bumping the MSRV because we would certainly prefer using the standard utilities over `zerocopy` where possible. Using `zerocopy` for small benefit now means we'll be doing another one soon (compatibility with `1.24.1` was kept over ~1.5 years).

I lean towards implementing the conversions in `utils`, but this conclusion and reasoning should be made clear in the implementation. It should be possible to re-evaluate these argument for any future extension made, we do not want to expand to duplicate `zerocopy` and if it proves necessary that any of this becomes dependent on a user provided type parameter we should switch, imho.

Before the implementation, cc @fintelia Thoughts? I know you lean towards less unsafety here so I'd be interested in knowing which of these options you regard as safer.

Shnatsel:
If zerocopy requires too new compiler and/or is too complex, you can use https://github.com/Lokathor/bytemuck instead. It is rather simple internally and passes [MIRI](https://github.com/rust-lang/miri) on its test suite, which checks for alignment issues as well as a number of Rust-specific rules such as `&mut` exclusivity, pointer provenance, etc.

HeroicKatora:
Additional argument for it: `bytemuck` has reached `1.0`.

aschampion:
`bytemuck` looks great, I'll give it a go.

aschampion:
Switch to bytemuck done, removed the 1.35 bump commit and tested that it builds on 1.34.2.

PR #1087, opened 2019-11-28
Replace Lenna test images

aschampion:
These images are now discouraged for use by the [Nature Publishing Group, Optical Society, and SIAM](https://www.nature.com/articles/s41565-018-0337-2), as well as having been removed from other major image processing libraries including Matlab and [FIJI/ImageJ](https://forum.image.sc/t/request-to-remove-lena-as-a-sample-image/24498).

Instead the Mandrill image (SIPI image 4.2.03) is used for tiff integration testing, benchmarking, and examples, and a reference image from PNGSuite is used for interlaced tests.


fintelia:
Didn't realize we included these. Thanks for fixing

PR #1088, opened 2019-11-29
Add limits to decoder

### LINKED ISSUE #938: Library wide memory limits
By now most decoder libraries have some sort of `Limits` struct or sanity checks that stops them from trying to decode images that would make them run out of memory. However this library doesn't have anything like it that could be used to configure these limits of the individual decoders. We should also use them to set a limit to the size of `ImageBuffer` and alike, with https://github.com/image-rs/image-png/pull/115 for example the png decoder will happily decode an image that is 100 000 x 100 000 pixels large as long as it doesn't use to much memory. In these cases we should then have a second check that we are not allocating a `ImageBuffer` of that size.

birktj:
This PR adds limits to the decoders that support it and adds a method for setting these in the `io::Reader` interface.

This fixes #938

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -26,7 +26,7 @@ use image;
 use image::{
     GenericImage, GenericImageView, ImageDecoder, ImageError, ImageFormat, ImageOutputFormat, ImageResult,
 };
-use io::free_functions;
+use io::{DecodingLimits, free_functions};
 use imageops;
 
 /// A Dynamic Image
@@ -133,7 +133,14 @@ impl DynamicImage {
     pub fn from_decoder<'a>(decoder: impl ImageDecoder<'a>)
         -> ImageResult<Self>
     {
-        decoder_to_image(decoder)
+        decoder_to_image(decoder, DecodingLimits::default())
+    }
+
+    /// Decodes an encoded image into a dynamic image.
+    pub(crate) fn from_decoder_limits<'a>(decoder: impl ImageDecoder<'a>, limits: DecodingLimits)
+        -> ImageResult<Self>
+    {
+        decoder_to_image(decoder, limits)
     }
 
     /// Returns a copy of this image as an RGB image.
@@ -639,9 +646,14 @@ impl GenericImage for DynamicImage {
 }
 
 /// Decodes an image and stores it into a dynamic image
-fn decoder_to_image<'a, I: ImageDecoder<'a>>(decoder: I) -> ImageResult<DynamicImage> {
+fn decoder_to_image<'a, I: ImageDecoder<'a>>(decoder: I, limits: DecodingLimits) -> ImageResult<DynamicImage> {
     let (w, h) = decoder.dimensions();
     let color_type = decoder.color_type();
+    
+    if ((color_type.bits_per_pixel() + 7) / 8) as u64 * w as u64 * h as u64 > limits.buffer_limit as u64 {
+        return Err(ImageError::InsufficientMemory)
+    }
+
     let buf = image::decoder_to_vec(decoder)?;
 
     let image = match color_type {
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -56,6 +56,8 @@ impl<R: Read> GifDecoder<R> {
     pub fn new(r: R) -> ImageResult<GifDecoder<R>> {
         Self::new_with_limits(r, DecodingLimits::default())
     }
+
+    /// Same as `new`, but allows you to set limits.
     pub fn new_with_limits(r: R, limits: DecodingLimits) -> ImageResult<GifDecoder<R>> {
         let gif_limits = gif::MemoryLimit(min(u32::max_value() as usize, limits.buffer_limit) as u32);
         let mut decoder = gif::Decoder::new(r);
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -43,6 +43,7 @@ use animation;
 use buffer::{ImageBuffer, Pixel};
 use color::{self, Rgba};
 use image::{self, AnimationDecoder, ImageDecoder, ImageError, ImageResult};
+use io::DecodingLimits;
 use num_rational::Ratio;
 
 /// GIF decoder
@@ -53,7 +54,12 @@ pub struct GifDecoder<R: Read> {
 impl<R: Read> GifDecoder<R> {
     /// Creates a new decoder that decodes the input steam ```r```
     pub fn new(r: R) -> ImageResult<GifDecoder<R>> {
+        Self::new_with_limits(r, DecodingLimits::default())
+    }
+    pub fn new_with_limits(r: R, limits: DecodingLimits) -> ImageResult<GifDecoder<R>> {
+        let gif_limits = gif::MemoryLimit(min(u32::max_value() as usize, limits.buffer_limit) as u32);
         let mut decoder = gif::Decoder::new(r);
+        decoder.set(gif_limits);
         decoder.set(ColorOutput::RGBA);
 
         Ok(GifDecoder {
--- a/src/io/free_functions.rs
+++ b/src/io/free_functions.rs
@@ -63,25 +63,25 @@ pub(crate) fn load_with_limits<R: BufRead + Seek>(r: R, format: ImageFormat, lim
     // Default is unreachable if all features are supported.
     match format {
         #[cfg(feature = "png_codec")]
-        image::ImageFormat::Png => DynamicImage::from_decoder(png::PngDecoder::new_with_limits(r, limits)?),
+        image::ImageFormat::Png => DynamicImage::from_decoder_limits(png::PngDecoder::new_with_limits(r, limits)?, limits),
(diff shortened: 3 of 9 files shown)


birktj:
This needs slightly more work before it is completely ready, but I would like some input on the design.

src/gif.rs
```diff
@@ -53,7 +54,14 @@ pub struct GifDecoder<R: Read> {
...
     /// Creates a new decoder that decodes the input steam ```r```
     pub fn new(r: R) -> ImageResult<GifDecoder<R>> {
+        Self::new_with_limits(r, DecodingLimits::default())
+    }
+
+    /// Same as `new`, but allows you to set limits.
+    pub fn new_with_limits(r: R, limits: DecodingLimits) -> ImageResult<GifDecoder<R>> {
+        let gif_limits = gif::MemoryLimit(min(u32::max_value() as usize, limits.buffer_limit) as u32);
         let mut decoder = gif::Decoder::new(r);
+        decoder.set(gif_limits);
```
HeroicKatora: Here we have `new_with_limits`, in the case of `io::Reader` there is `set_limits`. These should agree on one kind of interface.

src/io/free_functions.rs
```diff
@@ -102,13 +107,13 @@ pub(crate) fn image_dimensions_with_format_impl<R: BufRead + Seek>(fin: R, forma
         #[cfg(feature = "jpeg")]
         image::ImageFormat::Jpeg => jpeg::JpegDecoder::new(fin)?.dimensions(),
```
HeroicKatora: These match arms ignore the limits. That seems contradictory. We might think about ignoring defaulted limits but if a limit has been set *explicitely* then this should not happen and I'd rather outright refuse to decode than ignore them.

src/io/limits.rs
```diff
@@ -0,0 +1,15 @@
+/// Limits on how much memory that can be used in the decoding process. These limits are "best
+/// effort" limits and there are no strict guarantees that they will be followed at all times.
+#[derive(Debug, Copy, Clone)]
+pub struct DecodingLimits {
+    /// A limit on how large buffers a decoder is allowed to allocate. The default is 512MiB.
+    pub buffer_limit: usize,
```
HeroicKatora: Some formats already demonstrated the potential usefulness of more fine-grained limits. There should be some plan on extending the limit system, so this should not be a completely public struct.

src/io/limits.rs
```diff
@@ -0,0 +1,15 @@
+/// Limits on how much memory that can be used in the decoding process. These limits are "best
+/// effort" limits and there are no strict guarantees that they will be followed at all times.
+#[derive(Debug, Copy, Clone)]
+pub struct DecodingLimits {
```
HeroicKatora: A limit on `width` and `height` would also make sense. That would also be supported by all formats.

src/dynimage.rs
```diff
@@ -133,7 +133,14 @@ impl DynamicImage {
     pub fn from_decoder<'a>(decoder: impl ImageDecoder<'a>)
         -> ImageResult<Self>
     {
-        decoder_to_image(decoder)
+        decoder_to_image(decoder, DecodingLimits::default())
+    }
+
+    /// Decodes an encoded image into a dynamic image.
+    pub(crate) fn from_decoder_limits<'a>(decoder: impl ImageDecoder<'a>, limits: DecodingLimits)
```
fintelia: What is the point of this method if it just directly calls decoder_to_image? Is there some issue with the generics?

src/gif.rs
birktj: I have used `new_with_limits` when creating the decoders in order to set the limits as early as possible when starting decoding, most of the decoders start reading data in their `new` functions and I thought it was a good idea to set the limits before this.

src/dynimage.rs
birktj: I just copied the `from_decoder` method, but you are right. This would be simpler with just calling `decoder_to_image`

src/io/free_functions.rs
birktj: I see where you are coming from, but currently only three formats support limits. I think we should think a bit about this first.

birktj: Actually, by incorporating @fintelia's suggestion with a `ImageDecoder::set_working_bytes` method this could be implemented quite elegantly.

HeroicKatora: I'm aware that only three format support limits and we have only fuzzed png to ensure that the support is actually somewhat reasonable. But that's precisely my point, setting limits might be used as a level of sandboxing and it is completely unfit for that usecase when we then ignore the limits.

src/gif.rs
HeroicKatora: So because decoders can't delay reading metadata they have to do this in fallible constructors. This is apparently good enough so we don't have to modify it. I'm mostly wondering if the reasoning to make the metadata queries non-fallible in 2957fc25ba798fe42a661dccd9544c3e2a9a3167 still holds true.

cc: @fintelia

fintelia: Are there any formats were the resource use from decoding metadata is a concern? It generally just amounts to reading constant size headers, right?

The motivation for requiring metadata decoding in the constructor was that it streamlined the rest of the interface. If the constructor was infallible instead, then implementors would have to do lazy loading of metadata and any users would have to deal with possible I/O errors on every metadata call. Any failure to retrieve the basic image metadata necessarily means that loading the rest of the image cannot succeed.

And once you have the image metadata, a lot of the limits logic can actually live in the caller. Want to make sure you don't end up with an image that is too large? Just check the dimensions before calling decode. Intermediate buffers are still a concern, but long term I think we could have decoders for all the formats that use only a constant amount of additional space.

HeroicKatora: > Are there any formats were the resource use from decoding metadata is a concern? It generally just amounts to reading constant size headers, right?

I can't think of an immediate concern, we don't track constant space usage in any case since that would include the stack, local variables, etc. which is infeasible and not as important. I was just curious. But it is not entirely clear to me if *all* formats have a constant resource usage for reading metadata. They have `io::Seek` but need to refrain from reading any variable length metadata (EXIF comments, etc.) in the constructor.

src/io/limits.rs
Shnatsel: Prior art from libpng includes width/height in its limits, but it does not provide a rationale. It seems to do so for legacy reasons: https://libpng.sourceforge.io/decompression_bombs.html

Aside of width/height there is also frame count to consider for GIF and APNG, but I'm not convinced these need to be separate knobs and not a global limit on the in-memory size of the decompressed image.

HeroicKatora:
@birktj Triage: Can you address some of the comments and rebase on recent work?

birktj:
@HeroicKatora I will try to go for @fintelia's suggestion, however this is sufficiently different from the current path that the most efficient would probably be to do a new PR.

src/dynimage.rs
HeroicKatora: Doesn't this duplicate implementations? It would introduce an array of different `set_*` methods that will need to be added to both `Reader` and `ImageDecoder`. Would grouping those parameters into the `Limits` struct not be equivalent but require less interfaces? Chances are that the decoders will store many of the same resource restrictions internally.

HeroicKatora:
I'm not sure yet which approach has more technical merrit. My goal here was to invigorate the discussions itself :slightly_smiling_face:

fintelia:
> Doesn't this duplicate implementations? It would introduce an array of different set_* methods that will need to be added to both Reader and ImageDecoder. Would grouping those parameters into the Limits struct not be equivalent but require less interfaces? Chances are that the decoders will store many of the same resource restrictions internally.

Not all of the limits make sense on both. Any limit based on image metadata should be enforced by the Reader by calling the associated getter function on the inner ImageDecoder and validating the return value is in the allowable range. Meanwhile the actual ImageDecoders themselves shouldn't have to track and enforce these same limits. *set_max_working_bytes* is an exception because the amount of working space needed may not be known before decoding starts.

This design also has the property that new limits can be added in a backwards compatible way while still being clear to the user whether the requested limits are being enforced.

lilith:
What's the status on this change?

Gelbpunkt:
Hi, this would be very important for us to have in the library and I'd love to pick up the work from here, but my Rust is way too limited. Are there any plans to add this in the near future? :smile:

fintelia:
@Gelbpunkt I don't think there's any immediate plans to implement this. However, depending on which limits you wanted to enforce, it might be possible to do the necessary checks yourself without having to wait on this feature. What exactly is your use case?

Gelbpunkt:
@fintelia I need to be able to detect decompression bombs when I draw onto a base image provided by a user. The best I could do for now was using a Reader and into_dimensions to check the actual image size and prohibit a specific size, but I'm not sure if that catches all cases. Having the library handle this would take a lot of work and worry from me.

birktj:
I am going to close this for now. The design isn't perfect and by now the PR would probably require a full rewrite anyways. If anybody feel like picking up the torch that would be great.

PR #1089, opened 2019-11-30
Document and expose all new pub structures

HeroicKatora:
Of note, these were previously hidden despite `#[deny(unreachable_pub)]` because they were no completely unreachable, just unnamable. Note that the `ImageDecoder` utilizes them both in function signatures and through some trait trickery it was theoretically possible to type-def them (or so I think).

Also, `Progress` was completely opaque and missing any accessors, contrary to its use in the callback since it could not be inspected. Sorry :cry: .

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -379,8 +379,8 @@ pub(crate) fn decoder_to_vec<'a>(decoder: impl ImageDecoder<'a>) -> ImageResult<
 /// Represents the progress of an image operation.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Progress {
-    current: u64,
-    total: u64,
+    pub current: u64,
+    pub total: u64,
 }
 
 /// The trait that all decoders implement
--- a/src/image.rs
+++ b/src/image.rs
@@ -383,15 +383,23 @@ pub(crate) fn decoder_to_vec<'a>(decoder: impl ImageDecoder<'a>) -> ImageResult<
 /// progress `(0, 0)` if progress is unknown, without violating the interface contract of the type.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Progress {
-    /// A measure of completed decoding.
-    pub current: u64,
-    /// A measure of all necessary decoding work.
-    ///
-    /// This is in general greater or equal than `current`.
-    pub total: u64,
+    current: u64,
+    total: u64,
 }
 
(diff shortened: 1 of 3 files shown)


src/image.rs
```diff
@@ -377,12 +377,35 @@ pub(crate) fn decoder_to_vec<'a>(decoder: impl ImageDecoder<'a>) -> ImageResult<
...
+/// Note that this is not necessarily accurate and no change to the values passed to the progress
+/// function during decoding will be considered breaking. A decoder could in theory report the
+/// progress `(0, 0)` if progress is unknown, without violating the interface contract of the type.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Progress {
     current: u64,
     total: u64,
 }
 
+impl Progress {
```
fintelia: Instead of this impl block, I'd rather just have the two fields be marked pub.

src/image.rs
```diff
@@ -377,12 +377,35 @@ pub(crate) fn decoder_to_vec<'a>(decoder: impl ImageDecoder<'a>) -> ImageResult<
 }
 
 /// Represents the progress of an image operation.
+///
+/// Note that this is not necessarily accurate and no change to the values passed to the progress
+/// function during decoding will be considered breaking. A decoder could in theory report the
+/// progress `(0, 0)` if progress is unknown, without violating the interface contract of the type.
```
fintelia: I don't see the point of saying this. If a decoder doesn't support reporting progress, it can just do a callback with (0, total_bytes) before it starts and (total_bytes, total_bytes) after it finishes.

src/image.rs
HeroicKatora: Not really, as explain in the commit comment I'm not sure if we want to add different fields in the future? Like you said, it *might* mean bytes decoded but it could also be pixels. Until that is clear, I'd rather have the members be private so we can switch at any point.

src/image.rs
HeroicKatora: What you describe *is* a form or progress reporting and it requires the decoder to know `total_bytes`. For a compressed stream, which are the total bytes? The bytes to represent the image or the bytes of the compressed stream?

I just want to say that you should be extra cautious and leave us more freedom on whether we implement this for a decoder. Note that if we were to not leave this open, we need to asnwer the above question to provide progress reports for decoders that do not have such a callback in their own implementation.

Issue #1073, opened 2019-11-03
Release 0.23.0-preview0


fintelia:
The `next` branch has accumulated a bunch of changes that improve various aspects of the library, but require a major version bump. We should release a preview version of the crate soon so that we can test them.

Are there any remaining patches we should wait on before a first preview release?

HeroicKatora:
Of those outstanding PRs that should go into `next`:

- [ ] #1066 is a consistency change, necessary to start working on error types
- [ ] #1062 is rather small in scope, but with definitive impact on crate users. I'd rather put it into the first preview to set clearer expectations.

Other changes seem less consquential.

- #928 is an ambitious work but needs time to find the correct set of variants anyways. It's not highly critical.
- #930, #1060 show that we might want to dedicate some thought into making the `ImageFormat` enum non-exhaustive, potentially other types as well. However, that's not decided and delaying it until a finalized `0.23` seems just fine.
- #1036 likewise will not likely affect too many users
- No tracking issue but we would want to update some dependencies. That would mean ensuring that there aren't any issues blocking us from doing so. However, that's no breaking concern for the preview release.

CI indicates the branch is theoretically ready

CryZe:
image seems to use tiff 0.3.1 at the moment. It seems tiff 0.4.0 just got released. Will that also go into 0.23.0 of image? The reason why I'm asking is that tiff 0.3.1 seems to be the only crate pulling syn <1.0 into my dependency tree, which slows down compile times significantly.

nox:
Any ETA for 0.23?

Shnatsel:
I'd like to point out that while #1036 is not that critical in and of itself, it blocks further fuzzing. So it may be preventing discovery of other failure conditions on crafted inputs - panics, memory exhaustion, infinite loops, etc.

nox:
Seems like the biggest blocker is now #928.

Manishearth:
Could we perhaps have a patch release on 0.22 that uses the newer tiff (so that others can use the new num_derive and thus syn)? Duplicated syn dependencies really slow down build times

nox:
tiff is exposed from the public API which is why we are waiting on 0.23, AFAIK.

Manishearth:
Ah, makes sense.

PR #1091, opened 2019-12-05
Update release and notes for 0.23 preview

HeroicKatora:
Closes: #1073

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,11 +24,14 @@ version.
 - Heavily reworked `ColorType`:
   - This type is now used for denoting formats for which we support operations
       on buffers in these memory representations. Particularly, all channels in
-      pixel types are assumed to be sized in units of bytes.
+      pixel types are assumed to be an integer number of bytes (In terms of the
+      Rust type system, these are `Sized` and one can crate slices of channel
+      values).
   - An `ExtendedColorType` (WIP: do You have better name?) is used to express
-      more generic color formats from which the library can potentially
-      convert/scale/map into a `ColorType` buffer. This operation might be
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -8,6 +8,52 @@ Rust image aims to be a pure-Rust implementation of various popular image format
...
+WIP: This preview version is intended to allow an evaluation and feedback on
+the new interface. It is not yet perfect with regards to color spaces but it
+was designed mainly as an improvement over the current interface with regards
+to in-memory color formats, first. We'll get to color spaces in a later major
+version.
+
+- Heavily reworked `ColorType`:
+  - This type is now used for denoting formats for which we support operations
+      on buffers in these memory representations. Particularly, all channels in
+      pixel types are assumed to be sized in units of bytes.
```
fintelia: "assumed to be sized in units of bytes" sounds like RGB8 might be 8 bytes for each of the red green and blue channels. "assumed to be an integer number of bytes" might be more clear

CHANGES.md
```diff
@@ -8,6 +8,52 @@ Rust image aims to be a pure-Rust implementation of various popular image format
...
+  - An `ExtendedColorType` (WIP: do You have better name?) is used to express
+      more generic color formats from which the library can potentially
+      convert/scale/map into a `ColorType` buffer. This operation might be
```
fintelia: "convert/scale/map into a `ColorType` buffer" --> "has more limited support" ?

CHANGES.md
HeroicKatora: I wanted to capture a relation to Rust's type system here but the wording trailed off into 'clever' instead of 'understandable' :grinning:

PR #1093, opened 2019-12-07
Simplify cargo toml

fintelia:
This series of patches makes the Cargo.toml more concise, updates it in a few places, and removes the "gif_codec" and "png_codec" features in favor of just referring to them via "gif" and "png" respectively.

The first two items should have no end user impact, but renaming features represents a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,12 +37,10 @@ glob = "0.3"
 quickcheck = "0.6.2"
 
 [features]
-default = ["gif_codec", "jpeg", "ico", "png_codec", "pnm", "tga", "tiff", "webp", "bmp", "hdr", "dxt", "jpeg_rayon"]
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -62,18 +37,15 @@ glob = "0.3"
...
-png_codec = ["png"]
+ico = ["bmp", "png"]
 pnm = []
 tga = []
 webp = []
 bmp = []
 hdr = ["scoped_threadpool"]
 dxt = []
-jpeg_rayon = ["jpeg-decoder/rayon"]
+jpeg_rayon = ["jpeg/rayon"]
```
HeroicKatora: Can we remove `jpeg` from the name of this feature, at least in the long term? I know it's a bit hypothetical but I'd want to have other decoders utilize SIMD and not only `jpeg`. Simply `rayon` might conflict with the package if we ever include it directly, do you have ideas for a better name?

fintelia: I don't know any better options

HeroicKatora: It's fine for now, just would have been a nice-to-have.

PR #1094, opened 2019-12-08
Move core types to separate image-core crate

fintelia:
This patch should have minimal user visible changes,* but internally restructures the code to depend on a separate _image-core_ crate. This way we can evolve the core types and traits independently of the main crate, hopefully reaching 1.0 within a reasonable time span.

Once the ImageDecoder trait (and related types) are split off, it will be possible for the individual image format crates to implement it directly. That will remove a lot of boiler plate, and make the image ecosystem must more interoperable overall. (Since any breaking changes to image-core will require all downstream crates to bump their major version numbers, we should hold off doing this until it it has gotten close to if not reached 1.0 status.)

For now I use a github dependency on _image-core_, but that should of course be replaced by a published version on crates.io before merging.

*The main user visible changes I had to make were removing a couple From implementations on ImageError, and making Progress construable by downstream crates (since _image_ is a downstream crate of _image-core_)


HeroicKatora:
I have the impression that this is too hasty. At least due to the unclear issues of color support I had always expected proving the interface good enough by reworking (maybe all, but at least some) decoder *internals* first. As you said yourself, the decoders publicly relying on `image-core` is a major hazard for breaking changes.

HeroicKatora:
Nevertheless, this shows a few rough edges anyways. The public 
 >`impl  From<jpeg_decoder::PixelFormat> for ColorType`

is not intentional and only did not cause previous concerns because `jpeg` turned out to be *very* stable.

PR #1096, opened 2019-12-13
Add DynamicImage::into_* functions.

de-vri-es:
This PR adds `into_*` functions to `DynamicImage`. These are like `to_*`, but they avoid a copy in case the image was already in the correct format.

Interesting future work could be to reuse the existing buffer when possible.


de-vri-es:
Thanks for the quick review and merge :)

PR #1097, opened 2019-12-13
Fixes #1090.

### LINKED ISSUE #1090: Encoding huge BMP panics/creates corrupted file
I’m using _image 0.22.3_ with _rustc 1.39_ on 64-bit Windows 10.

```rust
let img = RgbImage::new(40_000, 40_000);
img.save_with_format("image.bmp", ImageFormat::BMP).unwrap();
```

This code panics in debug mode and creates an (assumably corrupted) file in release mode.

`thread 'main' panicked at 'attempt to multiply with overflow', bmp\encoder.rs:36:26`

The BMP format might not be able to store such a large image, but the encoder should at least return an `Err()` in both debug and release modes in such a case.

Mike-Neto:
Fixes #1090.

Adds checked_mul when computing img size.

Not really that happy about the solution I implemented. Open for feedback :)

### DIFF
--- a/src/bmp/encoder.rs
+++ b/src/bmp/encoder.rs
@@ -33,15 +33,14 @@ impl<'a, W: Write + 'a> BMPEncoder<'a, W> {
 
         let (dib_header_size, written_pixel_size, palette_color_count) = get_pixel_info(c)?;
         let row_pad_size = (4 - (width * written_pixel_size) % 4) % 4; // each row must be padded to a multiple of 4 bytes
-        let image_size = width.checked_mul(height);
-        if image_size.is_none() {
-            return Err(ImageError::DimensionError);
-        }
-        let image_size = image_size.unwrap().checked_mul(written_pixel_size);
-        if image_size.is_none() {
-            return Err(ImageError::DimensionError);
-        }
-        let image_size = image_size.unwrap() + (height * row_pad_size);
+        let image_size = width
+            .checked_mul(height)
+            .ok_or(ImageError::DimensionError)?
+            .checked_mul(written_pixel_size)
+            .ok_or(ImageError::DimensionError)?
+            .checked_add(height * row_pad_size)
+            .ok_or(ImageError::DimensionError)?;
+
         let palette_size = palette_color_count * 4; // all palette colors are BGRA
         let file_size = bmp_header_size + dib_header_size + palette_size + image_size;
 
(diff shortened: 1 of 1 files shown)


src/bmp/encoder.rs
```diff
@@ -33,8 +33,15 @@ impl<'a, W: Write + 'a> BMPEncoder<'a, W> {
...
-        let image_size = width * height * written_pixel_size + (height * row_pad_size);
+        let image_size = width.checked_mul(height);
+        if image_size.is_none() {
+            return Err(ImageError::DimensionError);
+        }
+        let image_size = image_size.unwrap().checked_mul(written_pixel_size);
+        if image_size.is_none() {
+            return Err(ImageError::DimensionError);
+        }
+        let image_size = image_size.unwrap() + (height * row_pad_size);
```
HeroicKatora: Instead of bool `is_none` checks and `unwrap`, it is more idiomatic to use `match` or the `?` (called try) operator. The former is for cases where error handling needs explicit control flow, the latter should be sufficient here. Also, it is easy to chain these operators together so there are no intermediate results that might get accidentally used later (e.g. if one deletes the last assignment during a refactoring).

A sketch of the lines here with this style looks like so:
```
let image_size = width.checked_mul(height)
    .ok_or(ImageError::DimensionError)?
    .checked_mul(written_pixel_size)
    .ok_or(ImageError::DimensionError)?;
```


And finally, note both the computation part `+ (height * row_pad_size)` and `width * written_pixel_size` during padding calcuation itself could also overflow. Fixing these as well would make the code even more stable.

Mike-Neto: Thanks, I knew there's got to be a better way to do it, I got confused by checked_mul returning an option instead of a result so I wasn't aware that I could map from none using ok_or right away 👍 .

Updated the other addition as well but didn't check the inner most multiplication as if height * width didn't trigger overflow it's very unlikely that height * [0-3] would.

Thanks 😄

HeroicKatora:
These commits are based on `next`. Since there are no breaking changes introduced here can you rebase them onto `master` and also edit the target of the PR? We'll bring them to `next` as well but a (final) bugfix release for the current version wouldn't hurt.

Mike-Neto:
I'd rather do it in a follow up PR as in `master` the return value for the function is not an `ImageResult` meaning I can't reuse this implementation.

Is that OK?

HeroicKatora:
> Is that OK?

Sure, why not.

PR #1100, opened 2019-12-15
Fixes #1090. Checks for overflowing dimensions.

### LINKED ISSUE #1090: Encoding huge BMP panics/creates corrupted file

Mike-Neto:
Refactored for master.

### DIFF
--- a/src/bmp/encoder.rs
+++ b/src/bmp/encoder.rs
@@ -33,22 +33,15 @@ impl<'a, W: Write + 'a> BMPEncoder<'a, W> {
         let (dib_header_size, written_pixel_size, palette_color_count) = get_pixel_info(c)?;
         let row_pad_size = (4 - (width * written_pixel_size) % 4) % 4; // each row must be padded to a multiple of 4 bytes
 
+        let make_invalid_input_err =
+            || io::Error::new(io::ErrorKind::InvalidInput, "Overflowing dimensions");
         let image_size = width
             .checked_mul(height)
-            .ok_or(io::Error::new(
-                io::ErrorKind::InvalidInput,
-                "Overflowing dimensions",
-            ))?
+            .ok_or_else(make_invalid_input_err)?
             .checked_mul(written_pixel_size)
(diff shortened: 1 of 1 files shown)


src/bmp/encoder.rs
```diff
@@ -33,7 +33,22 @@ impl<'a, W: Write + 'a> BMPEncoder<'a, W> {
         let (dib_header_size, written_pixel_size, palette_color_count) = get_pixel_info(c)?;
         let row_pad_size = (4 - (width * written_pixel_size) % 4) % 4; // each row must be padded to a multiple of 4 bytes
 
-        let image_size = width * height * written_pixel_size + (height * row_pad_size);
+        let image_size = width
+            .checked_mul(height)
+            .ok_or(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "Overflowing dimensions",
+            ))?
```
HeroicKatora: Compared to the error code on the `next` branch, this performs some real work and an allocation. Providing these arguments by value means that *all* code paths, not only the erroring ones, will instantiate these three times. This should utilize a closure and `ok_or_else` instead.

```rust
let make_invalid_input_err = || io::Error::new(
    io::ErrorKind::InvalidInput,
    "Overflowing dimensions");

let image_size = width
    .checked_mul(height)
    .ok_or_else(make_invalid_input_err)?
```

Mike-Neto: Great comment thanks. Updated as requested.

PR #1101, opened 2019-12-15
Remove deprecated Error::description

kornelski:
Rust is about to hard-deprecate Error::description, which has been soft deprecated for over a year.

Also switches `cause` to the newer `source` (requires 1.30.0).


HeroicKatora:
I reviewed these commits since CI showed `master` as broken afterwards. Somehow, the compiler got confused with very odd behaviour [in these builds](https://travis-ci.org/image-rs/image/builds/625418415?utm_source=github_status&utm_medium=notification). Note that some builds fail in the `quickcheck` dependency although they all use the same crate version, and in some cases there are other builds with the same compiler version and OS as well :fearful:

PR #1103, opened 2019-12-26
trimmed whitespace

adamnemecek:


HeroicKatora:
Usually simple formatting PRs without CI integration or some other way of ensuring they will be consistent are noise, sometimes hindering merges. However, it appears the bulk of those are in code that I touched so I will instead adjust my editor config and ask: Are those all whitespace errors?

adamnemecek:
Yes.

PR #1104, opened 2019-12-26
tiff: don't apply decoding_buffer_size limit

fintelia:
In the image::Decoder API, the user supplies the buffer to decode into. Since we aren't the ones allocating the buffer, we as a library don't need to worry about denial of service from allocating too much memory. TiffDecoder does not currently have any way to set the limits on the underlying tiff::decoder::Decoder, so this patch is necessary to load Tiff files > 256MB at all.

### DIFF
--- a/src/tiff.rs
+++ b/src/tiff.rs
@@ -33,7 +33,12 @@ impl<R> TiffDecoder<R>
 {
     /// Create a new TiffDecoder.
     pub fn new(r: R) -> Result<TiffDecoder<R>, ImageError> {
-        let mut inner = tiff::decoder::Decoder::new(r).map_err(ImageError::from_tiff)?;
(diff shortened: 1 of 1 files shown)


HeroicKatora:
The limits set in  `tiff` *should* only count buffers *used* during decoding. It is correct that this should *not* count the user provided buffer but only those allocated by the library itself.  That said, I don't quite understand what removing the limits for all uses completely has to do with that reasoning. We eventually call [`Decoder::read_image`](https://docs.rs/tiff/0.4.0/tiff/decoder/struct.Decoder.html#method.read_image) which *does* allocate a lot of memory on its own before copying samples to the output vector. This *is* a potential DOS vector. I'd be glad if we were to address this in `tiff` by adding is a method to read a full image into a pre-allocated buffer and removing the limit requirement in it.

PR #1105, opened 2019-12-30
Remove some unneeded scopes

euank:
Now that non-lexical-lifetimes have landed, we can drop some unneeded
indentation!

For reviewing this, I recommend using the ['ignore whitespace' diff](https://github.com/image-rs/image/pull/1105/files?utf8=%E2%9C%93&diff=split&w=1); it's a lot easier to see how trivial the change is that way.

(License blurb)

<Details>

</Details>

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -1119,134 +1119,131 @@ impl<R: Read + Seek> BmpDecoder<R> {
         let mut delta_pixels_left = skip_pixels;
         let mut eof_hit = false;
 
-        // Scope the borrowing of pixel_data by the row iterator.
-        {
-            // Handling deltas in the RLE scheme means that we need to manually
-            // iterate through rows and pixels.  Even if we didn't have to handle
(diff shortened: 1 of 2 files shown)


JohnTitor:
Note that NLL for 2015 has been enabled since 1.36.0.

euank:
It turns out the travis.yml configuration explicitly includes rust 1.34.2:

https://github.com/image-rs/image/blob/22adfbb92b09cb4c5fef3a84be62f840282ed135/.travis.yml#L13

Because of that, CI fails on this.

I'm happy to close this PR if we still want to support older versions. I think the main argument in favor of closing this for now is that some distros, e.g. debian stable, still ship an older version of rustc that couldn't compile the code after this change.

HeroicKatora:
Indeed, the argument was always having Debian Stable in mind.

euank:
I'll go ahead and close this. We can drop them once debian buster becomes oldstable I think. Sorry for the noise!

Issue #987, opened 2019-07-11
implement imageops::replace for overlapping regions of an image


lovasoa:
Because of rust's borrowing limitations, it is currently not possible to use the various functions of this crate that  take two image references, one mutable and the other immutable with data that is coming from a single image. However, it is a completely reasonable operation to, for instance, replace a region of an image by another region of the same image, that may be overlapping.

This function could have the following form

```rust 
fn copy_to<I>(patch: SubImage<I>, destination: (u32,u32))
```

Having such a primitive would help me with my [seamcarving](https://github.com/lovasoa/seamcarving) implementation.

HeroicKatora:
So like [`slice::copy_within`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_within) from which I would borrow the naming as well. It also provides a clear interface pattern of the form `copy_within(&mut self, from_region, to)` which would translate to

> `copy_within(&mut self, from: (u32, u32, u32, 32), to: (u32, u32))`

But there should problably be some proper `Rect` type for the first argument instead of that ad-hoc 4-tuple with unclear components. The problem with taking a subimage in the first argument is that this would be an active borrow of `self`and not possible.

HeroicKatora:
Also, the simple implementation is just a loop in the correct order. `ImageBuffer` could provide a specialization to speed-up the process with the slice `copy_within` implementation but unfortunately we do not have actual specialization yet.

lovasoa:
Why would it require specialization ? Couldn't it be done with a default trait method in `GenericImage`, and an implementation of this method in ImageBuffer ?

HeroicKatora:
Sure, some better implementation than the default one is already an improvement for `ImageBuffer`.  I thought there might be additional speedup that could be gained from specializing for some `Pixel` types but you can probably subvert it and work on the raw `Subpixel` samples anyways which are required to be [`Copy`](https://docs.rs/image/0.21.2/image/trait.Primitive.html). So should be all good.

HeroicKatora:
Closed in #1107.

PR #1107, opened 2020-01-03
Implement copy_within for GenericImage

Veykril:
Attempts to implement #987 with a specialized version for `ImageBuffer`

I unfortunately couldn't come up with a way of implementing this without basically copy pasting the copy logic 4 times while still using static dispatch due to there being 4 different cases to handle that determine in what order to copy the pixels. Hence I wrote a small macro to make the implementation more readable(even though macro normally does the opposite). If you got any ideas for how to improve this I'd be happy to change this.

Regarding the the function signature I currently chose to use a tuple for the two corners of the source and destination rectangles with a width and height parameter, this should obviously thought about a bit more. The other alternatives being two tuples one describing the source rectangle and the other describing the top left corner of the destination rectangle or instead of tuples using proper rectangle structs for the source rectangle.

Sample on how those signatures would look like when calling them:
```rust
image.copy_within((0, 0), (1, 1), 20, 13);
image.copy_within((0, 0, 20, 13), (1, 1));
image.copy_within(Rect { x: 0, y: 0, width: 20, height: 13) }, (1, 1));
```

I also made it return a bool since thats what `copy_from` currently does(https://github.com/image-rs/image/issues/901).

Edit: I forgot that [`slice::copy_within`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_within) became stable on 1.37, so assuming that this crate has MSRV 1.34 the specialized version of `ImageBuffer` has to be changed first, the only efficient way that comes to mind right now would be `ptr::copy`(given thats how copy_within is implemented), but given it's an unsafe operation I dont know how well received that would be.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -921,6 +921,35 @@ where
         self.get_pixel_mut(x, y).blend(&p)
     }
 
+    fn copy_within(&mut self, from: (u32, u32), to: (u32, u32), width: u32, height: u32) -> bool {
+        let (fx, fy) = from;
+        let (tx, ty) = to;
+        if tx.max(fx) + width > self.width() || ty.max(fy) + height > self.height() {
+            return false;
+        }
+
+        let px_size = <P as Pixel>::CHANNEL_COUNT as usize * std::mem::size_of::<<P as Pixel>::Subpixel>();
+        let width = px_size * width as usize;
+        if from.1 < to.1 {
+            for y in (0..height).rev() {
+                let sy = fy + y;
+                let dy = ty + y;
+                let src = px_size * (sy * self.width + fx) as usize;
+                let dst = px_size * (dy * self.width + tx) as usize;
+                (&mut **self).copy_within(src..src + width, dst);
+            }
+        } else {
+            for y in 0..height {
+                let sy = fy + y;
+                let dy = ty + y;
+                let src = px_size * (sy * self.width + fx) as usize;
+                let dst = px_size * (dy * self.width + tx) as usize;
+                (&mut **self).copy_within(src..src + width, dst);
+            }
+        }
+        true
+    }
+
     fn inner_mut(&mut self) -> &mut Self::InnerImage {
         self
     }
@@ -1081,7 +1110,8 @@ pub(crate) type BgraImage = ImageBuffer<Bgra<u8>, Vec<u8>>;
 #[cfg(test)]
 mod test {
 
-    use super::{ImageBuffer, RgbImage};
+    use super::{ImageBuffer, RgbImage, GrayImage};
+    use super::GenericImage;
     use color;
     #[cfg(feature = "benchmarks")]
     use test;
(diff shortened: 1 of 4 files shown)


HeroicKatora:
I wouldn't have a huge problem *copying* the implementation of `copy_within` from `std` if it is done with a large `FIXME` (and tracking issue, we'd want to have a generic tracking-issue for issues with MSRV anyways). It might not be the most comfortable solution but the implementation is short and not controversial itself. It does not even need to be *as* generic if that improves confidence?

src/buffer.rs
```diff
@@ -921,6 +921,35 @@ where
...
+        let (tx, ty) = to;
+        if tx.max(fx) + width > self.width() || ty.max(fy) + height > self.height() {
+            return false;
+        }
+
+        let px_size = <P as Pixel>::CHANNEL_COUNT as usize * std::mem::size_of::<<P as Pixel>::Subpixel>();
+        let width = px_size * width as usize;
+        if from.1 < to.1 {
+            for y in (0..height).rev() {
+                let sy = fy + y;
```
HeroicKatora: For all these index calculations, please use `pixel_indices_unchecked(x, y).{start,end}`. The current implementation has overflow problems due to the cast to `usize` happening too late. Very similar problems were once fixed by introducing the `pixel_indices` methods :)

src/buffer.rs
```diff
@@ -921,6 +921,35 @@ where
         self.get_pixel_mut(x, y).blend(&p)
     }
 
+    fn copy_within(&mut self, from: (u32, u32), to: (u32, u32), width: u32, height: u32) -> bool {
+        let (fx, fy) = from;
+        let (tx, ty) = to;
+        if tx.max(fx) + width > self.width() || ty.max(fy) + height > self.height() {
```
HeroicKatora: This is also prone to overflow, consider `tx = 2`, `width = u32::MAX`. I don't have a problem with panicking on such input, `std::slice::copy_within` also does this, but it should happen more predictably. For example, first assert that both coordinates lie within the boundaries. Then the calculation `self.width() - tx.max(fx)` is safe and can be compared to `width` as an alternative check.

Veykril: Ah, I completely forgot about overflow becoming a problem, addressed the points you mentioned.

Veykril:
Reimplemented a less generic version of [copy_wthin](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_within) and put a FIXME on it, so CI should hopefully succeed now. I'll create an issue regarding that comment once this PR gets merged(unless you wanna make the MSRV issue).

HeroicKatora:
>  I'll create an issue regarding that comment once this PR gets merged(unless you wanna make the MSRV issue).

I used some additional powers to create a label under which we can group similar issues. Separate tracking issues should work out slightly better than a single listing. I'll create the actual issue when the PR is merged.

> Small side note: I am not sure if I am mistaken but the implementation of pixel_indices{_unchecked doesn't seem to account for the Subpixel size, which normally is always 1 due to it being an u8 in all current image types. But from what I've seen the Subpixel could in theory be any other of the integer primitives which would mean that the function would not work correctly in those cases.

Thank you for noting these thoughts. It's fortunately not as bad, the index calculated here is counted in instances of `Subpixel` and not of the underlying bytes. A different set of pending changes (see preview version) support subpixels (which will also be renamed to `channels`) other than `u8` in the decoders better.

Veykril:
Ah right, my bad, for some reason in my mind I thought the ImageBuffer had a Deref implementation for `&[u8]` and not `&[P::Subpixel]`, then what I said obviously isn't true.

src/buffer.rs
```diff
@@ -921,11 +921,60 @@ where
         self.get_pixel_mut(x, y).blend(&p)
     }
 
+    fn copy_within(&mut self, from: (u32, u32), to: (u32, u32), width: u32, height: u32) -> bool {
```
HeroicKatora: Bikeshedding time. That's a lot of very similar parameters that could easily be switched up by accident. I kind of like the `Rect { }` idea most. The reasons are:

* It is consistent with the two-parameter style of the standard library.
* It groups source selection and destination point which makes it harder to swap them.
* `Rect` can be reused (e.g. [`ImageDecoder::read_rect`](https://docs.rs/image/0.23.0-preview.0/image/trait.ImageDecoderExt.html#method.read_rect), [`sub_image`](https://docs.rs/image/0.23.0-preview.0/image/struct.SubImage.html#method.sub_image))
* A separate struct such as `Rect` offers more extension points for named constructors to further improve clarity but also to add diversity.

Most similar might be the `trait` route, mimicking `slice::copy_within` even more closely. The `Rect` could be simply one type that implements the trait in practice. It might even be that the trait is `Into<Rect>` (although that feels too much like mixing concerns).

Veykril: After a bit of thinking I also prefer the `Rect { }` version the most, while it does require one to type a bit more. It seems to have the most advantages compared to the alternatives. You already gave a lot of good reasons like reuseability and readability. For the other two options, which are both really similar, the only thing I see about them is that they aren't really intuitive making it easy to mix them up. You can only guess what tuple means what and are most likely to get it wrong if you don't look at the parameter names.

HeroicKatora: Do you have the time to implement this version? There does not seem to be any reason not to go with this option then.

Veykril: I can implement it later today, two small questions regarding this then. Where should I put the `Rect` definition and should the target point by a tuple or two seperate args as in:
`image.copy_within(Rect { x: 0, y: 0, w: 20, h: 13) }, (1, 1));`
vs
`image.copy_within(Rect { x: 0, y: 0, w: 20, h: 13) }, 1, 1);`
?

HeroicKatora: Both the pixel related functions as well as operations such as `overlay` take two arguments instead of a tuple to represent coordinates. Maybe it is worth doing an overhaul at some point and unify this to a newtype but it seems more consistent to use two arguments right now.

To me there seem to be two sensible choices to place the implementation:
* in `src/image.rs` and export it at the top level
* in `src/math/rect.rs` and export it in `image::math`

HeroicKatora: A potential counter argument to having the target point as two arguments: It does not mirror `slice::copy_within` as closely. I'd say that it would still be sufficiently similar but we should consider it. Note that the same is true for indexing and the missing `Coord` type is a reason why we do have `ops::Index` implementations... But let's not get ahead of ourselves in this PR.

Veykril: I went with the two arguments approach given most functions use this too as said. Seems better to be uniform for this in my eyes. Put the `Rect` struct into `src/math/rect.rs` since it can, and probably will at some point, be used for more things than just this operation.

HeroicKatora:
Thank you for implementing this. If you interested in contributing more to Rust's image libraries you might be interested in joining `image-rs`. If you want to then read [guidelines](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md) and I can send you an invite.

Veykril:
Yes, I'd like to join the image-rs org. I'll definitely keep looking on things to help out with.

PR #1110, opened 2020-01-09
Move error types to seperate module

HeroicKatora:
This resolves a few rebasing issues that came up while rebasing #928.

At least it is supposed to. I'm expecting a few pushes.


HeroicKatora:
This should not yet do any semantic changes. If you spot any let me know.

PR #1117, opened 2020-01-21
Update `next` from master

HeroicKatora:


HeroicKatora:
Note: benchmarks have apparently not been updated to `edition = 2018`. Putting this to another PR as this is an issue only on `next`.

Issue #896, opened 2019-04-12
Please derive Clone, Debug, PartialEq and other traits on the ImageError


fschutt:
It's very annoying not being able to use a type in a unit test because the maintainer forgot to derive PartialEq, Clone and Debug - please derive all possible traits (Ord, Eq, Hash, Clone, etc.) on the `ImageError` type. There is no reason not to derive them, but it unnecessarily restricts what you can do with the error type, i.e. I can't derive Clone on an `enum MyError { Image(ImageError) }` because the image crate doesn't `#[derive(Clone)]` - which is IMO a very annoying mistake.

I can't fix it myself right now because there seem to be hiccups with this git repo (git clone succeds, but git checkout fails on Windows, due to weird things in the afl-tests, maybe someone rewrote git history in a weird way).

Thanks in advance.

HeroicKatora:
I can't reproduce the checkout problems with a clean clone of `master`. Could it may be automatic line ending conversion?  Anyways, `ImageError` is not `Clone` since it has the variant `IoError(io::Error)`. So I can only advise you to create either a simple newtype wrapper around the `ImageError` or recreate the cloneable variants, both with proper `From<ImageError>` to allow ergonomic conversion.

Regarding the other possible derived traits, it is indeed not completely consistent and we'd be happy with a PR adding missed implementations!

fschutt:
Yeah, well it's the usual problem - because `io::Error` derives absolutely nothing (a mistake in the standard library in my opinion), all wrapping errors can also not `#[derive(...)]` anything.

But please stop using Strings as errors or even worse, wrapping strings as `Io::Error(Box<dyn Error + 'static>)`. Right now there's a `UnsupportedImageFormat(ImageFormat)` error where the error could just wrap the error, but for some reason it gets turned into a `io::Error(ErrorKind::Other, format!("Unsupported format: {}", format))`:

https://github.com/image-rs/image/blob/cb485c0d41f28321900899a253756cabf2d26089/src/dynimage.rs#L850-L854

This is not how you do proper error handling! Just use `ImageError::UnsupportedError(format)`, no dynamic allocation necessary! Formatting the error should only be done in the `Display` function, not earlier and stop using `io::Error` as a wrapper for a `String`, it's not what `io::Error` was meant for.

HeroicKatora:
> Yeah, well it's the usual problem - because io::Error derives absolutely nothing (a mistake in the standard library in my opinion), all wrapping errors can also not #[derive(...)] anything.

[...]

PR #1119, opened 2020-01-26
Error rework

HeroicKatora:
Moves away from a single, open representation of errors to an extensible and precise one. This considers a previously identified few issues (in #928):

* The new variants are somewhat actionable since they at least identify the
  cause more closely.
* They are extensible as most state is opaque to some degree and key
  enumerations are marked as non-exhaustive which allows us to extend
  it later, for example to capture the difference between dimension 
  mismatch errors.
* They are more precise as the opaque state allows storing specialized
  variants for supported formats while still maintaining a uniform
  interface and permitting extension formats outside the library.
* Handling of 'stringification' is more principled as error values can
  be stored as-is instead of converting them on the spot. This might
  also improve the performance when the conversion is not required.

Issues: #921, #938, #896, #823, #741, 
PRs: #1066, #928,

Closes: #928

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -99,7 +99,9 @@ pub enum UnsupportedErrorKind {
     Color(ExtendedColorType),
     /// An image format is not supported.
     Format(ImageFormatHint),
-    Feature,
+    /// Some feature specified by string.
+    /// This is discouraged and is likely to get deprecated (but not removed).
+    GenericFeature(String),
     #[doc(hidden)]
     __NonExhaustive,
 }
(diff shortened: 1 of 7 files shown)


HeroicKatora:
The most recent commit added something invaluable that previous version of error types just could not support: A useful implementation of `source`. This standard mechanisms makes it possible for (mainly) decoders etc. to add additional information that can be queried opportunistically. Note that this does *not* commit to a specific version of `png` in the stable interface.

```rust
if let Err(err) = PngDecoder::new(&not_png[..]) {
    if let Some(source) = error.source() {
        if let Some(png) => source.downcast_ref::<png::DecodingError>() {
            // Do something special, may fail if that was the wrong version of `png` or other reasons.
        }
    }
}
```

birktj:
This looks very nice! The implementation for `source` is also nice and we should add it to the other decoders as well.

src/error.rs
```diff
@@ -1,84 +1,515 @@
...
+            underlying: None,
+        });
+
+    pub(crate) const ImageEnd: Self =
+        ImageError::Parameter(ParameterError {
+            kind: ParameterErrorKind::FailedAlready,
+            underlying: None,
+        });
+
+    pub(crate) fn UnsupportedError(message: String) -> Self {
```
birktj: Very nice trick! This should allow to incrementally improve errors in the rest of the library, however we must ensure that these functions are not misused.

HeroicKatora: Should they be private as here, or part of the public interface and deprecated, and if so, maybe with a changed name? It's hard to say if it not offering them makes it more enticing to properly switch, or to copy the internal implementation anyways but be mad about it not being available..

HeroicKatora:
> The implementation for source is also nice and we should add it to the other decoders as well.

Defer this to follow up PRs. I would rather not stall this one on the issue, the main part is the interface change where png is mostly a PoC.

src/error.rs
```diff
@@ -1,84 +1,515 @@
...
+    /// This means that it might be possible for the operation to succeed in general but
+    /// * it requires a disabled feature,
+    /// * the implementation does not yet exist, or
+    /// * no abstraction for a lower level could be found.
+    Unsupported(UnsupportedError),
 
-    /// An I/O Error occurred while decoding the image
+    /// An error occurred while interacting with the environment.
     IoError(io::Error),
+}
```
HeroicKatora: @Nemo157 remarked on #code-review

> how big is that error type?
might be worth adding some tests like assert!(size_of::<Error>() < 24); to ensure it doesn't become too big and make copying it too slow (though I don't know what the limit at which it's worth boxing the internals is)

This points to it making a real difference in some cases: https://github.com/serde-rs/json/issues/184

src/error.rs
```diff
@@ -1,84 +1,515 @@
...
+    /// This is discouraged and is likely to get deprecated (but not removed).
+    GenericFeature(String),
+    #[doc(hidden)]
+    __NonExhaustive,
+}
+
+#[derive(Debug)]
+pub struct EncodingError {
+    format: ImageFormatHint,
+    message: String,
```
HeroicKatora: Also by @Nemo157

> one minor optimization is to replace String variants with Box<str> since you don't need to mutate them, that saves a third of the data size in those variants

HeroicKatora: This would be used in the internal representations only, not those `*Kind` enums exposed.

src/error.rs
HeroicKatora: (Note: we can use the `[0][usize::from(inverted_condition)]` trick as a static assertion).

HeroicKatora: The clippy link `large_enum_variant` fires at `200`, I'd suggest using the same or a similar limit.

src/error.rs
HeroicKatora: In 7df7d73b2c3c1dba7621b9bd1dc99dd1bbe39898, where I added the public constructors for the error representations, I chose the names to encourage providing an underlying error with the short name `new` and gave a longer name to all other constructors. This basically resolves this question as keeping these functions private.

PR #1126, opened 2020-02-02
Interface restrictions for gif and jpeg

HeroicKatora:
Hide some public interface dependencies on gif and jpeg. Note that frame
oriented functions are now unused and will likely be brought back in a
(slightly less generic) manner that is more uniform over other multi
frame formats as well.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -728,12 +728,7 @@ impl DynamicImage {
             #[cfg(feature = "gif")]
             image::ImageOutputFormat::Gif => {
                 let mut g = gif::Encoder::new(w);
-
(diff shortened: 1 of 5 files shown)


HeroicKatora:
I just realized that there's a bug hidden in the interface. Note how several interfaces such as the io in `free_functions.rs` calls `gif::Frame::from_rgb` directly and without checking the color type and lossy-casting the width and height with `as u16`.

https://github.com/image-rs/image/blob/737a05d685810457d4279c4bbe098dbb70771d6e/src/io/free_functions.rs#L146

So, I'll rewrite the interface to align to `png::PngEncoder::write_image` and make this a bug-fix as well.

PR #1132, opened 2020-02-07
Promotes next to master branch

HeroicKatora:
After this, a release PR for `0.23` will be done and the old version will get a `version-0.22` branch (in case we have to do any safety critical bug fixes).


HeroicKatora:
Travis network failures are far from spurious, it's getting somewhat annoying.. This was the third time or so I had to manually restart a failing task due to it in the last week..

fintelia:
Does anywhere in this merge include a `rustfmt`? If not, we should probably run one now that we don't have two branches around

HeroicKatora:
I didn't do any. Yes, seems like a good idea to do one. Especially seeing if we can configure `rustfmt` in such a ways as to avoid the aggressive features so that we can put it on CI.

PR #1140, opened 2020-02-14
Optimize some functions in vp8.rs

philippeitis:
This uses slices to index into the relevant portions of the array for top_pixels and edge_pixels, which helps decrease runtime by ~100%, as seen below:

```
test webp::vp8::bench::bench_edge_pixels                       ... bench:           8 ns/iter (+/- 0)
test webp::vp8::bench::bench_edge_pixels_new                   ... bench:           0 ns/iter (+/- 0)
test webp::vp8::bench::bench_top_pixels                        ... bench:           5 ns/iter (+/- 0)
test webp::vp8::bench::bench_top_pixels_new                    ... bench:           0 ns/iter (+/- 0)
```

I also tweaked avg2 to avoid u16 conversions. 
I added tests to cover all changes made, (including tests which fully cover the output of avg2 and avg3 to ensure that they remain consistent). The tests for edge_pixels and top_pixels also help demonstrate how the functions work.

### DIFF
--- a/src/webp/vp8.rs
+++ b/src/webp/vp8.rs
@@ -2097,7 +2097,7 @@ mod test {
 
     #[test]
     fn test_avg2_specific() {
-        assert_eq!(255, avg2(255, 255), "avg2(255, 255), expected 255, got {}.", avg2(2, 1));
+        assert_eq!(255, avg2(255, 255), "avg2(255, 255), expected 255, got {}.", avg2(255, 255));
         assert_eq!(1, avg2(1, 1), "avg2(1, 1), expected 1, got {}.", avg2(1, 1));
         assert_eq!(2, avg2(2, 1), "avg2(2, 1), expected 2, got {}.", avg2(2, 1));
     }
--- a/src/webp/vp8.rs
+++ b/src/webp/vp8.rs
@@ -1786,14 +1786,16 @@ fn topleft_pixel(a: &[u8], x0: usize, y0: usize, stride: usize) -> u8 {
 }
 
 fn top_pixels(a: &[u8], x0: usize, y0: usize, stride: usize) -> (u8, u8, u8, u8, u8, u8, u8, u8) {
-    let a0 = a[(y0 - 1) * stride + x0];
-    let a1 = a[(y0 - 1) * stride + x0 + 1];
-    let a2 = a[(y0 - 1) * stride + x0 + 2];
-    let a3 = a[(y0 - 1) * stride + x0 + 3];
-    let a4 = a[(y0 - 1) * stride + x0 + 4];
-    let a5 = a[(y0 - 1) * stride + x0 + 5];
-    let a6 = a[(y0 - 1) * stride + x0 + 6];
-    let a7 = a[(y0 - 1) * stride + x0 + 7];
+    let pos = (y0 - 1) * stride + x0;
+    let a_slice = &a[pos..pos+8];
+    let a0 = a_slice[0];
+    let a1 = a_slice[1];
+    let a2 = a_slice[2];
+    let a3 = a_slice[3];
+    let a4 = a_slice[4];
+    let a5 = a_slice[5];
+    let a6 = a_slice[6];
+    let a7 = a_slice[7];
 
     (a0, a1, a2, a3, a4, a5, a6, a7)
 }
@@ -1813,15 +1815,17 @@ fn edge_pixels(
     y0: usize,
     stride: usize,
 ) -> (u8, u8, u8, u8, u8, u8, u8, u8, u8) {
-    let e8 = a[(y0 - 1) * stride + x0 + 3];
-    let e7 = a[(y0 - 1) * stride + x0 + 2];
-    let e6 = a[(y0 - 1) * stride + x0 + 1];
-    let e5 = a[(y0 - 1) * stride + x0];
-    let e4 = a[(y0 - 1) * stride + x0 - 1];
-    let e3 = a[y0 * stride + x0 - 1];
-    let e2 = a[(y0 + 1) * stride + x0 - 1];
-    let e1 = a[(y0 + 2) * stride + x0 - 1];
-    let e0 = a[(y0 + 3) * stride + x0 - 1];
+    let pos = (y0 - 1) * stride + x0 - 1;
+    let a_slice = &a[pos..=pos+4];
+    let e0 = a[pos + 4 * stride];
+    let e1 = a[pos + 3 * stride];
+    let e2 = a[pos + 2 * stride];
+    let e3 = a[pos + stride];
+    let e4 = a_slice[0];
+    let e5 = a_slice[1];
+    let e6 = a_slice[2];
+    let e7 = a_slice[3];
+    let e8 = a_slice[4];
 
     (e0, e1, e2, e3, e4, e5, e6, e7, e8)
 }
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Runtime of `0` means that llvm has constant inlined everything. You need to wrap both input and output for such small test in `test::black_box` to avoid this.

philippeitis:
I added some black_box calls and got these benchmarks:

```
test webp::vp8::bench::bench_edge_pixels                       ... bench:           8 ns/iter (+/- 0)
test webp::vp8::bench::bench_edge_pixels_new                   ... bench:           1 ns/iter (+/- 0)
test webp::vp8::bench::bench_top_pixels                        ... bench:           5 ns/iter (+/- 0)
test webp::vp8::bench::bench_top_pixels_new                    ... bench:           1 ns/iter (+/- 0)
```

src/webp/vp8.rs
```diff
@@ -1665,8 +1665,7 @@ fn avg3(left: u8, this: u8, right: u8) -> u8 {
 }
 
 fn avg2(this: u8, right: u8) -> u8 {
-    let avg = (u16::from(this) + u16::from(right) + 1) >> 1;
-    avg as u8
+    (this>>1) + (right>>1) + ((this | right) & 1)
```
HeroicKatora: Do you also have some data showing that this is actually an improvement?

HeroicKatora: And please don't simply benchmark this function alone, this is sure to get inline and hopefully optimized heavily at the inline site. Instead, benchmark some function that makes use of `avg2`.

HeroicKatora: Edit: better yet, benchmark `predict_4x4` directly under different modes. It would probably be difficult (and mask the influence of the change) to move the benchmark up to `intra_predict` so that might be the best target for a benchmark.

philippeitis:
Old:
```
test webp::vp8::bench::bench_predict_4x4                       ... bench:         600ns/iter (+/- 87)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         619 ns/iter (+/- 15)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         585 ns/iter (+/- 44)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         615 ns/iter (+/- 25)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         639 ns/iter (+/- 63)
```

New:
```
test webp::vp8::bench::bench_predict_4x4                       ... bench:         589 ns/iter (+/- 26)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         627 ns/iter (+/- 15)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         626 ns/iter (+/- 7)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         604 ns/iter (+/- 9)
test webp::vp8::bench::bench_predict_4x4                       ... bench:         583 ns/iter (+/- 22)
```

The benchmarks seem to favour the new version slightly, but given the high error, I can't make any conclusive statements on which version of avg2 is better, so I suppose it's a matter of personal taste.

HeroicKatora:
The median absolute deviation of those samples (73 and 81 respectively) is several times the difference between their average estimates, I don't think we can draw any definitive conclusions based on those. Note that `avg2` is only used in two prediction methods and likely a small contribution to the total runtime. Also note that the input is not masked with `black_box` so maybe even some inopportune inlining of prediction methods and sample values is involved?

fintelia:
Based on the [generated assembly](https://rust.godbolt.org/z/kLUG38) for the two versions, I'm rather skeptical that the new version is faster. If any differences are detectable beyond just being noise, then it looks like the new one will actually be a couple cycles _slower_.

HeroicKatora:
The assembly of a single call isn't reallly important here. In a single function, of course the compiler will take advantage of the full register width and there is no advantage to keeping the values small. However, for the purpose of SIMDification it might require double the instruction count due to halving the number of vector elements. However, the bounds checks might present a problem for that optimization route (addressing this is the part of the change here).

philippeitis:
Given the assembly and unreliability of the tests, I've decided to just revert the function.

philippeitis:
Benchmarks for predict_bvepred consistently show a 6% speedup, with no margin of error. (Assembly is 30% smaller as well).
```
test webp::vp8::bench::bench_predict_bvepred                   ... bench:          16 ns/iter (+/- 0)
test webp::vp8::bench::bench_predict_bvepred_new               ... bench:          15 ns/iter (+/- 0)
```

HeroicKatora:
> Benchmarks for predict_bvepred consistently show a 6% speedup, with no margin of error. (Assembly is 30% smaller as well).

Microbenchmarking like that can be very misleading. On that scale, you would need to have cycle counts, register pressure, and other CPU stats to mae any definitive statements as those are just a few hunder cycles. Please note that the use of `black_box` is also not quite correct. The compiler will search for constants or duplicate calculations. You need to ensure that *each* iteration gets 'fresh' inputs by wrapping them in `black_box` within the loop. The other ensures that the input is not completely constant folded but does not prevent the compiler from finding that each loop calculates the same result. It should read something like:

```rust
    fn bench_edge_pixels(b: &mut Bencher) {
        let v = make_sample_image();

        b.iter(|| {
            black_box(edge_pixels(black_box(&v), 5, 5, W * 2));
        });
    }
```

src/webp/vp8.rs
```diff
@@ -1813,42 +1815,53 @@ fn edge_pixels(
...
-    a[(y0 + 3) * stride + x0 + 2] = avg3(a1, a2, a3);
-
-    a[y0 * stride + x0 + 3] = avg3(a2, a3, a4);
-    a[(y0 + 1) * stride + x0 + 3] = avg3(a2, a3, a4);
-    a[(y0 + 2) * stride + x0 + 3] = avg3(a2, a3, a4);
-    a[(y0 + 3) * stride + x0 + 3] = avg3(a2, a3, a4);
+    let avg_1 = avg3(p, a0, a1);
+    let avg_2 = avg3(a0, a1, a2);
+    let avg_3 = avg3(a1, a2, a3);
+    let avg_4 = avg3(a2, a3, a4);
```
HeroicKatora: Could we improve readability in the process?

```rust
let avg = [avg_1, avg_2, avg_3, avg_4];

// [...]
let mut pos = y0 * stride + x0;
for _ in 0..4 {
    a[pos..=pos + 3].copy_from_slice(&avg);
    pos += stride;
}

```

philippeitis:
Updated edge_pixel benchmarks with more black boxing:
```
test webp::vp8::bench::bench_edge_pixels                       ... bench:           7 ns/iter (+/- 0)
test webp::vp8::bench::bench_edge_pixels_new                   ... bench:           2 ns/iter (+/- 0)
```

HeroicKatora:
>  Updated edge_pixel benchmarks with more black boxing:

Indeed, the function alone does less bounds checks. This is not news. Your micro benchmarks are getting spammy though, these are of little value to the performance of the whole decoding pipeline. The point of the optimizer is to be able to prove some of these bounds checks unecessary and remove them by inlining those small utilities so those benchmarks add little knowledge for the performance of the prediction stage.

philippeitis:
That's fair - I won't include any more micro-benchmarks, but the performance of predict_4x4 has improved by 15% - (the baseline time is 492ns +\- 5ns to 422ns +\- 4ns) with all of the changes I have included here - however, using only one optimized function at a time degrades performance (except for add_residue, which provides a 20ns improvement when used).

HeroicKatora:
That's a good find! Providing more strict type information definitely seems to help here. Can you add a comment to the function (and/or the local copies) to explain this? Thanks for transferring the structural improvements to the other places as well.

And if you feel comfortable with it, I wouldn't mind a somewhat cleaned up commit history, force pushing allowed. Specifically, anything that makes it easier to cherry-pick the test suite and test the complete speedup of all changes together.

philippeitis:
I've added comments and I've cleaned up the commit history. I've kept the optimizations in their separate commits, and the last commit contains all tests, optimizations and benchmarks you need to test the complete speed up.

src/webp/vp8.rs
```diff
@@ -1701,7 +1705,11 @@ fn predict_4x4(ws: &mut [u8], stride: usize, modes: &[IntraMode], resdata: &[i32
                 IntraMode::HU => predict_bhupred(ws, x0, y0, stride),
             }
 
-            add_residue(ws, rb, y0, x0, stride);
+            // Create a [i32; 16] array for add_residue by copying the
+            // slice from resdata into rb (slices do not work).
+            let mut rb = [0i32; 16];
+            rb.copy_from_slice(&resdata[i * 16..i * 16 + 16]);
```
HeroicKatora: The change here left aliased the `rb` a few lines above which is now unused. Could you remove it?

HeroicKatora:
Some numbers from my end as well:
```rust
// pre
test webp::vp8::test::bench_edge_pixels                        ... bench:           6 ns/iter (+/- 0)
test webp::vp8::test::bench_predict_4x4                        ... bench:         554 ns/iter (+/- 22)
test webp::vp8::test::bench_predict_bhepred                    ... bench:          12 ns/iter (+/- 0)
test webp::vp8::test::bench_predict_bldpred                    ... bench:          14 ns/iter (+/- 0)
test webp::vp8::test::bench_predict_brdpred                    ... bench:          16 ns/iter (+/- 2)
test webp::vp8::test::bench_predict_bvepred                    ... bench:          13 ns/iter (+/- 0)
test webp::vp8::test::bench_top_pixels                         ... bench:           4 ns/iter (+/- 0)

// post
test webp::vp8::test::bench_edge_pixels                        ... bench:           2 ns/iter (+/- 0)
test webp::vp8::test::bench_predict_4x4                        ... bench:         429 ns/iter (+/- 12)
test webp::vp8::test::bench_predict_bhepred                    ... bench:          10 ns/iter (+/- 0)
test webp::vp8::test::bench_predict_bldpred                    ... bench:          12 ns/iter (+/- 0)
test webp::vp8::test::bench_predict_brdpred                    ... bench:          16 ns/iter (+/- 1)
test webp::vp8::test::bench_predict_bvepred                    ... bench:           9 ns/iter (+/- 1)
test webp::vp8::test::bench_top_pixels                         ... bench:           1 ns/iter (+/- 0)
```

Issue #1134, opened 2020-02-07
Tracking issue: Converting error representations


HeroicKatora:
The `0.23` brought a more detailed and strictly typed error representation. We kept an internal compatibility interface [here](https://github.com/image-rs/image/blob/d0fde846621a1bdf212a4f718ab9c296ad177e0e/src/error.rs#L196-L228) to reduce the rework effort required. The plan is to incrementally port these to use more of the new, shiny features. For a basic example, [see the png module](https://github.com/image-rs/image/blob/d0fde846621a1bdf212a4f718ab9c296ad177e0e/src/png.rs#L264-L288). To help, claim one of the image formats by a dedicated tracking PR or issue and rework its decoder and encoder implementation or claim a legacy constructors and deprecate and remove it.

Formats:
- [x] bmp: #1141
- [x] dds: #1147
- [x] dxt: #1148
- [x] gif: #1196
- [x] hdr: #1196
- [x] ico: #1193 
- [x] jpeg: #1144
- [x] pnm: #1149, #1203
- [x] png: #1204
- [x] tga: #1145
- [x] tiff: #1150
- [x] webp: #1146 
- [x] flat: #1204
- [x] farbfeld: #1204

Constructors:
- [x] `InsufficientMemory` #1190
- [x] `DimensionError` #1189
- [x] `ImageEnd`: #1142 
- [ ] `UnsupportedError(String)`
- [x] `UnsupportedColor(ExtendedColorType)`: #1204
- [x] `FormatError(String)`: #1204

nabijaczleweli:
Does this solution make sense? (diff based on 3a94fc97b148992c55ae20eafcf52d7c376df8ac)

```diff
diff --git a/src/dynimage.rs b/src/dynimage.rs
index 4415fe6e..4163e6e6 100644
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -758,7 +758,9 @@ impl DynamicImage {
             }
 
             image::ImageOutputFormat::Unsupported(msg) => {
-                Err(ImageError::UnsupportedError(msg))
+                Err(ImageError::UnsupportedError(UnsupportedError::from_format_and_kind(
+                    ImageFormatHint::Unknown,
+                    UnsupportedErrorKind::Format(ImageFormatHint::Name(msg)))))
             },
 
             image::ImageOutputFormat::__NonExhaustive(marker) => match marker._private {},
diff --git a/src/image.rs b/src/image.rs
[...]

PR #1141, opened 2020-02-15
Use detailed errors for BMP encoder and decoder

lo48576:
This is part of #1134, rework for `bmp` module.

Most of the errors were `FormatError` and can be converted simply to `DecodingError`.
However, I'm not sure how `ImageError::UnsupportedError` should be translated:

* Which should be used for "Unsupported Bitmap Header", `DecoderError` or `UnsupportedError`?
    + This is emitted when BMP header type (more precisely, its size) is unknown.
* Which should be used for "Unsupported image type", `DecoderError` or `UnsupportedError`?
    + This is emitted when the BMP image uses PNG or JPEG compression, or other unknown image types.

Additionally, if `UnsupportedError` should be used for them, what error kind (variants of `UnsupportedErrorKind`) should be used for them?

* `Color` is not appropriate for them.
* `Format` would be better, but I'm not sure the errors are really "unsupported image formats" (because the decoder is sure it is reading BMP!), and it loses precision of error information.
* `GenericFeature` can be used to translate the error naively, but it is "discouraged and is likely to get deprecated".

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -9,8 +9,10 @@ use std::cmp::Ordering;
 use byteorder::{LittleEndian, ReadBytesExt};
 
 use crate::color::ColorType;
-use crate::error::{ImageError, ImageResult};
-use crate::image::{self, ImageDecoder, ImageDecoderExt, Progress};
+use crate::error::{
+    DecodingError, ImageError, ImageResult, UnsupportedError, UnsupportedErrorKind,
+};
+use crate::image::{self, ImageDecoder, ImageDecoderExt, ImageFormat, Progress};
 
 const BITMAPCOREHEADER_SIZE: u32 = 12;
 const BITMAPINFOHEADER_SIZE: u32 = 40;
@@ -113,9 +115,10 @@ fn check_for_overflow(width: i32, length: i32, channels: usize) -> ImageResult<(
     num_bytes(width, length, channels)
         .map(|_| ())
         .ok_or_else(|| {
-            ImageError::FormatError(
+            ImageError::Decoding(DecodingError::with_message(
+                ImageFormat::Png.into(),
                 "Image would require a buffer that is too large to be represented!".to_owned(),
-            )
+            ))
         })
 }
 
@@ -367,12 +370,16 @@ impl Bitfield {
         let mut shift = mask.trailing_zeros();
         let mut len = (!(mask >> shift)).trailing_zeros();
         if len != mask.count_ones() {
-            return Err(ImageError::FormatError(
-                "Non-contiguous bitfield mask".to_string(),
-            ));
+            return Err(ImageError::Decoding(DecodingError::with_message(
+                ImageFormat::Bmp.into(),
+                "Non-contiguous bitfield mask".to_owned(),
+            )));
         }
         if len + shift > max_len {
-            return Err(ImageError::FormatError("Invalid bitfield mask".to_string()));
+            return Err(ImageError::Decoding(DecodingError::with_message(
+                ImageFormat::Bmp.into(),
+                "Invalid bitfield mask".to_owned(),
+            )));
         }
         if len > 8 {
             shift += len - 8;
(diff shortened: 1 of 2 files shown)


src/bmp/decoder.rs
```diff
@@ -670,42 +700,46 @@ impl<R: Read + Seek> BmpDecoder<R> {
...
-                    ))
+                    )))
                 }
             },
             // PNG and JPEG not implemented yet.
             _ => {
-                return Err(ImageError::UnsupportedError(
+                return Err(ImageError::Decoding(DecodingError::with_message(
+                    ImageFormat::Bmp.into(),
                     "Unsupported image type".to_string(),
```
lo48576: Which should be used, `DecodingError` or `UnsupportedError`?

Possible candidate:

```rust
                return Err(ImageError::Unsupported(
                    UnsupportedError::from_format_and_kind(
                        ImageFormat::Bmp.into(),
                        UnsupportedErrorKind::GenericFeature(format!(
                            "image type {}",
                            image_type_u32
                        )),
                    ),
                ));
```

src/bmp/decoder.rs
```diff
@@ -770,9 +804,10 @@ impl<R: Read + Seek> BmpDecoder<R> {
                 BITMAPV4HEADER_SIZE => BMPHeaderType::V4,
                 BITMAPV5HEADER_SIZE => BMPHeaderType::V5,
                 _ => {
-                    return Err(ImageError::UnsupportedError(
+                    return Err(ImageError::Decoding(DecodingError::with_message(
+                        ImageFormat::Bmp.into(),
                         "Unsupported Bitmap Header".to_string(),
```
lo48576: Which should be used, `DecodingError` or `UnsupportedError`?

Possible candidate:

```rust
                    return Err(ImageError::Unsupported(
                        UnsupportedError::from_format_and_kind(
                            ImageFormat::Bmp.into(),
                            UnsupportedErrorKind::GenericFeature(
                                "Unsupported Bitmap Header".to_owned(),
                            ),
                        ),
                    ))
```

src/bmp/decoder.rs
HeroicKatora: This should be split into two: `UnsupportedError` for values `BI_JPEG = 4` and `BI_PNG = 5` and then a `DecodingError` for all other values.

src/bmp/decoder.rs
HeroicKatora: Similar deal. There are technically more header sizes allowed corresponding to other header types which we do not precisely know (for the future and because there are unofficial extensions...). I think in this case it should always fail with `Unsupported` if the header size is larger than `BITMAPCOREHEADER_SIZE` and fail with `DecodingError` if it is smaller.

lo48576:
Updated (and fixup'ed) the commit to use both `DecodingError` and `UnsupportedError`.

I noticed that [there are other compression types (0xb, 0xc, 0xd) for CMYK](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/4e588f70-bd92-4a6f-b77f-35d0feaf7a57), so I added match arms for them to emit more precise error message.

lo48576:
Sorry, force-pushed [little update](https://github.com/image-rs/image/compare/c7f751f426b09edc5c61f7a88f9d3ee71c594596..b84de256b4dbe1047045b772d347d877f62bed09): added image type to the error message for unknown image type.

src/bmp/decoder.rs
```diff
@@ -670,42 +702,73 @@ impl<R: Read + Seek> BmpDecoder<R> {
...
+                ));
+            }
             _ => {
-                return Err(ImageError::UnsupportedError(
-                    "Unsupported image type".to_string(),
-                ))
+                // Unknown compression enumeration.
+                return Err(ImageError::Decoding(DecodingError::with_message(
+                    ImageFormat::Bmp.into(),
+                    format!("Unknown image type {}", image_type_u32),
```
fintelia: Would it be accurate to say "unknown image compression type {}"?

lo48576: Should "Unknown" be all lowercase?

HeroicKatora: It's more consistent to have an upper case at the start, also with other errors. So no :smiley:

HeroicKatora:
Thank you! One format down.

PR #1149, opened 2020-02-16
Use detailed errors for PNM encoder and decoder

lo48576:
This is part of #1134, rework for `pnm` module.

### DIFF
--- a/src/pnm/decoder.rs
+++ b/src/pnm/decoder.rs
@@ -166,8 +166,7 @@ trait HeaderReader: BufRead {
     /// Reads the two magic constant bytes
     fn read_magic_constant(&mut self) -> ImageResult<[u8; 2]> {
         let mut magic: [u8; 2] = [0, 0];
-        self.read_exact(&mut magic)
-            .map_err(ImageError::IoError)?;
+        self.read_exact(&mut magic)?;
         Ok(magic)
     }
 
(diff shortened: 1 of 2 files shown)


lo48576:
Some format errors have underlying errors (mainly parse error of simple value or custom type).

```rust
// Old code.
    string
        .parse()
        .map_err(|err| ImageError::FormatError(format!("Error parsing {} as a sample: {}", string, err)))
```

`DecodingError` itself can have error message and such underlying errors at the same time.
However, `DecodingError` does not have constructor functions that receives both message and underlying errors.

I think it is the best if we can use `DecodingError::with_error_and_message(String, impl Into<Box<dyn Error + Send + Sync>>)` or something like that.
For now I changed the code as below, but I'm not sure it is OK, so this PR is draft.

```rust
// New code?
    string.parse().map_err(|err| {
        ImageError::Decoding(DecodingError::with_message(
            ImageFormat::Pnm.into(),
            format!("Error parsing {} as a sample: {}", string, err),
        ))
    })
```

HeroicKatora:
You could add a (currently private) internal `enum` representing the decoding errors which implements `Error`. Then its formatting could provide the same functionality as passing a message but with only a single error instance.

lo48576:
Added `PrimitiveParseError` to contain contextual message, invalid input string, and underlying parse error.

If necessary, I'll `git fixup` the commit (unify <https://github.com/image-rs/image/pull/1149/commits/7c1bc2b4cbd12ac65cb8770eaf23690403e8b7bd> and <https://github.com/image-rs/image/pull/1149/commits/f98b540a3ac0fcf54b42f717bd26b1c3551c9a24>).

PR #1151, opened 2020-02-18
Double performance of rgb / rgba encode methods in BMPEncode

philippeitis:
Original:
```
test bmp::encoder::tests::bench_encode_rgb    ... bench:   5,782,530 ns/iter (+/- 218,122)
test bmp::encoder::tests::bench_encode_rgba   ... bench:   8,998,140 ns/iter (+/- 16,662,076)
```

New:
```
test bmp::encoder::tests::bench_encode_rgb    ... bench:   3,316,475 ns/iter (+/- 209,397)
test bmp::encoder::tests::bench_encode_rgba   ... bench:   4,440,780 ns/iter (+/- 436,233)
```

For some reason, the rgba benchmark seems to have a large +/- of ~17M ns, but the actual timings are always ~9M ns.

I've included benchmarks, and I've tested against the BMP tests, which seem to work fine.

### DIFF
--- a/src/bmp/encoder.rs
+++ b/src/bmp/encoder.rs
@@ -151,24 +151,21 @@ impl<'a, W: Write + 'a> BMPEncoder<'a, W> {
         row_pad_size: u32,
         bytes_per_pixel: u32,
     ) -> io::Result<()> {
-        let x_stride = bytes_per_pixel;
+        let width = width as usize;
+        let height = height as usize;
+        let x_stride = bytes_per_pixel as usize;
         let y_stride = width * x_stride;
-        for row in 0..height {
+        for row in (0..height).rev() {
             // from the bottom up
-            let row_start = (height - row - 1) * y_stride;
-            for col in 0..width {
-                let pixel_start = (row_start + (col * x_stride)) as usize;
-                let r = image[pixel_start];
-                let g = image[pixel_start + 1];
-                let b = image[pixel_start + 2];
-                let a = image[pixel_start + 3];
+            let row_start = row * y_stride;
+            for px in image[row_start..row_start + y_stride].chunks_exact(x_stride) {
+                let r = px[0];
+                let g = px[1];
+                let b = px[2];
+                let a = px[3];
                 // written as BGRA
-                self.writer.write_u8(b)?;
-                self.writer.write_u8(g)?;
-                self.writer.write_u8(r)?;
-                self.writer.write_u8(a)?;
+                self.writer.write(&[b, g, r, a])?;
             }
-
             self.write_row_pad(row_pad_size)?;
         }
 
@@ -266,6 +263,9 @@ mod tests {
     use crate::image::ImageDecoder;
     use std::io::Cursor;
 
+    #[cfg(feature = "benchmarks")]
+    use test::{Bencher};
+
     fn round_trip_image(image: &[u8], width: u32, height: u32, c: ColorType) -> Vec<u8> {
         let mut encoded_data = Vec::new();
         {
@@ -280,6 +280,17 @@ mod tests {
         let mut buf = vec![0; decoder.total_bytes() as usize];
         decoder.read_image(&mut buf).expect("failed to decode");
         buf
+    }
+
+    #[cfg(feature = "benchmarks")]
+    #[bench]
+    fn bench_encode_rgba(b: &mut Bencher) {
+        let mut v= Vec::with_capacity(2000 * 2000);
+        let mut x = BMPEncoder::new(&mut v);
+        let mut im = vec![0; 2000 * 2000];
+        b.iter(|| {
+            x.encode_rgba(&mut im, 500, 2000, 0, 4);
+        });
     }
 
     #[test]
--- a/src/bmp/encoder.rs
+++ b/src/bmp/encoder.rs
@@ -258,6 +258,8 @@ mod tests {
 
     #[cfg(feature = "benchmarks")]
     use test::{Bencher};
+//     #[cfg(feature = "benchmarks")]
+//     use std::fs::File;
 
     fn round_trip_image(image: &[u8], width: u32, height: u32, c: ColorType) -> Vec<u8> {
         let mut encoded_data = Vec::new();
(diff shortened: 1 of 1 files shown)


src/bmp/encoder.rs
```diff
@@ -120,23 +120,20 @@ impl<'a, W: Write + 'a> BMPEncoder<'a, W> {
...
+        for row in (0..height).rev() {
             // from the bottom up
-            let row_start = (height - row - 1) * y_stride;
-            for col in 0..width {
-                let pixel_start = (row_start + (col * x_stride)) as usize;
-                let r = image[pixel_start];
-                let g = image[pixel_start + 1];
-                let b = image[pixel_start + 2];
+            let row_start = row * y_stride;
+            for px in image[row_start.. row_start + y_stride].chunks_exact(x_stride) {
```
fintelia: I'm a fan of using `image[row_start..][..y_stride]` for cases like these.

philippeitis:
I've made the requested changes.
I also simplified encode_gray (writing with a buffer of values and using the reverse iterator).

An additional optimization which could be particularly effective is writing to an intermediate buffer for each row to avoid calling write_all() / write_u8 in a tight loop, but this depends on the size of the buffer, and could be abused to allocate excessively large buffers.

HeroicKatora:
@philippeitis An intermediate buffer is already possible by using a `BufWriter`, there's shouldn't be a need to worry about it in a structure that's generic over the underlying writer. Maybe that means that we should benchmark over different writers, to get a more representative picture.

philippeitis:
Given that the writer used in the benchmarks is a raw vector (to avoid filesystem effects on writing), I think the speedup comes from reducing the number of calls in tight loops - each call to write_u8()? or write_all()? introduces additional overhead (simply calling the function, boundary checks, checking that a write is successful) which can be condensed into a single write at the end of the loop. A `bufWriter` would help in the case that we're writing to a filesystem, which is not what I am testing in the benchmarks.

philippeitis:
```
```
```   fn encode_gray(
        &mut self,
        image: &[u8],
        width: u32,
        height: u32,
        row_pad_size: u32,
        bytes_per_pixel: u32,
    ) -> io::Result<()> {
        // write grayscale palette
        for val in 0u8..=255 {
            // each color is written as BGRA, where A is always 0 and since only grayscale is being written, B = G = R = index
            self.writer.write_all(&[val, val, val, 0])?;
        }

        // write image data
        let width = width as usize;
        let height = height as usize;
        let x_stride = bytes_per_pixel as usize;
        let y_stride = width * x_stride;
        let mut buf = Vec::with_capacity(width);
        for row in (0..height).rev() {
            // from the bottom up
            let row_start = row * y_stride;
            for col in 0..width {
                let pixel_start = (row_start + (col * x_stride)) as usize;
                // color value is equal to the palette index
                buf.push(image[pixel_start]);
            }
            self.writer.write_all(&buf)?;
            buf.clear();

            self.write_row_pad(row_pad_size)?;
        }

        Ok(())
    }
```

This runs in
```
test bmp::encoder::tests::bench_encode_gray   ... bench:   1,970,567 ns/iter (+/- 1,026,174)
```
vs. the previous version's
```
test bmp::encoder::tests::bench_encode_gray   ... bench:   4,316,750 ns/iter (+/- 567,438)
```

(The large benchmark times are an artifact of my laptop's clock performance - for comparision, when underclocked, the original encode_rgba function runs in ~18M ns and the new one runs in ~7.5M ns, while the overclocked times are as in the original pull comment.)

philippeitis:
Alternative, if we want to ensure that this intermediate buffer never exceeds a certain capacity, we could use this implementation:
```
    fn encode_gray(
        &mut self,
        image: &[u8],
        width: u32,
        height: u32,
        row_pad_size: u32,
        bytes_per_pixel: u32,
    ) -> io::Result<()> {
        // write grayscale palette
        for val in 0u8..=255 {
            // each color is written as BGRA, where A is always 0 and since only grayscale is being written, B = G = R = index
            self.writer.write_all(&[val, val, val, 0])?;
        }
        // write image data
        let max_width = 4096usize;
        let width = width as usize;
        let height = height as usize;
        let x_stride = bytes_per_pixel as usize;
        let y_stride = width * x_stride;
        let buf_width = if width < max_width {width} else {max_width};
        let mut buf = vec![0u8; buf_width];
        // from the bottom up
        for row in (0..height).rev() {
            let mut col = 0;
            let mut pos = row * y_stride;
            let mut num_elems = 0;
            while col < width {
                while num_elems < buf_width && col < width {
                    buf[num_elems] = image[pos];
                    num_elems += 1;
                    pos += x_stride;
                    col += 1;
                }
                self.writer.write_all(&buf[..num_elems])?;
                num_elems = 0;
            }
            self.write_row_pad(row_pad_size)?;
        }

        Ok(())
    }
```

Old:
```
test bmp::encoder::tests::bench_encode_gray   ... bench:   5,069,330 ns/iter (+/- 453,906)
```
vs. new
```
test bmp::encoder::tests::bench_encode_gray   ... bench:   1,425,107 ns/iter (+/- 210,943)
```

HeroicKatora:
That's *very* curious because it is precisely what `BufWriter` should provide. Have you *tried* using a `BufWriter<Vec<u8>>` as the output writer of the benchmark, either by replacing the type in the benchmark or by wrapping the provided writer within `encode_gray`? It seems plausible that IO performance would also in general improve from grouping writes into larger ones but very weird for in-memory IO. We really should be careful to not optimize with a bad metric at the cost of complicating the logic and inhibiting future optimization.

philippeitis:
Benchmark times (best of 5 in both cases)
Old:
```
test bmp::encoder::tests::bench_encode_gray          ... bench:   2,700,717 ns/iter (+/- 209,294)
test bmp::encoder::tests::bench_encode_gray_buf      ... bench:   7,854,635 ns/iter (+/- 1,675,822)
test bmp::encoder::tests::bench_encode_gray_buf_file ... bench:  15,644,970 ns/iter (+/- 2,933,322)
```
New:
```
test bmp::encoder::tests::bench_encode_gray          ... bench:   2,058,517 ns/iter (+/- 238,720)
test bmp::encoder::tests::bench_encode_gray_buf      ... bench:   1,440,242 ns/iter (+/- 195,000)
test bmp::encoder::tests::bench_encode_gray_buf_file ... bench:   8,846,160 ns/iter (+/- 1,525,621)
```

The buffered writer does seem to help, in the case that we also buffer our own writes to it, which was quite surprising. It seems that using this intermediate buffer is effective for most writers..

philippeitis:
I've included the benchmarks (without the implementation - which is as it appears in the comments), but have left the file benchmarks commented out as I'm not sure how you'd want to handle the temporary file creation and deletion.

philippeitis:
And for completeness' sake, I tried this for encode_rgb and encode_rgba:
```
test bmp::encoder::tests::bench_encode_rgb               ... bench:   3,270,550 ns/iter (+/- 396,434)
test bmp::encoder::tests::bench_encode_rgb_int_buf       ... bench:   3,310,370 ns/iter (+/- 261,876)
test bmp::encoder::tests::bench_encode_rgb_ext_buf       ... bench:   8,389,970 ns/iter (+/- 329,949)
test bmp::encoder::tests::bench_encode_rgb_dub_buf       ... bench:   3,405,115 ns/iter (+/- 563,924)
test bmp::encoder::tests::bench_encode_rgb_buf_file      ... bench:  11,335,310 ns/iter (+/- 2,549,880)
test bmp::encoder::tests::bench_encode_rgb_dub_buf_file  ... bench:   6,084,135 ns/iter (+/- 2,581,271)
test bmp::encoder::tests::bench_encode_rgba              ... bench:   4,059,840 ns/iter (+/- 257,285)
test bmp::encoder::tests::bench_encode_rgba_int_buf      ... bench:   5,357,430 ns/iter (+/- 11,073,799)
test bmp::encoder::tests::bench_encode_rgba_ext_buf      ... bench:   9,692,920 ns/iter (+/- 16,647,828)
test bmp::encoder::tests::bench_encode_rgba_dub_buf      ... bench:   3,667,185 ns/iter (+/- 14,063,418)
test bmp::encoder::tests::bench_encode_rgba_buf_file     ... bench:  12,659,300 ns/iter (+/- 4,234,445)
test bmp::encoder::tests::bench_encode_rgba_dub_buf_file ... bench:   8,553,975 ns/iter (+/- 6,130,925)
```
encode_rgb(a): New implementation with no internal buffer
encode_rgb(a)_int_buf: Implementation with an internal buffer
encode_rgb(a)_ext_buf: New implementation with no internal buffer, but we wrap our vector with a buffer before calling the function
encode_rgb(a)_dub_buf: Implementation with an internal buffer and the vector is wrapped with a buffer.
encode_rgb(a)_buf_file: New implementation with no internal buffer writing to buffered file.
encode_rgb(a)_dub__buf_file: Implementation with an internal buffer writing to buffered file

As we can see, it seems that an internal buffer helps to mitigate a massive slowdown that occurs when we wrap our writers with a `BufWriter`, which I expect is the typical use case for this crate, and doesn't particularly impact performance when writing to raw vectors.

HeroicKatora:
Thanks for all the style improvements in the code as well :+1:

PR #1161, opened 2020-03-10
Require jpeg-decoder >= 0.1.17 for Send + Sync error

qnighy:
## Description

#1144 (image 0.23.1) requires https://github.com/image-rs/jpeg-decoder/pull/119 (jpeg-decoder 0.1.17) to compile, so clarify that in Cargo.toml.

## License


HeroicKatora:
Thanks, that seems right. I'm just surprised that https://github.com/image-rs/jpeg-decoder/pull/119 was not strictly semver compatible and yet there has no fallout. (Possibly because no one use the variant to construct an Error, only matches on them).

PR #1166, opened 2020-03-22
Add TGA to list of Supported Image Formats in README.md

pix64:
TGA is missing from the list of supported image formats


HeroicKatora:
Hm, the one build failure is very likely spurious, so let's ignore it. But nightly seems to fail on the default-features builds, I've definitely seen the exact same thing happening in previous commits (ignored it as spurious nightly failure). It seems it is not so spurious, might be worth investigating.

Issue #1167, opened 2020-03-23
Detect if dynamic image has alpha channel


fishrockz:
How does one ... ?
Tell if a dynamic image has a alpha channel?

Overall, I'm trying to achieve ...
I want to be able to tell if i am better off calling `to_rgb()` to save space if there is no alpha or `to_rgba()` to take advantage if there is alpha.

Additional context:
https://github.com/xi-editor/druid/pull/747/files/c45a2437dfb17db156f58ecb9c599c90182bd5e2#diff-6bd43626335f6f0943c43faf8b0ecdd7

fintelia:
Not a great option, but you can just match on the variants of DynamicImage:
```rust
match img {
    DynamicImage::ImageLuma8(..) => false,
    DynamicImage::ImageLumaA8(..) => true,
    DynamicImage::ImageRgb8(..) => false,
    DynamicImage::ImageRgba8(..) => true,
    ...
}

**Comment by HeroicKatora** [comment]
I'd rather match on `DynamicImage::color_type` as previously. You correctly observe, though, that this has become more brittle as the enum is now non-exhaustive and new variants might be either of the two possibilities. It should be fine to do so for now, there are no immediate plans to introduce new types that violate the assumption that channel count 2 or 4 is not an alpha color type.

So the immediate plan would be to introduce  `has_alpha(self: ColorType) -> bool` before any such color type where this is violated, that should be perfectly doable?

**Comment by fishrockz** [comment]
So we went with
```
fn has_alpha_channel(image: &image::DynamicImage) -> bool {
    use image::ColorType::*;
    match image.color() {
        La8 | Rgba8 | La16 | Rgba16 | Bgra8 => true,
        _ => false,
    }
}
```
But I don't like how fragile this is now.
I think `has_alpha(self: ColorType) -> bool` would be great
Should this be simple to add?
I can take a look if you'll be happy to have it? is the a trait somewhere that can enforce this?

HeroicKatora:
@fishrockz I'd be happy to have it. With `0.23` we've adopted the stance that `ColorType` is more or less the list of colors supported by `ImageBuffer` (or `DynamicImage` respectively). So adding an inherent method to that type may be the most straightforward way to implement this feature and should ensure it's implemented for future `DynamicImage` additions.

fishrockz:
I finally got round to having a go at this, sorry it has been so long, any you have already done it! Thank you @HeroicKatora !

PR #1172, opened 2020-03-26
Implement ColorType::has_alpha

HeroicKatora:
Since the enumeration is non-exhaustive, implementing this method might
be more brittle than necessary. This commits to having such a judgement
available for all future color types but that should not be problematic.

Closes: #1167

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -48,6 +48,16 @@ impl ColorType {
         }
     }
 
+    /// Returns if there is an alpha channel.
+    pub fn has_alpha(self) -> bool {
+        use ColorType::*;
+        match self {
+            L8 | L16 | Rgb8 | Bgr8 | Rgb16 => false,
+            La8 | Rgba8 | Bgra8 | La16 | Rgba16 => true,
+            __NonExhaustive(marker) => match marker._private {},
+        }
+    }
+
     /// Returns the number of bits contained in a pixel of `ColorType` ```c``` (which will always be
     /// a multiple of 8).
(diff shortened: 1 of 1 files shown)


src/color.rs
```diff
@@ -48,6 +48,16 @@ impl ColorType {
         }
     }
 
+    /// Returns if there is an alpha channel.
+    pub fn has_alpha(self) -> bool {
+        use ColorType::*;
+        match self {
+            L8 | L16 | Rgb8 | Bgr8 | Rgb16 => false,
+            La8 | Rgba8 | Bgra8 | La16 | Rgba16 => true,
+            __NonExhaustive(marker) => match marker._private {},
```
fishrockz, line 57: What dose `__NonExhaustive(marker) => match marker._private {},` do?

HeroicKatora: The `__NonExhaustive` variant emulates the `#[non_exhaustive]` attribute, which we can't use due to the minimum supported compiler version policy. It's defined [here in utils](https://github.com/image-rs/image/blob/95202a4af089ed22c17348992d2678dab77451c0/src/utils/mod.rs#L78-L95) which has an explanation for how this works and what alternatives were considered.

HeroicKatora: The explicit match here is used to prove to the compiler that the variant can in fact never be constructed, without having to rely on `unreachable` (which may compile to a panic) or the unsafe `unreachable_unchecked`.

fishrockz: Nice, Thanks for the explanation.

Issue #1069, opened 2019-10-27
Why are the `Pixels`, `EnumeratePixels` (etc) iterators not public types?


dlight:
How does one store on a struct the iterator that [`ImageBuffer::enumerate_pixels`](https://docs.rs/image/0.22.3/image/struct.ImageBuffer.html#method.pixels) returns? Is storing a trait object the only way?

I am trying to implement my own iterator over a grayscale image, and wrapping an iterator inside is much more convenient for that.

dlight:
A related question: is there any guarantee about the order of iteration? Can the programmer select the row major order?

I want to build an iterator where each yielded value depends on the pixel being iterated and the pixel below it. To exploit the cache more effectively, I want to iterate the matrix on row major order. So I will process the whole first row, saving it to a buffer, then start yielding values from the second row onwards.

dlight:
I'm doing this way because I suppose the `SampleLayout` will be row major.. (it's just `GrayImage`)

fintelia:
There isn't a way to select iteration order, but it is currently row-major and unlikely to change. I think you are right that a trait object is the easiest way to store the iterator because the return type isn't public. Another option is to disregard the iterator interface completely and just deref your `ImageBuffer` into a `&[T]` and either iterate over that slice or index into it as required.

This part of the API isn't the cleanest at the moment so I'd be curious to hear feedback on ideas if you think there would be a better way to structure it.

HeroicKatora:
The enumeration order is the memory layout order, which is row-major (I've never heard the term used to describe iteration order by itself). This won't change in any minor release as the fact that this is the most cache-friendly iteration order is probably codified in the existence of `enumerate_rows` where no `enumerate_columns` is provided. Also, [`Hyrum's Law`](https://www.hyrumslaw.com/) tells us there is already a project depending on that order without being explicit about it, as you.

On the issue of `pub`, that's an oversight I would say. They should likely be exported not in the top-level but a `buffer` module or similar.

dlight:
> This part of the API isn't the cleanest at the moment so I'd be curious to hear feedback on ideas if you think there would be a better way to structure it.

[...]

Issue #549, opened 2016-06-15
Iterator types are hidden in the documentation


retep998:
While the iterator types are marked as `pub` (thereby bypassing the private in public checks in Rust), they are inside a module which is private, and are not re-exported into a public location. Therefore when viewing the documentation, I cannot view these iterator types, so it makes it somewhat difficult to figure out what the iterators output.

nwin:
About which iterator types are you talking? Pixels and friends? They are not bypassing any checks. They are re-exported in lib.rs, that's a bug in rustdoc. Afaik that issue is already in the bugtracker.

retep998:
Yes Pixels. And yes I now see that they are indeed re-exported, which means rustdoc is just being terrible as usual.

nwin:
Just out of curiosity, what about http://www.piston.rs/image/image/struct.Pixels.html? It’s there…

retep998:
But they're not linked on http://www.piston.rs/image/image/struct.ImageBuffer.html

birktj:
Since this is a rustdoc bug and probably fixed by now, I will close this issue.

PR #1176, opened 2020-03-27
(Internally) restructure the buffer module

HeroicKatora:
Moves `Pixel` into the trait module. Exposes a public `buffer` module where iterators and other auxiliary definitions relating to the `ImageBuffer` are exported. The main `ImageBuffer` is _not_ re-exported within that module, for now. At some point we might want to make the type less special and then change the structure further and deprecate old export points.

Closes: #1069 
Closes: #549 (which apparently was either re-broken or never fixed)

### DIFF
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 
 use num_rational::Ratio;
 
-use crate::buffer::RgbaImage;
+use crate::RgbaImage;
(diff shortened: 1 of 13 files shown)


HeroicKatora:
Reviews should mainly focus on the public interface which __should__ not be affected by the change except by adding a completely new module and exporting some types within it.

Issue #1174, opened 2020-03-27
Contract for and details of ImageDecoderExt confusing


nabijaczleweli:
This happens in 0.23.2, as seen on docs.rs.

## Expected

Relatively important questions that popped into my head as I was (trying to) implement this to be answered:
  * When is "periodically"?
  * Is the output buffer always to size?
  * Who calls this?
  * What's the output format?

## Actual behaviour

Having the benefit of implementing this from inside `image` I could use `load_image()` that all existing implementations use to gleam some sort of data out of this:
  * "Periodically" seems to be once at the start, then after each chunk (can users depend on this?).
  * I'm still not sure, because
  * Nobody in `image` actually calls `read_rect{,_with_progress}()`
    ```
    Searching 2379 files for "read_rect"

    T:\image\src\bmp\decoder.rs:
     1417  
     1418  impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for BmpDecoder<R> {
     1419:     fn read_rect_with_progress<F: Fn(Progress)>(
     1420          &mut self,
     1421          x: u32,

    T:\image\src\dxt.rs:
      148  
      149  impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for DxtDecoder<R> {
      150:     fn read_rect_with_progress<F: Fn(Progress)>(
      151          &mut self,
      152          x: u32,

    T:\image\src\farbfeld.rs:
      145  
      146  impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for FarbfeldDecoder<R> {
      147:     fn read_rect_with_progress<F: Fn(Progress)>(
      148          &mut self,
      149          x: u32,
      ...
      234  
      235      #[test]
      236:     fn read_rect() {
      237          static RECTANGLE_IN: &[u8] =  b"farbfeld\
      238                                          \x00\x00\x00\x02\x00\x00\x00\x03\
      ...
      248          FarbfeldDecoder::new(Cursor::new(RECTANGLE_IN))
      249              .unwrap()
      250:             .read_rect_with_progress(1, 1, 1, 2, &mut out_buf, |p| println!("{:?}", p))
      251              .unwrap();
      252          assert_eq!(out_buf, RECTANGLE_OUT);

    T:\image\src\hdr\decoder.rs:
      100  
      101  impl<'a, R: 'a + BufRead + Seek> ImageDecoderExt<'a> for HDRAdapter<R> {
[...]

PR #1177, opened 2020-03-27
ImageDecoderExt documentation

nabijaczleweli:
Following discussion in #1174

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -435,9 +435,9 @@ pub trait ImageDecoder<'a>: Sized {
     }
 }
 
-/// ImageDecoderExt trait
+/// Specialized image decoding not be supported by all formats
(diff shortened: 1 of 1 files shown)


src/image.rs
```diff
@@ -430,9 +430,9 @@ pub trait ImageDecoder<'a>: Sized {
     }
 }
 
-/// ImageDecoderExt trait
+/// More exotic image parsing
```
nabijaczleweli: This might not be the perfect phrasing, but I didn't come up with anything closer to "The trait that all decoders implement" from `ImageDecoder` that conveys the right thought

HeroicKatora: > Specialized image decoding that need not be supported by all formats.

nabijaczleweli: Great! Added.

PR #1189, opened 2020-04-09
Remove DimensionError constructor

Veykril:
Removal of the DimensionError constructor listed in #1134.

There is one instance where I don't think the error is fitting though I wasn't sure what to replace it with either which is in the gif module where it is being used for erroring on the `frame_delay` being bigger than a `u16`.

### DIFF
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -387,13 +387,11 @@ impl<W: Write> Encoder<W> {
 
         // Create the gif::Frame from the animation::Frame
         let mut frame = Frame::from_rgba(width, height, &mut *rbga_frame);
-        frame.delay = (frame_delay / 10)
-            .try_into()
-            .map_err(|_| {
-                ImageError::Parameter(ParameterError::from_kind(
-                    ParameterErrorKind::DimensionMismatch
-                ))
-            })?;
+        // Saturate the conversion to u16::MAX instead of returning an error as that
+        // would require a new special cased variant in ParameterErrorKind which most
+        // likely couldn't be reused for other cases. This isn't a bad trade-off given
+        // that the current algorithm is already lossy.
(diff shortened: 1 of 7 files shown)


src/gif.rs
```diff
@@ -387,7 +387,13 @@ impl<W: Write> Encoder<W> {
 
         // Create the gif::Frame from the animation::Frame
         let mut frame = Frame::from_rgba(width, height, &mut *rbga_frame);
-        frame.delay = (frame_delay / 10).try_into().map_err(|_|ImageError::DimensionError)?;
```
Veykril, line 390: Here the `Option::None` is being mapped into a DimensionError if the `frame_delay / 10` expression of type `u32` doesn't fit into a `u16`. I believe this should have a different error kind returned as it is not really dimension related?

HeroicKatora: That's precisely why we're doing these, to uncover the misuse in some places.

HeroicKatora: There are two ways to go about this, I'm not quite sure which one is better.

* Treat this as a saturating conversion and use `u16::MAX`. This doesn't sound too bad given that the current algorithm is lossy in any case and then this would justifiably be the closest possible approximation. This is only relevant for frame durations above 10 minutes in any case.
* .. OR add a new variant to the `ParameterErrorKind` here.

Veykril: I would be in favor of the saturating conversion, as you already said that the algorithm is already lossy and because a new `ParameterErrorKind` for this seems like it would be solely used for this specific instance. At least I can't think of a name that could possibly have more applications than this instance.

HeroicKatora: Agreed then, be sure to add a comment with the explanation for why the saturation is performed.

PR #1193, opened 2020-04-11
Replace legacy error constructors in ICO decoder. Make DecodingError::message a Cow to reduce allocations

nabijaczleweli:
Also fixes a formatting bug in what is now `ImageEntryDimensionMismatch` with `format: Bmp`.

This knocks off `ico` from the list in #1134 and gets rid of one simple inline TODO item

### DIFF
--- a/src/ico/decoder.rs
+++ b/src/ico/decoder.rs
@@ -2,11 +2,11 @@ use byteorder::{LittleEndian, ReadBytesExt};
 use std::convert::TryFrom;
 use std::io::{self, Cursor, Read, Seek, SeekFrom};
 use std::marker::PhantomData;
-use std::mem;
+use std::{error, fmt, mem};
 
 use crate::color::ColorType;
-use crate::error::{ImageError, ImageResult};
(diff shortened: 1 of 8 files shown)


src/ico/decoder.rs
```diff
@@ -16,6 +16,85 @@ use crate::png::PngDecoder;
 // The first eight bytes of a PNG file always contain the following (decimal) values:
 const PNG_SIGNATURE: [u8; 8] = [137, 80, 78, 71, 13, 10, 26, 10];
 
+/// Errors that can occur during decoding and parsing an ICO image or one of its enclosed images.
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum IcoDecoderError {
```
HeroicKatora, line 21: Please keep this type private for now. Exposing it publicly makes it harder to move this into a separate crate or change it. We can still decide whether to expose it (and in what form) in a follow-up.

nabijaczleweli: Done

PR #1194, opened 2020-04-11
Destringtypify the BMP decoder

nabijaczleweli:
Opening the WIP because ~~(a) I'd like to rebase on #1193 when that's merged and (b)~~ I started doing this when i started drinking and I'd like to write some of this down before I pass out and forget.

Are `BitfieldMasksMissing` and `BitfieldMaskMissing` different or the same error conditions, considering the similar but differing original strings (i.e. can I merge the former into the latter)?

Also fixes https://github.com/image-rs/image/commit/183aab34ccb57374b0acac732ac7f681a17df842#r38436431
Vaguely ref: #1134
Part of: #1195

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -3,8 +3,8 @@ use std::io::{self, Cursor, Read, Seek, SeekFrom};
 use std::iter::{repeat, Iterator, Rev};
 use std::marker::PhantomData;
 use std::slice::ChunksMut;
-use std::{cmp, mem};
-use std::cmp::Ordering;
+use std::{error, fmt, mem};
+use std::cmp::{self, Ordering};
 
 use byteorder::{LittleEndian, ReadBytesExt};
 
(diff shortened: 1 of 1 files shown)


src/bmp/decoder.rs
```diff
@@ -109,17 +109,129 @@ impl<'a> Iterator for RowIterator<'a> {
     }
 }
 
+/// All errors that can occur when attempting to parse a BMP
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum BmpDecoderError {
```
HeroicKatora, line 114: Same concerns with instant  `pub` here as in the other PR. The string-errors was essentially the leanest interface possible, allowing arbitrary SemVer compatible additions.

HeroicKatora: Also this is in a module `bmp`, might as well drop the `Bmp`-prefix.

src/bmp/decoder.rs
```diff
@@ -109,17 +109,129 @@ impl<'a> Iterator for RowIterator<'a> {
...
+                f.write_fmt(format_args!("Palette size {} exceeds maximum size for BMP with bit count of {}", colors_used, bit_count)),
+        }
+    }
+}
+
+impl error::Error for BmpDecoderError {}
+
+/// Image types with distinct errors currently supported by `BmpDecoder`
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum BmpDecoderErrorInvalidChannelWidthImageType {
```
HeroicKatora, line 204: Waaaaaay too long ;)

HeroicKatora: `ChannelWidthError` essentially conveys the same information and is much shorter.

nabijaczleweli: Renamed to `DecoderErrorImageType`

src/bmp/decoder.rs
nabijaczleweli: Done.

Also of note: this is still perfectly opaque to the user – `DecodingError` stores `Option<Box<dyn Error + Send + Sync>>` which is perfectly unuseful, except to use it as `Display` (this transition should be transparent, too, since all those fields are private).

nabijaczleweli:
Made the overflow errors into Unsupporteds, as discussed in https://github.com/image-rs/image/pull/1196#discussion_r407252797

PR #1196, opened 2020-04-12
Remove legacy constructors from GIF and HDR decoders

nabijaczleweli:
I'm also a tad unsure of the import of the `limit_string_len()` function in HDR.

Ref: #1134

### DIFF
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -39,8 +39,8 @@ use num_rational::Ratio;
 use crate::animation;
 use crate::ImageBuffer;
 use crate::color::{ColorType, Rgba};
-use crate::error::{ImageError, ImageResult, ParameterError, ParameterErrorKind};
-use crate::image::{self, AnimationDecoder, ImageDecoder};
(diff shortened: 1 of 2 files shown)


src/gif.rs
```diff
@@ -123,8 +141,9 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for GifDecoder<R> {
                 // See the comments inside `<GifFrameIterator as Iterator>::next` about
                 // the error handling of `from_raw`.
                 let image = ImageBuffer::from_raw(f_width, f_height, &mut *buf).ok_or_else(
-                    || ImageError::UnsupportedError("Image dimensions are too large".into())
-                )?;
+                    || ImageError::Decoding(DecodingError::new(
```
HeroicKatora, line 144: This is not a decoding error. It means that the width and height are within specification but the platform we run on can not support the buffer space we would require.

HeroicKatora: It seems I missed this in `bmp`.

nabijaczleweli: Fixed, will do the same to `bmp`

PR #1203, opened 2020-04-18
Destringtypify PNM, WebP, WebP's VP8 decoders. Remove legacy constructor from PNM decoder. Add Error::source() to HDR's DecoderError. Remove ImageError::with_message(). Remove needless allocations from WebP decoder. Fix PNM header writing with custom TUPLTYPE

nabijaczleweli:
`ArbitraryTuplType::name()` – name? publicity?

Will go 'round the rest of `DecoderError`s and bring them back in line with the `Into<ImageError>` thing after this, it helps verbosity a lot.

### DIFF
--- a/src/pnm/decoder.rs
+++ b/src/pnm/decoder.rs
@@ -18,6 +18,173 @@ use crate::utils;
 
 use byteorder::{BigEndian, ByteOrder, NativeEndian};
 
+/// All errors that can occur when attempting to parse a PNM
+#[derive(Debug, Clone)]
+enum DecoderError {
+    /// PNM's "P[123456]" signature wrong or missing
+    PnmMagicInvalid([u8; 2]),
+    /// Couldn't parse the specified string as an integer from the specified source
+    UnparsableValue(ErrorDataSource, String, ParseIntError),
+
+    /// More than the exactly one allowed plane specified by the format
+    NonAsciiByteInHeader(u8),
+    /// The PAM header contained a non-ASCII byte
+    NonAsciiLineInPamHeader,
+    /// A sample string contained a non-ASCII byte
+    NonAsciiSample,
+
+    /// The byte after the P7 magic was not 0x0A NEWLINE
+    NotNewlineAfterP7Magic(u8),
+    /// The PNM header had too few lines
+    UnexpectedPnmHeaderEnd,
+
+    /// The specified line was specified twice
+    HeaderLineDuplicated(PnmHeaderLine),
+    /// The line with the specified ID was not understood
+    HeaderLineUnknown(String),
+    /// At least one of the required lines were missing from the header (are `None` here)
+    ///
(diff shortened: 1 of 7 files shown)


src/pnm/decoder.rs
```diff
@@ -18,6 +18,171 @@ use crate::utils;
...
+                f.write_fmt(format_args!("Invalid depth ({}) or maxval ({}) for tuple type {}", depth, maxval, tuple_type.name())),
+            DecoderError::InvalidDepth { tuple_type, depth } =>
+                f.write_fmt(format_args!("Invalid depth ({}) for tuple type {}", depth, tuple_type.name())),
+            DecoderError::TupleTypeUnrecognised =>
+                f.write_str("Tuple type not recognized"),
+        }
+    }
+}
+
+impl From<DecoderError> for ImageError {
```
HeroicKatora, line 126: Could you add a note here, saying that this needs to be hidden if `pnm` becomes a separate crate, until it is `1.0`? It works for private types though, which is very nice.

src/webp/decoder.rs
```diff
@@ -37,44 +72,38 @@ impl<R: Read> WebPDecoder<R> {
...
 
         Ok(size)
     }
 
     fn read_vp8_header(&mut self) -> ImageResult<u32> {
         loop {
-            let mut chunk = Vec::with_capacity(4);
-            self.r.by_ref().take(4).read_to_end(&mut chunk)?;
+            let mut chunk = [0; 4];
+            self.r.read_exact(&mut chunk)?;
```
HeroicKatora, line 95: This is a bugfix as well, right? Because before it could silently truncate the chunk name by a temporarily exhausted stream. Not sure how one could test for that bug though.

nabijaczleweli: I *think* the `take()`/`read_to_end()` combo worked, too:
> All bytes read from this source will be appended to the specified buffer. This function will continuously call `read()` to append more data until `read()` returns either `Ok(0)` or an error of non-`ErrorKind::Interrupted` kind.

Which sounds equivalent to `read_exact()` when paired with `take()`, but I'm knackered and might very well be missing something.

src/pnm/decoder.rs
nabijaczleweli: Done

src/webp/decoder.rs
HeroicKatora: It may read _less_ bytes than `take` if the input is shorter. Since `Ok(0)` might be a temporary condition, it's only a convention to interpret it as eof, it _could_ in theory stop reading and still succeed in the following calls. In any case, it would silently succeed and return an erro from an incorrect line.

nabijaczleweli: Oh, indeed! Will add that to the commit message.

PR #1204, opened 2020-04-18
Remove `ImageError::UnsupportedColor`, improve readability of DecoderErrors

nabijaczleweli:
Remove `ImageError::UnsupportedColor` usage from DynImage, farbfeld, flat, PNG and the constructor. Add `.into() -> ImageError` for the rest of *private* DecoderErrors.

Only TGA left, with #1145.

Part of #1134
Ref: #1203

### DIFF
--- a/src/bmp/decoder.rs
+++ b/src/bmp/decoder.rs
@@ -192,6 +192,12 @@ impl fmt::Display for DecoderError {
     }
 }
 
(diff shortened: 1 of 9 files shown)


HeroicKatora:
In the future, a title with the length of a full paragraph indicates that breaking the PR into independent pieces would be easier to review.

PR #1206, opened 2020-04-19
Berid of most (all?) warnings

nabijaczleweli:
Mainly the vomit from `ExtendedColorType` in each and every build, but tested with no features, one feature at a time, and all features, all of which have just the two warnings fixed by #1205.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -80,27 +80,49 @@ impl ColorType {
 /// decoding from and encoding to such an image format.
 #[derive(Copy, PartialEq, Eq, Debug, Clone, Hash)]
 pub enum ExtendedColorType {
+    /// Pixel is 1-bit luminance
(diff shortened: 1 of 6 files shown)


src/color.rs
```diff
@@ -78,6 +78,7 @@ impl ColorType {
 ///
 /// Another purpose is to advise users of a rough estimate of the accuracy and effort of the
 /// decoding from and encoding to such an image format.
+#[allow(missing_docs)]
```
JohnTitor, line 81: I'd like to add doc comments here like `ColorType` rather than suppress warnings.

nabijaczleweli: Done

HeroicKatora:
Looks like the main crate is free of warnings. Very nice :tada:

PR #1215, opened 2020-05-11
Use enlargeable types in to_luma functions.

### LINKED ISSUE #1214: to_luma methods can have floating point issues
In the `rgb_to_luma` and `bgr_to_luma` methods, the floating point arithmetic can result in incorrect results. For example:

```rust
use image::Pixel;

fn main() {
    let pixel = image::Rgb::from([13, 13, 13]);
    
    println!("{:?}", pixel.to_luma()); // Prints `Luma([12])`; should print `Luma([13])`
}
```
This is because in `rgb_to_luma`, `l` is equal to `12.99999904632568359375` before being converted back to a `u8`.

kaj:
This is an attempt to fix #1214, using the `Enlargeable` trait to select a suitable type for the calculation in `rgb_to_luma` and `bgr_to_luma`.

Integer pixel types will use suitable integer pixel types for the calculation, while floating point pixel types will use floating point types.

This PR also provides the `Enlargeable` trait for all `Primitive` types (luckily, `Primitive` is not implemented for 128 bit numeric types).

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -379,22 +379,23 @@ where
 }
 
 /// Coefficients to transform from sRGB to a CIE Y (luminance) value.
-const SRGB_LUMA: [f32; 3] = [0.2126, 0.7152, 0.0722];
+const SRGB_LUMA: [u32; 3] = [2126, 7152, 722];
+const SRGB_LUMA_DIV: u32 = 10000;
 
 #[inline]
 fn rgb_to_luma<T: Primitive>(rgb: &[T]) -> T {
-    let l = SRGB_LUMA[0] * rgb[0].to_f32().unwrap()
-        + SRGB_LUMA[1] * rgb[1].to_f32().unwrap()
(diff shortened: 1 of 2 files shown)


CrackedP0t:
I'll post my benchmark results here too:

[Here's a Gist with the code I used](https://gist.github.com/CrackedP0t/a4c76188baef4361d96e9974c7f12719)

With this benchmark, the old float method takes ~70 ms, and both the int and enlarge methods take ~20 ms, with the enlarge method usually running a few ms faster.

I think this PR is an excellent solution, but I'm still very interested in the maintainers' opinions

HeroicKatora:
Maintainers opinion: This is a breaking change. There should be a note somewhere so that something similar can be adopted for `0.24`, especially since it also improves performance, but I don't think the accuracy fix itself is worth a major version on its own. In particular it would imply that downstream users wouldn't automatically receive the fix which is half of the motivation, right?

PR #1221, opened 2020-05-22
Do not notify Gitter from fork repos

JohnTitor:


JohnTitor:
This was discussed on Gitter channel a while back and I think it makes sense, so I'll merge it.

PR #1222, opened 2020-05-24
Update the jpeg-decoder dependecy to 0.1.19

ArturKovacs:
I think I gave my consent for the MIT/Apache license before, but better safe than sorry.


Shnatsel:
Note that the latest available version is already used by default by the Cargo dependency resolution algorithm. The only thing this change accomplishes is making `image` strictly require at least this version, and fail to build with earlier ones. Is there a reason for doing this?

ArturKovacs:
Indeed, I didn't know about this so I'm closing this as it's not required.

PR #1228, opened 2020-05-28
Fix bug in gif encoding of rgba8 frames

### LINKED ISSUE #1227: gif::Encoder::encode() uses the wrong method for ColorType::Rgba8
Notice how Rgba8 calls Frame::from_rgb(), instead of Frame::from_rgba()
```rust
    /// Encode a single image.
    pub fn encode(
        &mut self,
        data: &[u8],
        width: u32,
        height: u32,
        color: ColorType,
    ) -> ImageResult<()> {
        let (width, height) = self.gif_dimensions(width, height)?;
        match color {
            ColorType::Rgb8 => self.encode_gif(Frame::from_rgb(width, height, data)),
            ColorType::Rgba8 => {
                self.encode_gif(Frame::from_rgb(width, height, &mut data.to_owned()))
            },
            _ => Err(ImageError::Unsupported(UnsupportedError::from_format_and_kind(
                ImageFormat::Gif.into(),
                UnsupportedErrorKind::Color(color.into())
            ))),
        }
    }
```

JMS55:
I believe this fixes #1227 

---

### DIFF
--- a/src/gif.rs
+++ b/src/gif.rs
@@ -316,7 +316,7 @@ impl<W: Write> Encoder<W> {
         match color {
             ColorType::Rgb8 => self.encode_gif(Frame::from_rgb(width, height, data)),
             ColorType::Rgba8 => {
-                self.encode_gif(Frame::from_rgb(width, height, &mut data.to_owned()))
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Thank you for the fix.

PR #1223, opened 2020-05-24
make the jpeg encoder work with non-memory and non RGB8 images

lovasoa:
Change the JPEG encoder to work on a GenericImageView instead of a byte slice

This change allows users to encode jpeg images from other sources
than memory, thus finally allowing the encoding of images that
do not fit in memory.

This also allows users to make the encoder work in a streaming fashion.
(see #1219)

This also finally allows encoding images from other pixel formats than
8-bit RGB. The conversion to 8bit YCbCr (used in JPEG) is done on the fly
during encoding.

I need this change for [dezoomify-rs](https://lovasoa.github.io/dezoomify-rs/).

I license past and future contributions under the dual MIT/Apache-2.0 license,
allowing licensees to chose either at their option.

PR #1229, opened 2020-05-28
Introduce bounds check pixel get/get_mut

HeroicKatora:
As discussed in #1223.

The intention here is to help the optimizer. For methods where access
out-of-bounds is not fatal the bounds checking code appears to be
repeated unnecessarily in many cases—once for a manual check to avoid
the panic and once in the impl in the path that would lead to such
panic.

This is backwards compatible in that the new methods are default
implemented on the traits. In the long term it seems nevertheless
advisable to flip this around and default implement the panicking method
instead. This would also improve possibilities for indexing and
adherence to the API guidelines.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -775,6 +775,14 @@ where
         *self.get_pixel(x, y)
     }
 
+    fn opt_pixel(&self, x: u32, y: u32) -> Option<P> {
+        if let Some(idx) = self.pixel_indices(x, y) {
+            Some(*P::from_slice(&self.data[idx]))
+        } else {
+            None
+        }
+    }
+
     /// Returns the pixel located at (x, y), ignoring bounds checking.
     #[inline(always)]
     unsafe fn unsafe_get_pixel(&self, x: u32, y: u32) -> P {
@@ -797,6 +805,14 @@ where
 
     fn get_pixel_mut(&mut self, x: u32, y: u32) -> &mut P {
         self.get_pixel_mut(x, y)
+    }
+
+    fn opt_pixel_mut(&mut self, x: u32, y: u32) -> Option<&mut P> {
+        if let Some(idx) = self.pixel_indices(x, y) {
+            Some(P::from_slice_mut(&mut self.data[idx]))
+        } else {
+            None
+        }
     }
 
     fn put_pixel(&mut self, x: u32, y: u32, pixel: P) {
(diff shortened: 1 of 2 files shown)


lovasoa:
`try_get_pixel` is not perfect (it may indeed suggest that the return type is a `Result`), but I think it conveys the meaning of what the function does better than `opt_get_pixel`. I wouldn't have guessed what the function behavior is just by its name.

src/buffer.rs
```diff
@@ -775,6 +775,14 @@ where
         *self.get_pixel(x, y)
     }
 
+    fn opt_pixel(&self, x: u32, y: u32) -> Option<P> {
+        if let Some(idx) = self.pixel_indices(x, y) {
+            Some(*P::from_slice(&self.data[idx]))
```
lovasoa, line 780: @HeroicKatora : Is the bound check on `self.data` correctly eliminated here ?

HeroicKatora: It would require the optimizer to note that the length of the slice is guaranteed to be longer than the maximum pixel index. I'm not entirely sure it does or can. I feel slightly uncomfortable replacing it with an `unsafe` version since the `ImageBuffer` is parameterized over a user provided buffer but a fallible indexing could avoid the panic path and instead have a `cmov` that is never taken. That might vectorize better than a panic path.

lovasoa: I don't think it's possible to create an `ImageBuffer` where the backing buffer is too short, is it ? This function is likely to be called in a tight loop in a performance-sensitive part of a program, I think using a single line of unsafe should be worth the performance gains.

HeroicKatora: It's certainly not possible to create a `ImageBuffer<_, &mut [u8]>` where the backing buffer is too short but the buffer offers customization of the container type. In those cases, there is nothing preventing an incorrectly implemented container from lying to the constructor to fake the correct length and subsequently shortening its internal length. It wouldn't even be `unsafe` to do so.

I think to fully use `unsafe` for this access we'd need to create a small wrapper type that can only be constructed with a buffer that is one of the standard types. Similar to `Pin<&mut T>` this could be constructed from any `ImageBuffer` by dereferencing the container.

lovasoa: Yes, this would be better, but this would be specific to the *standard* types. I was thinking of something that would still work for all `Container: Deref<Target = [P::Subpixel]>` and yet prevent the duplicated bound checks: we could check only whether `x < width`, then compute the pixel indices, and then use [`get`](https://doc.rust-lang.org/std/primitive.slice.html#method.get) on the slice (which return an option, that would be None if `y < height`). Would that work ?

HeroicKatora: Yep, the option trick would work for all types. The `get_unchecked` trick that completely eliminates the bounds check only works for standard types.

Issue #1240, opened 2020-05-29
Rayon feature broken on `1.34.2`


HeroicKatora:
When using the `jpeg_rayon` feature on a new project, cargo version `1.34.2` is no longer able to compile this project with the newest dependencies. This is because the transitive dependency `crossbeam-queue` has published a new minor version `0.2.2` that requires `MaybeUninit`, a feature that was stabilized in `1.36.0`.

Workarounds, and if this will get a backwards fix, will be filled in once decisions have been made. For now this is very similar to #975.

HeroicKatora:
This also affects the `criterion` dev-dependency that depends on `rayon` as well. Since dev-dependencies can not be made optional this breaks all test and bench builds on the old version and this can not be easily hidden behind a feature flag. A workaround would be to disable those builds.

HeroicKatora:
This does not work but I believe the informational purpose of this issue has run its course.

PR #1241, opened 2020-05-29
Disable test execution on old Rust version in CI

HeroicKatora:
Addresses the CI portion of #1240, until we have either reached a definitive decision or the next MSRV bump.


HeroicKatora:
This did not trigger a build, I'm unsure why Travis didn't pick this up.

HeroicKatora:
Well, CI is broken on master in any case so merging.

Issue #1239, opened 2020-05-29
No way to get a suitable extension from an ImageFormat


PurpleMyst:
I would like to be able to get a suitable extension for each variant of the `ImageFormat` enum.

My specific use case for this functionality is saving downloaded blobs to a file without format conversion.

This is more generally applicable to avoiding format conversions.

## Draft

https://github.com/image-rs/image/blob/8a4de735642e090889446f3f822e90cb643cb4ec/src/io/free_functions.rs#L223-L252 practically already implements the opposite of what I want, so I think it'd be possible to re-use much of this code to do the requested feature.

tripulse:
For JPEG (aka. JFIF) has multiple extensions.

PurpleMyst:
What about ImageFormat to MimeType as some people have suggested on Discord?

Nemo157:
If there was integration with [`mime`](https://crates.io/crates/mime) specifically then [`mime_guess`](https://crates.io/crates/mime_guess) already has mappings from mime-type to extension

```rust
dbg!(mime_guess::get_mime_extensions(&mime::IMAGE_JPEG));

    [
        "jpe",
        "jpeg",
        "jpg",
    ],
```

tripulse:
> What about ImageFormat to MimeType as some people have suggested on Discord?

Yeah, we can just include a crate for that.

PurpleMyst:
What do you mean by "include a crate for that"?

HeroicKatora:
The [`mime_guess`](https://docs.rs/mime_guess/2.0.3/mime_guess/) crates offers methods to get a list of known extension for a particular mime type.

HeroicKatora:
You could either use that, or we use it as an inspiration for the API that offers a similar functionality.  Particularly we want to ensure that formats that have no (officially) recognized file extension are not required to provide a bogus one. The main question seems like a good idea to me as well, we'll need roundtrip tests to ensure that the implementations don't diverge here but it seems quite feasible in general.

PR #1244, opened 2020-05-29
Add a reverse mapping of ImageFormat::from_path

HeroicKatora:
This is added specifically to list extension that are surely recognized when used with `from_path`. This same guarantee would not be given if using some external list of names. You can open another issue/PR if you feel that some extensions are missing from the list of a format.

@PurpleMyst, @tripulse, @Nemo157

Closes: #1239


PurpleMyst:
That's great! :confetti:

Issue #697, opened 2017-11-26
ImageBuffer should also derive Hash, PartialEq and Eq


yoanlcq:
Greetings,

Title says it all - I think this would be convenient for APIs that expose structs which happen to contain an `ImageBuffer` (as is the case for me).
Since it is a slight, non-breaking change, I didn't bother to do a PR - but it might also be an opportunity to dicuss this for other structs such as `Frame`.

HeroicKatora:
This was fixed in #1248.

PR #1248, opened 2020-05-31
Add derives for `Hash`, `PartialEq` and `Eq`.

soruh:
This provides a (very trivial) implementation for #697 but is also meant to encourage discussion about this Issue or make private decisions about it public.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -449,7 +449,7 @@ where
 ///
 /// image::imageops::overlay(&mut img, &on_top, 128, 128);
 /// ```
-#[derive(Debug)]
+#[derive(Debug, Hash, PartialEq, Eq)]
 pub struct ImageBuffer<P: Pixel, Container> {
     width: u32,
     height: u32,


src/buffer.rs
```diff
@@ -449,7 +449,7 @@ where
 ///
 /// image::imageops::overlay(&mut img, &on_top, 128, 128);
 /// ```
-#[derive(Debug)]
+#[derive(Debug, Hash, PartialEq, Eq)]
```
HeroicKatora, line 452: So the the unanswered question, which I did not immediately thought of when the issue was opened and subsequently forgot to write down, was whether it be useful to compare the _complete_ container as it is performed with the derive or if the `PartialEq`, `Eq`, `Hash` should only look at the channels that are part of the logical image given by `width` and `height`.

soruh: I think looking at the logical image would make more sense, if someone really wants to compare/hash the raw Bytes of an image I'd assume they could just call into raw. This behavior should obviously be documented though...

HeroicKatora: Can you recall any an precedent of a similar data structure, preferrably in the standard library or another project closely associated with the working groups?

soruh: Not from the top of my head, but I'll take a quick look right now.

HeroicKatora: Maybe even something simple as `Vec` counts. Comparing two `Vec`s only compares the initialized content and length by dereferencing to the slice, but not the capacity.

soruh: The only datastructure i can think of that seems similar is a (string) slice, which is a pointer to a buffer, which could be bigger than what the slice actually uses.

[Slices hash their contents as well as their lengths](https://doc.rust-lang.org/stable/src/core/hash/mod.rs.html#665-670), so our `derive` implementation follows the same Idea.

soruh: It looks like [`bitflags` hashes the whole container](https://docs.rs/bitflags/1.2.1/src/bitflags/lib.rs.html#406) and not just the used bits.

soruh: Maybe even something simple as `Vec` counts. Comparing two `Vec`s only compares the initialized content and length [by dereferencing to the slice](https://doc.rust-lang.org/stable/src/alloc/vec.rs.html#2269-2283), but not the capacity.

I don't think that is comparable, since comparing the whole buffer would be undefined behaviour and could (and would) produce different results each time, which is obviously undesireable

_Originally posted by @HeroicKatora in https://github.com/image-rs/image/pull/1248_

HeroicKatora: Obviously the undefined portion couldn't hash any instances of `T` but it's interesting nevertheless that a slice hashes its length but the vector does not hash (nor compare) its capacity even though it is part of their state.

But you're right that the example isn't perfect. In theses cases we need to be a little careful though as their motivation is likely the [`Borrow`](https://doc.rust-lang.org/stable/std/borrow/trait.Borrow.html) trait. Since `Vec<T>: Borrow<[T]>` it must behave appropriately:

> In particular Eq, Ord and Hash must be equivalent for borrowed and owned values: x.borrow() == y.borrow() should give the same result as x == y.

soruh: Ah sorry, I seem to have had misread your comment. I agree that they probably do this because of Borrow, but I think from a users point of view it also wouldn't make much sense if two Vecs with equal contents, but different capacities weren't equal <=> produce the same hash.

soruh: By the way: I am not that familiar with this libraries internals, how often would these two alternatives actually produce different results?

fintelia: The normal case is that an ImageBuffer is backed by a Vec sized exactly to the dimensions of the image, so there wouldn't be a difference. I kind of assumed (incorrectly based on checking the code) that we'd actually fail on ImageBuffer creation if the underlying container was too big.

HeroicKatora: No, we fail when the buffer is too small, not when it is too large. See the recent fix #1249 with a test case for that :smiley:

soruh: Is there a reason to not fail on oversized Buffers (except for API stability)?
When creating an image from a Raw Buffer I'd personally rather get an error than a broken image...

HeroicKatora: It doesn't give a broken image. Au contraire, it simply uses the prefix that belongs to the specified dimensions. We might have gone with resizing the container appropriately but there was no common trait available and the API is extremely general, and that is useful. Hence why it is so tricky to change.

HeroicKatora: In particular, it is lossless. You can pass a buffer that is too long and later unwrap it again, retrieving the fully initialized buffer. If the vector were shortened this would drop some initialized elements.

HeroicKatora: I think I've partially answered my own concern. Since the extra state of the buffer is part of the considered state, we should just use it when hashing. For users that only use fresh images and exactly sized buffers there is no difference, that seems to be a sizable portion of the user base in any case.

So just derive it.

Issue #195, opened 2014-12-13
Support APNG


retep998:
This issue is to track the state of Animated Portable Network Graphics support.
- [ ] Animated Portable Network Graphics support.
- [x] Introduce an API for animated images, see #145

More information:
https://wiki.mozilla.org/APNG_Specification

abonander:
cc PistonDevelopers/image-png#93

I've run into a use-case for this. I'm generating some demo animations for [img_hash](https://github.com/abonander/img_hash) but they're not really suited for GIF encoding for a number of reasons. I'd mostly just like an easy way to losslessly export a sequence of frames to a format that FFMPEG can ingest.

HeroicKatora:
After long waiting we finally have #1250 :tada: It needs a small amount of precursory work in `png` but has already been tested to work with a modified build of `emulsion`.

PR #1250, opened 2020-06-01
Add an AnimationDecoder adaptor for APNG

HeroicKatora:
We do not yet handle all color types. This conveniently reminds us to
make the transformation into a proper animated iterator a fallible one
as adding limits later also influences this interface. In particular we
must at least remember (read: allocate) one additional frame but
potentially more. Furthermore the animation interface does yet allow the
caller to allocate target buffers.

Unanswered questsions:

- [x] Check color support, are we okay with not supporting 16-bit samples and does it behave correctly here?
- [x] Is the interface good? 
  - What if one wants to fallback to simple `png` on unsupported colors images?
  - Should 'thumbnails' be handled explicitly? (The basic IDAT image can have no fcTL chunk in which case it is formally not part of the animation. This implementation currently yields it anyways).

Closes: #195


HeroicKatora:
Whoops, converted to draft as it more or less depends on unpublished pr in `png`: https://github.com/image-rs/image-png/pull/210

Issue #1252, opened 2020-06-05
Converting `RgbaImage` to `GrayImage`


robyoung:
What is the recommended way to convert between different types of `ImageBuffer`?

The simplest approach I have used is:
```rust
GrayImage::from_vec(
  width,
  height,
  rgba_image.pixels()
            .flat_map(|p| Vec::from(p.to_luma().0).into_iter())
            .collect())
```
However that is pretty slow because of allocating the `Vec` for each pixel.

I have also seen:
```rust
GrayImage::from_fn(
  width,
  height,
  |x, y| unsafe {
    rgba_image.unsafe_get_pixel(x, y).to_luma()
  })
```
But I would rather avoid unsafe if possible.

I am currently using [arrayvec::ArrayVec](https://docs.rs/arrayvec/0.5.1/arrayvec/struct.ArrayVec.html) which seems to be about as fast as the `unsafe` approach although I haven't benchmarked it.
```rust
GrayImage::from_vec(
  width,
  height,
  rgba_image.pixels()
            .flat_map(|p| ArrayVec::<[u8; 1]>::from(p.to_luma().0).into_iter())
            .collect())
```

It seems to me that this approach would be a bit easier if `Pixel` implemented `IntoIterator` with an iterator over it's `SubPixel`s. We could then `flat_map` directly on `p.to_luma()`. If that sounds like a good idea I would be very keen to submit a PR.  However, it all feels like I'm missing something obvious and there is a much better way of converting between types of `ImageBuffer`. So what is the recommended way to convert between different types of `ImageBuffer`?

HeroicKatora:
The best approach is to go through `DynamicImage`:

```rust
DynamicImage::ImageRgba8(rgba_image).into_luma()
```

Since this has not been the first time it has come up, do you have tips on how this could be made more obvious? Are the names simply not consistent enough? Do you think documentation on `ImageBuffer` could have helped? Top-level documentation? Something else?

robyoung:
Oh brilliant, that is perfect thank you.

In terms of the docs, I think something on the `ImageBuffer` would have helped me personally as that is where I quickly ended up when making changes to an image. However, my experience may not be the most common as my first exposure to this library was being dropped into the middle while trying to understand some client code.

HeroicKatora:
@robyoung Thanks, that's a lot better than no feedback :+1:

PR #1253, opened 2020-06-05
Direct user to DynamicImage for converting

robyoung:
Attempts to address the documentation suggestion in #1252

A lot of the methods for modifying an image are found in the `ImageBuffer` so that is where I was looking when trying to convert an image to a different Pixel type.


robyoung:
How does that look?

HeroicKatora:
> How does that look?

Yes, quite good.

HeroicKatora:
Thank you!

Issue #786, opened 2018-06-26
Color space information


casey:
It would be nice to be able to access image color space information, which is often stored in image metadata.

[Ogeon's palette crate](https://github.com/Ogeon/palette) has a good overview of color spaces as they relate to image formats.

For background, I'm writing a simple game framework, and I'd like to be able to correctly convert to a linear color space when loading assets, which I'm unable to do without knowing the image's color space.

Thanks!

HeroicKatora:
Based on `ffmpeg` I had an idea how this could be implemented internally, based on `palette` but without exposing that detail directly.

> For example, let's have a look at `ffmpeg` for reference:
> * There is a [`Context`](https://docs.rs/ffmpeg/0.2.0-alpha.1/ffmpeg/software/scaling/context/struct.Context.html) that describes the color transformation without any buffer involved.
> * Although it uses a completely different and dynamic [`Pixel`](https://docs.rs/ffmpeg/0.2.0-alpha.1/ffmpeg/util/format/pixel/enum.Pixel.html) so it may only be of limited use as an example.
> * Here is the [`ColorSpace`](https://docs.rs/ffmpeg/0.2.0-alpha.1/ffmpeg/software/scaling/color_space/enum.ColorSpace.html) enum for reference.

The similar interface in terms of Rust but also type safe would looks like so:

```rust
// Define an opaque converter struct, which models `Context`.
struct ColorConverter<From, To> {
    // Internals still unsure, but possibly:
    map: Box<dyn Fn(From) -> To + Send + Sync 'static>,
}

// The color space enum.
enum ColorSpace {
    // probably the same variants
}

// A trait to create a converter for those types for which we have it implemented.
trait MakeConverter<From, To> {
    fn make_converter(self) -> Result<ColorConverter<From, To>, ImageError>;
}

impl<From, To> ColorConverter<From, To> {
    pub fn new(seed: impl MakeConverter<From, To>)
        -> Result<Self, ImageError>
    {
        seed.make_converter()
    }

    fn apply(&self, from: &impl GenericImageView<Pixel=From>)
        -> ImageBuffer<To>
    { }
}

// And implement `MakeConverter` on `ColorSpace` for specific pixel types:
impl MakeConverter<Rgb<u8>, Luma<u8>> for ColorSpace {
    fn make_converter(self) -> Result<ColorConverter<From, To>, ImageError> {
        match self {
            ITU709 => /* some code to create it */
[...]

PR #1254, opened 2020-06-07
Make FromColor public

robyoung:
It is useful to have this trait public to convert an image buffer into an existing buffer rather than creating a new one.

For example (taken from ImageBuffer::convert):
```rust
let mut my_existing_buffer: GrayImage;
let my_rgba_buffer: RgbaImage;
from (to, from) in my_existing_buffer.pixels_mut().zip(src.pixels()) {
  to.from_color(from);
}
```

An alternative solution to my current problem would be to introduce a method on `ImageBuffer` that allows you to convert into an existing `ImageBuffer` however that seemed less flexible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,7 +114,7 @@ extern crate quickcheck;
 
 use std::io::Write;
 
-pub use crate::color::{ColorType, ExtendedColorType};
+pub use crate::color::{ColorType, ExtendedColorType, FromColor};
 
 pub use crate::color::{Luma, LumaA, Rgb, Rgba, Bgr, Bgra};
 


HeroicKatora:
The traits `FromColor` and `IntoColor` is somewhat private by design, though that is up for discussion. The current reasoning is that they are more or less a bug in the current treatment of color spaces (that was preserved for complete images due to compatibility and simplicity). See [this comment](https://github.com/image-rs/image/blob/master/src/color.rs#L395-L396) and #855 in particular.

> // FIXME: this trait should be removed and replaced with real color space models
// rather than assuming sRGB.

Encouraging the use of these imprecise methods further is not the way I'd like to evolve, it may also introduce more interfaces and obstacles to replacing them. Instead, the alternative might not be such a bad idea.

Introducing a color space aware wrapper (similar to `sub_image`?) that wraps a `GenericImage` could make this at least correct. What I'd really like to do is prototype this outside of `image` though and simply utilize its trait.

robyoung:
Most of what you said in your last comment was a bit over my head I'm afraid. I do not have a lot of experience in image processing. However, I did pick up on this.

> Instead, the alternative might not be such a bad idea.

If I were to introduce it as a method on `ImageBuffer` would `convert_into(&self, other: &mut T)` be a good name for it? I was struggling to come up with anything better.

HeroicKatora:
Crash course on color spaces: The term RGB describes only a very rough method of displaying/sampling color. Namely to have greenish, a redish, and a blueish light source and blend them together to have the same appearance as some reference color. It however does not on its own have any information on which colors to use and in what relation, which is what the _color space_ such as specified as `sRGB` or `Rec.709` defines. This is achieved by having the three exact _primary_ colors defined through their own typical simulus to each of the human eye's three different color receptors. More on [wikipedia](https://en.wikipedia.org/wiki/CIE_1931_color_space).

Due to its popularity, and push by Microsoft, the [`w3` standard](https://www.w3.org/Graphics/Color/sRGB.html) and de-facto default is `sRGB` in many printers, displays and other color devices as well as the internet. (Curiously that is actually not quite correct. Of all image formats, `jpeg` happens to be specified in terms of `Rec.601` which has a _very slightly_ different rgb from `sRGB`/`Rec.709` though that correction is sometimes more expensive to perform than it is worth..).

I hope that is enough information to understand why `RGB -> Luma` is not a well-determined conversion, only by enriching it with the particular information that the source color space is `sRGB` does it become specifiable.

With regards to the interface, it does not seem appropriate to have a new interface on `ImageBuffer` itself without any more parameters. For example, let's have a look at `ffmpeg` for reference:
* There is a [`Context`](https://docs.rs/ffmpeg/0.2.0-alpha.1/ffmpeg/software/scaling/context/struct.Context.html) that describes the color transformation without any buffer involved.
* Although it uses a completely different and dynamic [`Pixel`](https://docs.rs/ffmpeg/0.2.0-alpha.1/ffmpeg/util/format/pixel/enum.Pixel.html) so it may only be of limited use as an example.
* Here is the [`ColorSpace`](https://docs.rs/ffmpeg/0.2.0-alpha.1/ffmpeg/software/scaling/color_space/enum.ColorSpace.html) enum for reference.

Continued the actual idea for the implementation in #786 since it's a bit more and likely not very helpful if it stays in this immediate PR.

PR #1262, opened 2020-06-18
Fix loss of brightness in the JPEG encoder

### LINKED ISSUE #1260: Round-tripping JPEGs makes them darker
This happens in image 0.23.5

## Expected
The output JPEGs look like the input JPEGs

## Actual behaviour
The output JPEGs got darker with each round-trip, even with quality 100

## Reproduction steps
```rust
use image::ImageOutputFormat;
use std::{
    fs::File,
    io::{BufWriter, Write},
};

fn main() -> Result<(), anyhow::Error> {
    for i in 1..10 {
        let img = image::open(format!("image{}.jpg", i))?;
        let mut writer = BufWriter::new(File::create(format!("image{}.jpg", i + 1))?);
        img.write_to(&mut writer, ImageOutputFormat::Jpeg(100))?;
        writer.flush()?;
    }

    Ok(())
}
```

image

VVatashi:
Fixes #1260

### DIFF
--- a/src/jpeg/encoder.rs
+++ b/src/jpeg/encoder.rs
@@ -805,6 +805,7 @@ fn rgb_to_ycbcr<P: Pixel>(pixel: P) -> (u8, u8, u8) {
     let g: f32 = g.to_f32().unwrap();
     let b: f32 = b.to_f32().unwrap();
 
+    // Coefficients from JPEG File Interchange Format (Version 1.02), multiplied for 255 maximum.
     let y = 76.245 / max * r + 149.685 / max * g + 29.07 / max * b;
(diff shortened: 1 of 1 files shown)


VVatashi:
It's just the coefficients from the https://www.w3.org/Graphics/JPEG/jfif.pdf, multiplied by 255, so after dividing by the `max`, they will be exactly the same as there.

HeroicKatora:
That's already good information.

> Coefficients from JPEG File Interchange Format (Version 1.02), multiplied for 255 maximum.

would be a perfectly fine comment

HeroicKatora:
Thanks! Now just waiting on CI :slightly_smiling_face:

HeroicKatora:
:tada:

PR #1268, opened 2020-06-27
Fix a few minor grammar issues in docs

benfrankel:


HeroicKatora:
Thank you!

PR #1269, opened 2020-06-27
imageops/colorops: add `lookup` to `trait ColorMap`.

wathiede:
This change allows users to expand the indexed `Luma<u8>` returned by `imageops::colorops::index_colors` back into a true color image.

I also added type level documentation for `imageops::colorops::BiLevel`
and `math::nq::NeuQuant` showing how to use `ColorMap`.

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -3,19 +3,17 @@
 use num_traits::{Num, NumCast};
 use std::f64::consts::PI;
 
-use crate::ImageBuffer;
 use crate::color::{Luma, Rgba};
 use crate::image::{GenericImage, GenericImageView};
 use crate::math::nq;
 use crate::math::utils::clamp;
 use crate::traits::{Pixel, Primitive};
+use crate::ImageBuffer;
 
 type Subpixel<I> = <<I as GenericImageView>::Pixel as Pixel>::Subpixel;
 
 /// Convert the supplied image to grayscale
-pub fn grayscale<I: GenericImageView>(
-    image: &I,
-) -> ImageBuffer<Luma<Subpixel<I>>, Vec<Subpixel<I>>>
+pub fn grayscale<I: GenericImageView>(image: &I) -> ImageBuffer<Luma<Subpixel<I>>, Vec<Subpixel<I>>>
 where
     Subpixel<I>: 'static,
     <Subpixel<I> as Num>::FromStrRadixErr: 'static,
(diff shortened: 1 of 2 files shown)


src/imageops/colorops.rs
```diff
@@ -181,14 +179,45 @@ where
 pub trait ColorMap {
     /// The color type on which the map operates on
     type Color;
-    /// Returns the index of the closed match of `color`
+    /// Returns the index of the closest match of `color`
     /// in the color map.
     fn index_of(&self, color: &Self::Color) -> usize;
+    /// Looks up color by index in the color map.  If `idx` is out of range for the color map,
+    /// `None` is returned.
+    fn lookup(&self, idx: usize) -> Option<Self::Color>;
```
HeroicKatora, line 187: Looks good, except for the breaking change here. Can we add some default implementation returning `None`?

Possibly by adding a `has_lookup(&self) -> bool { false }` default implemented method as well?

HeroicKatora: See [RFC 1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#minor-change-adding-a-defaulted-item) for which changes are considered minor (i.e. don't require a major version bump).

wathiede: Thanks for the feedback, I meant to ask what is the best way to change APIs in the initial PR.  Thank you for the link to RFC 1105.

src/imageops/colorops.rs
```diff
@@ -181,14 +179,51 @@ where
 pub trait ColorMap {
     /// The color type on which the map operates on
     type Color;
-    /// Returns the index of the closed match of `color`
+    /// Returns the index of the closest match of `color`
     /// in the color map.
     fn index_of(&self, color: &Self::Color) -> usize;
+    /// Looks up color by index in the color map.  If `idx` is out of range for the color map, or
+    /// ColorMap doesn't implement `lookup` `None` is returned.
+    fn lookup(&self, _idx: usize) -> Option<Self::Color> {
```
HeroicKatora, line 187: This name is fine, nevertheless methods of the form `fn (&Self, usize) -> Option<_>` are often named `get`. However, that has a larger risk associated with it by colliding with the standard name (see considerations in the rfc). Another idea would be to name it `get_color` which should be much less risky. At the same time though, the name association to `has_lookup` is nice, we would have to find another name for that method then.

I don't want to spend too much energy on bikeshedding the name here, it's already pretty good. So, if there is no immediate idea for another pair of names then this is perfectly fine.

HeroicKatora: I didn't notice this concern on the first review either, sorry for any delay.

wathiede: I can do `get_color`/`has_get_color` if that seems better, or I'm okay leaving it as-is?  I have no strong preference, and I'm happy to let you decide on the style as the library maintainer.

Reading https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter I'm not sure `get` would be a good name, as I think there are other intuitive color map methods that could be named `get`, i.e. "return the underlying palette for the color map.  They point out `first` as an example where it is not named `get_first`, so I think `lookup` can be considered sufficient.  In less functional language `map` would have also been a good name, but that's a terrible name in rust due to the pervasive use of `map` for functional-style programming.

HeroicKatora: Let's leave it as is then, those are subtle differences that shouldn't be surprises.

Issue #669, opened 2017-07-16
TGA encoder


lukors:
There is a TGA decoder in the library, but no TGA encoder. From my experience, TGA is a pretty common format in game development, and it's also relatively simple. So should be a good time investment.

Josh015:
Is anyone else looking into this? If not, I might take a stab at it over my extended vacation next week.

HeroicKatora:
Feel free to go ahead :+1:

PR #1270, opened 2020-06-28
Added a TgaEncoder

Josh015:

This code implements #669.


src/tga/encoder.rs
```diff
@@ -0,0 +1,125 @@
...
+            }
+            _ => {}
+        }
+
+        self.writer.write_all(&image)?;
+        Ok(())
+    }
+
+    fn write_dimension(&mut self, value: u32) -> io::Result<()> {
+        if value < 1 || value > MAX_IMAGE_DIMENSION {
```
HeroicKatora, line 102: This could be checked before writing any data at all to avoid writing partial data. The violation should produce an `ImageError::EncodingError`, so checking it early would mean the writing can be left at a `io::Result` as well.

src/tga/encoder.rs
```diff
@@ -0,0 +1,125 @@
...
+        self.writer.write_u8(num_alpha_bits + other_channel_bits)?; // Bits per pixel.
+        self.writer
+            .write_u8((num_alpha_bits & ALPHA_BIT_MASK) | SCREEN_ORIGIN_BIT_MASK)?; // Upper left origin.
+
+        // Write out Bgr(a)8 or L(a)8 image data.
+        let mut image = Vec::from(buf);
+
+        match color_type {
+            ColorType::Rgb8 | ColorType::Rgba8 => {
+                for chunk in image.chunks_mut(color_type.bytes_per_pixel() as usize) {
```
HeroicKatora, line 90: Should be `usize::from(color_type.bytes_per_pixel())`, since `as ` casts are more brittle than using the exact conversion.

HeroicKatora:
>  The error handling/representation could use a bit of work, which is–admittedly—true for many of the existing implementations as well :)

Regarding this, I'm aware that we're probably not very clear about this for new contributors. Where did you look for error handling and how could it have been made clear from the start? I'd like to address this discoverability issue but I fear that adding it just where I feel like would be somewhat hard considering my own curse of knowledge in that regard.

Josh015:
I'm embarrassed to admit this but I just kinda looked at all the other encoders and pulled bits and pieces from them on the unscientific metric of "it looks right". Also, I'm relatively new to Rust development, so I'm sorta using this as a learning experience. If you want, I'd be happy to revise my simple TgaEncoder until we have a solid example of best practices.

HeroicKatora:
Nothing to be embarrased about, that's both what I considered most likely and what usually gives reasonable results given that consistent code style is generally considered good. So, it may then help to sprinkle in `TODO` comments in non-idiomatic implementations, with a note to change them and referring to good examples. I'll address this soon. If there is anything that is unclear about best practices here, feel free to ask/ping me.

src/tga/encoder.rs
Josh015: I'm a little unclear on this one. I believe you're saying you want me to make something like a "check_dimension(value: u32) -> ImageResult" function and move the calls up to encode(), and you want me to refactor it to return something like this:
ImageError::Encoding(EncodingError::new(
            ImageFormat::Tga.into(),
            err,
        ));

So what kind of value will I be putting in "err"?

HeroicKatora: That is what I intended to say, yes. First you would introduce a new private error enum with one variant for this kind of error. You could split it into two variants, one for width and height each, or handle both in the same variant. It would contain slightly more details to do two variants.

```rust
enum EncoderError {
    InvalidDimension(u32),
}
```

Then implement `fmt::Display` and `std::error::Error` for it. The displayed message would be what is currently being formatted. This allows you to pass an error of that type to `EncodingError::new`. The generic parameter and trait bound can be a little confusing at first (And for some reason the [`From` impl](https://doc.rust-lang.org/stable/src/std/error.rs.html#163) that enables this does currently not appear in the documentation of `std`. That's certainly not discoverable.).

Josh015: I did a little digging in the other encoders and found this pattern in a couple of them:
```rust
            u16::try_from(image.width()).map_err(|_| {
                ImageError::Parameter(ParameterError::from_kind(
                    ParameterErrorKind::DimensionMismatch,
                ))
            })?,
```
Are you sure I shouldn't be using something like this?

HeroicKatora: It's more of a guideline, that's true. I imagine it's different in some places because previously that error was previously a top-level variant of `ImageError` and reworking it was a significant amount of work. The intended use case is:
* `EncodingError`: The format doesn't support some parameters.
* `UnsupportedError`: We don't support some parameters, but the format would.
* `ParameterError`: Everything not covered by the above. For example, encoding a bitmap only color values 0 and 255 must be provided.
Anyways, that's the intention. If you see any that you find differ from this explanation you can open an issue or add a TODO :)

Josh015: Okay, let me push my recent round of updates so you can take a look at that while I make those changes.

src/tga/encoder.rs
```diff
@@ -123,3 +77,141 @@ impl<W: Write> ImageEncoder for TgaEncoder<W> {
...
+        let mut buf = vec![0; decoder.total_bytes() as usize];
+        decoder.read_image(&mut buf).expect("failed to decode");
+        buf
+    }
+
+    // FIXME: Taking a really long time to run. Tried switching to Bgr8, but then it just panics.
+    // #[test]
+    // fn huge_files_return_error() {
+    //     let mut encoded_data = Vec::new();
+    //     let image = vec![0u8; 3 * 40_000 * 40_000]; // 40_000x40_000 pixels, 3 bytes per pixel, allocated on the heap
```
HeroicKatora, line 141: Some other tests for dimension errors test them individually. So you'd check that `1×65536` errors and then that `65536×1` errors. This avoids the extremely large single allocation.

src/tga/encoder.rs
```diff
@@ -123,3 +77,141 @@ impl<W: Write> ImageEncoder for TgaEncoder<W> {
...
+    //     let encoder = TgaEncoder::new(&mut encoded_data);
+    //     let result = encoder.encode(&image, 40_000, 40_000, ColorType::Rgb8);
+    //     assert!(result.is_err());
+    // }
+
+    #[test]
+    fn round_trip_single_pixel_rgb() {
+        let image = [0, 1, 2];
+        let decoded = round_trip_image(&image, 1, 1, ColorType::Rgb8);
+        assert_eq!(3, decoded.len());
```
HeroicKatora, line 151: You could compare it to a slice directly.
```suggestion
        assert_eq!(decoded.as_slice(), [0, 1, 2]);
```

src/tga/encoder.rs
Josh015: Yeah, I came to that conclusion as well. As for the slowness, I've determined that it's the byte swap step that occurs when encoding an Rgb8 or Rgba8 image. Is there a faster way to do that operation?

HeroicKatora: I can't say for sure. The binary produced isn't exactly optimal but I can't find an easy way to produce something more optimal quickly either.

Josh015:
Okay, so this most recent check in takes the encoder in a different direction. I decided that since I was basically constructing a header to write out, why not make it an actual constructor and share the Header struct between the decoder and encoder. I also realized a NoImageData tga is technically valid, so I added support for that.

Thoughts?

src/tga/header.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub(crate) y_origin: u16,      // y-origin of image
+    pub(crate) image_width: u16,   // width of image
+    pub(crate) image_height: u16,  // height of image
+    pub(crate) pixel_depth: u8,    // bits per pixel
+    pub(crate) image_desc: u8,     // image descriptor
+}
+
+impl Header {
+    /// Create a header with all values set to zero.
+    pub(crate) fn new() -> Header {
```
HeroicKatora, line 90: If we're touching this anyways then you can simply derive `Default` here instead.

src/tga/header.rs
```diff
@@ -0,0 +1,169 @@
...
+            map_length: r.read_u16::<LittleEndian>()?,
+            map_entry_size: r.read_u8()?,
+            x_origin: r.read_u16::<LittleEndian>()?,
+            y_origin: r.read_u16::<LittleEndian>()?,
+            image_width: r.read_u16::<LittleEndian>()?,
+            image_height: r.read_u16::<LittleEndian>()?,
+            pixel_depth: r.read_u8()?,
+            image_desc: r.read_u8()?,
+        })
+    }
```
HeroicKatora, line 123: You may want to move writing into a method here as well for consistency.

src/tga/encoder.rs
Josh015: Sorry, totally forgot about this part. I believe I was able to implement this the way you wanted, but I'll let you be the judge. Though it broke the ```test_image_too_large()``` function and I'm not familiar enough with Rust errors to fix it. Any useful insights about what to do there, or should I just leave it for now?

HeroicKatora: Implemention looks indeed good. Let me get that test for you :)

README.md
```diff
@@ -31,7 +31,7 @@ https://docs.rs/image
 | WebP   | Lossy(Luma channel only) | No |
 | PNM    | PBM, PGM, PPM, standard PAM | Yes |
 | DDS    | DXT1, DXT3, DXT5 | No |
-| TGA    | Yes | No |
+| TGA    | Yes | Yes |
```
Josh015, line 34: I had a question on this part. Since I don't support all of the input ColorType enums, should I explicitly say which ones are supported (ie. ```RGB(8), RGBA(8), Gray(8), GrayA(8)```)

Josh015:
One other question. Given how much encoding-specific vs decoding-specific code is mixed into the ```header.rs``` file I'm debating just compacting that plus ```encoder.rs``` and ```decoder.rs``` down into a top-level ```tga.rs``` file. 

Any thoughts?

HeroicKatora:
The structure seems fine, I wouldn't change it. I find splitting modules into encoder and decoder slightly easier to navigate but the difference isn't large. The decoder has a few more imports as well at the moment.

README.md
HeroicKatora: That wouldn't hurt, please do.

Josh015:
As near as I can tell, that's the last of it. Please let me know if there's anything else that needs to be addressed.

Josh015:
Actually, I think I want to make one final change. I won't combine all the files, but I don't like how much encoder-specific and decoder-specific code is living in ```header.rs```. Is it okay if I move those methods and errors into ```decoder.rs``` and ```encoder.rs``` where appropriate?

HeroicKatora:
> Is it okay if I move those methods and errors into decoder.rs and encoder.rs where appropriate?

If you move them like written and keep them methods I don't see why not.

Josh015:
Okay, the code is in. I have no other plans for it unless you have feedback.

Josh015:
Scratch that, I think I have a compromise idea. What if I do the dimension validation in the ```encoder.rs```, so that the EncoderError code can stay there? The rest of the Header code can then live in ```header.rs" and I won't have any problem with it.

What say you?

HeroicKatora:
I've sent you an invite to the collaborators team. If you accept, this allows you to edit labels, merge PRs, push to branches as well as *create new repositories* under the `image-rs` org and comes without any responsibilities other than your observance of the [Code of Conduct](https://github.com/image-rs/organization/blob/master/CODE_OF_CONDUCT.md).

Josh015:
I'm honored! I'll try to contribute where I can, but I can't guarantee that I'll be super active in development. My forte is mostly web development with Typescript and C#, and I was only able to move quickly on this one because of past work with game development in C++. 

First order of business, I'll close #669 and add it to the version 0.2.3 project.

Issue #1271, opened 2020-06-30
DXT -> BCN?


Josh015:
I was just wondering if it would be a good idea to switch from ```Dxt*``` naming scheme to ```Bcn*``` in order to match Microsoft? Also, are there any plans to support ```BC4``` and ```BC5``` since ```L8``` and ```La8``` would map well to them?

fintelia:
My preference would be to create a new `image-bcn` crate and move all the dxt encoding/decoding to there. We could then deprecate and eventually remove all the `Dxt*` code from `image` and retain only support for the DDS format.

I don't personally have plans to add support for other block compression formats, but PRs are welcome!

HeroicKatora:
Starting the revision from a clean slate in a new crate sounds preferrable for me as well. Note it's possible retro-fit the current implementation to instead use a crate. Basically only implementation that we (sadly) can't readily change is the `DXTVariant` enum as it wasn't defined to be non-exhaustive (quite an oversight tbh). But we could add a _new_ enum instead and provide a converter from the old. That should leave many, if not all, interfaces intact. The DdsEncoder is entirely opaque as well. Also, we can defined _type_ aliases if we truly want to change the names without any major breakage (sadly we can't do the same for enum variants).

A new crate could expose many more internals, as long as it's not a public dependency of `image` the need for a stable interface isn't as large. It could experiment a lot more as well. At some point when format support gets detailed enough, we move them out into separate crates. For example, this happened with `png`, `gif`, `tiff`.

Waelwindows:
I think it's better to map BC5 to R8G8, because not all textures use the "green" channel as alpha. e.g. normal maps

fintelia:
@Waelwindows thanks for catching that! It probably isn't worth adding R8G8 as a format just to handle BC5, given that for decoder output La8 actually just means "a two channel format where even bytes are one channel and odd bytes the other". But you are right that the documentation should say that clearly somewhere so that users don't get confused

Issue #1217, opened 2020-05-13
YCbCr (YUYV)


raymanfx:
I am currently writing a cross-platform image streaming I/O library mainly focused on webcam capture. Most webcams do not provide RGB buffers, but almost all of the cheap ones usually provide YUYV. While I could write my own YUYV -> RGB routines and handle conversion in the library, I wanted to ask how you feel about supporting YUV family formats in this crate.

## Draft

8 and 16 bit YUV (single-planar aka interleaved) support would look just like the existing `Rgb` struct:
`pub struct Yuy2<T: Primitive>(pub [T; 4]);`

Notes:
 * YUYV == YUY2
 * 4 components describe a macropixel, which is two image pixels

I think it would make sense to focus on interleaved formats in the first step and add planar formats later on.

tylerhawkes:
I would love to have YUV support. I currently get frame buffers from a camera in the yuv_nv12 format and having support for them here would allow me to stop depending on opencv all together.

raymanfx:
I evaluated adding YUV support to this crate, but found that its internal architecture is too inflexible. Adding a new format requires touching many internal structures (cannot be done in an external crate) and there's no multi-planar API at all afaik.

I started developing a new crate for foreign function image conversion: https://github.com/raymanfx/ffimage. The design is flexible and I have a second subcrate to handle YUV formats explicitly (not on GitHub right now, but I can push it). Conversion is parallelized by Rayon, so it's quite efficient.

I was focussing on the packed API for now, but if you'd like to help, we can flesh out the multi-planar API and add support for your usecase.

HeroicKatora:
Indeed, the design of `ImageBuffer` is quite dated and large portions of it haven't been changed motivated by compatibility and effort. The most glaring issues are that it hadn't been able to consider safe byte casting when it was first implemented and retro-fitting is (as you've observed) almost impossible.

[...]

PR #1272, opened 2020-07-02
Add BCn and YUV444/YUYV formats to ExtendedColorType

fintelia:
This patch adds a couple of extended color types where a single "macro pixel" covers multiple pixels in the actual image.

cc: #1271 #1217

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -136,6 +136,22 @@ pub enum ExtendedColorType {
     /// Pixel is 8-bit BGR with an alpha channel
     Bgra8,
 
+    /// BLock compression format BC1
+    Bc1,
+    /// BLock compression format BC2
+    Bc2,
+    /// BLock compression format BC3
+    Bc3,
+    /// BLock compression format BC4
+    Bc4,
+    /// BLock compression format BC5
+    Bc5,
+
+    /// Packed Y, Cb and Cr channels, each 8-bits.
+    Yuv444,
+    /// YUYV format, also known as YUV4:2:2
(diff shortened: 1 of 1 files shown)


src/color.rs
```diff
@@ -157,27 +173,50 @@ impl ExtendedColorType {
...
+            ExtendedColorType::Yuyv => 3,
             ExtendedColorType::Rgba1 |
             ExtendedColorType::Rgba2 |
             ExtendedColorType::Rgba4 |
             ExtendedColorType::Rgba8 |
             ExtendedColorType::Rgba16 |
-            ExtendedColorType::Bgra8 => 4,
+            ExtendedColorType::Bgra8 |
+            ExtendedColorType::Bc2 |
+            ExtendedColorType::Bc3 => 4,
```
HeroicKatora, line 200: I'm not too sure about these values. Yes, there are 4 channels if seen as single values but wasn't this supposed to be used for finding the length of a slice of samples? In a macroblock there is only one sample, the block itself, since it is indivisible without converting it.

fintelia: Yeah, I wasn't sure either. We should probably add a comment explaining what "channels" actually means here.

HeroicKatora: We defined it to agree with `Pixel::CHANNEL_COUNT` and that is used to select a range of same length from a slice of channels. Since we can't subdivide the macro pixel, the most consistent option in my opinion according to that usage would be to define that the 'channel' in that type is the macro pixel itself or at least treat it that way.

PR #1279, opened 2020-07-12
Update meta data and changes for 0.23.7

HeroicKatora:


fintelia:
This looks good to me. Lets merge #1280 first, but otherwise I think this is good to go

PR #1162, opened 2020-03-15
Format code using 'cargo fmt'

Atul9:
License: I license past and future contributions under the dual MIT/Apache-2.0 license,
allowing licensees to chose either at their option.

PR #1284, opened 2020-07-19
imageops: introduce tile function

MicroJoe:

Example usage:

```rust
pub fn example_tile() {
    let mut img = RgbaImage::new(300, 170);
    let background_color = Rgba::from_slice(&[20, 20, 20, 255]);
    for p in img.pixels_mut() {
        *p = *background_color;
    }

    let tile = image::open("spinner.png").unwrap();

    image::imageops::tile(&mut img, &tile);
    img.save("test.png").unwrap();
}
```

Example output image:

test

### DIFF
--- a/src/imageops/mod.rs
+++ b/src/imageops/mod.rs
@@ -164,6 +164,32 @@ where
     }
 }
 
+/// Tile an image by repeating it multiple times
+///
+/// # Examples
+/// ```no_run
+/// use image::{RgbaImage};
+///
+/// fn main() {
+///      let mut img = RgbaImage::new(1920, 1080);
+///      let tile = image::open("tile.png").unwrap();
+///
+///      image::imageops::tile(&mut img, &tile);
+///      img.save("tiled_wallpaper.png").unwrap();
+/// }
+/// ```
+pub fn tile<I, J>(bottom: &mut I, top: &J)
+where
+    I: GenericImage,
+    J: GenericImageView<Pixel = I::Pixel>,
+{
+    for x in (0..bottom.width()).step_by(top.width() as usize) {
+        for y in (0..bottom.height()).step_by(top.height() as usize) {
+            overlay(bottom, top, x, y);
+        }
+    }
+}
+
 /// Replace the contents of an image at a given coordinate (x, y)
 pub fn replace<I, J>(bottom: &mut I, top: &J, x: u32, y: u32)
 where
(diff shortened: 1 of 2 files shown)


src/imageops/mod.rs
```diff
@@ -164,6 +164,21 @@ where
     }
 }
 
+/// Tile an image by repeating it multiple times
```
HeroicKatora, line 167: Could you add the example here? Marked with `no_run` since it tries to load from disk.

src/imageops/mod.rs
```diff
@@ -164,6 +164,21 @@ where
...
+pub fn tile<I, J>(bottom: &mut I, top: &J)
+where
+    I: GenericImage,
+    J: GenericImageView<Pixel = I::Pixel>,
+{
+    for i in 0..(bottom.width() / top.width() + 1) {
+        let x = i * top.width();
```
HeroicKatora, line 174: If both images have the same width then this loop runs twice, the second time not adding any pixels. You can use the iterator combinators to avoid this off-by-one:

```suggestion
    for x in (0..bottom.width()).step_by(top.width()) {
```

HeroicKatora:
@MicroJoe Do you you'll get to addressing the remarks this weekend? No pressure, I'm just trying to plan the potential `0.23.8` release.

MicroJoe:
> @MicroJoe Do you think you'll get to addressing the remarks this weekend? No pressure, I'm just trying to plan the potential `0.23.8` release.

Thanks for the reminder, applied proposed changes as well as added an example. Please let me know if anything else is missing.

By the way, my setup did want to reformat the whole `mod.rs` using `rustfmt` as it looks like this file was making rustfmt unhappy. Do you know if there is any plan to force apply rustfmt on the whole project?

HeroicKatora:
Thank you!

HeroicKatora:
> By the way, my setup did want to reformat the whole mod.rs using rustfmt as it looks like this file was making rustfmt unhappy. Do you know if there is any plan to force apply rustfmt on the whole project?

Certainly not 'the default' settings. It is quite opinionated and that in many regards can decrease legibility of code. [See here for some more](https://github.com/image-rs/image/pull/1162#pullrequestreview-374794859). That said, it does come up often enough that new contributors have auto-formatting on save enabled and then it can be harder to contribute. If there is some way to _selectively_ enable only small parts of the formatting (and disable the rest) through a configuration file, I'd certainly be interested.

PR #1289, opened 2020-07-24
imageops: Add `?Sized` bound to generic type `Map`.

wathiede:
This allows unsized types like those returned by `Box<dyn ColorMap<_>>>.as_ref()` to be passed into these functions.

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -318,7 +318,7 @@ macro_rules! do_dithering(
 /// Floyd-Steinberg dithering to improve the visual conception
 pub fn dither<Pix, Map>(image: &mut ImageBuffer<Pix, Vec<u8>>, color_map: &Map)
 where
-    Map: ColorMap<Color = Pix>,
+    Map: ColorMap<Color = Pix> + ?Sized,
     Pix: Pixel<Subpixel = u8> + 'static,
 {
     let (width, height) = image.dimensions();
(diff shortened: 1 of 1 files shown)


src/imageops/colorops.rs
```diff
@@ -318,7 +318,7 @@ macro_rules! do_dithering(
 /// Floyd-Steinberg dithering to improve the visual conception
 pub fn dither<Pix, Map>(image: &mut ImageBuffer<Pix, Vec<u8>>, color_map: &Map)
 where
-    Map: ColorMap<Color = Pix>,
+    Map: ColorMap<Color = Pix> + ?Sized,
```
HeroicKatora, line 321: So technically this might lead to _worse_ type inference. I don't consider this a problem, this is definitely a change for the better. The [evolution rfc](https://github.com/aturon/rfcs/blob/api-evolution/text/0000-api-evolution.md#signatures-in-functions) allows similar and more sweeping changes. Maybe I'll run [`rustwide`](https://github.com/rust-lang/rustwide) with some downstream crates before merging.

wathiede: I can't find much documentation on rustwide.  Are you going to author a program like https://github.com/rust-lang/rustwide/blob/master/examples/docs-builder.rs and run it against everything listed at https://crates.io/crates/image/reverse_dependencies?

HeroicKatora: Something like that, yes.

HeroicKatora: I should say, it's just an idea and an experiment at the same time to see if this is even feasible. Might want to add it to the release process in case it's easy.

HeroicKatora:
So I did all reverse dependencies down to `graphics_buffer` without problems. Code will come in time :)

PR #1290, opened 2020-07-24
describe the pixel enumeration order in the doc comment to avoid confusion

sassman:
- main change is the doc string for the enumerator, because I always get confused about the iteration order is it first rows then columns or the other way around. Maybe that helps other for confused people like me.
- the other stuff is just apply cargo fmt

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -535,6 +535,7 @@ where
     }
 
     /// Returns an iterator over the pixels of this image.
+    /// The iteration order is x = 0 to width then y = 0 to height
     pub fn pixels(&self) -> Pixels<P> {
         Pixels {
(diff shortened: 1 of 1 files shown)


sassman:
> Remove the fmt commit, it makes review near impossible.

Are ok with separating formatting and change so that there are 2 commits or generally not interested in the reformatting?

HeroicKatora:
Don't provide the formatting in this PR. We had several discussion on it and it's far too large to do likely in a single PR and without _any_ consideration for refining some of the `rustfmt.toml` settings.

sassman:
@HeroicKatora here you go, it's very tiny and not much.

HeroicKatora:
Thanks :slightly_smiling_face:

PR #1295, opened 2020-07-26
Fixed Inconsistent capitalization of Encoder/Decoder structs.

### LINKED ISSUE #1258: Inconsistent capitalization of PNG
https://github.com/image-rs/image/blob/ff6c4d9de05bf71385ca043ab8d67ba66760b70f/src/png.rs

Publicly exposed types `PNGReader`, `PngDecoder`, `PNGEncoder` have inconsistent capitalization.

Changelog entry 0.23.0 suggests that the preferred form in this crate would be `Png`.

ZackJorquera:
fixes #1258 

I created a type definition for each previous struct name and marked them as deprecated. In order to export these typedefs, I needed to use `#[allow(deprecated)]` in some of the `mod.rs` files. Instead, typedef definitions could be in the `mod.rs` files to not have to do this. Or they could be not marked as deprecated. Although, I think the `#[allow(deprecated)]` tags are fine given that the typedefs will be removed in the next version (I assume). 

All the names that were changed:
- `PNGReader` -> `PngReader`
- `PNGEncoder` -> `PngEncoder`
- `BMPEncoder` -> `BmpEncoder`
- `DXTReader` -> `PngReader`
- `DXTEncoder` -> `DxtEncoder`
- `DXTVariant` -> `DxtVariant`
- `gif::Encoder` -> `gif::GifEncoder`
- `ICOEncoder` -> `IcoEncoder`
- `HDRAdapter` -> `HdrAdapter`
- `HDREncoder` -> `HdrEncoder`
- `HDRImageDecoderIterator` -> `HdrImageDecoderIterator`
- `HdrMetadata` -> `HdrMetadata`
- `RGBE8Pixel` -> `Rgbe8Pixel`
- `JPEGEncoder` -> `JpegEncoder`
- `PNMEncoder` -> `PnmEncoder`
- `PNMHeader` -> `PnmHeader`
- `PnmSubtype` -> `PnmSubtype`

### DIFF
--- a/src/bmp/mod.rs
+++ b/src/bmp/mod.rs
@@ -8,16 +8,8 @@
 //!
 
 pub use self::decoder::BmpDecoder;
-pub use self::encoder::BmpEncoder;
-
-/// BMP Encoder
-///
-/// An alias of [`BmpEncoder`].
-///
-/// [`BmpEncoder`]: struct.BmpEncoder.html
-#[allow(dead_code)]
-#[deprecated(note = "Use `BmpEncoder` instead")]
-pub type BMPEncoder<'a, W> = BmpEncoder<'a, W>;
+#[allow(deprecated)]
+pub use self::encoder::{BmpEncoder, BMPEncoder};
 
 mod decoder;
 mod encoder;
--- a/src/bmp/mod.rs
+++ b/src/bmp/mod.rs
@@ -8,7 +8,7 @@
 //!
 
 pub use self::decoder::BmpDecoder;
-#[allow(deprecated)]
+#[allow(deprecated)] // TODO: when `BMPEncoder` is removed, remove this flag
 pub use self::encoder::{BmpEncoder, BMPEncoder};
 
 mod decoder;
--- a/src/bmp/mod.rs
+++ b/src/bmp/mod.rs
@@ -8,7 +8,16 @@
 //!
 
 pub use self::decoder::BmpDecoder;
-pub use self::encoder::BMPEncoder;
+pub use self::encoder::BmpEncoder;
+
+/// BMP Encoder
+///
+/// An alias of [`BmpEncoder`].
+///
+/// [`BmpEncoder`]: struct.BmpEncoder.html
+#[allow(dead_code)]
+#[deprecated(note = "Use `BmpEncoder` instead")]
+pub type BMPEncoder<'a, W> = BmpEncoder<'a, W>;
 
 mod decoder;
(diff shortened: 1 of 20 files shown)


src/bmp/mod.rs
```diff
@@ -8,7 +8,8 @@
 //!
 
 pub use self::decoder::BmpDecoder;
-pub use self::encoder::BMPEncoder;
+#[allow(deprecated)]
+pub use self::encoder::{BmpEncoder, BMPEncoder};
```
HeroicKatora, line 12: Since you commented on this in the description, I think it is perfectly fine to have the attribute ourselves currently.

src/dxt.rs
```diff
@@ -18,7 +18,7 @@ use crate::image::{self, ImageDecoder, ImageDecoderExt, ImageReadBuffer, Progres
 /// Note that DXT2 and DXT4 are left away as they're
 /// just DXT3 and DXT5 with premultiplied alpha
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum DXTVariant {
+pub enum DxtVariant {
```
HeroicKatora, line 21: So have missed something somewhat important in the description of the issue, and sadly did not check this sooner. The ability to use enum variants through type aliases got introduced [in `1.37`](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1370-2019-08-15). Since our MSRV policy is `1.34.2` support this would technically be a breaking change to rename enums. I think we sadly have to keep the orginal names for now for enums, and the alias can have the new capitalization. Clarification: The struct types should be fine.

ZackJorquera: I completely forgot to change my rustc version, sorry about that.

I can change that.

ZackJorquera: I also changed `PnmSubtype`. I can fix that too.

HeroicKatora: Please do, yes. And add a `TODO` comment referencing either the fact or the comment.

src/pnm/header.rs
```diff
@@ -11,8 +11,14 @@ pub enum SampleEncoding {
 }
 
 /// Denotes the category of the magic number
+///
+/// DEPRECATED: The name of this enum will be changed to [`PnmSubtype`].
```
ZackJorquera, line 15: I opted to use a comment to say that this was deprecated as it does not require any `#[allow(deprecated)]` tags in the rest of the code.

PR #1296, opened 2020-07-29
Fix clippy warnings

TaKO8Ki:
Good day!

I've fixed some clippy warnings. There are still three warnings like the following.

clippy 0.0.21

```sh
warning: All variants have the same postfix: `Invalid`
  --> src/dds.rs:22:1
   |
22 | / enum DecoderError {
23 | |     /// Wrong DDS channel width
24 | |     PixelFormatSizeInvalid(u32),
25 | |     /// Wrong DDS header size
...  |
31 | |     DdsSignatureInvalid,
32 | | }
   | |_^
   |
   = note: `#[warn(clippy::enum_variant_names)]` on by default
   = help: remove the postfixes and use full paths to the variants instead of glob imports
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names

warning: needless `fn main` in doctest
   --> src/imageops/mod.rs:171:4
    |
171 | /// use image::{RgbaImage};
    |    ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(clippy::needless_doctest_main)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_doctest_main

warning: called `.as_ref().map(String::as_str)` on an Option value. This can be done more directly by calling `ext.as_deref()` instead
   --> src/io/free_functions.rs:234:15
    |
234 |       let ext = ext.as_ref()
    |  _______________^
235 | |         .map(String::as_str);
    | |____________________________^ help: try using as_deref instead: `ext.as_deref()`
    |
    = note: `#[warn(clippy::option_as_ref_deref)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_as_ref_deref

warning: 3 warnings emitted
```

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -619,7 +619,11 @@ pub trait GenericImageView {
     /// Returns the pixel located at (x, y)
     ///
     /// This function can be implemented in a way that ignores bounds checking.
-    #[allow(clippy::missing_safety_doc)]
+    /// # Safety
+    ///
+    /// The coordinates must be [`in_bounds`] of the image.
+    ///
+    /// [`in_bounds`]: #method.in_bounds
     unsafe fn unsafe_get_pixel(&self, x: u32, y: u32) -> Self::Pixel {
         self.get_pixel(x, y)
     }
--- a/src/image.rs
+++ b/src/image.rs
@@ -619,6 +619,7 @@ pub trait GenericImageView {
     /// Returns the pixel located at (x, y)
     ///
     /// This function can be implemented in a way that ignores bounds checking.
+    #[allow(clippy::missing_safety_doc)]
     unsafe fn unsafe_get_pixel(&self, x: u32, y: u32) -> Self::Pixel {
         self.get_pixel(x, y)
     }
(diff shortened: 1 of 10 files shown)


src/image.rs
```diff
@@ -619,6 +619,7 @@ pub trait GenericImageView {
     /// Returns the pixel located at (x, y)
     ///
     /// This function can be implemented in a way that ignores bounds checking.
+    #[allow(clippy::missing_safety_doc)]
```
HeroicKatora, line 622: These ones just make us forget that a safety doc should actually be added.

```suggestion
    /// # Safety
    ///
    /// The coordinates must be [`in_bounds`] of the image.
    ///
    /// [`in_bounds`]: #method.in_bounds
```

src/image.rs
```diff
@@ -672,6 +673,7 @@ pub trait GenericImage: GenericImageView {
     /// Puts a pixel at location (x, y)
     ///
     /// This function can be implemented in a way that ignores bounds checking.
+    #[allow(clippy::missing_safety_doc)]
```
HeroicKatora, line 676:
```suggestion
    /// # Safety
    ///
    /// The coordinates must be [`in_bounds`] of the image.
    ///
    /// [`in_bounds`]: traits.GenericImageView.html#method.in_bounds
```

HeroicKatora, line 676: Shouldn't similar warnings appear in `ImageBuffer`? If so then please add the safety sections there as well.

src/image.rs
TaKO8Ki: Sure, I've accepted your commit suggestion.

src/image.rs
TaKO8Ki: I've the added safety section as well!!

PR #1308, opened 2020-08-25
Update README.rs

SuperSamus:
GenericImage was split into two since commit 85f2a7d but Readme wasn't updated in section 4.1, so I did it.
However this section is now much longer. I trimmed some comments, removed all the unsafe functions and all default implementations, but in my opinion it's still too long. Feel free to trim it even more before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Maintainers: @HeroicKatora, @fintelia
 
 This crate provides basic imaging processing functions and methods for converting to and from image formats.
 
-All image processing functions provided operate on types that implement the `GenericImage` trait and return an `ImageBuffer`.
+All image processing functions provided operate on types that implement the `GenericImageView` and `GenericImage` traits and return an `ImageBuffer`.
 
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Thanks for keeping it updated. With regards to the section being too long, I actually thought of moving it to the traits. Then these sections would be an overview of the functionality (good call on removing the `unsafe` code I would say) and only keeping links to those documentation pages in the Readme. I'm curious what you think of that?

SuperSamus:
You mean something like section 2.2? Looks good to me. (Then for consistency the other sections should have their links to the documentation)

HeroicKatora:
> You mean something like section 2.2?

Precisely. And if you'd add the link to the other section I would also be glad if it doesn't bother you.

SuperSamus:
Done. I renamed the PR because I did some extra stuff.

HeroicKatora (approved):
Thank you! I really appreciate the effort for more consistency.

PR #1311, opened 2020-09-01
Remove deprecated type definitions of encoder/decoder structs

orhun:
Removes (and renames the original struct if necessary) the deprecated type definitions that are created at #1295 due to fixing inconsistent capitalization.

Following type definitions are removed:
* `PNGReader`
* `PNGEncoder`
* `BMPEncoder`
* `DXTReader`
* `DXTEncoder`
* `DXTVariant`
* `gif::Encoder`
* `ICOEncoder`
* `HDRAdapter`
* `HDREncoder`
* `HDRImageDecoderIterator`
* `HDRMetadata`
* `RGBE8Pixel`
* `JPEGEncoder`
* `PNMEncoder`
* `PNMHeader`
* `PNMSubtype`


fintelia:
Thanks for making this PR! I agree that we want to merge this change at some point, but we should probably wait a little. The release that deprecated the old names for those items only was released two days ago... :stuck_out_tongue:

fintelia:
Merged via #1490. Thanks for the contribution!

PR #1313, opened 2020-09-10
Add non-consumable equivalent of into_raw to ImageBuffer

KernelErr0r:
I license past and future contributions under the dual MIT/Apache-2.0 license.

I needed a way in my game engine to share raw image data across multiple iterations of infinite loop (because obviously I don't want to load it every time or copy it) so I added one.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -646,6 +646,11 @@ where
         self.data
     }
 
+    /// Returns the underlying raw buffer
+    pub fn as_raw(&self) -> &Container {
+        &self.data
+    }
+
     /// The width and height of this image.
     pub fn dimensions(&self) -> (u32, u32) {
         (self.width, self.height)


HeroicKatora (approved):

HeroicKatora:
Thanks. That's pretty interesting how seemingly barely anyone noticed this wasn't a thing before :smile:

fintelia:
I'm not opposed to merging this, but do realize that `as_raw` provides very similar functionality to the existing Deref implementation. (For the common case of ImageBuffer<Vec<T>>, the difference would be that one returns a `&Vec<T>` and the other a `&[T]`).

KernelErr0r:
Oh, I didn't actually realise it has deref implemented that does the same thing because I just saw that there isn't anything similar to into_raw that doesn't consume, anyway, I think it might be useful to add as_raw to make API a little bit more consistent **_and friendlier to blind people like me_**

HeroicKatora:
Yep, I also find this much more obvious. If it's purely about getting a raw image buffer then another tip would be to use `as_flat_samples`. The rust guidelines also discourage the use of `Deref` for types with inherent methods in the first place and I do want to slowly transition away from it as it mixes badly with coordinate access as well. Well, carefully. This is still useful for the case of `Buffer = Arc<[T]>` or so, there you can clone the buffer without an allocation but can not do so with Deref alone.

HeroicKatora:
> friendlier to blind people like me

We try to be inclusive where we can :wink:

Issue #1152, opened 2020-02-20
Support AVIF (AV1 Image FIle Format)


jansol:
Spec at https://aomediacodec.github.io/av1-avif/

Since there is already a PR for FLIF support, it probably makes sense to have at least a tracking issue for AVIF as well.

kornelski:
For AVIF encoding, have a look at https://github.com/kornelski/cavif-rs 

It's a 100% pure Rust integration with rav1e.

HeroicKatora:
I tried using the encoder with the bindings and `imagemagick` thinks the produced file is not valid:

```text
$ cargo run --example convert --features=avif -- tests/reference/tiff/testsuite/mandrill.tiff.694c777d.png mandrill.avif
$ file mandrill.avif
mandrill.avif: ISO Media
$ convert -verbose mandrill.avif mandrill.jpeg
convert: Invalid input: No 'hdlr' box `mandrill.avif' @ error/heic.c/IsHeifSuccess/136.
convert: no images defined `mandrill.jpeg' @ error/convert.c/ConvertImageCommand/3285
```

I believe this refers to this statement [in the specification](https://aomediacodec.github.io/av1-avif/#image-sequences).
There are no mentions of `pict` in `avif-serialize`. Are there any reference, roundtrip or self-tests to execute?

kornelski:
It roundtrips with libavif.

est31:
Should this be reopened? There is no decoding support yet.

foresterre:
@kornelski 

I'm not always experiencing success with encoding AVIF in https://github.com/foresterre/sic which uses the latest https://github.com/image-rs/image (at time of writing v0.23.10) using the current implementation. Or, I am not sure whether it works or not, because Firefox seems unable to decode the resulting files (Firefox 81, x64 on Ubuntu):

![image](https://user-images.githubusercontent.com/5955761/95007226-dfd35a80-05fc-11eb-93d4-b94972ce7710.png)

with:

![image](https://user-images.githubusercontent.com/5955761/95006773-579e8680-05f7-11eb-8a03-3bfb519c28a9.png)

If I disable `use-dav1d` I get the same results.

I've made a minimal example using the `image` crate which can be found here: https://github.com/foresterre/avifexample/blob/main/src/main.rs#L24

I used the `wh1616.png` sample file: [avif_samples.zip](https://github.com/image-rs/image/files/5322653/avif_samples.zip) (includes my .avif results).

[...]

PR #1314, opened 2020-09-14
Rough draft including ravif for AVIF encoding

HeroicKatora:
Converts the data to the `Img` buffer and encodes it into an intermediate vector, then writes this into the output writer. It tries to avoid actually allocating for that conversion. It considers that `RGBA8` can be converted from any byte slice and uses the applicable conversion. Any non-rgba8 image is converted using an intermediate buffer. (cc: @fintelia) Some of this code may be used for the conversion feature as well.
However it does not support limits.

Closes: #1152

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,6 @@ env:
     - FEATURES='webp'
     - FEATURES='hdr'
     - FEATURES='farbfeld'
-    - FEATURES='ravif'
 matrix:
   allow_failures:
     - name: "Clippy"
@@ -39,6 +38,9 @@ matrix:
     - name: "Fuzz corpus"
     - rust: nightly
   include:
+    - os: linux
+      rust: stable
+      env: FEATURES='ravif'
     - os: osx
       rust: 1.34.2
     - os: windows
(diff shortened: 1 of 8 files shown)


HeroicKatora:
I have no idea how to test the encoding without depending on the availability of `libavif/avifdec` on the test system. For now, I've validated the images manually and it appears to work.. also note that `ravif` is of course tested itself.

kornelski:
For testing:

- to test just the MP4/HEIF structure, I wrote https://lib.rs/avif-parse It's pure Rust, cut-down fork of Firefox's mp4parse.

- to test pixels decoded from AV1, I wrote https://lib.rs/aom-decode It needs cmake and builds libaom, so it's a relatively heavy dependency. Sadly, there's no pure-Rust AV1 decoder.

HeroicKatora:
@kornelski The pixel decoding would be more interesting, of course then also with full bindings, and the pure structure seems like a duplicated effort. I suppose we just wait here, any updates on the situation are appreciated.

est31:
There is a partial rewrite of dav1d in https://github.com/rainliu/rav1d/ but it yielded an error no matter which image I gave it.

PR #1316, opened 2020-09-20
Update gif crate

HeroicKatora:
Well this is slightly awkward. In an effort to avoid denial-of-service, the `gif` crate now checks each frame descriptor to be inbounds. Apparently this is not quite expected and there's a test [preventing out-of-bound action](https://github.com/image-rs/image/commit/503690269091d47444e9c94ffbd8f394de897b0b) in case one of the descriptors is actually out of bounds. We'll revert this and do another try.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ byteorder = "1.3.2"
 num-iter = "0.1.32"
 num-rational = { version = "0.3", default-features = false }
 num-traits = "0.2.0"
-gif = { version = "0.10.0", optional = true }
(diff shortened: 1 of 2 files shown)


HeroicKatora:
Just clarifiying: the standard does _not_ allow this kind of use but it appears to be okay in practice.

> a. Description. Each image in the Data Stream is composed of an Image
      Descriptor, an optional Local Color Table, and the image data.  **Each
      image must fit within the boundaries of the Logical Screen, as defined
      in the Logical Screen Descriptor.**

PR #1322, opened 2020-10-03
Make functions in jpeg/encoder.rs more idiomatic, clean up todos

philippeitis:
This PR replaces the use of write_u8 and write_all on Vec's to push and extend_from_slice respectively, which resolves a few TODO's and removes the need to do `let _ =` to handle the errors that `write_u8` and `write_all` may throw.

### DIFF
--- a/src/jpeg/encoder.rs
+++ b/src/jpeg/encoder.rs
@@ -761,28 +761,24 @@ fn build_huffman_segment(
     m: &mut Vec<u8>,
     class: u8,
     destination: u8,
-    numcodes: &[u8],
+    numcodes: &[u8; 16],
     values: &[u8],
 ) {
     m.clear();
 
-    // TODO: More idiomatic would be pub, extend_from_slice
     let tcth = (class << 4) | destination;
-    let _ = m.write_u8(tcth);
+    m.push(tcth);
 
-    assert_eq!(numcodes.len(), 16);
+    m.extend_from_slice(numcodes);
 
-    let _ = m.write_all(numcodes);
-
-    let mut sum = 0usize;
-
-    for &i in numcodes.iter() {
-        sum += i as usize;
-    }
+    let sum: usize = numcodes
+                        .iter()
+                        .map(|&x| x as usize)
+                        .sum();
 
     assert_eq!(sum, values.len());
 
-    let _ = m.write_all(values);
+    m.extend_from_slice(values);
 }
 
 fn build_quantization_segment(m: &mut Vec<u8>, precision: u8, identifier: u8, qtable: &[u8]) {
@@ -898,6 +894,13 @@ mod tests {
     use crate::image::ImageDecoder;
 
     use super::{build_jfif_header, JpegEncoder, PixelDensity};
+    use crate::jpeg::encoder::{
+        build_huffman_segment,
+        DCCLASS,
+        LUMADESTINATION,
+        STD_LUMA_DC_CODE_LENGTHS,
+        STD_LUMA_DC_VALUES
+    };
     use super::super::JpegDecoder;
 
     #[cfg(feature = "benchmarks")]
@@ -1020,6 +1023,18 @@ mod tests {
         assert!(decoded[2] < 50, "bad blue channel in {:?}", &decoded);
     }
 
+    #[test]
+    fn bench_build_huffman_segment() {
+        let mut buf = vec![];
+        build_huffman_segment(
+            &mut buf,
+            DCCLASS,
+            LUMADESTINATION,
+            &STD_LUMA_DC_CODE_LENGTHS,
+            &STD_LUMA_DC_VALUES,
+        );
+        assert_eq!(buf, vec![0, 0, 1, 5, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]);
+    }
     #[cfg(feature = "benchmarks")]
     #[bench]
     fn bench_jpeg_encoder_new(b: &mut Bencher) {
@@ -1028,5 +1043,4 @@ mod tests {
             let x = JpegEncoder::new(&mut y);
         })
     }
-
 }
--- a/src/jpeg/encoder.rs
+++ b/src/jpeg/encoder.rs
@@ -330,7 +330,7 @@ pub struct JpegEncoder<'a, W: 'a> {
     writer: BitWriter<'a, W>,
 
     components: Vec<Component>,
-    tables: Vec<u8>,
+    tables: Vec<[u8; 64]>,
 
     luma_dctable: Vec<(u8, u16)>,
     luma_actable: Vec<(u8, u16)>,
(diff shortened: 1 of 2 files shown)


src/jpeg/entropy.rs
```diff
@@ -1,24 +1,26 @@
+#![feature(const_fn)]
```
HeroicKatora, line 1: We're not switching to nightly :confused:

src/jpeg/entropy.rs
```diff
@@ -1,24 +1,26 @@
...
 
     // Annex C.2
     // Figure C.1
     // Generate table of individual code lengths
-    for i in 0u8..16 {
-        j = 0;
-
-        while j < bits[usize::from(i)] {
+    let mut i = 0;
+    while i < 16 {
```
HeroicKatora, line 16: I wouldn't consider this more idiomatic.

src/jpeg/entropy.rs
```diff
@@ -1,24 +1,26 @@
...
+        let mut j = 0;
+        while j < bits[i as usize] {
```
HeroicKatora, line 18: Neither this. Casting with `as` works for more cases than `usize::from`, including some lossy cases. `clippy` also remarks about this.

src/jpeg/encoder.rs
```diff
@@ -736,47 +742,39 @@ fn build_frame_header(
 fn build_scan_header(m: &mut Vec<u8>, components: &[Component]) {
     m.clear();
 
-    // TODO: More idiomatic would be extend_from_slice, to_be_bytes
-    let _ = m.write_all(&[components.len() as u8]);
+    let _ = m.push(components.len() as u8);
```
HeroicKatora, line 745: `let _ = ` should no longer be necessary. I like this change.

src/jpeg/encoder.rs
```diff
@@ -736,47 +742,39 @@ fn build_frame_header(
...
-    let _ = m.write_all(&[0]);
+    m.extend_from_slice(&[0, 63, 0]);
 }
 
 fn build_huffman_segment(
     m: &mut Vec<u8>,
     class: u8,
     destination: u8,
-    numcodes: &[u8],
+    numcodes: &[u8; 16],
```
HeroicKatora, line 760: This is a nice find :+1:

HeroicKatora (commented):
Would you split the simple changes from those requiring `const fn`? We haven't discussed when to bump the MSRV next. It's fine if one of the PRs stays open to be used at a later point :)

src/jpeg/encoder.rs
HeroicKatora: I realize I might have been unclear here. Please remove the `let _ = ` part, keep the test.

src/jpeg/entropy.rs
philippeitis: Fair enough, I've reverted the changes.

src/jpeg/encoder.rs
philippeitis: I missed that, but there should be no more `let _ = ` in encoder.rs.

src/jpeg/encoder.rs
```diff
@@ -775,14 +767,13 @@ fn build_quantization_segment(m: &mut Vec<u8>, precision: u8, identifier: u8, qt
     assert_eq!(qtable.len() % 64, 0);
```
philippeitis, line 767: Shouldn't this be testing that qtable.len >= 64? Otherwise, it'd allow a qtable with len 0.

HeroicKatora: Yeah, that seems correct.

HeroicKatora: I think it could be even more specific and check for equality. It's only ever called from here

https://github.com/image-rs/image/blob/a890363ca2b6850bb9673ea9174d8042216efa0a/src/jpeg/encoder.rs#L532-L535

and that provides `chunks(64)`. That could then also be changed to `chunks_exact(64)` with its own assert?

philippeitis: It looks like tables is always exactly 128 u8's long:

`       
        let mut tables = Vec::new();
        let scale_value = |&v: &u8| {
            let value = (u32::from(v) * scale + 50) / 100;

            clamp(value, 1, u32::from(u8::max_value())) as u8
        };
        tables.extend(STD_LUMA_QTABLE.iter().map(&scale_value));
        tables.extend(STD_CHROMA_QTABLE.iter().map(&scale_value));
`

and I don't know if there's anything that would make it violate this constraint, given that the elements of JpegEncoder are not publically accessible and table is only modified at instantiation - so instead of using chunks_exact, we could just do `tables: [u8; 128]` and take the chunks manually.

HeroicKatora: I wouldn't rely on this here if it's not necessary. This method follows a part of the standard that describes how to encode the 8×8 quantization tables which justifies the length of 64. As an encoder we _can_ choose to only every use the same two tables, so that's fine. However, we do not want to unreasonably be more strict that necessary. It seems somewhat likely that we might want to use different tables in the future? If we introduce more invariants than the 64 byte blocks enshrined in the standard this would be unnecessarily ossified.

HeroicKatora: That said, I wouldn't be opposed to using `Vec<[u8; 64]>` instead of `Vec<u8>`! But it might be tricky and more verbose to apply `scale_value` there.

philippeitis: Not really - if you take a mutable reference and modify the values in place, it's relatively straightforwards:

```
        let mut tables = vec![STD_LUMA_QTABLE.clone(), STD_CHROMA_QTABLE.clone()];
        let scale_value = |v: &mut u8| {
            let value = (u32::from(*v) * scale + 50) / 100;

            *v = clamp(value, 1, u32::from(u8::max_value())) as u8;
        };
        tables[0].iter_mut().for_each(scale_value);
        tables[1].iter_mut().for_each(scale_value);
```

HeroicKatora: Good idea, that should work :+1:

src/jpeg/encoder.rs
```diff
@@ -405,14 +401,14 @@ impl<'a, W: Write> JpegEncoder<'a, W> {
...
+        let scale_value = |v: &mut u8| {
+            let value = (u32::from(*v) * scale + 50) / 100;
 
-            clamp(value, 1, u32::from(u8::max_value())) as u8
+            *v = clamp(value, 1, u32::from(u8::max_value())) as u8;
         };
-        tables.extend(STD_LUMA_QTABLE.iter().map(&scale_value));
-        tables.extend(STD_CHROMA_QTABLE.iter().map(&scale_value));
+        tables[0].iter_mut().for_each(scale_value);
+        tables[1].iter_mut().for_each(scale_value);
```
fintelia, line 411: This block of code could probably be cleaner with a bit of refactoring. Not a huge deal, but might be worth doing if it is being changed anyway

HeroicKatora: What kind of reworking do you mean? Extracting the `scale_value` as a function?

philippeitis:
Benchmarking the current changes with the encoding benchmarks using Criterion shows an average improvement of 6.35%, a minimum improvement of 2.48%, and a maximum of 10.53% on my computer. This improvement is consistent across several sessions.

The majority of the change seems to be from optimizing write_block to directly read values from UNZIGZAG and removing the `if k == 63 {break;}` statement, which wasn't really necessary since the loop exits at k=63 anyways.

src/jpeg/encoder.rs
fintelia: Mostly just removing duplication/making the code more concise. There's lots of possible variations, but one option would be something like:
```rust
let tables = vec![STD_LUMA_QTABLE.clone(), STD_CHROMA_QTABLE.clone()].into_iter().map(|t|t.iter_mut().for_each(|v|{
        *v = clamp((u32::from(*v) * scale + 50) / 100, 1, u32::from(u8::max_value())) as u8;
})).collect();

HeroicKatora: Ahh, collecting the tables in one step instead of having it mutable and changing its contents. Sounds good

philippeitis: I couldn't get it to be just one line without adding cloning, though I did remove the manual tables[0] / tables[1] scaling.  It's not particularly more concise, since maintaining readability also matters.

HeroicKatora (approved):

PR #1324, opened 2020-10-03
Add const implementation for build_huff_lut

philippeitis:
This PR adds a const implementation of derive_codes_and_sizes and build_huff_lut - as these functions are typically used on const data - this speeds up JpegEncoder::new / JpegEncoder::new_with_quality by 4x (2250ns -> 500ns), but this is only really significant if you're making many, many small images and creating new encoders for each.

This isn't currently compatible without using nightly on 1.34, but it does compile without nightly on 1.46, so merging this PR would require a bump in the minimum version required.


HeroicKatora:
Oops, I must have misclicked the merge button on the wrong PR..

HeroicKatora:
Github automatically closed this since all its commits were merged.. Well.. I'll rebase those commits.

Issue #1318, opened 2020-09-30
Content based file type guessing


jerry73204:
The impl of [ImageFormat::from_path](https://docs.rs/image/0.23.10/image/enum.ImageFormat.html#method.from_path) calls an internal method ([code](https://github.com/image-rs/image/blob/6f84d86302e63fdd80899333e5b5aeb6e278354c/src/io/free_functions.rs#L231-L260)) that guesses the file type by filename extension. It could produce false results in case of a PNG file named `xxx.jpg`. The [image::open()](https://docs.rs/image/0.23.10/image/fn.open.html) calls the method and produces error when it guesses the wrong type. Do you consider the content-based method? We already have a [tree\_magic](https://github.com/aahancoc/tree_magic/) crate that does the job.

fintelia:
The [`io::Reader`](https://docs.rs/image/0.23.10/image/io/struct.Reader.html) interface supports guessing the format based on the initial bytes of the file. Is that what you are looking for?

jerry73204:
Yes, I missed it. Sorry for inconvenience.

HeroicKatora:
It appears somewhat common to be classified as a documentation deficiency. Alternatively, I think we could slowly start to deprecate the free functions in the root and nudge people over to `io::Reader`.

Issue #1291, opened 2020-07-25
`load_from_memory*` functions don't work with data returned from `DynamicImage.to_bytes()`


biot023:
The `load_from_memory` and `load_from_memory_with_format` functions are failing with bytes returned from `DynamicImage#to_bytes()`.

There are two tests that show this:

Test one:

```rust
#[test]
fn from_load_from_memory() {
    let img = open("features/fixtures/img01.jpg")
        .expect("File should load");

    let img = load_from_memory(&img.to_bytes())
        .expect("File should load from its own data");

    assert!(true);
}
```

Fails with this:

> ---- from_load_from_memory stdout ----
> thread 'from_load_from_memory' panicked at 'File should load from its own data: Unsupported(UnsupportedError { format: Unknown, kind: Format(Unknown) })', tests/transformers_test.rs:50:15

Test two:

```rust
#[test]
fn from_load_from_memory_with_format() {
    let img = open("features/fixtures/img01.jpg")
        .expect("File should load");

    let img = load_from_memory_with_format(&img.to_bytes(), ImageFormat::Jpeg)
        .expect("File should load from its own data and format hint");

    assert!(true);
}
```

Fails with this:

> ---- from_load_from_memory_with_format stdout ----
> thread 'from_load_from_memory_with_format' panicked at 'File should load from its own data and format hint: Decoding(DecodingError { format: Exact(Jpeg), underlying: Some(Format("first two bytes is not a SOI marker")) })', tests/transformers_test.rs:61:15

As best as I can tell, I'm reading `Vec<u8>` data from a successfully opened `DynamicImage` back into two functions that take a `Vec<u8>` and return a `DynamicImage`. But it's failing.
There's obviously some nuance I'm missing!

If anybody could point me in the right direction to accomplish what I'm trying to achieve, here, it would be *really* appreciated.
Thanks,
   Doug.

HeroicKatora:
The `to_bytes` method does not encode the image, it returns to you the bytes making up the pixels. You shouldn't decode those bytes but you can use `ImageBuffer::from_raw` to restore an image. This requires you to save all the metadata separately. If you don't want to do that manually, i.e. for process-to-process communication or very temporary storage, then you may consider using one of the `pnm` formats which takes care of prepending the the image size and color type to the (basically) raw bytes with without any of the computation overhead (compression, quantization, ..) of jpeg and png.

biot023:
Thanks @HeroicKatora !
What I need to do is:

[...]

Issue #1232, opened 2020-05-28
"Malformed GIF header" error on a valid image


Shnatsel:
This happens in `image` from git, revision 092db1e21b3e8736711e2cbe1d9b5245613756c1

## Expected

The attached image should decode successfully

## Actual behaviour

`Decoding(DecodingError { format: Exact(Gif), underlying: Some(Format("malformed GIF header")) })`

## Reproduction steps

![4235a](https://user-images.githubusercontent.com/291257/83196151-aca72900-a13b-11ea-9c88-45fbfa800f66.gif)

```rust
fn main() -> std::io::Result<()> {
    let path = std::env::args().skip(1).next().unwrap();
    let image = image::io::Reader::open(path)?.guess_format().decode().unwrap();
    Ok(())
}
```

Shnatsel:
Another file triggering the same error:

![nc_69](https://user-images.githubusercontent.com/291257/83201161-34456580-a145-11ea-895d-7ae018a21a6b.gif)

Shnatsel:
Another one

![2006 02 17_VDayEbbyRaed06](https://user-images.githubusercontent.com/291257/83209097-b6d72080-a157-11ea-852c-ad00d8fa67e8.gif)

Shnatsel:
Also a case of wrong extension on the files. Contents-based file type guess works correctly. My apologies.

Issue #1231, opened 2020-05-28
Documentation suggests using guess_format() but it's private


Shnatsel:
This happens in `image` from git, revision 092db1e21b3e8736711e2cbe1d9b5245613756c1

https://docs.rs/image/0.23.4/image/io/struct.Reader.html#method.open suggests using the following code, if I'm reading it right: 
```rust
image::io::Reader::open(path)?.guess_format();
```
but when I try it in image I get an error that the `guess_format()` method is private.

## Reproduction steps

```rust
fn main() -> std::io::Result<()> {
    let path = std::env::args().skip(1).next().unwrap();
    let image = image::io::Reader::open(path)?.guess_format().decode().unwrap();
    Ok(())
}
```

HeroicKatora:
The link and text should be `with_guessed_format`, the exposed method of the builder struct.

PR #1327, opened 2020-10-05
Deprecate compound input fn in favor of Reader

197g:
Any comments on the questions below are welcome.

In light of #1318 and previously #1291 , #1232 , #1231 , #1072 , #792 (possibly more), I conclude that opening files with the functions in the main module is at least misunderstood but quite possibly a bad interface in itself. It's often not clear how to do combine aspects of these methods by performing parts of them manually—such as how to open a file with a known image type. In that sense these functions are too numerable and too specific at the same time.

`io::Reader` intends to fix this by encapsulating the known inputs as an object and then presenting builder-style methods for adjusting the details, independent of whether one opens a file or reads a slice of data from memory.

This PR intends to deprecate the utility wrappers in `dynimage`, not the primitive methods in `io/free_functions.rs`—`guess_format` and `load`. This was always planned.

The main concern is to find the sweet spot for the scope of the deprecation.
* Are we okay with the set of methods being deprecated?
* Are the replacements actually easy enough to be discovered? Should we add some more documentation concerning this? One particular sore spot might be loading from memory since __no__ method in `Reader` takes a byte slice as input. For new Rusteceans the generics might be unecessarily confusing.
* Should we delay this until a writer has been added?

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -988,12 +988,13 @@ fn image_to_bytes(image: &DynamicImage) -> Vec<u8> {
 /// content before its path.
 ///
 /// [`io::Reader`]: io/struct.Reader.html
+#[deprecated = "Use `io::Reader`, it combines input functions in more flexible and consistent interface."]
 pub fn open<P>(path: P) -> ImageResult<DynamicImage>
 where
     P: AsRef<Path>,
 {
-    // thin wrapper function to strip generics before calling open_impl
-    free_functions::open_impl(path.as_ref())
+    crate::io::Reader::open(path)?
+        .decode()
 }
 
 /// Read the dimensions of the image located at the specified path.
@@ -1003,12 +1004,13 @@ where
 /// content before its path or manually supplying the format.
 ///
 /// [`io::Reader`]: io/struct.Reader.html
+#[deprecated = "Use `io::Reader`, it combines input functions in more flexible and consistent interface."]
 pub fn image_dimensions<P>(path: P) -> ImageResult<(u32, u32)>
 where
     P: AsRef<Path>,
 {
-    // thin wrapper function to strip generics before calling open_impl
-    free_functions::image_dimensions_impl(path.as_ref())
+    crate::io::Reader::open(path)?
+        .into_dimensions()
 }
 
 /// Saves the supplied buffer to a file at the path specified.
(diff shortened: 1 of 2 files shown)


fintelia:
I'm very much in favor of deprecating these functions, but I also think we should rename `io::Reader` at the same time since it would be much harder to do later. (We should of course keep the old name as an alias for a while to have backwards compatibility). 

My concern with the current name is that it is too generic. It only barely manages to avoid conflicting with the _Read_ and _BufReader_ names in stdlib, and the module scope doesn't disambiguate because the standard library also has an io module. Worse people commonly `use std::io;` so that they can disambiguate _std::io::Error_ from other errors so they then can't also `use image::io`.

My preference would be to switch to calling it _ImageReader_ and making the io module internal to the crate.

HeroicKatora:
I don't share that sentiment of name conflicts outside the library at all. Namespaces exist to avoid that exact problem of complicating the choice for names by collisions. True glob-imports are discouraged in any case. You're free to use any of:

```
use image::io as iio;
```

HeroicKatora:
Perhaps the examples should use this renaming import though. That would be easier to copy&paste which is part of the motivation for having examples in the first place.

fintelia:
Using renaming imports would be an improvement, but I want to think through a bit more. I guess my concerns are:

1. _Discoverability:_ The generated docs are too cluttered, but despite being one of the most important types in the crate, io::Reader is particularly hidden. 
2. _Code readability:_ Seeing `Reader::new(...).decode()` tells you almost nothing about what it being read. Even the io prefix doesn't make things more clear. Your suggestion of using renaming imports largely resolves this point.
3. _Consistency:_ Other names like _ImageDecoder_ include the Image prefix in them. Probably the least important item.

ripytide:
I agree with renaming `Reader` -> `ImageReader` as it a more explicit name, the `io` module also only contains three structs at the moment so I don't feel that constitutes a public module so I'd support making that private too. In fact, I feel the same way about `image::math` which only contains the `Rect` type at the moment, that could easily be exported at the top-level.

I'd also support deprecating the free-standing image opening functions. To me, this is analogous to rust itself where users are forced to be more explicit about the operations they are doing which means they must learn more about what exactly they are doing in the first place.

PR #1328, opened 2020-10-06
Move image file encoders/decoders to codecs module

fintelia:
This PR moves all the modules for specific image file formats into a new _codecs_ module. Due to https://github.com/rust-lang/rust/issues/47236 I don't think there is a way to show a deprecation notice to people still using the old path.

With this change, the modules section of the generated docs are now more reasonable:


HeroicKatora (commented):
Looks good. But I'd like to give it a day or two on fear of having missed any details.

HeroicKatora (approved):
I can't thing of any downside. Plus the module documentation of `codecs` could be dedicated towards more documentation on the traits and style at some point.

PR #1329, opened 2020-10-06
Streamline docs.rs documentation and move some content to README

fintelia:
Not super polished, but hopefully an improvement over the current text.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,13 +1,9 @@
-//! This crate provides native rust implementations of
-//! image encoders and decoders and basic image manipulation
-//! functions.
-//!
-//! Additional documentation can currently also be found in the
-//! [README.md file which is most easily viewed on github](https://github.com/image-rs/image/blob/master/README.md).
-//!
-//! [Jump forward to crate content](#reexports)
-//!
 //! # Overview
+//!
+//! This crate provides native rust implementations of image encoding and decoding as well as some
+//! basic image manipulation functions. Additional documentation can currently also be found in the
+//! [README.md file which is most easily viewed on
+//! github](https://github.com/image-rs/image/blob/master/README.md).
 //!
 //! There are two core problems for which this library provides solutions: a unified interface for image
 //! encodings and simple generic buffers for their content. It's possible to use either feature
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1,122 +1,100 @@
...
+//! # use std::io::Write;
+//! # use image::ImageOutputFormat;
+//! # use image::DynamicImage;
+//! # fn main() -> Result<(), image::ImageError> {
+//! # let img: DynamicImage = unimplemented!();
+//! # let img2: DynamicImage = unimplemented!();
+//! img.save("empty.jpg")?;
+//!
+//! let mut bytes: Vec<u8> = Vec::new();
+//! img2.write_to(&mut bytes, image::ImageOutputFormat::Png)?;
```
HeroicKatora, line 42: Unfortunately not available except with the `png` feature. Actually, why is that? We don't do the same with `ImageFormat` and it's always possible to instead error. Or a method for querying this dynamically..

HeroicKatora (commented):
Just a bit of reorganization, I think.

README.md
```diff
@@ -250,3 +250,25 @@ fn main() {
...
+1. A standard specifies that configuration _x_ is required for decoders/encoders and there
+   exists an essentially canonical way to control it.
+2. At least two different implementations agree on some (sub-)set of features in practice.
+3. A technical argument including measurements of the performance, space benefits, or otherwise
+   objectively quantified benefits can be made, and the added interface is unlikely to require
+   breaking changes.
+
+Features that fulfill two or more criteria are preferred.
+
+Re-exports of dependencies that reach version `1` will be discussed when it happens.
```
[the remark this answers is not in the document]
HeroicKatora: The `codecs` move was landed so this would now be possible.

HeroicKatora (approved):

src/lib.rs
```diff
@@ -1,123 +1,87 @@
...
-//! * [`DynamicImage::write_to`] can be used to encode an image into any writer, for example into a
-//!   vector of bytes in memory.
-//! * [`save_buffer`], [`save_buffer_with_format`] provide a low-level interface for saving an image
-//!   in the file system, where the library initializes the chosen encoder.
-//! * [`ImageEncoder`] is a trait for encoding a byte buffer of image data and the inverse of the
-//!   `ImageDecoder` interface.
+//! let mut bytes: Vec<u8> = Vec::new();
+//! img2.write_to(&mut bytes, image::ImageOutputFormat::Png)?;
+//! # Ok(())
+//! # }
```
HeroicKatora, line 45:
```suggestion
//! # }
//! # #[cfg(not(feature = "png"))] fn main() {}
```

HeroicKatora (commented):
Maybe it works like that?

PR #1331, opened 2020-10-09
Fix clippy warnings

TianyiShi2001:
Fixing these clippy warnings makes the code cleaner and more readable, and in some cases improves performance (particularly, fixing unnecessary lazy evaluation by changing `ok_or_else` to `ok_or` (well, maybe not, because the compiler may be smart enough to optimise them out)).

### DIFF
--- a/src/hdr/decoder.rs
+++ b/src/hdr/decoder.rs
@@ -884,10 +884,10 @@ fn parse_dimensions_line(line: &str, strict: bool) -> ImageResult<(u32, u32)> {
     const DIMENSIONS_COUNT: usize = 4;
 
     let mut dim_parts = line.split_whitespace();
-    let c1_tag = dim_parts.next().ok_or_else(|| DecoderError::DimensionsLineTooShort(0, DIMENSIONS_COUNT))?;
-    let c1_str = dim_parts.next().ok_or_else(|| DecoderError::DimensionsLineTooShort(1, DIMENSIONS_COUNT))?;
-    let c2_tag = dim_parts.next().ok_or_else(|| DecoderError::DimensionsLineTooShort(2, DIMENSIONS_COUNT))?;
-    let c2_str = dim_parts.next().ok_or_else(|| DecoderError::DimensionsLineTooShort(3, DIMENSIONS_COUNT))?;
+    let c1_tag = dim_parts.next().ok_or(DecoderError::DimensionsLineTooShort(0, DIMENSIONS_COUNT))?;
+    let c1_str = dim_parts.next().ok_or(DecoderError::DimensionsLineTooShort(1, DIMENSIONS_COUNT))?;
+    let c2_tag = dim_parts.next().ok_or(DecoderError::DimensionsLineTooShort(2, DIMENSIONS_COUNT))?;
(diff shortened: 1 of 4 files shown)


src/io/free_functions.rs
```diff
@@ -235,8 +235,7 @@ pub(crate) fn guess_format_from_path_impl(path: &Path) -> Result<ImageFormat, Pa
         .and_then(|s| s.to_str())
         .map(str::to_ascii_lowercase);
 
-    let ext = ext.as_ref()
-        .map(String::as_str);
+    let ext = ext.as_deref();
```
kaj, line 238: I think this requires rust 1.40.0, while `image` still promises backwards compatibility to 1.36.0.

src/pnm/decoder.rs
```diff
@@ -616,10 +616,7 @@ impl<R: Read> PnmDecoder<R> {
 fn read_separated_ascii<T: FromStr<Err = ParseIntError>>(reader: &mut dyn Read) -> ImageResult<T>
     where T::Err: Display
 {
-    let is_separator = |v: &u8| match *v {
-        b'\t' | b'\n' | b'\x0b' | b'\x0c' | b'\r' | b' ' => true,
-        _ => false,
-    };
+    let is_separator = |v: &u8| matches! { *v, b'\t' | b'\n' | b'\x0b' | b'\x0c' | b'\r' | b' '};
```
kaj, line 619: Very nice, but requires rust 1.42.0 (of course, this is about all uses of `matches!`).

kaj (commented):
Nice stuff, but some requires rust 1.40 or 1.42, and I think image still promises compatibility with rust 1.36.

src/io/free_functions.rs
HeroicKatora: With `1.34.2` to be exact.

TianyiShi2001:
I've reverted the incompatible fixes (will redo them at some point in the future). Can the remaining fixes be merged?

src/io/free_functions.rs
TianyiShi2001: Thank you! I've reverted this.

src/pnm/decoder.rs
TianyiShi2001: Thank you! I've reverted this.

HeroicKatora:
Can you revert them by rebasing and removing the commits and then force pushing instead? Github is weird in that it closes pending PRs if they only contain commits of another PR.. Regardless of whether that other PR contains commits that revert them. We can also squash merge.

> Can the remaining fixes be merged?

Yes, I think so.

> Will redo them at some point in the future

If you open another PR with only those fixes then I'll tag it appropriately. The number of good changes that would require a new MSRV is a factor in the consideration of when to bump it.

kaj (approved):

TianyiShi2001:
> Can you revert them by rebasing and removing the commits and then force pushing instead? 

Thank you! This has been done.

HeroicKatora (approved):

HeroicKatora:
@TianyiShi2001 I see you're becoming quite interested here and you already have a few PRs merged now. Time for a formal invite. Hi :wave: [The collaborator role is explained here](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md#membership-and-teams). Basically, you can merge/tag/review PRs and you only have to stick to the code of conduct.

PR #1335, opened 2020-10-09
Reduce copying in src/png.rs

TianyiShi2001:
Most of the time we read data into an empty vector. So intead of always copying by `extend_from_slice` and emptying `self.buffer`, just `mem::swap()` them when `buf` is empty (which is the common case).

### DIFF
--- a/src/png.rs
+++ b/src/png.rs
@@ -93,8 +93,13 @@ impl<R: Read> Read for PngReader<R> {
 
     fn read_to_end(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
         let mut bytes = self.buffer.len();
-        buf.extend_from_slice(&self.buffer);
-        self.buffer = Vec::new();
+        if buf.is_empty() {
+            std::mem::swap(&mut self.buffer, buf);
+        } else {
+            buf.extend_from_slice(&self.buffer);
+            self.buffer.clear();
+        }
+
         self.index = 0;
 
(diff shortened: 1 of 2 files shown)


TianyiShi2001:
BTW, is my understanding about `image::png::PngReader` correct:

- The `image::png::PngReader` is a simplified wrapper around `png::Reader`
- `png::Reader` can read:
    - normal PNG either row-by-row or in one step
    - interlaced PNG in one step only
- However `image::png::PngReader` wants a unified API that allows either reading row-by-row or in one step, so:
    - when reading common PNG, `png::Reader.next_row` is used to read row-by-row
    - when reading interlaced PNG, "real" row-by-row reading cannot be achieved, but it pretend to allow this by pre-loading the entire PNG using `png::Reader.next_frame`, and read this "intermediate" buffer row-by-row

fintelia (approved):

fintelia:
Your understanding is mostly correct. For interlaced images, I believe `png::Reader` actually does support "row-by-row" reading, but the actual order that it outputs rows in is  jumbled (something like 0, 4, 2, 6, 1, 3, 5, 7...). Since `image::png::PngReader` promises to produce rows in order, this isn't very helpful.

HeroicKatora:
> BTW, is my understanding about image::png::PngReader correct:

Yes, that's all accurate. At least that is the intention if there are no bugs.

HeroicKatora (approved):

PR #1336, opened 2020-10-09
Provide scale method of JpegDecoder.

kaj:
Add a `scale(&mut self, requested_width, requested_height)` to `JpegDecoder`, exposing same functionality in `jpeg_decoder::Decoder`.

Question: As it is allowed for this function to do nothing, maybe it should be added to trait `ImageDecoder`?  If would be possibe to add a meaningfull implementation to any "progressive" format, such as png and gif, and there could be a do-nothing default implementation in the trait.

### DIFF
--- a/src/jpeg/decoder.rs
+++ b/src/jpeg/decoder.rs
@@ -33,6 +33,33 @@ impl<R: Read> JpegDecoder<R> {
             metadata,
         })
     }
+
+    /// Configure the decoder to scale the image during decoding.
+    ///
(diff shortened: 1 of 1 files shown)


HeroicKatora (commented):
That's a pretty nifty way of asking for a different result. To answer the question, yes, I could definitely see this being very useful for the trait as well. The default implementation would merely return an error. With that insight, I could see a similar technique being used to request blocks of the image or even different color types.

Issue #1338, opened 2020-10-09
Bench `math::nq`


TianyiShi2001:
The complex but stable (API unlikely to change) algorithms in `math::nq` seems to be good subjects of benchmarking. Benchmarks will help those who want to improve their performance.

TianyiShi2001:
In my experience, for example, the compiler isn't always smart enough to elide the bound checks for indexing where all indexing operations are garanteed to be in-bound within a function, and using `unsafe {x.get_unchecked()}` can be used to improve performance, if you can tolerate `unsafe`s in your code in exchange for perf improvement.

fintelia:
Do we know if anyone is actually using `math::nq`?

HeroicKatora:
Trading performance for some well-reviewed `unsafe` can be okay, `imageproc` needs to do it as well. It is, however, not the preferred resolution. In my experience, a large part of the work is finding the cause of a particular slowdown. Here, temporarily removing some bounds checks by-hand using `unsafe` is a useful tool for finding missed optimization. In many cases there are alternatives. The `llvm` optimizer is quite powerful when given the right inputs.

For example, bounds checks on slices can often be removed by using iteration instead. Or by doing a check once at the start of a local scope then relying on llvm deducing that all indices are smaller than the initial check.  When in doubt, the [secure code working group](https://rust-lang.zulipchat.com/#narrow/stream/146229-wg-secure-code) or `#black-magic` in [the community Discord](https://discord.gg/aVESxV8) can also answer some questions and propose alternative.

HeroicKatora:
[We can't say that it is not used at all](https://github.com/search?q=%22image%3A%3Amath%3A%3Anq%22&type=code&l=rust).

That said, the code was at some point extracted as a separate crate: [`color_quant`](https://crates.io/crates/color_quant/reverse_dependencies).

HeroicKatora:
Which was transferred to `image-rs` not too long ago, by the way:
https://github.com/image-rs/color_quant

TianyiShi2001:
> Which was transferred to `image-rs` not too long ago, by the way:
> https://github.com/image-rs/color_quant

Then, how about replacing the implementation of the `math::nq` module with the corresponding implementations in the `color_quant` crate, and any benchmarks and further developments on this algorithm will go in the `color_quant` crate.

HeroicKatora:
[...]

PR #1339, opened 2020-10-09
deprecate math::nq; alias to color_quant

TianyiShi2001:
As per https://github.com/image-rs/image/issues/1338

TODO:

- [x] release the new version of `color_quant`: https://github.com/image-rs/color_quant/pull/11
- [X] resolve potential conflicts


src/math/nq.rs
```diff
@@ -98,353 +69,5 @@ type Color = Quad<i32>;
...
-                    }
-                    j -= 1;
-                }
-            }
-        }
-        best
-    }
-}
+#[deprecated(note = "Use the `color_quant` crate instead")]
+pub type NeuQuant = color_quant::NeuQuant;
```
HeroicKatora, line 73: That is a breaking change. We need to supply an empty shell and dispatch to the `color_quant` methods.

HeroicKatora (changes requested):

src/math/nq.rs
HeroicKatora: Note that another crate might already depend on both crates and have individual trait impls for both types. This doesn't seem likely but let's rather avoid the risk.

TianyiShi2001: As long as use of deprecated funtion is not denied, this will not break? See: https://github.com/image-rs/image/pull/1339/commits/83a761a8ee9b697ff0c45bdd95c8a89ba6abc2a3

Running `cargo test` produces

```
...
test src/math/nq.rs - math::nq::NeuQuant (line 36) ... ok
```

HeroicKatora: It won't break this crate but is nevertheless a breaking change according to SemVer and the Rust RFCs on the topic.

src/lib.rs
```diff
@@ -121,7 +121,7 @@
 #![warn(missing_docs)]
 #![warn(unused_qualifications)]
 #![deny(unreachable_pub)]
-#![deny(deprecated)]
+//#![deny(deprecated)]
```
HeroicKatora, line 124: Please annotate the individual tests/use sites with `#[allow(deprecated)]` instead. That way, one can easily tally all the deprecated code with `grep`.

src/math/nq.rs
TianyiShi2001: I have no experience in dealing with this kind of scenario, so could you please make the edits that are needed?

src/lib.rs
TianyiShi2001: Great to learn this technique!

src/math/nq.rs
HeroicKatora: Sure, no problem.

TianyiShi2001:
It was a great lesson to learn, thank you!

PR #1344, opened 2020-10-12
Upgrade tiff to 0.6

HeroicKatora:
I'm not sure if we should forbid the floating point color types `Rgb{,a}{F32,F64}` more explicitly for now but `bytemuck` makes it painless to defined the buffer conversion, even though our own `ColorType` does not support those buffers yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+dist: focal
 sudo: false
 notifications:
   webhooks:
@@ -41,6 +42,12 @@ matrix:
     - os: linux
       rust: stable
       env: FEATURES='ravif'
+      before_install:
+        - sudo apt-get update && 
+          sudo apt-get -y install nasm;
+      script:
+        - cargo test -v --no-default-features --features "$FEATURES" &&
+          cargo doc -v --no-default-features --features "$FEATURES";
     - os: osx
       rust: 1.34.2
     - os: windows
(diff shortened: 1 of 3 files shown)


fintelia (commented):
Shouldn't the colortype conversion code also have to be updated to handle floats?

HeroicKatora:
We'd need to introduce floating point `ColorType` in the first place, I think. Plus I think the `SampleFormat` is not actually correctly exposed in `tiff`. Good point.

HeroicKatora:
> Shouldn't the colortype conversion code also have to be updated to handle floats?

I'm not even sure how exactly how to do this. I've added checks such that this surely returns an error early for now.

HeroicKatora:
The build failure is:

> thread 'main' panicked at 'This version of NASM is too old: No such file or directory (os error 2)', /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/rav1e-0.4.0-alpha/build.rs:244:7
>
>  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

We might need to adjust the build to either install `nasm` or to skip it.

HeroicKatora:
Seems like it needs `2.14` but the latest package in the Ubuntu 18.04 environment is `2.11`. I'm going to disable that build for now...

paolobarbolini:
> Seems like it needs 2.14 but the latest package in the Ubuntu 18.04 environment is 2.11. I'm going to disable that build for now...

You can run this on `ubuntu-20.04`

HeroicKatora:
Correction: the image is based on `rust: latest` which is based on Debian buster. So the distribution is not directly in our control unless we bootstrap the Rust installation as well. But the latest package is `nasm: 2.14`. Ah, I just missed updating the package cache first :slightly_smiling_face:

HeroicKatora:
There is a way to run it on 20.04, that is by explicitly setting the distribution to `focal`. This has the required version of `nasm`.

PR #1345, opened 2020-10-14
Improve safety of ICO BMP mask length check

mfro:
If mask_end is invalid and ends up being less than the current stream position, the decoder currently panics.

I've attached an example seen in the wild (Logitech G HUB icon). The attached ICO is technically invalid and does not display properly in at least Sublime Text 3 and the windows explorer preview. It does display properly in windows preview when it is used as an icon for a lnk or exe or when opened in GIMP.

[lghub_1.zip](https://github.com/image-rs/image/files/5376709/lghub_1.zip)


src/codecs/ico/decoder.rs
```diff
@@ -332,13 +332,15 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for IcoDecoder<R> {
                 let mask_start = r.seek(SeekFrom::Current(0))?;
                 let mask_end =
                     u64::from(self.selected_entry.image_offset + self.selected_entry.image_length);
-                let mask_length = mask_end - mask_start;
 
-                if mask_length > 0 {
+                if mask_end > mask_start {
```
HeroicKatora, line 336: Forgive me the potentially naive question, I don't know much about the `ico` format (nor bmp really). What about the case where `mask-end == mask_start`? Is it allowed? Should it enter this block?

HeroicKatora (commented):
Nice catch. Do you happen to have a non-copyright protected test case? Also quite funny how different decoders react to it. `firefox` display an icon but it's broken. `imlib2` decodes the probably intended result, at least judging from the look of the result.

Could you add a `TODO` to hint that the other libraries provide some sort of recovery instead of a hard failure?

src/codecs/ico/decoder.rs
mfro: No problem at all, I'm not 100% sure myself but I think `mask-end == mask-start` is what is expected if there is no mask. I've added a comment and changed the check a bit to be more precise.

mfro:
I added a comment explaining the situation. Unfortunately, as I thought about it more I realized that the `BmpIcoMaskTooShortForImage` error case didn't really make much sense. Because there is no flag or length which directly indicates the presence of the mask, the error wasn't really serving its purpose. The more general and accurate error case is probably `image_length` doesn't match the expected length either with or without a mask.

I wasn't sure what the best approach for dealing with that error variant was. I changed it to deprecated with a brief explanation.

mfro:
I did some more research about what these values should be, and as best I can tell the `mask_end == mask_start` (now `image_end == data_end`) case is technically invalid. In that case, windows explorer preview will instead treat the end of the image data as the mask, and show a corrupted icon. GIMP, Chrome, Firefox, Sublime, will all load the image and not apply any mask. Given that, I think the best we can do is treat `image_end == data_end` as a special case, and otherwise load the mask.

I'm guessing the Wikipedia article misinterpreted the addition of PNG support, where the mask is not required, as meaning the mask is not required for BMP either. That's purely speculation on my part.

Here is a screenshot of the windows explorer preview for an ico file with no mask:
capture

And here is the same ico file in chrome:
cap2

src/codecs/ico/decoder.rs
```diff
@@ -31,7 +31,9 @@ enum DecoderError {
     /// The enclosed PNG is not in RGBA, which is invalid: https://blogs.msdn.microsoft.com/oldnewthing/20101022-00/?p=12473/.
     PngNotRgba,
 
-    /// The optional mask row, containing 1 bit per pixel, padded to 4 bytes, was too short for this image.
+    /// This could be replaced with some error about an invalid image_length in the header
+    #[allow(dead_code)]
+    #[deprecated = "Because there is no encoding for the length of the mask this error corresponded to a different case than it claimed"]
     BmpIcoMaskTooShortForImage,
```
HeroicKatora, line 37: This is a private enum so you can just remove unused variants. That said, should't it still check the length in case there is _some_ mask to ensure it's not only a partial one? Oh how I wish the test suite was larger..

mfro: Oh yeah of course. I also see what you mean about the error. Do we want to attempt to load the ICO any way if the `image_length` is not one of the two expected values? If not I can make the check more strict again and remove the test case I added or keep it as an example of a malformed icon. Either way some fix is still needed for the decoder to avoid a panic.

This does raise a question: What is this library's attitude towards decoding strictness? I haven't looked at what this would mean for other codecs, but it would be useful for my application to be able to load the original logitech icon despite the malformed header. There are also some other `DecoderError` variants that are validation beyond what is necessary to load the image data. For example, some other icons I've come across use a technically invalid non-RGBA8 PNG color type, but are otherwise valid.

HeroicKatora: As you can imagine, the mix of base crates for individual formats gives a range of approaches to the problem and there is no single consistent solution. I'll just explain how I would approach this. As a general guideline, I'd resist the temptation to guess. For technically unsupported color types there can hardly be any different semantics if it was allowed hence there isn't a lot of harm in supporting them.

However for partial masks this seems less clear. For too large values it seems okay again, it's also more compatible with previous releases which is an upside given the small test coverage. Since you found relatively consistent behaviour for empty masks, I'd definitely include that. Definitely keep the extra test here! GIMP, Chrome, Firefox, Sublime is also enough to say that this is more or less consistent where supported. Windows appears like a bug to me since it is apparently interpreting _some other_ data as a mask.. That's very fishy and uncontrollable.

mfro: Ok, I've pushed another commit that I think should match the original strictness but report an error instead of panic on the invalid case in question.

How about making another constructor for `IcoDecoder` that has a bool strictness parameter? The default behavior of the decoder would be as it is now, but a user could construct an `IcoDecoder` directly to decode icons such as the logitech one? If that would be acceptable I could open another pull request.

HeroicKatora: Unless you encounter an image in the while I wouldn't worry about the case too much. Plus I'd discuss it as part of another PR. The three basic options are a separate constructor or a bool field or a bool parameter.

HeroicKatora (approved):

Issue #1347, opened 2020-10-25
Inconsistency with DynamicImage conversion methods


aloucks:
The `as` methods (e.g. `as_rgba8`, `as_luma16`) all contain the component bit depth, while the `into` and `to` method do not. For example, there's no way to unwrap to `Luma<u16>` without a match statement.

I'd like to propose deprecating all of the current `into` and `to` methods in favor of adding new ones with the explicit bit depth (e.g. have both `into_luma8` and `into_luma16`). 

Finally, some signatures return type aliases where as others do not. I think the type aliases are less helpful than just showing that these all return `ImageBuffer` with with different pixel types.

I'd like to put in a PR for these changes but wanted to gather feedback first.

fintelia:
That all sounds reasonable to me

PR #1349, opened 2020-10-25
Add bit-depth specific `into_` and `to_` DynamicImage conversion methods

aloucks:

---

This PR introduces additional color conversion when the source and destination bit-depth differs. The `DynamicImage` `to_` and `into_` methods have been made consistent by specifying the bit-depth in the name and deprecating the originals. Notably, some of the `Bgr[a]16` conversion methods are missing due to `DynamicImage` having no variant for them.

This addresses the primary concern from #1347, but does not address the return type inconsistency. Some of the type aliases used in the method signatures are private, which results in the documentation being inconsistent. I've decided to leave that change for a separate PR.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -194,42 +194,118 @@ impl DynamicImage {
     }
 
     /// Returns a copy of this image as an RGB image.
+    #[deprecated = "replaced by `to_rgb8`"]
     pub fn to_rgb(&self) -> RgbImage {
         dynamic_map!(*self, ref p -> {
             p.convert()
         })
     }
 
+    /// Returns a copy of this image as an RGB image.
+    pub fn to_rgb8(&self) -> RgbImage {
+        dynamic_map!(*self, ref p -> {
+            p.convert()
+        })
+    }
+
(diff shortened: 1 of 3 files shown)


fintelia (approved):

src/dynimage.rs
```diff
@@ -194,110 +194,306 @@ impl DynamicImage {
...
+    /// If the image was already the correct format, it is returned as is.
+    /// Otherwise, a copy is created.
+    pub fn into_bgr8(self) -> BgrImage {
+        match self {
+            DynamicImage::ImageBgr8(x) => x,
+            x => x.to_bgr8(),
         }
     }
 
+    // NOTE: DynamicImage::ImageBgr16 variant does not currently exist, so there is no `into_bgr16`
```
fintelia, line 405: This variant will probably never exist. It is kind of unfortunate that we have BGR at all, but adding it in more places would probably require a strong reason beyond just being consistent.

HeroicKatora: At least it serves as a reminder for any plan of a replacement that this needs to be considered early. I wouldn't change the comment actually.

PR #1352, opened 2020-11-01
Add copyless DynamicImage to byte slice/vec conversion

aloucks:
Adds `DynamicImage::into_bytes(self)` and `DynamicImage::as_bytes(&self)`. 

The primary motivation for this was needing the underlying storage buffer as a `byte` vector without having to create a copy using `to_bytes`. I added `as_bytes` as an extra convenience.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -694,7 +694,17 @@ impl DynamicImage {
         }
     }
 
+    /// Return this image's pixels as a byte slice.
+    pub fn as_bytes(&self) -> &[u8] {
+        image_as_bytes(self)
+    }
+
     /// Return this image's pixels as a byte vector.
+    pub fn into_bytes(self) -> Vec<u8> {
+        image_into_bytes(self)
+    }
+
+    /// Return a copy of this image's pixels as a byte vector.
     pub fn to_bytes(&self) -> Vec<u8> {
         image_to_bytes(self)
     }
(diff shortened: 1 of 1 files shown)


src/dynimage.rs
```diff
@@ -1177,6 +1187,52 @@ fn image_to_bytes(image: &DynamicImage) -> Vec<u8> {
...
+        DynamicImage::ImageRgba16(a) => cast_vec(a.into_raw()),
+    }
+}
+
+fn image_as_bytes(image: &DynamicImage) -> &[u8] {
+    fn cast_slice(src: &[u16]) -> &[u8] {
+        unsafe {
+            let len = 2 * src.len();
+            let ptr = src.as_ptr();
+            std::slice::from_raw_parts(ptr as *const u8, len)
```
HeroicKatora, line 1219: Can you use `bytemuck` for this? I would rather have the least amount of unsafe here that's possible.

src/dynimage.rs
```diff
@@ -1177,6 +1187,52 @@ fn image_to_bytes(image: &DynamicImage) -> Vec<u8> {
...
 }
 
+fn image_into_bytes(image: DynamicImage) -> Vec<u8> {
+    fn cast_vec(mut src: Vec<u16>) -> Vec<u8> {
+        unsafe {
+            let len = 2 * src.len();
+            let cap = 2 * src.capacity();
+            let ptr = src.as_mut_ptr();
+            std::mem::forget(src);
+            Vec::from_raw_parts(ptr as *mut u8, len, cap)
```
HeroicKatora, line 1197: This is unsound. [In particular](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.from_raw_parts):

> * T needs to have the same size and alignment as what ptr was allocated with. (T having a less strict alignment is not sufficient, the alignment really needs to be equal to satsify the dealloc requirement that memory must be allocated and deallocated with the same layout.)

The size with which it was allocated is `2` and the size you intend to construct is `1`.

HeroicKatora (changes requested):

HeroicKatora:
`DynamicImage::into_bytes(self)` isn't possible, at least not with the standard provided `Vec`. The borrowing version however, is.

src/dynimage.rs
```diff
@@ -694,7 +694,17 @@ impl DynamicImage {
         }
     }
 
+    /// Return this image's pixels as a byte slice.
```
fintelia, line 697: This should say that the provided pixel data is in native endian.

src/dynimage.rs
aloucks: Ah, that's a good catch. I mistakenly believed that the less strict alignment requirement allowed it to be safe. 

This was the primary use case that I had :(

Perhaps a signature of `into_bytes(self) -> Result<Vec<u8>, Self>` would be suitable? I think that would allow something like `img.into_bytes().map_err(|this| this.to_bytes())` so that the user can get the free conversion when possible and have the copy as a fallback.

HeroicKatora: An alternative would be to use something similar to `take` here.

```rust
fn take_bytes(&mut self) -> Result<Vec<u8>, NotBytesError>
```

That is, on success we leave an empty image and on failure we return some potentially descriptive error. In particular that would allow `From<NotBytesError> for ImageError` as well as a proper `Debug`/`Display` impl which would be more ergonomic in some situations that want to use `?` operator.

aloucks: Wouldn't this have to allocate a new buffer for `self`'s storage?

fintelia: I'm not sure this makes sense to add to DynamicImage if it is going to fail for half the formats and is relatively straightforward to implement in user code.

src/dynimage.rs
```diff
@@ -1177,6 +1187,46 @@ fn image_to_bytes(image: &DynamicImage) -> Vec<u8> {
...
+        DynamicImage::ImageLumaA16(a) => cast_vec(a.into_raw()),
+        DynamicImage::ImageRgb16(a) => cast_vec(a.into_raw()),
+        DynamicImage::ImageRgba16(a) => cast_vec(a.into_raw()),
+    }
+}
+
+fn image_as_bytes(image: &DynamicImage) -> &[u8] {
+    use bytemuck::cast_slice;
+    match image {
+        DynamicImage::ImageLuma8(a) => &a.as_flat_samples().samples,
```
fintelia, line 1217: I think `&a.as_flat_samples().samples` could just be `&*a`

src/dynimage.rs
HeroicKatora: > Wouldn't this have to allocate a new buffer for selfs storage?

No, `Vec::new` is basically free and returns basically a constant.

HeroicKatora: > And is relatively straightforward to implement in user code.

Is it, though? It's an operation that takes the memory from all representations that use bytes as samples. Since the enum is non-exhaustive you can't possibly implement this downstream for all future possibilities. And even for the current ones it's somewhat tedious.

aloucks: > > Wouldn't this have to allocate a new buffer for selfs storage?
> 
> No, `Vec::new` is basically free and returns basically a constant. The trick would be to really empty the buffer such that it has size 0×0.

Yes, `Vec::new` is free, but wouldn't `set_pixel` and `get_pixel` panic if the buffer/vec is empty? In other words, is replacing the internal buffer with a zero-length vec still considered "valid" ?

aloucks:
@HeroicKatora The latest commit will return the internal container if it's already a `Vec<u8>`, otherwise a copy is made.

HeroicKatora (approved):

HeroicKatora:
Thanks!"

Issue #1354, opened 2020-11-04
Consider making dynimage::resize_dimensions function public


filnet:
I would like to be able to use the `dynimage::resize_dimensions` functions directly.

My specific use case for this functionality is provisioning sizes in UI layouts.
I am a user of the image crate and need a similar feature. At the moment I use a copy of the code.

I understand that I am asking for a shift of maintenance burden...

HeroicKatora:
Sounds cool. I think the main concerns of such a free standing function would be 

* Call interface, my main concern being the `bool` parameter that, maybe, should be an enum in the exposed variant. Or whether dimensions should be passed as two pairs or as four numbers. Two pairs seems harder to accidentally misuse but is not consistent with the usage of `resize` of course. There are some tradeoffs both ways.
* name: `resize_dimensions` isn't the worst name but Feel free to suggest any that is more intuitive and more precise than that. It's not perfectly clear that this is about preserving aspect ratio.
* The module where the method lives. That seems pretty clear to me, it should be `math` or `math::utils` if it does not pertain to a particular instance of an image.

A PR with some choices and short (like, two sentences, maybe?) reasoning for those three would help greatly.

filnet:
Suggestions taken.

Additional suggestions:
* Change the existing `dynimage::resize_dimensions` function as little as possible and keep it private
* Expose it through one or more public functions with new (better) signatures.
* The image crate should continue to call the private `dynimage::resize_dimensions` function.

PS : while using `dynimage::resize_dimensions` I came across what I believe is an issue.
Scaling (4096, 1) to (128, 128) will give (128, 0) which is not a correct image size.
Should give (128, 1).

PR #1355, opened 2020-11-05
Move the dynimage::resize_dimensions function to math::utils

filnet:
Note that math::utils::resize_dimensions() is crate public.
We don't want to expose it outside of this crate in its current form.

Relates to https://github.com/image-rs/image/issues/1354


fintelia:
Currently the only function in `math::utils` is deprecated. In fact, `math::Rect` is the only thing in the entire math module that isn't deprecated, so I think it probably just makes sense to have the public export of resize_dimensions at the module root as well.

HeroicKatora (approved):
Better place than the current one. I do share the sentiment of not exposing it in `utils`, maybe it would work in `math` directly?

PR #1361, opened 2020-11-14
Add `ImageFormat::from_extension`

a1phyr:
I also moved the inner non-generic function form `io::free_functions` and removed the intermediate `PathError` enum.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::too_many_arguments)]
 use std::convert::TryFrom;
+use std::ffi::OsStr;
 use std::io;
 use std::io::Read;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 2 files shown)


a1phyr:
Not-so-related note: I think it would make more sense for `ImageFormatHint::PathExtension` to contain an `OsString` instead of a `PathBuf`, as it stores a mere extension.

This would be a breaking change, however.

HeroicKatora (changes requested):
:+1: on moving the method and making the error handling more concise. And for homomorphizing the body of the conversion. One small nit on the interface.

HeroicKatora:
> I think it would make more sense for ImageFormatHint::PathExtension to contain an OsString instead of a PathBuf, as it stores a mere extension.

That would indeed be breaking, but _adding_ a new variant shouldn't be.

src/image.rs
```diff
@@ -65,11 +65,46 @@ pub enum ImageFormat {
 }
 
 impl ImageFormat {
+    /// Return the image format specified by a path's file extension.
+    pub fn from_extension(ext: &str) -> Option<Self> {
```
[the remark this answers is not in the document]
a1phyr: I made the function return `None` if the conversion to `str` fails, because in this case it wouldn't match any extension anyway.

HeroicKatora (approved):

src/image.rs
```diff
@@ -65,11 +66,51 @@ pub enum ImageFormat {
 }
 
 impl ImageFormat {
+    /// Return the image format specified by a path's file extension.
```
fintelia, line 69: Could we distinguish how this is function is different from the one below? Right now they have identical doc comments...

a1phyr: I added examples, which are IMO the easiest to understand the difference.

HeroicKatora:
Nice, thank you!

PR #1364, opened 2020-11-16
Fix warnings in reference_images tests and unignore them

nabijaczleweli:
The second patch might be better served as just unignoring these tests, since I don't really see why they *are* ignored, it's not like they take an exorbitant amount of time to build or run (`time cargo test -- --ignored` says real 0m5.351s, user 0m14.907s, sys 0m1.250s for me on a pre-compiled run), and being missed in an API transition indicates that they will likely be missed again, later, with potentially more consequences.

### DIFF
--- a/tests/reference_images.rs
+++ b/tests/reference_images.rs
@@ -46,7 +46,7 @@ fn render_images() {
     process_images(IMAGE_DIR, None, |base, path, decoder| {
         println!("render_images {}", path.display());
         let img = match image::open(&path) {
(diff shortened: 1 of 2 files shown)


HeroicKatora (commented):
There really doesn't seem to be a good reason that I know of for ignoring some of the image loading tests given those runtimes. I had considered that it might depend on active features but am no longer sure.

nabijaczleweli:
Updated

fintelia (approved):

PR #1369, opened 2020-11-17
Fix the crash when decoding TGA  with color map.

### LINKED ISSUE #1363: TGA with indexed colors decode to crash!
Crash when decode tga with indexed colors

## Expected

Decode correctly

## Actual behaviour

Crash

## Reproduction steps
[](url)
1. use examples/opening.rs to read `indexed_color.tga`
2. run and crash

[indexed_color.tga..zip](https://github.com/image-rs/image/files/5546654/indexed_color.tga.zip)

dev-chee:
Fix #1363

### DIFF
--- a/examples/opening.rs
+++ b/examples/opening.rs
@@ -11,7 +11,8 @@ fn main() {
     let file = if env::args().count() == 2 {
         env::args().nth(1).unwrap()
     } else {
-        panic!("Please enter a file")
+        //panic!("Please enter a file")
+        "tests/images/tga/testsuite/ccm8.tga".to_string()
     };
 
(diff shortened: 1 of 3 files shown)


HeroicKatora:
Would you add a regression test? If it's easily possible then maybe modify the headers in one of the [images here](https://github.com/image-rs/image/tree/master/tests/images/tga/testsuite) or use the image in your issue if you have the copyright for it—it's 256Kb which is still small enough to add to the repo.

-----

Aside: I hadn't realized previously but we're linked-to on Wikipedia for hosting the TGA 'conformance suite' :eyes:

PR #1370, opened 2020-11-17
Add Default to the Primitive trait

FSMaxB-dooshop:

This is strictly speaking a breaking change, but since `Primitive` already requires `NumCast`, you can always implement `Default` by doing the equivalent of `<MyPrimitive as NumCast>::from(0)`.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -27,7 +27,7 @@ impl EncodableLayout for [u16] {
 }
 
 /// Primitive trait from old stdlib
-pub trait Primitive: Copy + NumCast + Num + PartialOrd<Self> + Clone + Bounded {}
+pub trait Primitive: Copy + NumCast + Num + PartialOrd<Self> + Clone + Bounded + Default {}
 
 impl Primitive for usize {}
 impl Primitive for u8 {}


FSMaxB-dooshop:
With this PR, you can write:
```rust
fn default_pixel<PixelType: Pixel>() -> PixelType {
    let zero = PixelType::Subpixel::default();
    PixelType::from_channels(zero, zero, zero, zero)
}
```

Right now it is:
```rust
fn default_pixel<PixelType: Pixel>() -> PixelType {
    let zero = <PixelType::Subpixel as NumCast>::from(0);
    PixelType::from_channels(zero, zero, zero, zero)
}
```

FSMaxB-dooshop:
Ideally the `Pixel` trait would also require `Default`. I can do that in a separate PR if you want.

fintelia:
Wouldn't this be a breaking change?

HeroicKatora:
I don't see the benefit of the breaking change. Firstly, the semantics of a `Default` trait may indeed be different than casting from `0` even if that implementation is possible. Consider for example that `Default` should never panic while casting might. This panic should remain explicit but your motivation essentially proposes that it may be implicit in the proposed implementation of `Default`.

The benefit for requiring the trait also seems unclear. There's nothing add to the library itself that uses it, so why do it here? Why not have a sub-trait that requires `Pixel + Default`?

Together with the sketched method this seems like an XY-problem. You seem to want `default_pixel` (or already have that in another library?). Have you considered add a defaulted method to the `Pixel` trait which would not be a breaking change? An alternative would also to utilize the more semantically appropriate [`Bounded::min_value()`](https://docs.rs/num-traits/0.2.11/num_traits/bounds/trait.Bounded.html) whose trait already is required.

FSMaxB-dooshop:
> Together with the sketched method this seems like an XY-problem.

It kind of is XY-problem since I am essentially trying to create an empty `ImageBuffer<PixelType, Vec<PixelType::Subpixel>` generically for all types of `Pixel`.

HeroicKatora:
Do you specifically need `ImageBuffer<_, _>: Default` or what if this is addressed by adding a new constructor? Another benefit of this option might be to clarify if there are any color semantics that should be attached to such a method. Such a constructor could be on `ImageBuffer` or on `Pixel` or both.

FSMaxB-dooshop:
In my case I just needed an empty `ImageBuffer` to swap out using `std::mem::swap`, which would require the `Container` to be constrained to `Default` through. But in general constructing a clean slate `ImageBuffer` with `width` and `height` for any Pixel type, at least would be quite useful, though probably only possible if Container is `Vec` or at least `FromIterator<Subpixel>`.

HeroicKatora:
A manual implementation of `Default` could produce a 0×0 image without any bound on `Subpixel`, only requiring the rather simple constraint `Container: Default` which is also available for `Container = Vec<_>`. A derive would be over-constrained though, it needs to be manual.

FSMaxB-dooshop:
> A manual implementation of Default could produce a 0×0 image without any bound on Pixel or Subpixel, only requiring the rather simple constraint Container: Default which is also available for Container = Vec<_>. A derive would be over-constrained though, it needs to be manual.

Sounds good. I can do that in a separate PR!

HeroicKatora:
I was thinking that it would replace this PR. It addresses the problem directly, and avoids the breaking change, and is more discoverable.

FSMaxB-dooshop:
Closing in favor of #1376

PR #1372, opened 2020-11-18
Fix tga decoder that decodes color map and 16-bit depth format

### LINKED ISSUE #1363: TGA with indexed colors decode to crash!

dev-chee:
Fix #1363 , and continue pr  #1369 

Fix tga decoder that decodes color map and 16-bit depth format.
Again, it passed the regression test :)

---

Sorry, I am not very skilled in operating the PR.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -938,6 +938,21 @@ where
     }
 }
 
+impl<P, Container> Default for ImageBuffer<P, Container>
+where
+    P: Pixel,
+    Container: Default,
+{
+    fn default() -> Self {
+        Self {
+            width: Default::default(),
+            height: Default::default(),
+            _phantom: PhantomData,
+            data: Default::default(),
+        }
+    }
+}
+
 impl<P, Container> Deref for ImageBuffer<P, Container>
 where
     P: Pixel + 'static,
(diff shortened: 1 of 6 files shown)


fintelia (commented):
This mostly looks quite reasonable to me.

My one big concern is that it seems this PR removes a bunch of error handling around invalid values in the TGA header. Just because the spec says that only certain values are possible, doesn't mean that we won't encounter corrupt files. In those situations, the decoder should be returning an error rather than silently doing the wrong thing.

HeroicKatora:
> I am not very skilled in operating the PR.

No worries. Three helpful tips that may be helpful for beginners, in case you haven't used Git a lot yet:

* You can *force push* (`git push -f`) to the branch to overwrite the state of a PR. This allows you to correct mistakes or even do complete rewrites. [Reference for the command is here](https://git-scm.com/docs/git-push#Documentation/git-push.txt---force) or `man git push`.
* You can `git rebase --interactive master` to reorder, group and remove commits. So you can experiment freely and later make it appear as one commit. [Reference is here](https://git-scm.com/docs/git-rebase) or `man git rebase`.
* Github allows collaborators of `image-rs` to push to the branch the PR is based on, in your repository. So if you need any help with a complicated git related task, you can always ask.

HeroicKatora:
As far as I can tell, the main concern of removing a bunch of error checks is not yet addressed. For example the check

`if self.header.pixel_depth <= 24`

is no fuzzy and would permit unexpected depths such as `23` as rgb or `25` as rgba where it previously matched exact bit depths. Both cases seem problematic.

PR #1376, opened 2020-11-19
Implement Default for ImageBuffer if Container implements Default

FSMaxB-dooshop:
Solution for the problem that lead to #1370 

This implements `Default` for any `ImageBuffer` that has a container type that also implements `Default`

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -938,6 +938,21 @@ where
     }
 }
 
+impl<P, Container> Default for ImageBuffer<P, Container>
+where
+    P: Pixel,
+    Container: Default,
+{
+    fn default() -> Self {
+        Self {
+            width: Default::default(),
+            height: Default::default(),
+            _phantom: PhantomData,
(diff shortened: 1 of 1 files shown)


src/buffer.rs
```diff
@@ -938,6 +938,21 @@ where
...
 
+impl<P, Container> Default for ImageBuffer<P, Container>
+where
+    P: Pixel,
+    Container: Default,
+{
+    fn default() -> Self {
+        Self {
+            width: Default::default(),
+            height: Default::default(),
```
fintelia, line 949: It might be more obvious to say `0` here instead of `Default::default()`

fintelia (approved):

FSMaxB-dooshop:
Not sure what failed here: https://travis-ci.com/github/image-rs/image/jobs/444179245

HeroicKatora:
I've restarted the job and it seems to ran without issue. Seems to have been spurious on Travis's end.

PR #1378, opened 2020-11-26
Add write_image_with_speed_quality function.

BezPowell:
A draft change to add an additional method to avif to allow writing an image while specifying the quality and speed. It's a bit messy with the number of arguments, but shouldn't break any existing apis.

### DIFF
--- a/src/codecs/avif/mod.rs
+++ b/src/codecs/avif/mod.rs
@@ -49,7 +49,7 @@ impl<W: Write> AvifEncoder<W> {
     pub fn write_image_with_speed_quality(mut self, data: &[u8], width: u32, height: u32, color: ColorType, quality: u8, speed: u8) -> ImageResult<()> {
         // Clamp quality and speed to range
         let quality = max(0, min(quality, 100));
-        let speed = max(1, min(speed, 10));
+        let speed = max(0, min(speed, 10));
 
         let config = self.config(color, quality, speed);
         // `ravif` needs strongly typed data so let's convert. We can either use a temporarily
--- a/src/codecs/avif/mod.rs
+++ b/src/codecs/avif/mod.rs
@@ -5,7 +5,7 @@
 /// [AVIF]: https://aomediacodec.github.io/av1-avif/
 use std::borrow::Cow;
 use std::io::Write;
-use std::cmp::{min, max};
+use std::cmp::min;
 
 use crate::{ColorType, ImageBuffer, ImageFormat, Pixel};
 use crate::{ImageError, ImageResult};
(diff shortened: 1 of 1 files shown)


src/codecs/avif/mod.rs
```diff
@@ -37,7 +38,20 @@ impl<W: Write> AvifEncoder<W> {
...
+
+    /// Encode image data with the indicated color type, quality and speed.
+    ///
+    /// The encoder currently requires all data to be RGBA8, it will be converted internally if
+    /// necessary. When data is suitably aligned, i.e. u16 channels to two bytes, then the
+    /// conversion may be more efficient.
+    pub fn write_image_with_speed_quality(mut self, data: &[u8], width: u32, height: u32, color: ColorType, quality: u8, speed: u8) -> ImageResult<()> {
+        // Clamp quality and speed to range
+        let quality = max(1, min(quality, 100));
+        let speed = max(1, min(speed, 10));
```
paolobarbolini, line 52: I haven't tested this but I think it should be possible to use 0 as the quality and the speed. In `libavif` quality 0 means lossless compression, speed 0 is very slow but still usable.

BezPowell: Thanks for pointing that out. I foolishly assumed that it would start at 1.
I've checked out ravif and it seems that quality is 0 - 100, but speed is 1 - 10.
```
/// 0-100 scale
pub quality: u8,
/// 0-100 scale
pub alpha_quality: u8,
/// rav1e preset 1 (slow) 10 (fast but crappy)
pub speed: u8,
```

paolobarbolini: 🤔 the rav1e docs say that they do have a speed 0 setting. I'm not sure why ravif starts from 1.

https://docs.rs/rav1e/0.4.0-alpha/rav1e/prelude/struct.SpeedSettings.html#method.from_preset

BezPowell: That is odd. I've just tried setting the minimum to 0 and running an encode through and nothing broke and the filesize was significantly smaller than 1, so I guess it must just be a typo in the ravif comments?

Apologies for all the back on forth with that one.

src/codecs/avif/mod.rs
```diff
@@ -37,7 +38,20 @@ impl<W: Write> AvifEncoder<W> {
...
-        let config = self.config(color);
+        self.write_image_with_speed_quality(data, width, height, color, 100, 1)
+    }
+
+    /// Encode image data with the indicated color type, quality and speed.
+    ///
+    /// The encoder currently requires all data to be RGBA8, it will be converted internally if
+    /// necessary. When data is suitably aligned, i.e. u16 channels to two bytes, then the
+    /// conversion may be more efficient.
+    pub fn write_image_with_speed_quality(mut self, data: &[u8], width: u32, height: u32, color: ColorType, quality: u8, speed: u8) -> ImageResult<()> {
```
HeroicKatora, line 49: Nit on the interface style, I'd prefer storing a `ravif::Config` in the encoder and having another constructor instead. This is already used for options in both `png` and `jpeg`. Another idea would be to provide setters for quality and speed instead of a separate constructor.

BezPowell: Definitely agree with you there. The reason I didn't do it was purely down to me being thick and not wanting to interfere with a public struct; it looks like I wasn't paying attention there though as I forgot that in rust properties on a public struct are private unless explicitly declared otherwise. I'll rework that now to use a separate constructor like jpeg - I'm afraid you are going to have quite a bit of commit history added to the repo by the time I'm done.

BezPowell: I've now reworked the code to set speed and quality on the constructor instead. Cargo check seems to have broken itself on my machine this morning, but it compiles and works when testing it on a simple application.

The old `config()` function called from `write_image()` was passed the `ColorType`, but made no use of it. I've created an empty function called `set_color()` to replicate this and serve as a reminder, but can delete it if deemed unnecessary.

HeroicKatora (approved):

src/codecs/avif/mod.rs
```diff
@@ -23,12 +24,33 @@ use rgb::AsPixels;
...
 impl<W: Write> AvifEncoder<W> {
     /// Create a new encoder that writes its output to `w`.
     pub fn new(w: W) -> Self {
-        AvifEncoder { inner: w, fallback: vec![] }
+        AvifEncoder::new_with_speed_quality(w, 1, 100)
+    }
+
+    /// Create a new encoder with specified speed and quality,
+    /// that writes its output to `w`.
+    pub fn new_with_speed_quality(w: W, speed: u8, quality: u8) -> Self {
```
fintelia, line 38: Could we document the allowed ranges for quality and speed?

BezPowell: Absolutely. I've added that now.

PR #1382, opened 2020-11-30
Less allocations in DynImage::write_to.

laurmaedje:
When looking at the `DynamicImage::write_to`, I noticed that there is a bit more cloning/allocation happening than necessary. In particular, the `bytes` variable created in the beginning is not used in some cases and `self.to_rgba().iter().cloned().collect()` converts and then iterates and collects again. The changed version borrows the bytes instead and only allocates when a conversion is required. I also shortened `a.iter().cloned().collect()` to `a.as_raw().clone()` in `image_to_bytes`.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -905,7 +905,7 @@ impl DynamicImage {
         // When no features are supported
         let w = w;
         #[allow(unused_variables,unused_mut)]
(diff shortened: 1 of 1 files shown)


HeroicKatora (approved):
Not sure what's up with CI but code looks good.

PR #1383, opened 2020-12-14
Convert Pixel types into corresponding types from rgb crate.

msmorgan:
This allows their components to be accessed by name. For example,
if `c` is `Rgb<u8>`, one can write `c.r` instead of `c.channels()[0]`.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -1,4 +1,5 @@
-use std::ops::{Index, IndexMut};
+use std::mem;
+use std::ops::{Deref, DerefMut, Index, IndexMut};
 
 use num_traits::{NumCast, ToPrimitive, Zero};
 
@@ -205,6 +206,7 @@ macro_rules! define_colors {
         $interpretation: expr,
         $color_type_u8: expr,
         $color_type_u16: expr,
+        $( $rgb_equiv:ident )::*,
         #[$doc:meta];
     )*} => {
 
@@ -282,13 +284,13 @@ impl<T: Primitive + 'static> Pixel for $ident<T> {
 
     fn to_luma(&self) -> Luma<T> {
         let mut pix = Luma([Zero::zero()]);
-        pix.from_color(self);
+        FromColor::from_color(&mut pix, self);
         pix
     }
 
     fn to_luma_alpha(&self) -> LumaA<T> {
         let mut pix = LumaA([Zero::zero(), Zero::zero()]);
-        pix.from_color(self);
+        FromColor::from_color(&mut pix, self);
         pix
     }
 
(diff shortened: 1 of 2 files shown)


HeroicKatora:
Interesting idea. Maybe we should deref into the types of the [`rgb`](https://docs.rs/rgb/0.8.25/rgb/) crate instead of defining our own? It's possibly one of the most stable crates for the domain and would improve interop, avoiding the situation of the xkcd comic in its Readme :) The biggest difference being that its gray scale types for some reason don't seem to use the same structure.

fintelia:
Worth pointing out that you can do `c.0[0]` instead of `c.channels()[0]`, but obviously `c.r` is nicer.

Not thrilled that this would add more unsafe to the crate. I don't see anything unsound though as far as I can tell.

msmorgan:
@HeroicKatora Done.

src/color.rs
```diff
@@ -364,18 +366,43 @@ impl<T: Primitive + 'static> From<[T; $channels]> for $ident<T> {
...
+    fn from(value: $( $rgb_equiv )::* <T>) -> Self {
+        unsafe { mem::transmute_copy(&value) }
+    }
+}
+
+impl<T: Primitive> From<$ident<T>> for $( $rgb_equiv )::* <T> {
+    fn from(value: $ident<T>) -> Self {
+        unsafe { mem::transmute_copy(&value) }
+    }
+}
```
HeroicKatora, line 393: If I'm not mistaken we can avoid all of the unsafety by using [`rgb::FromSlice`](https://docs.rs/rgb/0.8.25/rgb/trait.FromSlice.html) or some alternative. Particularly the `transmute_copy` is a very big and dangerous hammer that can easily be mistaken in macro code and copy and paste.

msmorgan: Done. It was a little fiddly though.

HeroicKatora (commented):
From an implementation perspective this PR has my approval. It's just that my experience with the technique of using `Deref` like that is limited, so I do wonder if there is any stability concern that we're overlooking? A run of [`rustwide`](https://github.com/rust-lang/rustwide) against downstream crates might give more certain results, I've wanted to write some integration for this for a while now.

msmorgan:
I was inspired by [nalgebra](https://nalgebra.org/) crate, which uses `Deref` to expose convenience properties [for `Matrix` (and vectors)](https://github.com/dimforge/nalgebra/blob/dev/src/base/coordinates.rs), and [for `Point`](https://github.com/dimforge/nalgebra/blob/dev/src/geometry/point_coordinates.rs).

fintelia:
I started looking at the code this generates and I'm starting to get worried this is too messy/confusing. After this change we'd have structs Rgb and RGB, crate rgb, enum variants Rgb8 and Rgb16, and type aliases RGB8 and RGB16. Yes, our pixel struct Rgb would transparently deref to RGB, but the reverse would require an into() call. To make matters worse, the change adds a couple methods to our pixel structs which return types from the rgb crate.

The core issue is that Rust doesn't really support the ability to have multiple structs that can be used interchangeably. If we want to support nice field accesses, I think the right way to do it would be to commit to a breaking change. It would take more discussion but I do think a strong case could be made for adopting the standard rgb crate types outright (though perhaps wait for them to reach 1.0 first?).

HeroicKatora:
>  It would take more discussion but I do think a strong case could be made for adopting the standard rgb crate types outright (though perhaps wait for them to reach 1.0 first?).

They have been on `0.8` for about three years now, iirc. It's slightly unfortunate but as a 'compatibility crate' they are more or less bound to that version for now but could re-export those types as `1.0` at some point by a dependency on the old version.

msmorgan:
@fintelia

> After this change we'd have structs Rgb and RGB, crate rgb, enum variants Rgb8 and Rgb16, and type aliases RGB8 and RGB16.

`rgb::RGB<T>` and `rgb::RGB8` are not publicly exported by `image`, so they shouldn't add too much confusion for users. In the docs, the associated type `<Rgb<T> as Deref>::Target` will show up as belonging to a different crate.

> Yes, our pixel struct Rgb would transparently deref to RGB, but the reverse would require an into() call.

This is true if you have a value of `rgb::RGB<T>`, but `Deref` produces a reference, `&rgb::RGB<T>`, with the same lifetime as the borrow by the `deref` function. There would be no need to convert back, since the  owned value is still an `image::Rgb`.

fintelia:
Here's a screenshot that shows what this looks like in the generated documentation. I don't see anything that indicates that `Rgb` and `RGB` / `RGBA` are from different crates:
image

fintelia:
Here's a screenshot that shows what this looks like in the generated documentation. I don't see anything that indicates that `Rgba` and `RGB` / `RGBA` are from different crates:

image

msmorgan:
Hmm, that is confusing. You'd be directed to the rgb crate's type if you click on "RGBA". I wonder if there's a way to hide all those methods though. The types from nalgebra that vector and point deref into have very few methods on them so they don't cause so much documentation noise.

msmorgan:
I added a commit that leaves the `From` impls intact for conversion to/from the `rgb` crate types, but adds simple proxy structs as `Deref` targets for component access. Let me know what you think.

Issue #820, opened 2018-10-07
Support ARGB ColorType


prasannavl:
Currently, there's no support for ARGB32. Cairo/Skia supports this, and is also often the format that X is going to return it's visuals in (for instance from xcb_get_image/shm_get, or xlib XGetImage, etc. 

Would be extremely helpful to have support for it. 

PS: Having conversions functions would be a good starting point, so that they can alteast be imported in, at a one time cost.

nox:
I also filed https://github.com/PistonDevelopers/image-png/issues/86.

PR #1385, opened 2020-12-15
Add ARGB ColorType

danielvschoor:
I needed an ARGB buffer for one of my projects and I decided to implement it in the image crate instead of doing it myself. I'm not sure if something like this is needed in the official crate (the only reference I could find to ARGB is #820), but I decided to open a PR anyway. The biggest change is to the `define_colors` macro, since it did not support alpha channels in any other position than the last position. 

I wasn't sure where I needed to add the new ColorType (i.e. encoding/decoding parts), so I kept it as basic as possible.

If you are a new contributor, consent to licensing by including this text:

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -1007,6 +1007,19 @@ impl FromColor<Argb<u16>> for Argb<u8> {
     }
 }
 
+impl FromColor<Argb<u8>> for Argb<u16> {
+    fn from_color(&mut self, other: &Argb<u8>) {
+        let argb = self.channels_mut();
+        let argb8 = other.channels();
+        argb[0] = upcast_channel(argb8[0]);
+        argb[1] = upcast_channel(argb8[1]);
+        argb[2] = upcast_channel(argb8[2]);
+        argb[3] = upcast_channel(argb8[3]);
+
+    }
+}
+
+
 impl FromColor<Rgba<u8>> for Argb<u16> {
     fn from_color(&mut self, other: &Rgba<u8>) {
         let argb = self.channels_mut();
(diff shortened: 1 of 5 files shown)


src/color.rs
```diff
@@ -1386,10 +1388,12 @@ macro_rules! downcast_bit_depth_early {
 downcast_bit_depth_early!(Luma, Luma, LumaA);
 downcast_bit_depth_early!(Rgb, Rgb, LumaA);
 downcast_bit_depth_early!(Rgba, Rgba, LumaA);
+downcast_bit_depth_early!(Argb, Argb, LumaA);
```
danielvschoor, line 1391: I'm unsure about these downcasts and if I'm doing it for the correct types

fintelia:
I'd personally lean against including more color channel permutations throughout the library. The conversion between RGBA and ARGB is cheap and lossless, so I'd rather only have people who need ARGB to worry about it. Other people might feel differently though

HeroicKatora:
Adding the structs for the purpose of implementing the `Pixel` trait seems quite useful in general. But do not that adding them to `DynamicImage` as a variant is a breaking change so I'd definitely separate those two changes.

saihaze:
I think ARGB is quite necessary. I am using wlroots (C API), and its renderer only consumes ARGB images when creating OpenGL ES textures. At least we should provide a way to convert RGBA buffers to ARGB ones.

fintelia:
That should just be a short `for` loop. Something like:

```rust
fn rgba_to_argb(img: &RgbaImage) -> Vec<u8> {
    let rgba_bytes = img.into_vec();
    for (rgba, argb) in rgba {
    } 
}

```

fintelia:
That should just be a short for loop. Something like:

```rust
fn rgba_to_argb(img: &RgbaImage) -> Vec<u8> {
    let bytes = img.into_vec();
    for pixel in bytes.chunks_exact_mut(4) {
        let alpha = pixel[3];
        pixel[3] = pixel[2]; // blue
        pixel[2] = pixel[1]; // green
        pixel[1] = pixel[0]; // red
        pixel[0] = alpha;
    }
    bytes
}

saihaze:
You are right. Could this be added to the library?
1. If breaking changes are avoided, it can be an optional feature.
2. You have said that conversations about ARGB and RGBA is cheap and lossless, so I think it is better not to prefer one of them.
By the way, I am reading wlroots code to see if I can make changes on their side.

fintelia:
Could you say more about why it would help to have that method as part of `image` rather than in your own code?

saihaze:
OK. As a general-purpose image library, being more adaptive is never a bad idea. Users may face similar situations and such utilities can help ease convertion between different formats. I surely can implement such functions in my own code (which is what I am actually doing now), but I still hope this functionality can be added to your great library.

2023年8月5日 12:31:18 Jonathan Behrens ***@***.***>:

PR #1387, opened 2020-12-17
GitHub actions ci

HeroicKatora:


HeroicKatora:
Now for the horrifing part, debugging why something runs or does not run in the cloud..

HeroicKatora:
fb32421 had already been queued. But did not again the next time I reverted to it, removing its marker. And then did again the third time.

Okay, we learn: 
* The task box at `Merge pull request` is outdated by up to 4 minutes with regards to queued checks.
* The 'Actions' tab needs to be fully reloaded for it to display new actions.
* New tasks do not cancel outdated ones automatically.
* Sometimes when force pushing quickly a task might get dropped (don't know if this was only due to restoring an old state or not).

It's a mess and there should be an offline validation tool.

HeroicKatora:
So, `ravif` and the other OS configurations are still missing. Both appear to stall or never trigger CI. I'm out of ideas as to why.

HeroicKatora:
Note: Some parts (`ravif` + windows + macos) are still not working but the available of some CI outweighs this. I'll try to re-enable the missing parts at a later date.

Issue #1373, opened 2020-11-18
Ergonomics of checking if a file is supported without reading it


Boscop:
It would be great to have a way to check if a file is supported without having to read the file. 

My specific use case for this functionality is:
I'm filtering the results of a glob search, and only want to retain paths to files that can be read by this crate (including formats supported in the future, so not hardcoding the list of image formats). Without having to read each one, since they are gigapixel images.

## Draft

Several ways of offering this functionality come to mind:
1. Provide an iterator over all image formats, so that I can flat_map extension_strs to get all supported extensions into a HashSet and lookup in that while filtering the paths from the glob search. 
2. A function is_extension_supported. Less flexible, very narrow. I think option 1 makes more sense. 

I see that you are preventing exhaustiveness checking on the enum, so you can't derive strum::EnumIter but it could be a manual `iter` method on the enum. This wouldn't go against the reasons who preventing exhaustiveness checking, since it's a legitimate desire to be able to iterate all supported formats.

Note: For being able to implement my filter correctly, I'd also need a way to filter the image format iterator by whether that format is supported for reading. So there could be a method `ImageFormat::can_read() -> bool`.

HeroicKatora:
Both parts seems reasonably small and useful. I can offer guidance on implementation if necessary.

> I see that you are preventing exhaustiveness checking on the enum, so you can't derive strum::EnumIter but it could be a manual iter method on the enum. This wouldn't go against the reasons who preventing exhaustiveness checking, since it's a legitimate desire to be able to iterate all supported formats.

Exactly. A downstream crate shouldn't depend on the exact set of formats that are known. The simplest way would be to provide an opaque struct that implements `Iterator<Item=ImageFormat>` and _internally_ iterates over a slice of all formats. A slight alteration of the concept would be an associated method `ImageFormat::iter() -> impl Iterator<Item=Self>`. It wouldn't be as easy to store that in a struct but would that be required?

Boscop:
Yeah, I think the struct that impls Iterator is the way to go. I think there's an unwritten rule that crates should only use `-> impl` for private functions, at least as long as existential types are not on stable. After that, it'd be easy to also store these in a struct, if the crate defines a `pub type FooIter = impl Iterator<Item = Foo>;`

fintelia:
[...]

PR #1388, opened 2020-12-20
Adds can_read on ImageFormat

Mike-Neto:
Progresses #1373

Adds can_read.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -134,76 +134,46 @@ impl ImageFormat {
     }
 
     /// Return if the ImageFormat can be decoded by the lib.
-    ///
-    /// # Example
-    ///
-    /// ```
-    /// use image::ImageFormat;
-    /// #[cfg(feature = "png")]
-    /// assert_eq!(ImageFormat::Png.can_read(), true);
-    /// ```
     #[inline]
     pub fn can_read(&self) -> bool {
         // Needs to be updated once a new variant's decoder is added to free_functions.rs::load
         match self {
-            #[cfg(feature = "png")]
             ImageFormat::Png => true,
-            #[cfg(feature = "gif")]
             ImageFormat::Gif => true,
-            #[cfg(feature = "jpeg")]
             ImageFormat::Jpeg => true,
-            #[cfg(feature = "webp")]
             ImageFormat::WebP => true,
-            #[cfg(feature = "tiff")]
             ImageFormat::Tiff => true,
-            #[cfg(feature = "tga")]
             ImageFormat::Tga => true,
-            #[cfg(feature = "dds")]
-            ImageFormat::Dds => true,
-            #[cfg(feature = "bmp")]
+            ImageFormat::Dds => false,
             ImageFormat::Bmp => true,
-            #[cfg(feature = "ico")]
             ImageFormat::Ico => true,
-            #[cfg(feature = "hdr")]
             ImageFormat::Hdr => true,
-            #[cfg(feature = "pnm")]
             ImageFormat::Pnm => true,
-            #[cfg(feature = "farbfeld")]
             ImageFormat::Farbfeld => true,
-            _ => false,
+            ImageFormat::Avif => false,
+            ImageFormat::__NonExhaustive(marker) => match marker._private {},
         }
     }
 
     /// Return if the ImageFormat can be encoded by the lib.
-    ///
-    /// # Example
-    ///
-    /// ```
-    /// use image::ImageFormat;
-    /// #[cfg(feature = "png")]
-    /// assert_eq!(ImageFormat::Png.can_write(), true);
-    /// #[cfg(feature = "pnm")]
-    /// assert_eq!(ImageFormat::Pnm.can_write(), false);
-    /// ```
     #[inline]
     pub fn can_write(&self) -> bool {
         // Needs to be updated once a new variant's encoder is added to free_functions.rs::save_buffer_with_format_impl
         match self {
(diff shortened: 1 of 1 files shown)


Mike-Neto:
This PR adds on `can_read` as I was not clear on how to implement `can_write` other than exactly the same way as I did `can_read`, is that what's intended or can someone point me towards what would that be.

src/image.rs
```diff
@@ -133,6 +133,23 @@ impl ImageFormat {
...
+    /// # Example
+    ///
+    /// ```
+    /// use image::ImageFormat;
+    ///
+    /// let result = ImageFormat::can_read("images/ferris.png");
+    /// assert_eq!(result, true);
+    /// ```
+    #[inline]
+    pub fn can_read<P>(path: P) -> bool where P : AsRef<Path> {
```
fintelia, line 147: I thought we were going to make this have signature `pub fn can_read(&self) -> bool`?

Mike-Neto: oh I misunderstood it then, I will update it soon 👍

Mike-Neto: While trying to implement it as `pub fn can_read(&self) -> bool` I'm even more confused on what to implement exactly, sorry for this but I will need a more concrete example on what to do.

From what I read from your first post, the problem we're trying to solve is filtering the results of a glob search into just the paths that are supported by the lib in a way that needs not to be modified if a new type is supported, which was way I implemented it as `pub fn can_read<P>(path: P) -> bool where P : AsRef<Path>` which is still wrong as it doesn't check for the existance of a decoder for the corresponding ImageFromat, only that the ImageFromat exists in the crate.

Now for it to be `pub fn can_read(&self) -> bool` it implies that you already have the ImageFromat (which is odd given you're trying to check from paths, this means you already had to run something to map the path to the ImageFromat) and are checking for support for encode/decode right? So what I would have to implement would be a way to match self to a decoder/encoder for can_read/can_write respectively.

However since the way it is done in `pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<DynamicImage>` it's trough a match statement I see no other option other than copy paste the match & update it to return what I can see maps to a decoder now, the problem with this is that if a current format is added without a decoder & then the decoder is added latter this match can get out of sync with the libs functionality, which would be troublesome to maintain.

Is this really what we want for this, or is it something else?

Thank you for the patience & all the help!

fintelia: > Now for it to be pub fn can_read(&self) -> bool it implies that you already have the ImageFromat (which is odd given you're trying to check from paths, this means you already had to run something to map the path to the ImageFromat) and are checking for support for encode/decode right?

Yes. To go from a path / file extension to an ImageFormat we have [`from_path`](https://docs.rs/image/0.23.12/image/enum.ImageFormat.html#method.from_path) / [`from_extension`](https://docs.rs/image/0.23.12/image/enum.ImageFormat.html#method.from_extension).

> I see no other option other than copy paste the match & update it to return what I can see maps to a decoder now, the problem with this is that if a current format is added without a decoder & then the decoder is added latter this match can get out of sync with the libs functionality, which would be troublesome to maintain.

Yes, I was just imagining a match statement. You are right that this adds one more place to get out of sync, but if we were able to do the kind of introspection necessary to avoid that, we probably wouldn't need this function in the first place :wink:

src/image.rs
```diff
@@ -133,6 +133,77 @@ impl ImageFormat {
...
+    /// # Example
+    ///
+    /// ```
+    /// use image::ImageFormat;
+    /// assert_eq!(ImageFormat::Png.can_write(), true);
+    /// assert_eq!(ImageFormat::Pnm.can_write(), false);
+    /// ```
+    #[inline]
+    pub fn can_write(&self) -> bool {
+        // Needs to be updated once a new variant's encoder is added to free_functions.rs::save_buffer_with_format_impl
```
Mike-Neto, line 187: I noticed that `free_functions.rs::save_buffer_with_format_impl` & `free_functions.rs::save_buffer_impl` have different match clauses as in `free_functions.rs::save_buffer_with_format_impl` has no references to `PnmEncoder` is that inteded? Can I fix that in a follow up PR?

Mike-Neto: Or should I just implement what https://github.com/image-rs/image#supported-image-formats says we support?

src/image.rs
```diff
@@ -133,6 +133,80 @@ impl ImageFormat {
...
+            ImageFormat::Bmp => true,
+            #[cfg(feature = "ico")]
+            ImageFormat::Ico => true,
+            #[cfg(feature = "hdr")]
+            ImageFormat::Hdr => true,
+            #[cfg(feature = "pnm")]
+            ImageFormat::Pnm => true,
+            #[cfg(feature = "farbfeld")]
+            ImageFormat::Farbfeld => true,
+            _ => false,
```
fintelia, line 173: Could you replace this line with a match on the non-exhaustive variant? This will trigger a compile error if a new variant is added without this function being updated, which is what we want.
```rust
ImageFormat::__NonExhaustive(marker) => match marker._private {},
```

src/image.rs
HeroicKatora: I'd keep it simple for now. The difference in the two methods seems to be indeed an error.

fintelia (approved):

src/image.rs
HeroicKatora: Opened #1390 for it.

HeroicKatora (approved):

PR #1393, opened 2020-12-31
Added in_place variants of colorops

vallentin:
On Stack Overflow a [question] asked in relation to the efficiency of the `colorops` module. I [answered][answer] implementing `*_in_place` variants of the functions, to avoid the needless cloning and heap allocations for temporary images.

I wasn't able to find any discussions of why the `colorops` module didn't have `*_in_place` variants of its functions, while the `imageops` module does. So if that is desired, then this PR implements those.

More information can be found in my Stack Overflow [answer], along with some benchmarking details.

[question]: https://stackoverflow.com/questions/65517715/fastest-way-to-perform-multiple-imageimageops-on-a-single-image-rust
[answer]: https://stackoverflow.com/a/65524735/2470818

-----

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -1,14 +1,14 @@
 //! Functions for altering and converting the color of pixelbufs
 
-use num_traits::{Num, NumCast};
+use num_traits::{Bounded, Num, NumCast};
 use std::f64::consts::PI;
 
 use crate::color::{Luma, Rgba};
 use crate::image::{GenericImage, GenericImageView};
 #[allow(deprecated)]
 use crate::math::nq;
-use crate::utils::clamp;
 use crate::traits::{Pixel, Primitive};
+use crate::utils::clamp;
 use crate::ImageBuffer;
 
 type Subpixel<I> = <<I as GenericImageView>::Pixel as Pixel>::Subpixel;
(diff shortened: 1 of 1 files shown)


fintelia:
Looking over your Stack Overflow answer, it looks like you included the JPEG decoding time (via `image::load_from_memory`) in your benchmark. Since that is likely a significant fraction of the total runtime it would be good to get numbers without that. 

The other question I have is how those functions would perform without all the layers of abstraction we have in the colorops module. If we're already missing out on tons of performance, then we might want to look into larger changes rather than just adding in place versions of the existing functions. Could you see how something like the following compares against the generic versions?
```rust
fn contrast(image: &mut RgbImage, contrast: f32) {
    let percent = ((100.0 + contrast) / 100.0).powi(2);
    (*image).iter_mut().for_each(|v|{
         let f = ((*v as f32 / 255.0 - 0.5) * percent + 0.5) * 255.0;
        *b = clamp(f, 0.0, 255.0) as u8;
    });
}
```

vallentin:
>  it looks like you included the JPEG decoding time (via `image::load_from_memory`) in your benchmark

Yes. I specially used `image::load_from_memory` compared to `image::open` to avoid the IO overhead. But yes, now that you mention it, it kinda is redundant. I think I was in my "testing mindset", and wanted all benchmarks to return the same final image.

I'll redo the tests. But that's gonna have to say until tomorrow 🎆 

> how those functions would perform without all the layers of abstraction we have in the colorops module

I did test `RgbImage` as well, just excluded it from the already long answer. The performance was similar to the `*_in_place` variants or slightly lower.

I'll redo the benchmarks tomorrow or in a 1-2 days.

HeroicKatora (approved):
Let's wait on the benchmarks. From a code perspective this looks very good to me.

PR #1394, opened 2020-12-31
codecs/gif: Add speed argument

Tuetuopay:
This is straight up forwarded to the underlying gif library, and allows setting the encoding quality to be able to get faster encodes.

This is especially useful for debug builds, where the absence of optimization from the compiler makes this awfully slow. For a simple test gif (260x160, 36 frames), it took more than 1 minute to encode it on a core i7 laptop. On release builds, the difference is much less perceptible, but this still gives the option to the user.

### DIFF
--- a/src/codecs/gif.rs
+++ b/src/codecs/gif.rs
@@ -293,6 +293,7 @@ impl<'a, R: Read + 'a> AnimationDecoder<'a> for GifDecoder<R> {
 pub struct GifEncoder<W: Write> {
     w: Option<W>,
     gif_encoder: Option<gif::Encoder<W>>,
+    speed: i32,
 }
 
 /// GIF encoder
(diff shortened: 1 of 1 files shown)


HeroicKatora:
This is similar to the `quality` parameter of jpeg/png. The interface of storing that parameter in the encoder struct and providing a constructor/setter seems more convenient, and there's no need to duplicate all methods.

HeroicKatora:
> This is especially useful for debug builds, where the absence of optimization from the compiler makes this awfully slow. For a simple test gif (260x160, 36 frames), it took more than 1 minute to encode it on a core i7 laptop. 

[cargo allows overriding the optimization level of a single dependency](https://doc.rust-lang.org/cargo/reference/profiles.html#build-dependencies).

Tuetuopay:
> This is similar to the `quality` parameter of jpeg/png. The interface of storing that parameter in the encoder struct and providing a constructor/setter seems more convenient, and there's no need to duplicate all methods.

Thanks, I will redo using this interface style, especially since I have another PR coming with this style.

> > This is especially useful for debug builds, where the absence of optimization from the compiler makes this awfully slow. For a simple test gif (260x160, 36 frames), it took more than 1 minute to encode it on a core i7 laptop.
> 
> [cargo allows overriding the optimization level of a single dependency](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides).

Maybe I'm very bad at configuring Cargo, but I haven't been able to get a debug configuration with performance even remotely similar to what I get in release mode :disappointed:

HeroicKatora (approved):
Last time (don't remember where) there was an idea of having a small enum instead of different quality arguments with different allowed intervals. That said, let's just have the same interface for all encoders right now and then fix them up in one go once such an enum is fleshed out.

PR #1395, opened 2021-01-01
frame: Add buffer_mut function

Tuetuopay:
This allows getting a mutable `RgbaImage` from a mutable `Frame` to easily edit a frame sequence (e.g. a GIF image).

### DIFF
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -83,6 +83,11 @@ impl Frame {
         &self.buffer
     }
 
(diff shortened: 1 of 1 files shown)


HeroicKatora (approved):
The only potential downside I see would be that one can adjust the size of the image afterwards but we don't make use of that as an invariant in any constructor, nor any encoder. The library already commits to having some `RgbaImage` representation due to `buffer`.

PR #1396, opened 2021-01-01
gif: Add set_repeat to GifEncoder

Tuetuopay:
This option allows controlling the GIF repetition options, like setting the GIF to loop indefinitely.

### DIFF
--- a/src/codecs/gif.rs
+++ b/src/codecs/gif.rs
@@ -289,11 +289,30 @@ impl<'a, R: Read + 'a> AnimationDecoder<'a> for GifDecoder<R> {
     }
 }
 
+/// Number of repetitions for a GIF animation
+#[derive(Clone, Copy)]
+pub enum Repeat {
+    /// Finite number of repetitions
+    Finite(u16),
+    /// Looping GIF
+    Infinite,
+}
+
+impl Repeat {
+    pub(crate) fn to_gif_enum(&self) -> gif::Repeat {
+        match self {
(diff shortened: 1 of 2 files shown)


src/codecs/gif.rs
```diff
@@ -289,10 +289,29 @@ impl<'a, R: Read + 'a> AnimationDecoder<'a> for GifDecoder<R> {
...
+}
+
+impl From<Repeat> for gif::Repeat {
+    fn from(repeat: Repeat) -> Self {
+        match repeat {
+            Repeat::Finite(n) => gif::Repeat::Finite(n),
+            Repeat::Infinite => gif::Repeat::Infinite,
+        }
+    }
+}
```
HeroicKatora: Including this conversion would make the exact version of `gif` a public dependency. Don't add it. (It actually makes CI fail, that's at least reassuring to see).

HeroicKatora (changes requested):

src/codecs/gif.rs
Tuetuopay: Ah yes, did not know this was one of your policies. I originally wanted to straight up use (and re-export) the `gif::Repeat`  enum, but it does not implement `Copy` nor `Clone`, which is kind of a pain. Changing this!

Tuetuopay: Updated. CI seems happy now, but I'll let you be the last judge on this.

HeroicKatora (approved):

PR #1398, opened 2021-01-14
Add an AVIF decoder

### LINKED ISSUE #1152: Support AVIF (AV1 Image FIle Format)

linkmauve:
This encoder so far only supports 8-bit data, and due to #1217 is much less efficient than it could be, as data is always converted to BGRA8888.  I could also add support for I8 and IA88, but those formats are uncommon enough we can probably ignore the inefficiency for a first version.

This is 5am quality code, so no error handling but `.unwrap()`, and there are some `unsafe` blocks which could possibly be avoided, I will get back to it after a first round of review.

This also depends on https://github.com/mozilla/mp4parse-rust/pull/266 to build.

Fixes #1152.


Cargo.toml
```diff
@@ -43,7 +46,7 @@ quickcheck = "0.9"
 criterion = "0.3"
 
 [features]
-default = ["gif", "jpeg", "ico", "png", "pnm", "tga", "tiff", "webp", "bmp", "hdr", "dxt", "dds", "farbfeld", "jpeg_rayon"]
+default = ["avif", "gif", "jpeg", "ico", "png", "pnm", "tga", "tiff", "webp", "bmp", "hdr", "dxt", "dds", "farbfeld", "jpeg_rayon"]
```
HeroicKatora, line 49: The feature can't be default, as it conflicts with the MSRV policy. I'd be happy for a note though and to add it in the next major release.

Cargo.toml
```diff
@@ -33,6 +33,9 @@ scoped_threadpool = { version = "0.1", optional = true }
 tiff = { version = "0.6.0", optional = true }
 ravif = { version = "0.6.0", optional = true }
 rgb = { version = "0.8.25", optional = true }
+mp4parse = { version = "0.11.4", path = "../mp4parse/mp4parse", optional = true }
```
HeroicKatora, line 36: A path dependency?

src/codecs/avif/decoder.rs
```diff
@@ -0,0 +1,156 @@
...
+        let mut primary_decoder = dav1d::Decoder::new();
+        primary_decoder.send_data(ctx.primary_item(), None, None, None).unwrap();
+        let picture = primary_decoder.get_picture().unwrap();
+        let alpha_picture = if let Some(alpha_item) = ctx.alpha_item() {
+            let mut alpha_decoder = dav1d::Decoder::new();
+            alpha_decoder.send_data(alpha_item, None, None, None).unwrap();
+            Some(alpha_decoder.get_picture().unwrap())
+        } else {
+            None
+        };
```
HeroicKatora, line 39: This should not `unwrap` nor `assert`, it handles potentially dangerous user input.

src/codecs/avif/decoder.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    fn color_type(&self) -> ColorType {
+        ColorType::Bgra8
+    }
+
+    fn into_reader(self) -> ImageResult<Self::Reader> {
+        let data = self.picture.plane_data_ptr(PlanarImageComponent::Y) as *mut _;
+        let (width, height) = self.picture.plane_data_geometry(PlanarImageComponent::Y);
+        let vec = unsafe { Vec::from_raw_parts(data, (width * height) as usize, (width * height) as usize) };
```
HeroicKatora, line 79: `Vec::from_raw_parts` is only safe when `data` was allocated with the Rust global allocator, and exactly with the size given in `capacity` and allign of its data type, that is 1. There is no proof of these facts, so allocate this into a new vector for exampling using `[u8]::to_vec`.

src/codecs/avif/decoder.rs
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    let data = self.picture.plane_data_ptr(PlanarImageComponent::U) as *mut _;
+                    let (width, height) = self.picture.plane_data_geometry(PlanarImageComponent::U);
+                    unsafe { std::slice::from_raw_parts(data, (width * height) as usize) }
+                },
+                {
+                    let data = self.picture.plane_data_ptr(PlanarImageComponent::V) as *mut _;
+                    let (width, height) = self.picture.plane_data_geometry(PlanarImageComponent::V);
+                    unsafe { std::slice::from_raw_parts(data, (width * height) as usize) }
+                },
```
HeroicKatora, line 124: The combination of `unsafe` and potentially overflowing multiplication `width * height` does _not_ bode well. Additionally, the unchecked cast with `as`? The length also seems to be reused several times over so validating it once initial by calculating the result with checked arithmetic would be preferred.

HeroicKatora (commented):
The amount of `unsafe` in decoder is not yet very reassuring, particularly without safety comments. If that combination, deriving a slice from the `plane_data_ptr` and `plane_data_geometry`, is always safe then I'm wondering why this is not an abstraction provided in the `dav1d` crate?

Cargo.toml
linkmauve: This crate hasn’t been released with https://github.com/mozilla/mp4parse-rust/pull/266 yet, we’ll need one before merging this PR, hence the Draft status.

Cargo.toml
linkmauve: Thanks, this is fixed in https://github.com/image-rs/image/pull/1398/commits/84c72fee412ca6f68e9e88d6199077ad72166a7e

src/codecs/avif/decoder.rs
linkmauve: Once https://github.com/rust-av/dav1d-rs/pull/42 is merged, this won’t have any `unwrap` anymore (see the current draft version).

src/codecs/avif/decoder.rs
linkmauve: I completely missed the `.plane()` API, which implements `AsRef<[u8]>` and is thus perfectly safe to use here. The latest version now removed all `unsafe` from this module.

src/codecs/avif/decoder.rs
linkmauve: Same here, latest version removed all `unsafe` blocks in favour of the `.plane()` API.

Cargo.toml
linkmauve: I’ve replaced it with a git dependency instead, so that CI can pass.

src/codecs/avif/decoder.rs
```diff
@@ -0,0 +1,160 @@
...
+            };
+            let dst_format = dcp::ImageFormat {
+                pixel_format: dcp::PixelFormat::Bgra,
+                color_space: dcp::ColorSpace::Lrgb,
+                num_planes: 1,
+            };
+            let (width, height) = self.dimensions();
+            let planes = &[
+                self.picture.plane(PlanarImageComponent::Y),
+                self.picture.plane(PlanarImageComponent::Y),
```
HeroicKatora, line 127: Probably a c&p error, right?
```suggestion
                self.picture.plane(PlanarImageComponent::Y),
                self.picture.plane(PlanarImageComponent::U),
```

linkmauve: Oh thanks, that’s exactly where my weird colours were coming from! :D

linkmauve:
There, both the mp4parse and dav1d crates got a release, I’ve updated this PR to use the non-git version.

src/codecs/avif/decoder.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn from(err: dav1d::Error) -> ImageError {
+        ImageError::Decoding(DecodingError::new(ImageFormat::Avif.into(), err))
+    }
+}
+
+impl From<dcp::ErrorKind> for ImageError {
+    fn from(err: dcp::ErrorKind) -> ImageError {
+        ImageError::Decoding(DecodingError::new(ImageFormat::Avif.into(), err))
+    }
+}
```
HeroicKatora, line 35: This turns all of those crates into public dependencies. We usually don't want that to happen for any image format whose encoder/decoder hasn't reached a stable version yet.

HeroicKatora: Hm, one of the CI jobs should have linted this: https://github.com/image-rs/image/runs/1772778509
I wonder why not. It might have to do with this not being a default feature.

HeroicKatora (approved):
The code looks good, there is only this one annoying interface detail we have to worry about.

HeroicKatora (changes requested):
Sorry for the confusion. Misclicked. The one review comment previously is already correct.

src/codecs/avif/decoder.rs
linkmauve: Hmm, should I do the conversion manually then, instead of using `?` in the code?  I previously had a quite heavy `.map_err()` doing the same conversion, but that made the code bothersome to read.

I could go back to the `.map_err()` way, but using a helper conversion function, would that be ok?

linkmauve: That’s what I did, with the new `error_map()` function which goes in the `.map_err()` to replace this `From` impl.

Cargo.toml
foresterre: > The feature can't be default, as it conflicts with the MSRV policy. I'd be happy for a note though and to add it in the next major release.

Making `avif` a default feature will require a modern version of `nasm` to be present aside from a rust toolchain, since AVIF encoding depends on `ravif`, which requires a recent version of `nasm` to be present (see https://github.com/image-rs/image/issues/1404#issuecomment-769117536), at least by default.

HeroicKatora (approved):

HeroicKatora:
Thanks a lot. ETA: next weekend.

PR #1403, opened 2021-01-17
Allow setting the AvifEncoder ColorSpace

paolobarbolini:

This change makes it possible to use the ravif YCbCr encoding option, which converts RGB to YCbCr, getting higher compression ratios in AVIF.

### DIFF
--- a/src/codecs/avif/mod.rs
+++ b/src/codecs/avif/mod.rs
@@ -3,9 +3,8 @@
 /// The [AVIF] specification defines an image derivative of the AV1 bitstream, an open video codec.
 ///
 /// [AVIF]: https://aomediacodec.github.io/av1-avif/
-
 pub use self::decoder::AvifDecoder;
-pub use self::encoder::AvifEncoder;
+pub use self::encoder::{AvifEncoder, ColorSpace};
 
 mod decoder;
 mod encoder;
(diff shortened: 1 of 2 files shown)


src/codecs/avif/mod.rs
```diff
@@ -15,9 +15,31 @@ use crate::error::{EncodingError, ParameterError, ParameterErrorKind, Unsupporte
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ColorSpace {
+    /// Encode as YCbCr
+    YCbCr,
+    /// Encode as RGB
+    RGB,
+
+    #[doc(hidden)]
+    __NonExhaustive(crate::utils::NonExhaustiveMarker),
+}
```
HeroicKatora, line 31: I'm a bit confused with this. Neither YCbCr nor RGB are color spaces, they are very rough methods of specifying an encoding. It seems that `ravif` translates these two specific 'Color config semantics' (quoting avif specification) and in particular uses `BT.709` and `sRGB`. I'd greatly prefer using the sames `Bt709` and `Srgb` for these variants.

HeroicKatora (commented):
It's in the 1.0 specification so I guess we can expect any encoder to support this extension.

paolobarbolini:
Rebased and addressed the code review comment

HeroicKatora (approved):

fintelia (approved):

PR #1407, opened 2021-01-23
Add dimension overflow check for decoders

### LINKED ISSUE #1399: attempt to multiply with overflow in image::load_from_memory
This happens in the function `image::load_from_memory`

## Expected

An error about that the image format was invalid

## Actual behaviour

Full output:

```
$ RUST_BACKTRACE=1 ./target/debug/image-reproduce 
thread 'main' panicked at 'attempt to multiply with overflow', /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/image.rs:485:9
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::panicking::default_hook::{{closure}}
   4: std::panicking::default_hook
   5: std::panicking::rust_panic_with_hook
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::panicking::panic
   9: image::image::ImageDecoder::total_bytes
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/image.rs:485
  10: image::image::decoder_to_vec
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/image.rs:421
  11: image::dynimage::decoder_to_image
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/dynimage.rs:1140
  12: image::dynimage::DynamicImage::from_decoder
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/dynimage.rs:194
  13: image::io::free_functions::load
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/io/free_functions.rs:85
  14: image::dynimage::load_from_memory_with_format
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/dynimage.rs:1323
  15: image::dynimage::load_from_memory
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/dynimage.rs:1308
  16: image_reproduce::main
             at src/main.rs:5
  17: std::rt::lang_start::{{closure}}
             at /usr/src/rustc-1.43.0/src/libstd/rt.rs:67
  18: std::panicking::try::do_call
  19: __rust_maybe_catch_panic
  20: std::rt::lang_start_internal
  21: std::rt::lang_start
             at /usr/src/rustc-1.43.0/src/libstd/rt.rs:67
  22: main
  23: __libc_start_main
  24: _start
```

## Reproduction steps

```
    let data = vec![0x66, 0x61, 0x72, 0x62, 0x66, 0x65, 0x6c, 0x64, 0xc9, 0xc9,
                    0xc9, 0x36, 0x39, 0x30, 0x38, 0x31];

    let _ = image::load_from_memory(&data);
```

In version: image = "0.23.12"

fseegraeber:
Decoder constructors now check if parsed image dimensions will
result in the image buffer exceeding u64::MAX.

Implemented for DDS, Farbfeld, HDR, PNM. Other formats either can not
overflow, are not contained in this repo or already implemented that check.

Fixes #1399.

#### Image Formats
- **avif** -> no decoder
- **bmp** -> already checks
- **hdr** -> check + test
- **ico** -> uses BMP or PNG decoders internally
- **jpeg** -> other crate/repository
- **pnm** -> check + test
- **tga** -> width and height are `u16`, byte per pixel is a maximum of 4, can not overflow with `u64`
- **webp** -> format specifies filesize as `u32::MAX - 10`. Can not overflow with `u64`
- **dds** -> check + test
- **farbfeld** -> check + test
- **gif** -> other crate/repository
- **png** -> other crate/repository
- **tiff** -> other crate/repository

Very new to Rust so let me know if this solution is valid. Especially the `utils::check_dimension_overflow` function. Not sure how you would prevent duplicating that code otherwise in Rust.

_I license past and future contributions under the dual MIT/Apache-2.0 license,
allowing licensees to chose either at their option._


src/utils/mod.rs
```diff
@@ -63,6 +63,28 @@ pub(crate) fn expand_bits(bit_depth: u8, row_size: u32, buf: &[u8]) -> Vec<u8> {
...
+/// 
+/// Returns the optionally provided `ImageError` on overflow or defaults to `LimitErrorKind::DimensionError`.
+#[allow(dead_code)]
+// When no image formats that use it are enabled
+pub(crate) fn check_dimension_overflow(
+    width: u32,
+    height: u32,
+    bytes_per_pixel: u8,
+    error: Option<crate::ImageError>,
+) -> crate::ImageResult<()> {
```
HeroicKatora, line 76: This requires that we always pass an `error` even when the check succeeds. (Also, the `None` case isn't actually used?). A slight more efficient interface would remove those cost by taking a constructor instead:

```suggestion
pub(crate) fn check_dimension_overflow(
    width: u32,
    height: u32,
    bytes_per_pixel: u8,
    error: impl FnOnce() -> crate::ImageError,
) -> crate::ImageResult<()> {
```

src/utils/mod.rs
```diff
@@ -63,6 +63,28 @@ pub(crate) fn expand_bits(bit_depth: u8, row_size: u32, buf: &[u8]) -> Vec<u8> {
...
+    use crate::error::{ImageError, LimitError, LimitErrorKind};
+
+    (width as u64)
+        .checked_mul(height as u64)
+        .and_then(|size| size.checked_mul(bytes_per_pixel as u64))
+        .map(|_| ())
+        .ok_or(error.unwrap_or(ImageError::Limits(LimitError::from_kind(
+            LimitErrorKind::DimensionError,
+        ))))
```
HeroicKatora, line 85: A bit too dense for my taste, particularly `map(|_| ())` is weird to read. I've written code like that but now prefer more explicit flow control. So I would split the three cases with `if` after the calculation chain :)

```suggestion
    let byte_len = (width as u64)
        .checked_mul(height as u64)
        .and_then(|size| size.checked_mul(bytes_per_pixel as u64));
        
    if byte_len.is_some() {
        Ok(())
    } else {
        // Or match on `error` with the current parameter type.
        Err(error())
    }
```

HeroicKatora (changes requested):
Looks very good overall, I just have some rather small nits on the code style of the check method itself.

src/utils/mod.rs
fintelia, line 85: A slightly simpler way to do the same check would be:
```rust
width as u64 * height as u64 > u64::MAX / bytes_per_pixel as u64
```

src/utils/mod.rs
fintelia, line 76: It might be better to have this function just return a bool, and make the caller deal with throwing the error.

fintelia (commented):
Thanks for this submission, it all looks quite good to me! Left a few comments about tweaking the interface, but I don't think they're strictly necessary.

src/utils/mod.rs
HeroicKatora: Hm, also a good idea to avoid the issue [mentioned here](https://github.com/image-rs/image/pull/1407#discussion_r563164600). Either way is fine for me.

src/utils/mod.rs
fseegraeber: Ah smart, I like it.

fseegraeber: I took inspiration from the BMP implementation that uses that `map(|_| ())` 😄 . But I agree the chain is a lot. I'll probably go with the [suggested](https://github.com/image-rs/image/pull/1407#discussion_r563167890):
```rust
width as u64 * height as u64 > u64::MAX / bytes_per_pixel as u64
```

src/utils/mod.rs
fseegraeber: Yea, I initially thought I would have some cases were it would make sense to provide a default error. Since that is not the case I would probably take @fintelia's [suggestion](https://github.com/image-rs/image/pull/1407#discussion_r563168041) and return a bool. That seems a bit more flexible and obvious.

fseegraeber:
Thanks for the reviews! I implemented the suggested changes.

HeroicKatora (approved):
Looks good to me, extra kudos for all the testing :+1:

PR #1408, opened 2021-01-23
Round float numbers before conversion in sampling

### LINKED ISSUE #1402: Shrinking image causes visual glitch
This happens in `image::imageops::resize`.

## Expected

Shrinking shouldn't add visual glitch

## Actual behaviour

It does happen

## Reproduction steps

Input image:

yuno

```rust
// let decoded = raw RGBA buffer
// let xsize = 622;
// let ysize = 622;

let rgba = image::RgbaImage::from_raw(info.xsize, info.ysize, decoded)
    .expect("Failed to consume the decoded RGBA buffer");

rgba.save(&format!("{}\\debug.png", env!("CARGO_MANIFEST_DIR"))).unwrap();

let resized =
    image::imageops::resize(&rgba, 256, 256, image::imageops::Lanczos3);

resized.save(&format!("{}\\debug-resized.png", env!("CARGO_MANIFEST_DIR"))).unwrap();
```

Output image:

debug

Glitch:

image

Shrinking with image editors e.g. Paint.NET doesn't add such glitches, is this expected here?

saschanaz:

Closes #1402

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -105,6 +105,20 @@ pub(crate) struct Filter<'a> {
     pub(crate) support: f32,
 }
 
+struct FloatNearest(f32);
+
+impl ToPrimitive for FloatNearest {
+    fn to_i64(&self) -> Option<i64> {
+        NumCast::from(self.0.round())
+    }
+    fn to_u64(&self) -> Option<u64> {
+        NumCast::from(self.0.round())
+    }
+    fn to_f64(&self) -> Option<f64> {
+        NumCast::from(self.0)
+    }
+}
+
 // sinc function: the ideal sampling filter.
 fn sinc(t: f32) -> f32 {
     let a = t * f32::consts::PI;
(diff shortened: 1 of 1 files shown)


src/imageops/sample.rs
```diff
@@ -261,10 +261,10 @@ where
...
             let (t1, t2, t3, t4) = (t.0 / sum, t.1 / sum, t.2 / sum, t.3 / sum);
             let t = Pixel::from_channels(
-                NumCast::from(clamp(t1, 0.0, max)).unwrap(),
-                NumCast::from(clamp(t2, 0.0, max)).unwrap(),
-                NumCast::from(clamp(t3, 0.0, max)).unwrap(),
-                NumCast::from(clamp(t4, 0.0, max)).unwrap(),
+                NumCast::from(clamp(t1, 0.0, max).round()).unwrap(),
+                NumCast::from(clamp(t2, 0.0, max).round()).unwrap(),
+                NumCast::from(clamp(t3, 0.0, max).round()).unwrap(),
+                NumCast::from(clamp(t4, 0.0, max).round()).unwrap(),
```
fintelia, line 267: This is going to be wrong if S is a floating point type. I'm not entirely sure how to resolve that

src/imageops/sample.rs
```diff
@@ -105,6 +105,20 @@ pub(crate) struct Filter<'a> {
...
+    fn to_i64(&self) -> Option<i64> {
+        NumCast::from(self.0.round())
+    }
+    fn to_u64(&self) -> Option<u64> {
+        NumCast::from(self.0.round())
+    }
+    fn to_f64(&self) -> Option<f64> {
+        NumCast::from(self.0)
+    }
+}
```
HeroicKatora, line 120: Very interesting solution :slightly_smiling_face: If I read it correctly then `to_f32` dispatches to `to_f64` hence no additional override is necessary. Maybe add this as a comment as it's not obvious at all that nothing was accidentally omitted.

HeroicKatora (approved):

PR #1409, opened 2021-01-24
Derive Debug, Hash, PartialEq, Eq for DynamicImage

Systemcluster:
All variants of `DynamicImage` (which are all `ImageBuffer` types) derive `Debug, Hash, PartialEq, Eq`.

`DynamicImage` not implementing them seems like an undesirable restriction, this makes it usable where those traits are required.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -39,7 +39,7 @@ use crate::math::resize_dimensions;
 use crate::traits::Pixel;
 
 /// A Dynamic Image
-#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


HeroicKatora (approved):

HeroicKatora:
Thank you!

fintelia:
Is `Eq` going to cause us problems if DynamicImage ever supports floating point images?

HeroicKatora:
@fintelia Since we can't add them in a stable version anyways, no. But maybe we'll need to rethink the addition then.

PR #1413, opened 2021-01-27
Optimize default implementation of copy_from

mbrubeck:
Copying by row instead of by column makes this function 5x to 10x faster for typical image types (like `ImageBuffer`) that store pixels by row.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -841,8 +841,8 @@ pub trait GenericImage: GenericImageView {
             )));
         }
 
-        for i in 0..other.width() {
-            for k in 0..other.height() {
+        for k in 0..other.height() {
+            for i in 0..other.width() {
                 let p = other.get_pixel(i, k);
                 self.put_pixel(i + x, k + y, p);
             }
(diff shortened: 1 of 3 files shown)


fintelia (approved):

src/image.rs
```diff
@@ -841,8 +841,8 @@ pub trait GenericImage: GenericImageView {
             )));
         }
 
-        for i in 0..other.width() {
-            for k in 0..other.height() {
+        for k in 0..other.height() {
+            for i in 0..other.width() {
```
HeroicKatora, line 845: Can we still get the same gains if we move this implementation to the impl on `ImageBuffer` as a override of the method? I'm slightly worried this might be depended upon somewhere. It's maybe a bit too cautious but…

mbrubeck: That would optimize the `ImageBuffer` case, but other implementers (like `SubImage` and `ViewMut` and any third-party types) would still use the default version, which would almost always be slower than optimal.

Note that other code in this crate (for example, a lot of the code in `imageops`) assumes that a `GenericImage` is most efficiently iterated row-by-row, so I think it's reasonable to use the same assumption here.

HeroicKatora: Okay. That makes it seem quite unlikely that someone depends on col-by-col and has not already specialized this.

HeroicKatora (approved):

PR #1416, opened 2021-01-29
Improved documentation of pixel enumeration and coordinate access.

faldor20:
It was unclear where the coordinate space in an image started from. I have added notes about it in a few key places.


fintelia (approved):

faldor20:
Actually. I would like to add a similar note about the positioning of the sub-image and cropping square. 
If anyone can think of any other places where such things should be clarified that would be great.

kaj (approved):
I like the clarification.

PR #1417, opened 2021-01-30
Enabled bounds check for gif frames

### LINKED ISSUE #1238: "Index out of range" panic on decoding GIF files (both valid and malformed)
This happens in `image` from git, revision 092db1e21b3e8736711e2cbe1d9b5245613756c1

## Expected

Image decodes successfully or returns error, but doesn't panic.

## Actual behaviour

When decoding the AFL-generated seed set for libgif, `image` panics on many of the images:

```
15141:thread 'main' panicked at 'index 724992 out of range for slice of length 4096', src/libcore/slice/mod.rs:2725:5
15142-Triggered by: "./afl_testcases/gif_im/full/images/id:000477,src:000000,op:havoc,rep:16.gif"
--
15193:thread 'main' panicked at 'index 4096 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15194-Triggered by: "./afl_testcases/gif_im/full/images/id:001432,src:000003+001096,op:splice,rep:2.gif"
--
15261:thread 'main' panicked at 'index 8 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15262-Triggered by: "./afl_testcases/gif_im/full/images/id:001950,src:000962,op:havoc,rep:4.gif"
--
15277:thread 'main' panicked at 'index 4096 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15278-Triggered by: "./afl_testcases/gif_im/full/images/id:001430,src:000003+001096,op:splice,rep:16.gif"
--
15303:thread 'main' panicked at 'index 4176 out of range for slice of length 4096', src/libcore/slice/mod.rs:2725:5
15304-Triggered by: "./afl_testcases/gif_im/full/images/id:000813,src:000481,op:arith8,pos:38,val:-12.gif"
--
15311:thread 'main' panicked at 'index 4096 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15312-Triggered by: "./afl_testcases/gif_im/full/images/id:001434,src:000003+001096,op:splice,rep:16.gif"
--
15317:thread 'main' panicked at 'index 4227136 out of range for slice of length 4096', src/libcore/slice/mod.rs:2725:5
15318-Triggered by: "./afl_testcases/gif_im/full/images/id:000517,src:000037,op:havoc,rep:2.gif"
--
15495:thread 'main' panicked at 'index 6604900 out of range for slice of length 4096', src/libcore/slice/mod.rs:2725:5
15496-Triggered by: "./afl_testcases/gif_im/full/images/id:000428,src:000000,op:havoc,rep:4.gif"
--
15505:thread 'main' panicked at 'index 8327168 out of range for slice of length 4096', src/libcore/slice/mod.rs:2725:5
15506-Triggered by: "./afl_testcases/gif_im/full/images/id:000397,src:000000,op:havoc,rep:16.gif"
--
15553:thread 'main' panicked at 'index 4227136 out of range for slice of length 4096', src/libcore/slice/mod.rs:2725:5
15554-Triggered by: "./afl_testcases/gif_im/full/images/id:000572,src:000080,op:havoc,rep:2.gif"
--
15599:thread 'main' panicked at 'index 8 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15600-Triggered by: "./afl_testcases/gif_im/full/images/id:001971,src:001950,op:havoc,rep:4.gif"
15601:thread 'main' panicked at 'index 1024 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15602-Triggered by: "./afl_testcases/gif_im/full/images/id:000686,src:000131,op:havoc,rep:16.gif"
--
15625:thread 'main' panicked at 'index 4096 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15626-Triggered by: "./afl_testcases/gif_im/full/images/id:001720,src:001434,op:flip1,pos:43.gif"
--
15657:thread 'main' panicked at 'index 4096 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15658-Triggered by: "./afl_testcases/gif_im/full/images/id:001429,src:000003+001096,op:splice,rep:32.gif"
--
15675:thread 'main' panicked at 'index 128 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15676-Triggered by: "./afl_testcases/gif_im/full/images/id:000589,src:000080,op:havoc,rep:8.gif"
--
15713:thread 'main' panicked at 'index 1024 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15714-Triggered by: "./afl_testcases/gif_im/full/images/id:001139,src:000679,op:flip1,pos:6.gif"
--
15819:thread 'main' panicked at 'index 4096 out of range for slice of length 0', src/libcore/slice/mod.rs:2725:5
15820-Triggered by: "./afl_testcases/gif_im/full/images/id:001433,src:000003+001096,op:splice,rep:2.gif"
--
15839:thread 'main' panicked at 'index 724992 out of range for slice of length 4096', src/libcore/slice/mod.rs:2725:5
15840-Triggered by: "./afl_testcases/gif_im/edges-only/images/id:000477,src:000000,op:havoc,rep:16.gif"
```

## Reproduction steps

[panicking_afl_gif_testcases.tar.gz](https://github.com/image-rs/image/files/4699108/panicking_afl_gif_testcases.tar.gz)

```rust
fn main() -> std::io::Result<()> {
    let path = std::env::args().skip(1).next().unwrap();
    let image = image::io::Reader::open(path)?.guess_format().decode().unwrap();
    Ok(())
}
```

fseegraeber:
Fixes #1238 

Reading a gif that contained frames exceeding its "Logical Screen" dimensions would panic. 
This PR simply enables the check that is provided by the underlying decoder.

Feel free to reject or comment if you want to handle out of bounds frames differently.

### DIFF
--- a/src/codecs/gif.rs
+++ b/src/codecs/gif.rs
@@ -94,51 +94,61 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for GifDecoder<R> {
     fn read_image(mut self, buf: &mut [u8]) -> ImageResult<()> {
         assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
 
-        let (f_width, f_height, left, top);
+        let frame = match self.reader.next_frame_info()
+            .map_err(ImageError::from_decoding)? {                
+            Some(frame) => FrameInfo::new_from_frame(frame),
+            None => {
+                return Err(ImageError::Parameter(ParameterError::from_kind(
+                    ParameterErrorKind::NoMoreData,
+                )))
+            }
+        };
 
-        if let Some(frame) = self.reader.next_frame_info().map_err(ImageError::from_decoding)? {
-            left = u32::from(frame.left);
-            top = u32::from(frame.top);
-            f_width = u32::from(frame.width);
-            f_height = u32::from(frame.height);
+        let (width, height) = self.dimensions();
+
+        if (frame.left, frame.top) == (0, 0) && (frame.width, frame.height) != self.dimensions() {
+            // If the frame matches the logical screen, directly read into it
+            self.reader.read_into_buffer(buf).map_err(ImageError::from_decoding)?;
         } else {
-            return Err(ImageError::Parameter(
-                ParameterError::from_kind(ParameterErrorKind::NoMoreData)
-            ));
-        }
+            // If the frame does not match the logical screen, read into an extra buffer
+            // and 'insert' the frame from left/top to logical screen width/height.
+            let mut frame_buffer = vec![0; self.reader.buffer_size()];
+            self.reader.read_into_buffer(&mut frame_buffer[..]).map_err(ImageError::from_decoding)?;
 
-        self.reader.read_into_buffer(buf).map_err(ImageError::from_decoding)?;
+            let frame_buffer = ImageBuffer::from_raw(frame.width, frame.height, frame_buffer);
+            let image_buffer = ImageBuffer::from_raw(width, height, buf);
 
-        let (width, height) = (u32::from(self.reader.width()), u32::from(self.reader.height()));
-        if (left, top) != (0, 0) || (width, height) != (f_width, f_height) {
-            // This is somewhat of an annoying case. The image we read into `buf` doesn't take up
-            // the whole buffer and now we need to properly insert borders. For simplicity this code
-            // currently takes advantage of the `ImageBuffer::from_fn` function to make a second
-            // ImageBuffer that is properly positioned, and then copies it back into `buf`.
-            //
-            // TODO: Implement this without any allocation.
-
-            // Recover the full image
-            let image_buffer = {
-                // See the comments inside `<GifFrameIterator as Iterator>::next` about
-                // the error handling of `from_raw`.
-                let image = ImageBuffer::from_raw(f_width, f_height, &mut *buf).ok_or_else(
-                    || ImageError::Unsupported(UnsupportedError::from_format_and_kind(
+            // `buffer_size` uses wrapping arithmetics, thus might not report the
+            // correct storage requirement if the result does not fit in `usize`.
+            // `ImageBuffer::from_raw` detects overflow and reports by returning `None`.
+            if frame_buffer.is_none() || image_buffer.is_none() {
+                return Err(ImageError::Unsupported(
+                    UnsupportedError::from_format_and_kind(
                         ImageFormat::Gif.into(),
-                        UnsupportedErrorKind::GenericFeature(format!("Image dimensions ({}, {}) are too large", f_width, f_height)))))?;
+                        UnsupportedErrorKind::GenericFeature(format!(
+                            "Image dimensions ({}, {}) are too large",
+                            frame.width, frame.height
+                        )),
+                    ),
+                ));
+            }
 
-                ImageBuffer::from_fn(width, height, |x, y| {
-                    let x = x.wrapping_sub(left);
-                    let y = y.wrapping_sub(top);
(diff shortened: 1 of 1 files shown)


fseegraeber:
On a note: I just discovered that, when running all test targets, the check_references & render_images tests fail with the tests\\images\\gif\\anim\\oob.gif. Not suprisingly I guess.

Image viewers I tested display that image regardless (interestingly Windows, IrfanView and VSCode all seem to display it differently). If this should be supported it must be defined how i guess.

HeroicKatora:
The common interpretation is to discard data outside the bounds and only blend the pixels that are inbounds. I'm quite worried why the CI does not show failure despite images failing to render?

fseegraeber:
To me it looks like if `--no-default-features` overrides the `--features "gif"`.

HeroicKatora:
@fseegraeber You're right. The reference tests need _both_ `gif` and `png` enabled as any unsupported image formats are skipped and `png` is required to load the references. It seems that no configuration has all features enabled.

fseegraeber:
> The common interpretation is to discard data outside the bounds and only blend the pixels that are inbounds. 

Enabling the bounds check is not the way to go then. 

What you describe seems to be the current behavior. Problematic are only frames that exceed the logical screen in size. These currently lead to a panic.

Should we 
- allocate the frame in an extra buffer and only copy the inbound pixels, or
- return with an error

Ideally the underlying gif decoder would directly support that scenario (transfering a frame rectangle into a target rectangle) making additional allocations unneccesary, but I'm not sure if that is desired.

HeroicKatora:
> Should we [..] allocate the frame in an extra buffer and only copy the inbound pixels

I believe the code already has the intention of doing that: https://github.com/image-rs/image/blob/b77d4f3698a6187a78a91a157df0b1b9fd7b8e2e/src/codecs/gif.rs#L264-L277

If there are any remaining panics (now I'm no longer sure) then I think we should fix those while keeping that same intention.

fseegraeber:
You'r right the `Iterator` implementation does it that way. Allocating frame + logical screen if they dont match.
The problem is in `read_image`:
https://github.com/image-rs/image/blob/b77d4f3698a6187a78a91a157df0b1b9fd7b8e2e/src/codecs/gif.rs#L110
`self.reader.read_into_buffer(buf)` expects `buf` to be atleast of frame size, but it is of size `self.total_bytes()` which uses the logical screen size.

It probably makes sense to consolidate those two methods.

HeroicKatora:
Well spotted. Yes, that seems like the right way to handle it.

fseegraeber:
`GifDecoder::read_image` does now handle frames bigger than the logical screen size by allocating it into an extra buffer and inserting the frame into the logical screen image at the correct position. This is basically what `next()` was doing.

I refactored `read_image` to better align with `GifFrameIterator::next` and got rid of one `copy_from_slice`.

The functionality is pretty similar, yet there are some differences, that made it hard to really consolidate them. `read_image` does not need to blend between different frames, while `next` needs to. I experimented with a 
```rust
fn read_next_frame(blend: Option<F>, ...) where F: Fn(...) {...
```
but couldn't get it to compile in the end😄 . I doubt it would really have improved readability.

I refactored some of the shared functionality into a helper struct, not sure how much that imroves it though.

HeroicKatora (approved):
That all looks good to me, including helper struct. Doing the infallible conversions to our datatypes in one place instead of interrupting code flow and assigning to previous declared stack slots improves readability.

PR #1418, opened 2021-01-31
Update readme, release notes, metadata for 0.23.13

HeroicKatora:
It's been a while since the last release but we have some major new features—notably `AVIF` support—as well as fixes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,30 @@ Rust image aims to be a pure-Rust implementation of various popular image format
 
 ## Changes
 
+### Version 0.23.13
+
+- Fix an inconsistency in supported formats of different methods for encoding
+  an image.
+- Fix `thumbnail` choosing an empty image. It now always prefer non-empty image
+  dimensions.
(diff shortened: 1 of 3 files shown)


HeroicKatora:
https://github.com/image-rs/image/pull/1417#issuecomment-771139188 On hold until we've figured out if there is a hole in our test coverage from the recent CI rework.

fseegraeber:
The `--no-default-features` seems to override single `--features` for `cargo test`.

The `println!` here is also kind of misleading:
https://github.com/image-rs/image/blob/6d55953830d5e70fceda8abb442f92cbf10d9522/tests/reference_images.rs#L153-L156
since if you run with e.g. `cargo test --no-default-features --features gif  -- --nocapture` it will print the path, but the actual test will not be executed:
https://github.com/image-rs/image/blob/6d55953830d5e70fceda8abb442f92cbf10d9522/tests/reference_images.rs#L193-L194

If I run all tests locally on the latest master there are no failures though.

PR #1419, opened 2021-01-31
Allow loading alpha-only TGAs as L8

### LINKED ISSUE #1392: Support for loading alpha-only TGA images
I would like to be able load TGA images that only contain alpha channel, i.e. num_alpha_bits=8, other_channel_bits=0, color=false. Some images in the wild use this encoding for whatever reason.

Potentially A8 could be added as an explicit ColorType so that alpha-only images could round-trip (decode and encode) without losing information. Perhaps alpha-only semantic information is in fact important for some applications, despite being somewhat esoteric.

The simplest/trivial solution is just https://github.com/kayru/image-rs-fixes/commit/a59fea12bfad77a2503f83cc2fbe84aaadb02b5d.
```
match (num_alpha_bits, other_channel_bits, color) {
    // ...
    (0, 8, false) => self.color_type = ColorType::L8,
    (8, 0, false) => self.color_type = ColorType::L8, // alpha-only image is treated as L8
    _ => {
    // ...
}
```

kayru:
Fixes #1392.

### DIFF
--- a/src/codecs/tga/decoder.rs
+++ b/src/codecs/tga/decoder.rs
@@ -54,6 +54,7 @@ pub struct TgaDecoder<R> {
 
     image_type: ImageType,
     color_type: ColorType,
+    original_color_type: Option<ExtendedColorType>,
 
     header: Header,
     color_map: Option<ColorMap>,
@@ -76,6 +77,7 @@ impl<R: Read + Seek> TgaDecoder<R> {
 
             image_type: ImageType::Unknown,
             color_type: ColorType::L8,
+            original_color_type: None,
 
             header: Header::default(),
             color_map: None,
@@ -153,7 +155,11 @@ impl<R: Read + Seek> TgaDecoder<R> {
             (0, 24, true) => self.color_type = ColorType::Rgb8,
             (8, 8, false) => self.color_type = ColorType::La8,
             (0, 8, false) => self.color_type = ColorType::L8,
-            (8, 0, false) => self.color_type = ColorType::L8, // alpha-only image is treated as L8
+            (8, 0, false) => { 
+                // alpha-only image is treated as L8
+                self.color_type = ColorType::L8;
+                self.original_color_type = Some(ExtendedColorType::A8);
+            },
             _ => {
                 return Err(ImageError::Unsupported(
                     UnsupportedError::from_format_and_kind(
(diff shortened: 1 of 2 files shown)


src/codecs/tga/decoder.rs
```diff
@@ -153,6 +153,7 @@ impl<R: Read + Seek> TgaDecoder<R> {
             (0, 24, true) => self.color_type = ColorType::Rgb8,
             (8, 8, false) => self.color_type = ColorType::La8,
             (0, 8, false) => self.color_type = ColorType::L8,
+            (8, 0, false) => self.color_type = ColorType::L8, // alpha-only image is treated as L8
```
HeroicKatora, line 156: This would be less confusing if the [`original_color_type`](https://docs.rs/image/0.23.12/image/trait.ImageDecoder.html#method.original_color_type) would indicate this as `A8` or a similarly named new variant of `ExtendedColorType` that clarifies that the color has been converted, or rather reinterpreted, as Luminance data.

Could you override that method in the impl of `ImageDecoder` below?

HeroicKatora (changes requested):

src/codecs/tga/decoder.rs
kayru: Will do. Originally I thought that adding A8 as a native format may set some expectations that I wasn't prepare to commit to implementing (such as full saving/loading) :) If it's acceptable to just add A8, do conversion and store original format, then I'm all for it.

HeroicKatora: Only add it to `ExtendedColorType`, not `ColorType`. That makes it clear that it is not fully and not directly supported but you can grab the information anyways for those image formats that care about the use case, including tga.

src/codecs/tga/decoder.rs
```diff
@@ -377,6 +383,13 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for TgaDecoder<R> {
         self.color_type
     }
 
+    fn original_color_type(&self) -> ExtendedColorType {
+        match self.original_color_type {
+            Some(original_color_type) => original_color_type,
+            None => self.color_type().into()
+        }
```
fintelia, line 390: This could just use `Option::unwrap_or` / `Option::unwrap_or_else`

HeroicKatora (approved):

PR #1421, opened 2021-02-02
Avif and default tests

HeroicKatora:
Note that this splits avif-feature into encoder and decoder.

They require different native dependencies in their build systems, one
of the available on Ubuntu-20.04 (the latest Github build environment)
while the other one is not.

Unblocks: #1418

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,6 +28,30 @@ jobs:
       if: ${{ matrix.rust != '1.34.2' }}
       env:
         FEATURES: ${{ matrix.features }}
+  test_defaults:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        override: true
+    - name: build
+      run: cargo build -v --release
+    - name: test
+      run: cargo test -v --release
+  test_avif:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: install-dependencies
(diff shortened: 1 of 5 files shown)


.github/workflows/rust.yml
```diff
@@ -28,6 +28,32 @@ jobs:
...
+      run: sudo apt update && sudo apt install nasm
+    - uses: actions/checkout@v2
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        override: true
+    - name: build
+      run: cargo build -v --no-default-features --features="avif"
+    # We don't test this yet, as it requires libdav1d which isn't in ubuntu 20.04
+    # run: cargo build -v --no-default-features --features="avif-decoding"
```
paolobarbolini, line 56: As described in the readme for dav1d-rs (https://github.com/rust-av/dav1d-rs#building) you may be able to work around this issue by setting the `SYSTEM_DEPS_DAV1D_BUILD_INTERNAL=always` environment variable to [make it git clone dav1d from source, compile it and link it statically](https://github.com/rust-av/dav1d-rs/blob/8d15e3a010412e3c8a98e3215f7af8d3ea0854f2/dav1d-sys/build.rs#L25-L55).

`ninja` and `meson` will have to be installed for this to work.

paolobarbolini (approved):
:+1:

.github/workflows/rust.yml
HeroicKatora: Ah, seems interesting. I'll try after the release.

Issue #1420, opened 2021-02-01
Panic with some TGA files


gsurrel:
This happens in the software:
- https://github.com/qarmin/czkawka/tree/master/czkawka_gui

Link to issue I opened there:
- https://github.com/qarmin/czkawka/issues/237#issuecomment-770440958

## Expected

The thread processing the images panics, which it should not.

## Actual behaviour

The thread panics, preventing the software to continue normally.

## Reproduction steps

Run czkawka_gui (available through cargo install, requires GTK libraries), and scan for "Broken files" in a firectory containing the [following zipped TGA image](https://github.com/image-rs/image/files/5902901/Aurinko_dazzle.tga.zip).

.

fintelia:
Could you capture a backtrace with the line of code where the panic is?

qarmin:
```
thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 2', /home/rafal/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/mod.rs:545:31
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/panicking.rs:495:5
   1: core::panicking::panic_fmt
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/panicking.rs:92:14
   2: core::panicking::panic_bounds_check
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/panicking.rs:69:5
   3: core::slice::<impl [T]>::swap
             at /home/rafal/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/mod.rs:545:31
   4: image::codecs::tga::decoder::TgaDecoder<R>::reverse_encoding
             at /home/rafal/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/codecs/tga/decoder.rs:292:21
   5: <image::codecs::tga::decoder::TgaDecoder<R> as image::image::ImageDecoder>::read_image
             at /home/rafal/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.12/./src/codecs/tga/decoder.rs:412:9
   6: image::image::decoder_to_vec
[...]

PR #1432, opened 2021-02-13
Issue 1428 tga

HeroicKatora:
See commit log for details. I will prepare some regression tests as well.

Fix: #1420 

This also found and fixed an additional crash where too much decoded image data was present.


HeroicKatora:
I started fuzzing and found a couple more panics.

PR #1437, opened 2021-02-14
Fix #1430

### LINKED ISSUE #1430: BmpDecoder::read_rect enters unreachable code
This happens in `image::load_rect` function because `BmpDecoder` implementation of `ImageDecoderExt` trait provides `seek_scanline` parameter which looks like this: `|_, _| unreachable!()`. But it is reachable in runtime.

## Expected
area of bmp image is read correctly by `image::load_rect` function

## Actual behaviour
Process panics with error:
```rust
thread 'main::tests::decode_bmp' panicked at 'internal error: entered unreachable code', /home/byshnev/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.23.13/./src/codecs/bmp/decoder.rs:1470:84
```

## Reproduction steps
```rust
    fn decode_bmp() {
        let f = File::open("path/to/any_rgba8.bmp").unwrap();
        let mut decoder = BmpDecoder::new(f).unwrap();

        let mut buf: Vec<u8> = vec![0; decoder.total_bytes() as usize];
        decoder.read_rect(0, 0, 8, 8, buf.as_bytes_mut()).unwrap(); // adjust width/height depending on image size
    }
```

fintelia:
Since `load_rect` unconditionally seeks back to the start of the image after loading, we can't have BmpDecoder panic when it is told to seek. Also adds a test to verify that this issue is fixed.

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -1467,7 +1467,7 @@ impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for BmpDecoder<R> {
         progress_callback: F,
     ) -> ImageResult<()> {
         let start = self.reader.seek(SeekFrom::Current(0))?;
-        image::load_rect(x, y, width, height, buf, progress_callback, self, |_, _| unreachable!(),
+        image::load_rect(x, y, width, height, buf, progress_callback, self, |_, _| Ok(()),
                          |s, buf| s.read_image_data(buf))?;
         self.reader.seek(SeekFrom::Start(start))?;
         Ok(())
@@ -1476,7 +1476,7 @@ impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for BmpDecoder<R> {
 
 #[cfg(test)]
 mod test {
-    use super::Bitfield;
+    use super::*;
 
     #[test]
     fn test_bitfield_len() {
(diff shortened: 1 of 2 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -1467,7 +1467,7 @@ impl<'a, R: 'a + Read + Seek> ImageDecoderExt<'a> for BmpDecoder<R> {
         progress_callback: F,
     ) -> ImageResult<()> {
         let start = self.reader.seek(SeekFrom::Current(0))?;
-        image::load_rect(x, y, width, height, buf, progress_callback, self, |_, _| unreachable!(),
+        image::load_rect(x, y, width, height, buf, progress_callback, self, |_, _| Ok(()),
                          |s, buf| s.read_image_data(buf))?;
         self.reader.seek(SeekFrom::Start(start))?;
```
HeroicKatora, line 1472: Hm, if `load_rect` wants to seek back to the start then I wonder if the callback might perform that relative seek, and these two lines be merged?

HeroicKatora: And wait a minute, does this work properly when you try to read any rect other than `x = 0`?

fintelia: The trick is that there's only 1 scanline total, so load_rect just skips the part of the image that isn't required

HeroicKatora: But what if it doesn't start at offset 0?

fintelia: Oh, I see what you're saying. What the code is _meant_ to do is read the scanline once and then copy it into the output buffer row by row. What it actually does is repeatedly read the same scanline (seeking first if necessary) and then copy one row out of it each time.

This part of the code could use significantly more testing...

HeroicKatora (approved):
Better, I guess? Oh man, we really need more exhaustive testing here.

PR #1445, opened 2021-03-05
Decoding limits try 2

birktj:
I have given decoding limits another try based on the latest discussion from #938. 

The current design uses a non-exhaustive `Limits` struct with public fields to provide an interface for setting decoding limits. It also defines a `LimitSupport` that decoders can use together with `Limits::check_support` in order to check if they support all of the limits that are set in a backwards-compatible way.

In the documentation of `Limits` I have written a little bit about the difference between *strict* and *non-strict* limits. The idea being that with non-strict limits decoders that support them should set them but without excluding decoders that do not support limits. For strict limits only decoders that can uphold the limits should be used.

This PR does intentionally not provide any `ImageDecoder::set_limit` implementations. This is to reduce its scope so we can keep the discussion centered around the `Limits` API as well as what kinds of guarantees should be provided by the `Limits` API. These implementations should be easy to add in follow-up PRs.

I also added a little commit to change the formatting of a couple of comments that my editors syntax-highlighting engine didn't like.

Some design questions:
- Should `Limits` and `LimitSupport` use methods instead of public fields to set decoding limits?
- Do the `max_image_width` and `max_image_height` limits make any sense?
- Does the current distinction between strict and non-strict limits make any sense, should there be a different set of guarantees?

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -754,7 +754,7 @@ impl DynamicImage {
     /// Resize this image using the specified filter algorithm.
     /// Returns a new image. The image's aspect ratio is preserved.
     /// The image is scaled to the maximum possible size that fits
-    /// within the bounds specified by ```nwidth``` and ```nheight```.
+    /// within the bounds specified by `nwidth` and `nheight`.
     pub fn resize(&self, nwidth: u32, nheight: u32, filter: imageops::FilterType) -> DynamicImage {
         let (width2, height2) =
             resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
@@ -764,7 +764,7 @@ impl DynamicImage {
 
     /// Resize this image using the specified filter algorithm.
     /// Returns a new image. Does not preserve aspect ratio.
-    /// ```nwidth``` and ```nheight``` are the new image's dimensions
+    /// `nwidth` and `nheight` are the new image's dimensions
     pub fn resize_exact(
         &self,
         nwidth: u32,
@@ -777,7 +777,7 @@ impl DynamicImage {
     /// Scale this image down to fit within a specific size.
     /// Returns a new image. The image's aspect ratio is preserved.
     /// The image is scaled to the maximum possible size that fits
-    /// within the bounds specified by ```nwidth``` and ```nheight```.
+    /// within the bounds specified by `nwidth` and `nheight`.
     ///
     /// This method uses a fast integer algorithm where each source
     /// pixel contributes to exactly one target pixel.
@@ -789,8 +789,8 @@ impl DynamicImage {
     }
 
     /// Scale this image down to a specific size.
-    /// Returns a new image. Does not preserve aspect ratio.
-    /// ```nwidth``` and ```nheight``` are the new image's dimensions.
+    /// Returns a new image. Does not peserve aspect ratio.
+    /// `nwidth` and `nheight` are the new image's dimensions.
     /// This method uses a fast integer algorithm where each source
     /// pixel contributes to exactly one target pixel.
     /// May give aliasing artifacts if new size is close to old size.
@@ -802,7 +802,7 @@ impl DynamicImage {
     /// Returns a new image. The image's aspect ratio is preserved.
     /// The image is scaled to the maximum possible size that fits
     /// within the larger (relative to aspect ratio) of the bounds
-    /// specified by ```nwidth``` and ```nheight```, then cropped to
+    /// specified by `nwidth` and `nheight`, then cropped to
     /// fit within the other bound.
     pub fn resize_to_fill(
         &self,
@@ -826,14 +826,14 @@ impl DynamicImage {
     }
 
     /// Performs a Gaussian blur on this image.
-    /// ```sigma``` is a measure of how much to blur by.
+    /// `sigma` is a measure of how much to blur by.
     pub fn blur(&self, sigma: f32) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::blur(p, sigma))
     }
 
     /// Performs an unsharpen mask on this image.
-    /// ```sigma``` is the amount to blur the image by.
-    /// ```threshold``` is a control of how much to sharpen.
+    /// `sigma` is the amount to blur the image by.
+    /// `threshold` is a control of how much to sharpen.
     ///
     /// See <https://en.wikipedia.org/wiki/Unsharp_masking#Digital_unsharp_masking>
     pub fn unsharpen(&self, sigma: f32, threshold: i32) -> DynamicImage {
@@ -850,14 +850,14 @@ impl DynamicImage {
     }
 
     /// Adjust the contrast of this image.
-    /// ```contrast``` is the amount to adjust the contrast by.
+    /// `contrast` is the amount to adjust the contrast by.
     /// Negative values decrease the contrast and positive values increase the contrast.
     pub fn adjust_contrast(&self, c: f32) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::contrast(p, c))
     }
 
     /// Brighten the pixels of this image.
-    /// ```value``` is the amount to brighten each pixel by.
+    /// `value` is the amount to brighten each pixel by.
     /// Negative values decrease the brightness and positive values increase it.
     pub fn brighten(&self, value: i32) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::brighten(p, value))
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -789,7 +789,7 @@ impl DynamicImage {
     }
 
     /// Scale this image down to a specific size.
-    /// Returns a new image. Does not peserve aspect ratio.
+    /// Returns a new image. Does not preserve aspect ratio.
     /// `nwidth` and `nheight` are the new image's dimensions.
     /// This method uses a fast integer algorithm where each source
     /// pixel contributes to exactly one target pixel.
--- a/src/image.rs
+++ b/src/image.rs
@@ -643,6 +643,7 @@ pub trait ImageDecoder<'a>: Sized {
     ///     decoder.read_image(buf.as_bytes());
     ///     buf
     /// }
+    /// ```
     fn read_image(self, buf: &mut [u8]) -> ImageResult<()> {
         self.read_image_with_progress(buf, |_| {})
     }
--- a/src/image.rs
+++ b/src/image.rs
@@ -686,14 +686,18 @@ pub trait ImageDecoder<'a>: Sized {
     /// limits that is possible to set.
     ///
     /// Note to implementors: make sure you call [`Limits::check_support`] so that 
-    /// decoding fails if any unsupported strict limits are set.
-    /// [`Limits::check_support`] will also check the `max_image_width` and
+    /// decoding fails if any unsupported strict limits are set. Also make sure
+    /// you call [`Limits::check_dimensions`] to check the `max_image_width` and
     /// `max_image_height` limits.
     ///
     /// [`Limits`]: ./io/struct.Limits.html
     /// [`Limits::check_support`]: ./io/struct.Limits.html#method.check_support
+    /// [`Limits::check_dimensions`]: ./io/struct.Limits.html#method.check_dimensions
     fn set_limits(&mut self, limits: crate::io::Limits) -> ImageResult<()> {
-        limits.check_support(self, crate::io::LimitSupport::default())?;
+        limits.check_support(&crate::io::LimitSupport::default())?;
+
+        let (width, height) = self.dimensions();
+        limits.check_dimensions(width, height)?;
 
         Ok(())
     }
--- a/src/image.rs
+++ b/src/image.rs
@@ -681,6 +681,22 @@ pub trait ImageDecoder<'a>: Sized {
 
         Ok(())
     }
+
+    /// Set decoding limits for this decoder. See [`Limits`] for the different kinds of 
+    /// limits that is possible to set.
+    ///
+    /// Note to implementors: make sure you call [`Limits::check_support`] so that 
+    /// decoding fails if any unsupported strict limits are set.
+    /// [`Limits::check_support`] will also check the `max_image_width` and
+    /// `max_image_height` limits.
+    ///
+    /// [`Limits`]: ./io/struct.Limits.html
+    /// [`Limits::check_support`]: ./io/struct.Limits.html#method.check_support
+    fn set_limits(&mut self, limits: crate::io::Limits) -> ImageResult<()> {
+        limits.check_support(self, crate::io::LimitSupport::default())?;
+
+        Ok(())
+    }
 }
 
 /// Specialized image decoding not be supported by all formats
--- a/src/io/free_functions.rs
+++ b/src/io/free_functions.rs
@@ -34,39 +34,76 @@ pub(crate) fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
 #[allow(unused_variables)]
 // r is unused if no features are supported.
 pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<DynamicImage> {
+    load_inner(r, super::Limits::default(), format)
+}
+
+pub(crate) trait DecoderVisitor {
+    type Result;
+    fn visit_decoder<'a, D: ImageDecoder<'a>>(self, decoder: D) -> ImageResult<Self::Result>;
+}
+
+pub(crate) fn load_decoder<R: BufRead + Seek, V: DecoderVisitor>(r: R, format: ImageFormat, visitor: V) -> ImageResult<V::Result> {
     #[allow(unreachable_patterns)]
     // Default is unreachable if all features are supported.
     match format {
         #[cfg(feature = "avif-decoder")]
-        image::ImageFormat::Avif => DynamicImage::from_decoder(avif::AvifDecoder::new(r)?),
+        image::ImageFormat::Avif => visitor.visit_decoder(avif::AvifDecoder::new(r)?),
         #[cfg(feature = "png")]
-        image::ImageFormat::Png => DynamicImage::from_decoder(png::PngDecoder::new(r)?),
+        image::ImageFormat::Png => visitor.visit_decoder(png::PngDecoder::new(r)?),
         #[cfg(feature = "gif")]
-        image::ImageFormat::Gif => DynamicImage::from_decoder(gif::GifDecoder::new(r)?),
+        image::ImageFormat::Gif => visitor.visit_decoder(gif::GifDecoder::new(r)?),
         #[cfg(feature = "jpeg")]
-        image::ImageFormat::Jpeg => DynamicImage::from_decoder(jpeg::JpegDecoder::new(r)?),
+        image::ImageFormat::Jpeg => visitor.visit_decoder(jpeg::JpegDecoder::new(r)?),
         #[cfg(feature = "webp")]
-        image::ImageFormat::WebP => DynamicImage::from_decoder(webp::WebPDecoder::new(r)?),
+        image::ImageFormat::WebP => visitor.visit_decoder(webp::WebPDecoder::new(r)?),
         #[cfg(feature = "tiff")]
-        image::ImageFormat::Tiff => DynamicImage::from_decoder(tiff::TiffDecoder::new(r)?),
+        image::ImageFormat::Tiff => visitor.visit_decoder(tiff::TiffDecoder::new(r)?),
         #[cfg(feature = "tga")]
-        image::ImageFormat::Tga => DynamicImage::from_decoder(tga::TgaDecoder::new(r)?),
+        image::ImageFormat::Tga => visitor.visit_decoder(tga::TgaDecoder::new(r)?),
         #[cfg(feature = "dds")]
-        image::ImageFormat::Dds => DynamicImage::from_decoder(dds::DdsDecoder::new(r)?),
+        image::ImageFormat::Dds => visitor.visit_decoder(dds::DdsDecoder::new(r)?),
         #[cfg(feature = "bmp")]
-        image::ImageFormat::Bmp => DynamicImage::from_decoder(bmp::BmpDecoder::new(r)?),
+        image::ImageFormat::Bmp => visitor.visit_decoder(bmp::BmpDecoder::new(r)?),
         #[cfg(feature = "ico")]
-        image::ImageFormat::Ico => DynamicImage::from_decoder(ico::IcoDecoder::new(r)?),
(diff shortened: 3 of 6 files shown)


src/io/mod.rs
```diff
@@ -1,5 +1,113 @@
...
+    /// The maximum allowed image width. This limit is strict. The default is no limit.
+    pub max_image_width: Option<u32>,
+    /// The maximum allowed image height. This limit is strict. The default is no limit.
+    pub max_image_height: Option<u32>,
+    /// The maximum allowed amount of memory to be allocated by the 
+    /// decoder at any one time. This limit is non-strict by default,
+    /// set `max_alloc_strict` to make this limit strict. The default 
+    /// is 512MiB. Note that limits below 1MiB are not supported, this 
+    /// is also the case if this limit is strict.
+    pub max_alloc: Option<u64>,
```
HeroicKatora, line 52: This is not entirely clear and it will be complicated with 'strict'. It says 'at any one time' but it omits whether this means 'calls to `alloc` specify at most size x' or 'calling decode will increased memory use by at most x'. 

The former is reasonbly easy and what decoders mostly implement at the moment—i.e. ancillary data, color tables, icc, are _not_ subtracted from the available budget. Just to be clear, I'd be in principle okay with calling all of these decoders non-strict!

However, the latter option is borderline impossible to implement and specify. Does it include memory reserved by the underlying allocator in anticipation of later use? Should it count stack usage? Is deallocated memory considered to be free again instantly? In case of an error, or an exceeded limit, should the use up to that point be report? This appears contradictory because the motivation for strictness is a guarantee in the first place but we can't specify the guarantee exactly. For strict limits there should also be an explanation that supporting them must come with 'essentially free' construction. The dilemma of `tiff` ifd-reader should not occur for very strict decoders.

The `Limits` struct is extensible so we could defer `strict` to a future iteration of limits?

src/image.rs
```diff
@@ -617,6 +618,22 @@ pub trait ImageDecoder<'a>: Sized {
...
+    /// [`Limits::check_support`] will also check the `max_image_width` and
+    /// `max_image_height` limits.
+    ///
+    /// [`Limits`]: ./io/struct.Limits.html
+    /// [`Limits::check_support`]: ./io/struct.Limits.html#method.check_support
+    fn set_limits(&mut self, limits: crate::io::Limits) -> ImageResult<()> {
+        limits.check_support(self, crate::io::LimitSupport::default())?;
+
+        Ok(())
+    }
```
HeroicKatora, line 636: The `Limits` struct could grow to quite a size, it should be passed by reference if this is feasible. Or did I overlook a good reason to pass it by value?

Also, what other kinds of implementations of this method and in particular the check might there be? It seems odd—but most flexible—that the decoder might be the responsible for performing it. Did you consider the alternative

```rust
/// Enable support for the limit types requested, return support matrix.
fn limit_support(&mut self, limits: &Limits) -> ImageResult<LimitSupport> {
    LimitSupport::default()
}
```

Note that I'm _not_ asking you to change it right away, I'm wondering about tradeoffs of either interface.

src/io/mod.rs
```diff
@@ -1,5 +1,113 @@
...
 pub(crate) mod free_functions;
 
 pub use self::reader::Reader;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+/// Set of supported strict limits for a decoder
+pub struct LimitSupport {
+    /// This field indicates whether the strict limit on the number of bytes is
+    /// supported.
+    pub max_alloc_strict: bool,
```
HeroicKatora, line 15: Notwithstanding the comment on strictness of memory allocation above, I think there should be some way to distinguish between a best-effort enforcement and disregarding the limit entirely. Should this field be a tri-state bool or, in case strictness is deferred to another PR, be split into two?

src/io/mod.rs
```diff
@@ -1,5 +1,113 @@
...
+impl Default for LimitSupport {
+    fn default() -> LimitSupport {
+        LimitSupport {
+            max_alloc_strict: false,
+            _non_exhaustive: (),
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
```
HeroicKatora, line 28: I know clippy will be annoying but I wouldn't derive `Copy` here, on purpose. Note that any future limit that requires an allocation space—consider a boxed callback—would be a breaking change otherwise.

HeroicKatora (commented):
This looks great, I like how `LimitSupport` moves the logic of comparing the requirements and guarantees into `image` itself.

> Should Limits and LimitSupport use methods instead of public fields to set decoding limits?

It seems perfectly fine as is. The limit parameter do not require maintaining an internal invariant and its field seem to be largely orthogonal.

> Do the max_image_width and max_image_height limits make any sense?

Yes

> Does the current distinction between strict and non-strict limits make any sense, should there be a different set of guarantees?

See below.

src/image.rs
fintelia: I'd imagine that a lot of decoders would want to hold onto a copy of limits, which would favor either an owned or boxed value.

src/io/mod.rs
```diff
@@ -1,5 +1,113 @@
...
+/// limits the library makes a stronger guarantee that the limit will not be exceeded.
+/// Exceeding a strict limit is considered a critical bug. If a decoder cannot 
+/// guaranree that it will uphold a strict limit it *must* fail with 
+/// `image::error::LimitErrorKind::Unsupported`.
+///
+/// The limit check should only ever fail if a limit will be exceeded or an unsupported strict
+/// limit is used.
+pub struct Limits {
+    /// The maximum allowed image width. This limit is strict. The default is no limit.
+    pub max_image_width: Option<u32>,
```
fintelia, line 44: It would save a some space to make these `Option<NonZeroU32>`s. Though it also would be slightly less ergonomic, so I could be persuaded either way

src/io/mod.rs
```diff
@@ -1,5 +1,113 @@
...
+use crate::{error, ImageError, ImageResult};
+
 mod reader;
 pub(crate) mod free_functions;
 
 pub use self::reader::Reader;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+/// Set of supported strict limits for a decoder
+pub struct LimitSupport {
```
fintelia, line 12: Should there be a 1-1 mapping with fields in the limits struct?

src/image.rs
```diff
@@ -617,6 +618,22 @@ pub trait ImageDecoder<'a>: Sized {
...
+    /// Note to implementors: make sure you call [`Limits::check_support`] so that 
+    /// decoding fails if any unsupported strict limits are set.
+    /// [`Limits::check_support`] will also check the `max_image_width` and
+    /// `max_image_height` limits.
+    ///
+    /// [`Limits`]: ./io/struct.Limits.html
+    /// [`Limits::check_support`]: ./io/struct.Limits.html#method.check_support
+    fn set_limits(&mut self, limits: crate::io::Limits) -> ImageResult<()> {
+        limits.check_support(self, crate::io::LimitSupport::default())?;
+
```
fintelia, line 634: This could/should also check the width and height, right?

src/io/mod.rs
birktj: The idea was that max_image_width are assumed strict by default and that all decoders support it. As far as I know `ImageDecoder::dimension` is assumed to be correct for all decoders? If so I don't really see any case where one is not able to set strict limits on image size.

src/image.rs
birktj: `Limits::check_support` takes the decoder by reference so that it can check width and height. I thought this would be simpler as this is a check all decoders would have to do.

src/io/mod.rs
birktj: Since this is part of the user facing api I don't think the extra 4 bytes saved is worth it.

src/io/mod.rs
birktj: This seems like a good point

src/image.rs
birktj: Passing by reference is probably a good idea, I think most decoders are going to construct their own limit struct and only need access to the values. 

The idea was of course that decoders would enable their own limits internal in `set_limits`, however there is no reason  this can't be done in `limit_support`. I did think about the possibility of having a method of type `fn limit_support(&self) -> LimitSupport`, but did not see enough of a reason to add it. This may be a better alternative.

src/io/mod.rs
birktj: What is was thinking was something among the lines of 'accumulated calls to malloc and free (where we think of free as reserving negative memory) will never grow past `max_alloc`'. I did think about saying something about it allowing usage of '1MiB or 10%, whatever is greater' outside of the specified amount, but then the decoder might as well just remove 10% on its own.

The idea is that strict memory limits are more or less **strict** in the sense that if it is important for you that  memory usage is limited you should be able to set strict and be confident that decoder will not potentially try to allocate more memory than specified. This would of course currently lead to only a subset of decoders supporting strict limits and the docs should probably reflect that, but I was afraid of adding a table that would quickly get outdated.

Deferring `strict` to later versions could be an idea, however this would mean that `LimitSupport` would be empty for now.

src/io/mod.rs
birktj: This may not have been entirely clear, but the idea was that you first set the limit and it is supported everywhere. However if you need to make sure that memory usage is limited for some reason you should be able to enable strict versions of limits. This means that if guaranteeing that usage is below the limit is not possible, decoding should fail, but *only* if the strict version is enabled. The best-effort enforcement is by default.

src/image.rs
fintelia: Ah, I see. Didn't realize that from the name

src/io/mod.rs
fintelia: Well, 16 bytes if you include the other fields which also have zero as a nonsensical limit.

src/image.rs
fintelia: Returning a LimitSupport more error-prone to me, and also kind of like leaking an "internal" implementation detail. I'd prefer if only implementors of ImageDecoder needed to think about forward compatibility of limit restrictions (and the associated LimitSupport type) while callers to set_limits just needed to know that they'd get an unsupported error if they tried to set a limit they weren't allowed to.

birktj: That is a very good point. I would think external users of this library are far more likely to want to use `set_limits` than they are to implement `ImageDecoder`. If `set_limits` returns the `LimitSupport` it becomes very awkward to use

HeroicKatora: Agreed, `set_limits` appears more ergonomic overall. With regards to copying, decoders will want to hold onto only those limits they support and not the full non-exhaustive struct. And the caller may want to reuse the limits. None of this is really consequential when the `Copy` bounds stay, but otherwise I'd rather have:

> fn set_limits(&mut self, limits: &crate::io::Limits) -> ImageResult<()>

src/io/mod.rs
HeroicKatora: What I was trying to say is, there are three versions of memory limit support:
* Decoders that don't care (the non-strict version here?). Using them to decode untrusted input is _never_ okay for a caller that is concerned about reliability of their programs.
* Decoders that haven't been written with a guarantee in mind, or that have been written to a slightly diverging specification of what the limit means, but will still use the limit as a guideline. A caller might decide that using those is fine with a decently large safety margin.
* Decoders that really _guarantee_ what we've specified that `strict` should mean.

birktj: Couldn't the decoder set their own safety margin in that case? If they set an internal limit of 1GiB but might end up allocating 5GiB they might as well not have any limit. However on the other hand if we are reasonably sure that the are not going use more than 2x the limit it could first divide its own limit by 2 before setting it.

Thinking about it some more we may need to do some redesign of most (all?) of the limits of the various decoder to support the strict limits in way I was thinking. If so I am supportive of only having non-strict limits for now. I did add a simple check for `total_bytes` in `load` which should give us a little protection to begin with and adding other limits for the decoders we already have in a 

A way forward would probably be to introduce some sort of `limited-allocator` crate that we can integrate into the various decoders. And also have a shared `Limits` struct in this crate so all decoders can use a similar API. Am I correct in assuming that we could switch `image::io::Limits` to one reexported from another crate in a backwards-compatible way as long as we design the new one in an identical way?

src/image.rs
birktj: Do we ever need to reuse the limits after calling `set_limits` though?

src/io/mod.rs
HeroicKatora: Not really, no. It's less about some factor but rather: The decoder calls into a decompression library or helper for X and it suspects that this lib will setup _some_ reasonable buffer for an operation but it won't or can't control the size of the buffer exactly. Then it can not argue for strict support but one can argue that it will reasonably ensure that the limit is adhered to. Note that it's not about a multiplicative margin but rather some fixed overhead that it might not have a lot of control over. (See also the stack-depth question).

> Am I correct in assuming that we could switch image::io::Limits to one reexported from another crate in a backwards-compatible way as long as we design the new one in an identical way?

No, strictly speaking we can't change a struct definition to a re-export without a breaking change as one might have local trait impls for both which would then conflict. We can however provide trivial converters as we do for `color_quant`.

HeroicKatora: Note that the argument of unstrict or unknown library behavior already applies to `Vec::extend_from_slice` or `Vec::push`. Both do _some_ allocation and it very likely doubles the memory usage but we can't rely on it with a guarantee. Now that gets even more complicated for `HashMap`s or anything more complicated. It's also very hard to control exactly. A library would be indeed quite interesting to have for that use case.

birktj:
I have updated the implementation to among other things remove the `max_alloc_strict`. It seems that it will be quite difficult to decide on what this is supposed to mean and the whole idea behind the design is that it should be able to add these sort of new limits in a backwards-compatible way.

src/io/mod.rs
```diff
@@ -1,5 +1,123 @@
...
+        Ok(())
+    }
+
+    /// This function increases the `max_alloc` limit with amount. Should only be used
+    /// togheter with [`reserve`].
+    ///
+    /// [`reserve`]: #method.reserve
+    pub fn free(&mut self, amount: u64) {
+        if let Some(max_alloc) = self.max_alloc.as_mut() {
+            *max_alloc -= std::cmp::min(*max_alloc, amount);
```
HeroicKatora, line 120: This contradicts the documentation. Did you mean: `max_alloc = max_alloc.saturating_add(amount)`

birktj: Oops

src/io/free_functions.rs
```diff
@@ -56,39 +56,67 @@ pub(crate) fn open_impl(path: &Path) -> ImageResult<DynamicImage> {
...
+        image::ImageFormat::Dds => visitor.visit_decoder(dds::DdsDecoder::new(r)?),
         #[cfg(feature = "bmp")]
-        image::ImageFormat::Bmp => DynamicImage::from_decoder(bmp::BmpDecoder::new(r)?),
+        image::ImageFormat::Bmp => visitor.visit_decoder(bmp::BmpDecoder::new(r)?),
         #[cfg(feature = "ico")]
-        image::ImageFormat::Ico => DynamicImage::from_decoder(ico::IcoDecoder::new(r)?),
+        image::ImageFormat::Ico => visitor.visit_decoder(ico::IcoDecoder::new(r)?),
         #[cfg(feature = "hdr")]
-        image::ImageFormat::Hdr => DynamicImage::from_decoder(hdr::HdrAdapter::new(BufReader::new(r))?),
+        image::ImageFormat::Hdr => visitor.visit_decoder(hdr::HdrAdapter::new(BufReader::new(r))?),
```
birktj, line 92: Is the reader supposed to be wrapped in a `BufReader` here? I see that the reader already has ` R: BufRead` bound. Might as well change it in this PR if it is not supposed to be this way?

src/io/mod.rs
```diff
@@ -1,5 +1,123 @@
...
+/// Resource limits for decoding.
+///
+/// Limits can be either *strict* or *non-strict*. Non-strict limits are best-effort
+/// limits where the library does not guarantee that limit wont be exceeded. Do note 
+/// that it is still considered a bug if a non-strict limit is exceeded, however as 
+/// some of the underlying decoders do not support not support such limits one cannot 
+/// rely on these limits being supported. For stric limits the library makes a stronger 
+/// guarantee that the limit will not be exceeded. Exceeding a strict limit is considered 
+/// a critical bug. If a decoder cannot guaranree that it will uphold a strict limit it 
+/// *must* fail with `image::error::LimitErrorKind::Unsupported`.
```
birktj, line 34: Any suggestions for alternate wording here. If we go ahead and merge this PR without any strict limits this may be slightly confusing?

HeroicKatora: Maybe this should allude to the potential to introduce strict limits in the future, noting that the `LimitSupport` will default to false, and that decoders should turn them to true if they can consider their violation a critical bug.

src/io/mod.rs
```diff
@@ -1,5 +1,123 @@
...
+        LimitSupport {
+            _non_exhaustive: (),
+        }
+    }
+}
+
+/// Resource limits for decoding.
+///
+/// Limits can be either *strict* or *non-strict*. Non-strict limits are best-effort
+/// limits where the library does not guarantee that limit wont be exceeded. Do note
```
fintelia, line 28: nit: wont -> won't

src/io/mod.rs
```diff
@@ -1,5 +1,123 @@
...
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+#[allow(missing_copy_implementations)]
+pub struct Limits {
+    /// The maximum allowed image width. This limit is strict. The default is no limit.
+    pub max_image_width: Option<u32>,
+    /// The maximum allowed image height. This limit is strict. The default is no limit.
+    pub max_image_height: Option<u32>,
+    /// The maximum allowed amount of memory to be allocated by the 
+    /// decoder at any one time. This limit is non-strict and some 
+    /// decoders may ignore it. The default is 512MiB.
```
fintelia, line 47: This text is a bit ambiguous about whether "at one time" means that each individual allocation must be less than the limit, or whether the sum across all allocations that haven't been freed yet must be below the limit.

src/image.rs
HeroicKatora: Not in the current `io::Reader` but I consider that a more general caller that uses the interface might want to.

src/io/mod.rs
fintelia: I quite like the idea of having each limit be a a tri-state bool. The non-strict variant could be called "best-effort" with a specification something like "exceeds the limit by no more than a small fixed (but unspecified) factor plus some additional constant amount".

fintelia:
I quite like this design. It combines a very logical end-user API for ImageDecoder with sufficient hooks for downstream libraries to provide forward compatibility.

The main think I'd to give further thought to is the idea of letting the user specify whether each limit is strict or best-effort.

fintelia:
What is still left on this? I did a quick skim through the code, and while the interface might not be perfect, it does seem good enough to be usable. Certainly far better than the current status of not having limits.

birktj:
I think it should be more or less good to go as long as the docs make sense and don't have any spelling mistakes. I interpreted the dying discussion to mean that this design was not quite what we wanted. Also deciding what guarantees we can possibly make proved difficult.

Note that as previously discussed this PR intentionally has a limited scope and more useful limits are intended to be added later.

fintelia:
I think there are a handful of typos, but we can do a proofreading pass once the main technical questions are resolved.

@HeroicKatora how do you feel about going forward with a design substantially like what's proposed now?

src/image.rs
HeroicKatora: My only outstanding concern is over deciding between:

* `ImageDecoder { fn limit_support(&mut self, limits: &Limits) }`
* `ImageDecoder { fn limit_support(&mut self, limits: Limits) }`

However, we've agreed on have mutation methods on `Limits`. This makes the first alternative more brittle since it allows a use where the decoder clones its instance while the caller reuses the original instance, effectively duplicating the resources and thus subverting the limit check. So, let's keep it precisely how it was proposed here.

HeroicKatora:
If you all agree on the resolution of my previously opened outstanding concern by just closing it then there shouldn't be any more API concerns left. So yes, we can go ahead.

src/dynimage.rs
```diff
@@ -788,8 +788,8 @@ impl DynamicImage {
     }
 
     /// Scale this image down to a specific size.
-    /// Returns a new image. Does not preserve aspect ratio.
-    /// ```nwidth``` and ```nheight``` are the new image's dimensions.
+    /// Returns a new image. Does not peserve aspect ratio.
```
fintelia, line 791: peserve -> preserve

src/io/mod.rs
```diff
@@ -1,5 +1,130 @@
...
+/// Limits can be either *strict* or *non-strict*. Non-strict limits are best-effort
+/// limits where the library does not guarantee that limit will not be exceeded. Do note 
+/// that it is still considered a bug if a non-strict limit is exceeded, however as 
+/// some of the underlying decoders do not support not support such limits one cannot 
+/// rely on these limits being supported. For stric limits the library makes a stronger 
+/// guarantee that the limit will not be exceeded. Exceeding a strict limit is considered 
+/// a critical bug. If a decoder cannot guarantee that it will uphold a strict limit it 
+/// *must* fail with `image::error::LimitErrorKind::Unsupported`.
+///
+/// Currently there are no strict limits supported, however they will be added in the
```
fintelia, line 36: This isn't true. max_image_{width, height} are strict

src/io/mod.rs
```diff
@@ -1,5 +1,130 @@
...
+    /// some decoders may ignore it. The default is 512MiB.
+    pub max_alloc: Option<u64>,
+    _non_exhaustive: (),
+}
+
+impl Default for Limits {
+    fn default() -> Limits {
+        Limits {
+            max_image_width: None,
+            max_image_height: None,
```
fintelia, line 63: Could we set these limits to say 32768?

fintelia (changes requested):
Add a couple requested tweaks. If you could address those and fix the merge conflicts I'm inclined to just go ahead with merging this. Don't want to let the perfect be the enemy of the good

src/io/mod.rs
birktj: Are there any good arguments for this?

birktj:
I think I addressed most of your concerns. I am not convinced having default limits on image size is that useful though, but if you have any good arguments I will easily change my mind.

src/io/mod.rs
fintelia: It would mean the default ImageDecoder::set_limits() would at least catch truly outrageous image dimensions

PR #1448, opened 2021-03-06
Metadata

197g:
Adds a common interface for decoding meta data.

The `Metagram` struct, storing all meta data, is partly public and has a `Default` implementation that should be interpreted as having no meta data. The goal is that the metadata can later again be used by encoders, and in any case there are no invariants between the fields. Also note that EXIF and ICC are treated as opaque blobs for now. We do _not yet_ interpret the color conversion and source information given in them to inform the conversion to the `rgb`ish types of `DynamicImage`¹.

This considers two decoder implementations: Either the data is immediately available or it will be supplied during the decoding process. In the first case the decoders has already decoded all data from a header and stores it directly into a `Recoder` borrowed from the caller while in the second case they want to keep an instance for the following decoding call, realized with a `SharedRecorder` that provides thread-safe shared access to a single target. 

The design also considers recursive use in that a `Recorder` can share the target of a `SharedRecorder`. This allows any decoder, holding on to a `SharedRecorder`, to create a new argument that can be passed to another decoder's `ImageDecoder::metagram` method.

The combined encoding-decoding/decoding-encoding process may be lossy in all instances, for the moment it is a best-effort. There is no attempt at adding any impls in this PR.

------

¹Reasoning: We don't have library integration for this yet. Furthermore, it would affect downstream users to change the output color types too suddenly. Instead, my traint of thought is to establish that in the future both original and output color spaces will be defined in `Metagram` and make those fields as accurate as possible. That preparse decoders for the flexibility necessary to transparently handle a change in the decoding interpretation, but the discussion on this is not part of the PR.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -3,6 +3,7 @@ use std::ops::{Index, IndexMut};
 use num_traits::{NumCast, ToPrimitive, Zero};
 
 use crate::traits::{Pixel, Primitive};
+use crate::utils::NonExhaustiveMarker;
 
 /// An enumeration over supported color types and bit depths
 #[derive(Copy, PartialEq, Eq, Debug, Clone, Hash)]
@@ -31,12 +32,11 @@ pub enum ColorType {
     Bgra8,
 
     #[doc(hidden)]
-    __NonExhaustive(crate::utils::NonExhaustiveMarker),
+    __NonExhaustive(NonExhaustiveMarker),
 }
 
 /// Color information of an image's texels.
 #[derive(Clone, Debug, PartialEq)]
-#[non_exhaustive]
 pub enum Color {
     /// The color space is given by an encoded ICC profile.
     ///
@@ -65,6 +65,9 @@ pub enum Color {
         /// The absolute luminance of the values in the color space.
         luminance: Luminance,
     },
+
+    #[doc(hidden)]
+    __NonExhaustive(NonExhaustiveMarker),
 }
 
 /// Transfer functions from encoded chromatic samples to physical quantity.
@@ -74,7 +77,6 @@ pub enum Color {
 /// describes how scene lighting is encoded as an electric signal. These are applied to each
 /// stimulus value.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-#[non_exhaustive]
 pub enum Transfer {
     /// Specified in ITU Rec.709.
     Bt709,
@@ -96,22 +98,26 @@ pub enum Transfer {
     Bt2100Pq,
     /// ITU Rec.2100 Hybrid Log-Gamma.
     Bt2100Hlg,
+
+    #[doc(hidden)]
+    __NonExhaustive(NonExhaustiveMarker),
 }
 
 /// The reference brightness of the color specification.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-#[non_exhaustive]
 pub enum Luminance {
     /// 100cd/m².
     Sdr,
     /// 10_000cd/m².
     /// Known as high-dynamic range.
     Hdr,
+
+    #[doc(hidden)]
+    __NonExhaustive(NonExhaustiveMarker),
 }
 
 /// The relative stimuli of the three corners of a triangular gamut.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-#[non_exhaustive]
 pub enum Primaries {
     Bt601_525,
     Bt601_625,
@@ -119,13 +125,18 @@ pub enum Primaries {
     Smpte240,
     Bt2020,
     Bt2100,
+
+    #[doc(hidden)]
+    __NonExhaustive(NonExhaustiveMarker),
 }
 
 /// The whitepoint/standard illuminant.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-#[non_exhaustive]
 pub enum Whitepoint {
     D65,
+
+    #[doc(hidden)]
+    __NonExhaustive(NonExhaustiveMarker),
 }
 
 impl ColorType {
@@ -247,7 +258,7 @@ pub enum ExtendedColorType {
     Unknown(u8),
 
     #[doc(hidden)]
-    __NonExhaustive(crate::utils::NonExhaustiveMarker),
+    __NonExhaustive(NonExhaustiveMarker),
 }
 
 impl ExtendedColorType {
--- a/src/color.rs
+++ b/src/color.rs
@@ -34,6 +34,100 @@ pub enum ColorType {
     __NonExhaustive(crate::utils::NonExhaustiveMarker),
 }
 
+/// Color information of an image's texels.
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum Color {
+    /// The color space is given by an encoded ICC profile.
+    ///
+    /// This is a superset of other options but the consumer must itself decode and extract the
+    /// values. They should indicate an error similar to a completely unsupported color space in
+    /// case this fails.
+    Icc {
+        /// The binary ICC data.
+        profile: Vec<u8>,
+    },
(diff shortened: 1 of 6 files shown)


jrmuizel:
How should something like png's sRGB chunk show up here? Also why `Metagram` instead of `Metadata`?

jrmuizel:
Maybe `color_profile` should be `color_space` which would be an enum something like:
```rust
enum ColorSpace {
   SRGB,
   GammaAndChromaticity(GammaAndChromaticity),
   ICC(Vec<u8>),
   Unknown
}

struct GammaAndChromaticity {
   gamma: f32,
   white: XY,
   red: XY,
   green: XY
}
```

HeroicKatora:
Color profile is modelled after the representation in AV1, the 'newest' of the image formats.

HeroicKatora:
Color profile refers to an ICC profile and we do not (yet) want to interpret it. Its internal format is a TIF-like directory structure encoded which takes quite a bit to decode fully. This can also be done afterwards so I didn't want to task the decoder itself with it.

jrmuizel:
Yeah, I'm familiar with ICC profiles (I wrote [qcms](https://github.com/FirefoxGraphics/qcms/)).

I wouldn't recommend having the decoders construct ICC profiles for color spaces that have a simpler representation. If an application ever wants to know or display the name of a color space it's lot easier do that with a named constant vs trying to reconstruct it from an ICC profile. For example, if you want to avoid the cost of color conversion it's a lot easier to compare names than it is ICC profiles.

FWIW, here are some examples of other applications representing color spaces by name:
[Chrome](https://source.chromium.org/chromium/chromium/src/+/master:ui/gfx/color_space.h;l=49;drc=befd94f013d07347860a9269ae5ad692cc319829)
[AVIF](https://github.com/AOMediaCodec/libavif/blob/741b2a050962a21d68b7c3819234ec88c110250a/include/avif/avif.h#L221)
[CoreGraphics](https://developer.apple.com/documentation/coregraphics/cgcolorspace/color_space_names?language=objc)

HeroicKatora:
Good to know, thanks for the input. Out of all the image formats, AVIF's representation seems most reasonable with regards to extensibility (although the enum approach as proposed could rule out some invalid representations). It introduces quite a lot of new enums for the representation. I already have a draft for this as well, for the description of texels to be specific, in a private project. Would you say it should be part of this PR, then I'd add it here.

jrmuizel:
Yeah, the AVIF approach seems reasonable. JPEG-XL uses something similar:
https://gitlab.com/wg1/jpeg-xl/-/blob/master/lib/include/jxl/color_encoding.h

src/color.rs
```diff
@@ -34,6 +34,100 @@ pub enum ColorType {
...
+    Bt601,
+    Smpte240,
+    /// Also known as the identity function.
+    Linear,
+    /// The common sRGB standard which is close to standard 'gamma correction'.
+    Srgb,
+    /// ITU Rec.2020 with 10 bit quantization.
+    Bt2020_10bit,
+    /// ITU Rec.2020 with 12 bit quantization.
+    Bt2020_12bit,
```
jrmuizel, line 94: Does Bt2020_10bit use a different transfer function than Bt2020_12bit?

HeroicKatora: Not mathematically, but the quantization is different. Anyways this is straightup copied from the `transfer_characteristics` field of AV1.

src/color.rs
```diff
@@ -31,7 +32,111 @@ pub enum ColorType {
...
+/// The relative stimuli of the three corners of a triangular gamut.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Primaries {
+    Bt601_525,
+    Bt601_625,
+    Bt709,
+    Smpte240,
+    Bt2020,
+    Bt2100,
+
```
johannesvollmer, line 128: Would it make sense to also (or even exclusively?) provide an unrestricted primaries type? In .exr images, the primaries are defined by exactly positioning the three triangle corners on the  CIE standard chromaticity diagram (using three Vec2f, one for each r, g, and b). 

(This is not the problem of the image crate, but extracting the current Primaries enum would require reverse-detecting the standard primaries from the Vec2f values, which might also happen in other formats)

HeroicKatora: I had considered this but didn't want to decide whether to have variants with fields, or to just put the fields on `Metagram` where they could be filled in automatically when one of the pre-defined variants is chosen. I'm also not exactly sure what the best representation would be (`xy`-factors? transformation matrix from CIE XYZ?) so I punted that decision.

fintelia (commented):
The `Color` handling part of this looks quite plausible to me, though I'm not an expert on color spaces so I can't actually say whether it is correct in that regard.

I'm less convinced by the `Recorder` aspect. Have you considered alternatives like say an `ImageDecoder::read_image_metadata` function that returned the image data and matadata at the same time?

src/io/metagram.rs
```diff
@@ -0,0 +1,220 @@
+use std::cell::RefCell;
+use std::sync::{Arc, Mutex};
+
+use crate::color::Color;
+
+// oder (Engram), Metagram,
```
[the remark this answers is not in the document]
HeroicKatora: Yes, comment is to be removed in a cleanup commit.

src/io/metagram.rs
```diff
@@ -0,0 +1,220 @@
...
+/// Collects some arbitrary meta data of an image.
+///
+/// Note that information collected here will, per default, appear opaque to the `image` library
+/// itself. For example, the width and height of an image might be recorded as one set of values
+/// that's completely different from the reported dimensions in the decoder. During decoding and
+/// writing of images the library may ignore the color profile and exif data. You should always
+/// recheck with the specific decoder if you require color accuracy beyond presuming sRGB. (This
+/// may be improved upon in the future, if you have a concrete draft please open an issue).
+#[derive(Clone, Default)]
+pub struct Metagram {
```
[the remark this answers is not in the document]
HeroicKatora: I debated this in my initial design. The `metadata` would refer to each of the individual pieces of information but not the whole grouping. This struct's name is supposed to suggest that it contains all structural and process metadata that has been available during the decoding process or is available for encoding. I don't know, naming is hard :joy:

src/io/metadata.rs
```diff
@@ -26,10 +25,37 @@ pub struct Metagram {
...
+/// commonly associated with decoding of additional chunks, validation, and additional resource
+/// usage. All of these can be opt-out by allowing such a customization. Note that, nevertheless,
+/// the absence of metadata is not definitive proof of its absence in the original file. It might
+/// be more recent than the decoder and not yet have support.
+#[derive(Clone, Default)]
+// We may want to add a non-copy field?
+#[allow(missing_copy_implementations)]
+pub struct RecorderConfig {
+    /// Whether the decoder should try to provide color profile data.
+    pub color_profile: DatumRequested,
```
fintelia, line 49: Any particular reason this couldn't be `pub skip_color_profile: bool`?

HeroicKatora: Yes, but it took me a while to make up my own mind. My vague idea was to have a third variant, `Required`, that would _fail_ decoding when no information is available. This could allow decoders to fail fast instead of needing to wait for the final check which would complement `Limits` nicely. However, there are some pain points:

* We will have to add a check _after_ the decoder in any case if we want to make the an actionable guarantee for the caller. But when those checks trigger we will merely discard already correct data which is rather odd and mostly detrimental.
* If we have any sort of recovery or fallback then we need to relax it to `Requested` in any case with similar concerns as above.
* There might be overlapping constraints such as 'ensure a correct, complete and accurate record of all color conversion and quantization steps in the image creation pipeline' which wouldn't fit in the tri-state idea in any case.
* They can always be converted to a flag on the recorded `MetadataContainer`. Consider the above then we could add a `bool` attribute that signals the completeness of the color information in other fields.

This leaves only code clarity as a potential benefit of an explicit new enum.

PR #1451, opened 2021-03-10
Update to num-rational 0.4

sdroege:

----

This might need special care if num-rational is used in any public API.


HeroicKatora (approved):
It should not be part of the public dependencies anymore.

PR #1457, opened 2021-03-23
Fix and improve gif decoding in GifDecoder::read_image

### LINKED ISSUE #1455: Panic when decoding a gif with oversized frame
Panic when decoding a gif with a frame that exceeds the logical screen size but doesn't have an offset.

## Expected

The resulting image should be frame content that lies in the logical screen dimensions.

## Actual behaviour

```
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca\/library\std\src\panicking.rs:495
   1: core::panicking::panic_fmt
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca\/library\core\src\panicking.rs:92
   2: core::slice::index::slice_end_index_len_fail
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca\/library\core\src\slice\index.rs:41
   3: core::slice::index::{{impl}}::index_mut<u8>
             at C:\Users\Felix\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\slice\index.rs:249
   4: core::slice::index::{{impl}}::index_mut<u8>
             at C:\Users\Felix\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\slice\index.rs:289
   5: core::slice::index::{{impl}}::index_mut<u8,core::ops::range::RangeTo<usize>>
             at C:\Users\Felix\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\slice\index.rs:26
   6: gif::reader::Decoder<std::io::cursor::Cursor<slice<u8>*>>::read_into_buffer<std::io::cursor::Cursor<slice<u8>*>>
             at C:\Users\Felix\.cargo\registry\src\github.com-1ecc6299db9ec823\gif-0.11.1\src\reader\mod.rs:273
   7: image::codecs::gif::{{impl}}::read_image<std::io::cursor::Cursor<slice<u8>*>>
             at .\src\codecs\gif.rs:111
   8: image::image::decoder_to_vec<u8,image::codecs::gif::GifDecoder<std::io::cursor::Cursor<slice<u8>*>>>
             at .\src\image.rs:498
   9: image::dynimage::decoder_to_image<image::codecs::gif::GifDecoder<std::io::cursor::Cursor<slice<u8>*>>>
             at .\src\dynimage.rs:1064
  10: image::dynimage::DynamicImage::from_decoder<image::codecs::gif::GifDecoder<std::io::cursor::Cursor<slice<u8>*>>>
             at .\src\dynimage.rs:181
  11: image::io::free_functions::load<std::io::cursor::Cursor<slice<u8>*>>
             at .\src\io\free_functions.rs:67
  12: image::dynimage::load_from_memory_with_format
             at .\src\dynimage.rs:1295
  13: image::dynimage::load_from_memory
             at .\src\dynimage.rs:1280
  14: image::dynimage::test::frames_exceeding_logical_screen_size2
             at .\src\dynimage.rs:1348
  15: image::dynimage::test::frames_exceeding_logical_screen_size2::{{closure}}
             at .\src\dynimage.rs:1337
  16: core::ops::function::FnOnce::call_once<closure-0,tuple<>>
             at C:\Users\Felix\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ops\function.rs:227
  17: core::ops::function::FnOnce::call_once
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca\library\core\src\ops\function.rs:227
```
## Reproduction steps

```rust
        let data = vec![
            0x47, 0x49, 0x46, 0x38, 0x39, 0x61, 0x0A, 0x00, 0x0A, 0x00, 0xF0, 0x00, 0x00, 0x00,
            0x00, 0x00, 0x0E, 0xFF, 0x1F, 0x21, 0xF9, 0x04, 0x09, 0x64, 0x00, 0x00, 0x00, 0x2C,
            0x00, 0x00, 0x00, 0x00, 0x10, 0x00, 0x10, 0x00, 0x00, 0x02, 0x23, 0x84, 0x8F, 0xA9,
            0xBB, 0xE1, 0xE8, 0x42, 0x8A, 0x0F, 0x50, 0x79, 0xAE, 0xD1, 0xF9, 0x7A, 0xE8, 0x71,
            0x5B, 0x48, 0x81, 0x64, 0xD5, 0x91, 0xCA, 0x89, 0x4D, 0x21, 0x63, 0x89, 0x4C, 0x09,
            0x77, 0xF5, 0x6D, 0x14, 0x00, 0x3B,
        ];

        let _ = super::load_from_memory(&data);
```

Originally from @albel727, who already correctly [identified](https://github.com/image-rs/image/commit/281e95e72c137878ebd59717936907c3a5e14aad#r48503391) the faulty line:
https://github.com/image-rs/image/blob/d2f9fe8026db226f3fa6ff4e2473aca5977e6d26/src/codecs/gif.rs#L109

I've only added the backtrace and test data.

albel727:
Fixes #1455

Check before entering the direct buffer writing fast path, that the frame actually fits into the logical screen.
Add issue_1455_oversized.gif file to tests, to ensure that we keep handling this correctly.

Fix handling of small frames that start at (0,0), but have smaller dimensions than the logical screen.
Previous code incorrecly chose the fast path for them, which led to line wraparound.
Add issue_1455_undersized.gif file to tests, to ensure that we keep handling this correctly.

Enable the fast path in more cases. Now all frames that fit vertically and touch the borders of the logical screen horizontally are handled by it.
Add border_touching_layers.gif file to tests, which illustrates the case and hopefully will catch bugs, if the relevant code changes.

I license this contribution under the dual MIT/Apache-2.0 license,


src/codecs/gif.rs
```diff
@@ -106,9 +106,29 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for GifDecoder<R> {
 
         let (width, height) = self.dimensions();
 
-        if (frame.left, frame.top) == (0, 0) && (frame.width, frame.height) != self.dimensions() {
-            // If the frame matches the logical screen, directly read into it
+        if (frame.left, frame.width) == (0, width) && (frame.top as u64 + frame.height as u64 <= height as u64) {
```
fintelia, line 109: How about `frame.left == 0 && frame.width == width && ...`?

HeroicKatora:
I wondered why I couldn't reproduce the issue with `emulsion` and it seems that the `AnimationDecoder` handles all of it correctly :joy: The failing test case was an accidental breaking change within `avif` and fixed in #1456

src/codecs/gif.rs
albel727: Yes, makes more sense to me too, now that I think of it, given that one is a coordinate and another is a dimension.

albel727:
If anyone wonders, why so many checked_mul()-s, it's because despite this check above
https://github.com/image-rs/image/blob/17db4620f2a4b228bfb83cecad124dc0215814f5/src/codecs/gif.rs#L95
I couldn't rely on width and height being small enough, because `total_bytes()` uses unchecked multiplication
https://github.com/image-rs/image/blob/17db4620f2a4b228bfb83cecad124dc0215814f5/src/image.rs#L561
and `u32 * u32 * 4` channels can actually overflow `u64`.

HeroicKatora (approved):

fintelia:
> I couldn't rely on width and height being small enough, because total_bytes() uses unchecked multiplication

Might be worth just to changing total_bytes() to panic if the multiplication would overflow. There's no point in spending too much time worrying about images that take anywhere close to 2^64 bytes...

albel727:
>Might be worth just to changing total_bytes() to panic if the multiplication would overflow.

Yeah, that's probably worth to do in any case. I just don't have a sufficient grasp of the code nor authority to make such far reaching decisions, and I'm not prepared to deal with the possible fallout in the context of this PR, which I envision as a contained change to quickly fix an urgent case of incorrect gif rendering.

>There's no point in spending too much time worrying about images that take anywhere close to 2^64 bytes...

Well, I'd leave the checked_mul()-s either way. Then there's no need to rely on the presence of such remote asserts for correctness, and I guess the compiler will optimize the checks out anyway, when possible.

HeroicKatora:
https://github.com/image-rs/image/pull/1407

This dealt with overflows in other decoders, and deferred gif due to being implemented in `gif` itself. I'm not sure if this is accurate but I wouldn't worry about it here.

albel727:
If it were accurate, one wouldn't have gotten a panic in #1455 on this line:

https://github.com/image-rs/image-gif/blob/3b2682920cf82176a86ae1861efd668ea7b3a484/src/reader/mod.rs#L273

I mean yes, the bound check on the slice access does "deal with it", and it doesn't become something more dangerous like a buffer overflow, but it never hurts to check earlier, so that one doesn't have to debug across crates to uncover the actual issue. And hopefully to avoid a silent rendering corruption like the one which happens now on undersized frames. The compiler will likely eliminate redundant checks, which are likely dwarfed by decoding itself anyway.

albel727:
Ironically, `buffer_size()` calculation in the gif crate is also an unchecked multiplication of `u16 * u16 * channels`, which could overflow `usize` if it's 32 bit.

https://github.com/image-rs/image-gif/blob/3b2682920cf82176a86ae1861efd668ea7b3a484/src/reader/mod.rs#L352-L363

PR #1462, opened 2021-04-12
Added #[non_exhaustive] attributes, and removed __NonExhaustive enum variants

cbrownsey:
While reading through the docs, I noticed that this crate was using the older form of non exhaustiveness involving a private enum variant that cannot be matched on. Moving this to the newer #[non_exhaustive] macro form is trivial, however it does increase the MSRV from 1.38.0 to 1.40.0. I do not know the current consensus for this crate on how the MSRV should be handled when it comes to version changes, so some guidance would be appreciated.


src/buffer.rs
```diff
@@ -939,6 +940,36 @@ where
...
+    Container: Deref<Target = [P::Subpixel]>,
+{
+    /// Writes the buffer to a writer in the specified format.
+    ///
+    /// See [`ImageOutputFormat`](../enum.ImageOutputFormat.html) for
+    /// supported types.
+    ///
+    /// **Note**: TIFF encoding uses buffered writing,
+    /// which can lead to unexpected use of resources
+    pub fn write_to<W, F>(&self, writer: &mut W, format: F) -> ImageResult<()>
```
johannesvollmer, line 956: Shouldn't this be called `save_to` to conform to the `save` method? If I understand correctly, `write` is only ever used internally, not in the public API

fintelia:
I think this PR picked up some extraneous conflicts, but it should be enough to just rebase https://github.com/image-rs/image/pull/1462/commits/6db376d9112e2e6ab63078135b28735c48c612ca on the next branch and then merge that.

PR #1463, opened 2021-04-14
DXT: Fix alpha handling take 2

CensoredUsername:
Rebased version of #1060 so it can be merged. It was already approved, but I haven't worked on this for a while so would appreciate if someone else can review this to make sure it's up to spec.


CensoredUsername:
Had some more time to look at this and it checks out fine, so merging these bugfixes.

Issue #1469, opened 2021-05-01
`SubImage::to_image` is unreasonably strict


Lucretiel:
Currently, the signature for `to_image` is:

```rust
   pub fn to_image(&self) -> ImageBuffer<DerefPixel<I>, Vec<DerefSubpixel<I>>>
    where
        I: Deref,
        I::Target: GenericImage + 'static,
```

However, there's no reason for `I::Target` to need to be a `GenericImage`, because the function simply copies the view to a new image. This should be changed to `GenericImageView`.

PR #1470, opened 2021-05-01
`to_image` now only requires `I::Target: GenericImageView`

Lucretiel:
This change downgrades the trait bound on `to_image` from `GenericImage` to `GenericImageView`, because it only requires immutable access to the inner image. `GenericImage: GenericImageView`, so this change should be fully backwards compatible.

Fixes https://github.com/image-rs/image/issues/1469


HeroicKatora (approved):
LGTM

PR #1475, opened 2021-05-10
OpenEXR basic support

johannesvollmer:
Hi! Just experimented with adding OpenEXR support. Does not compile yet, work in progress.

Here's a list of things to do:
- [x] Solve sendable `Read`/`Write` (exrs only supports sendable stuff due to multi threading for now, might be able to relax this in exrs)
- [x] Reduce memcpy calls (requires exrs refactor)
- [x] Try f32 support (using u16, clamping 0-1, for now)
- [x] Use exr meta information more precisely (do not allocate alpha channel if none was in the file etc)
- [x] Basic tests
- [x] Try seekable writer if possible (kind of checked: this change must be done in the `next` branch)
- [x] Not ignore data window position and display window (allocate transparent display window instead of data window, and load only display window blocks?). Reason: Auto cropped images should not be tiny.
- [x] Investigate color conversion linearrgb/srgb

Current limitations:
- only pixel type `Rgba32F` and `Rgba16F` are supported
- only non-deep rgb/rgba files supported, no conversion from/to YCbCr or similar
- only the first non-deep rgb layer is used
- only the largest mip map level is used
- pixels outside display window are lost
- meta data is lost
- dwaa/dwab compressed images not supported yet by the exr library
- (chroma) subsampling not supported yet by the exr library


johannesvollmer:
Ah wait I branched off master lol. Should I have used branch `next` instead`?

fintelia:
> Ah wait I branched off master lol. Should I have used branch next instead`?

They're mostly the same. `ColorType` is marked non-exhaustive so you might be able to avoid actually making any breaking changes, in which case the master branch is right.

src/codecs/openexr.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn scanline_bytes(&self) -> u64 {
+        // we cannot always read individual scan lines,
+        // as the tiles or lines in the file could be in random order.
+        // therefore we currently read all lines at once
+        // Todo: respect exr.line_order
+       self.total_bytes()
+    }
+
+    fn read_image(mut self, target: &mut [u8]) -> ImageResult<()> {
+        self.read_image_with_progress(target, |_|{})
```
fintelia, line 101: This is the default implementation, so you don't have to duplicate it unless you're planning on making an optimized version later.

johannesvollmer: I know. I implemented it because this should always be true, even if the default implementation changes in the future.

johannesvollmer:
@fintelia @HeroicKatora Are there any challenges with making the writer seekable? The reader already is. The user can always write to a byte vector first, if their writer is not seekable. As most writers are seekable anyways, this change would probably result in memory usage reduction. I can open a separate pull request for this change.

fintelia:
@johannesvollmer please go ahead and make a PR against the `next` branch (we can't include it in the 0.23.x series because it would be a breaking change).

johannesvollmer:
In the exr library, I tried relaxing the requirements of the `Read`, to remove the `Send` requirement. Frankly, after some time, I can't find a solution where the parallel decompression works without requiring the reader to be sendable. `File` is sendable, so that is not a problem. Do we want fast decompression or support non-sendable readers?

fintelia:
The more bounds we add to the interface the more restrictive it is. Could you go into more detail on why the library requires the reader to be`Send`?

johannesvollmer:
The exr library reads individual pixel tiles one after another from a byte source. Multiple tiles can be decompressed in parallel. This should happen on the fly, without allocating all tiles at once. As a consequence, multiple threads have to pull the next pixel tile from the file, which apparently requires the byte source to be sendable between the threads. 

I have tried it again today and I think I got it working such that the reading from the file is only ever done in the main thread, and the blocks are sent to different threads for processing. Not sure yet whether it works correctly 100% of the time though, still doing some tests.

johannesvollmer:
So, I have managed to not require sendable readers and writers. And I think it is possible to keep it this way in the future. I have released a new version of the exr library with the necessary changes. I'll try to get the seekable reader and writer PR done next.

src/codecs/openexr.rs
```diff
@@ -0,0 +1,185 @@
...
+                    position.1 = height - (position.1 + 1); // openexr stores +y upwards, but we need +y downwards
+                    let first_f32_index = position.flat_index_for_size(*size);
+                    buffer[first_f32_index*4 .. (first_f32_index + 1)*4].copy_from_slice(&[r, g, b, a_or_1]);
+                }
+            )
+            .first_valid_layer() // TODO select exact layer by self.header_index?
+            .all_attributes()
+            .on_progress(move |progress| {
+                let mut prog = progress_callback;
+                prog(Progress::new((progress*blocks_in_header as f64) as u64, blocks_in_header)); // TODO precision errors?
```
johannesvollmer, line 118: The `on_progress` function call does not compile, because it expects a function pointer, but a closure is given. But the very definition of the called function asks for a closure, and not for a function pointer?!

Definition:

```rust
/// Specify a function to be called regularly throughout the loading process.
/// Replaces all previously specified progress functions in this reader.
pub fn on_progress(self, on_progress: F) -> Self where F: FnMut(f64) { Self { on_progress, ..self } }
```

Call:
```rust
.on_progress(move |progress| {
    progress_callback(Progress::new(
        (progress*blocks_in_header as f64) as u64, blocks_in_header) // TODO precision errors?
    );
})
```

Error: 
```
error[E0308]: mismatched types
   --> src\codecs\openexr.rs:115:26
    |
115 |               .on_progress(move |progress| {
    |  __________________________^
116 | |                 progress_callback(Progress::new(
117 | |                     (progress*blocks_in_header as f64) as u64, blocks_in_header) // TODO precision errors?
118 | |                 );
119 | |             })
    | |_____________^ expected fn pointer, found closure
    |
    = note: expected fn pointer `fn(f64)`
                  found closure `[closure@src\codecs\openexr.rs:115:26: 119:14]`
```

Why does the compiler say it requires a function pointer when in fact it should indeed accept any closure? Any ideas?

HeroicKatora: That is because the type  `F` mentioned here is the type parameter of [`ReadImage`](https://docs.rs/exr/1.2.0/exr/image/read/image/struct.ReadImage.html) and not one that you choose in `on_progress`. And the `all_attributes` method chooses to return specifically a `ReadImage<fn(_: f64), Self>`.

I'd say that this is something that the `exr` library should improve on.

johannesvollmer: The function must have a separate type parameter, replacing the struct type parameter.

johannesvollmer:
I'd like to ask for some suggestions concerning tests. My initial idea was to read and compare an HDR and EXR file of the same image, any objections? Are there any tests that should be done for every image type (and can also be done for high dynamic range f32 images)?

johannesvollmer:
Also, is there no standardized way to obtain an image buffer from a decoder? I got stuck with the following code:

```rust
    let hdr: Vec<Rgb<f32>,> = image::codecs::hdr::HdrDecoder::new(
        io::BufReader::new(fs::File::open(&reference_path).unwrap())
    ).unwrap().read_image_hdr().unwrap();

    let exr_pixels: ImageBuffer<Rgba<f32>, Vec<f32>> = image::codecs::openexr::ExrDecoder::read(
        io::BufReader::new(fs::File::open(&exr_path).unwrap())
    ).unwrap().read_f32_image_buffer().unwrap();

    // TODO let hdr = ImageBuffer::from_raw(___)
    assert_eq!(exr_pixels.dimensions(), hdr.____);
    assert_eq!(exr_pixels.pixels(), hdr.____);
```

johannesvollmer:
Also, I couldn't find whether the image crate stores images upside down or upside up. Is positive y coordinate considered top of the image, or positive y considered the bottom of the image?

johannesvollmer:
Got the tests working, but it feels like there should be an easier way of doing it

HeroicKatora:
`(0, 0)` is considered top-left for the buffers in this library.

johannesvollmer:
Added tests:
- compare_exr_hdr
- roundtrip_rgba
- roundtrip_rgb
- compare_rgba_rgb

johannesvollmer:
The feature flag goes by the full name `openexr` (to avoid clashing with the dependency `exr`). But the file format variant equals the file extension, `Exr`, and not `OpenEXR`. Do you think this could lead to confusion? I could rename the file format variant to the full name.

johannesvollmer:
Is it required to actually implement an encoder object? It seems like it would already work with a simple function call.

This code seems pretty much like boilerplate:
```rust
#[Debug]
pub struct Encoder<W> (W);
impl<W> ImageEncoder for Encoder(W) where W: Write /*+ Seek*/ {
    fn write_image(self, buf: &[u8], width: u32, height: u32, color_type: ColorType) -> ImageResult<()> {
        write_buffer(self.0, buf, width, height, color_type)
    }
}
```

Or does it only exist as some kind of future-proof design which enables us to add other functions like `write_image_with_progress` at a later stage?

johannesvollmer:
@fintelia @heroickatora 

- are these tests fine?
- should I rename the ImageFormat variant from `Exr` to `OpenExr`?

HeroicKatora:
* are these tests fine?

Yes, looks like it.

* should I rename the ImageFormat variant from Exr to OpenExr?

Please do. And also remove the work in progress mark when you want a code review.

johannesvollmer:
> also remove the work in progress mark or ask explicitly when you want a code review

Oh yeah I just didn't realize that we are already at that point haha :)

src/buffer.rs
```diff
@@ -1340,6 +1344,14 @@ pub(crate) type Gray16Image = ImageBuffer<Luma<u16>, Vec<u16>>;
 /// Sendable 16-bit grayscale + alpha channel image buffer
 pub(crate) type GrayAlpha16Image = ImageBuffer<LumaA<u16>, Vec<u16>>;
 
+/// An image buffer for 32-bit float RGB pixels,
+/// where the backing container is a flattened vector of floats.
+pub(crate) type Rgb32FImage = ImageBuffer<Rgb<f32>, Vec<f32>>;
```
johannesvollmer, line 1349: Why are these `pub(crate)` and not `pub`? I would like them public to reduce the typing required by users

README.md
```diff
@@ -23,14 +23,15 @@ All image processing functions provided operate on types that implement the `Gen
 | PNG    | All supported color types | Same as decoding |
 | JPEG   | Baseline and progressive | Baseline JPEG |
 | GIF    | Yes | Yes |
-| BMP    | Yes | RGB(8), RGBA(8), Gray(8), GrayA(8) |
```
johannesvollmer, line 26: Documentation: Renamed these to camel case while I was already there

src/buffer.rs
```diff
@@ -948,6 +949,9 @@ where
 {
     /// Writes the buffer to a writer in the specified format.
     ///
+    /// Assumes the writer is buffered. In most cases,
+    /// you should wrap your writer in a `BufWriter` for best performance.
+    ///
```
johannesvollmer, line 954: Documentation: Clarified whether the writer being buffering or not buffered while I was there

src/codecs/openexr.rs
```diff
@@ -0,0 +1,433 @@
...
+
+use crate::{ImageDecoder, ImageResult, ColorType, Progress, ImageError, ImageFormat, ImageBuffer, Rgba, Rgb, ImageEncoder, ExtendedColorType};
+use std::io::{Write, Seek, BufRead, Cursor, BufReader, BufWriter};
+use crate::error::{DecodingError, ImageFormatHint, LimitError, LimitErrorKind, EncodingError};
+use crate::image::decoder_to_vec;
+use std::path::Path;
+use crate::buffer_::{Rgb32FImage, Rgba32FImage};
+
+
+// TODO this could be a generic function that works for any format
```
johannesvollmer, line 26: This (and the other functions) have the same purpose as `ImageBuffer::save<Q>(path: Q)`, but with two differences:
1. It works without having to implement `Pixel::COLOR_TYPE` for `Rgba<f32>`.
2. There is an additional version that works with generic readers and writers (`impl Read`), not only one version with a path

Instead of increasing the API surface of the OpenEXR module, we could have them as `pub(crate)` and wait until the `Pixel` trait is implemented for `Rgb<f32>`, and furthermore add the non-path variants to the `ImageBuffer` type.

src/codecs/openexr.rs
```diff
@@ -0,0 +1,433 @@
...
+    let height = height as usize;
+
+    let mut seekable_write = Cursor::new(Vec::with_capacity(width*height*3*4)); // TODO remove
+
+    match color_type {
+        ColorType::Rgb32F => {
+            let pixels: &[f32] = bytemuck::try_cast_slice(bytes)
+                .expect("image byte buffer must be aligned to f32");
+
+            debug_assert_eq!(pixels.len(), width * height * 3, "invalid dimensions for byte buffer length");
```
johannesvollmer, line 282: These debug assertions could also return an error. They are not expensive and only run once per image. What is the preferred solution?

src/image.rs
```diff
@@ -511,6 +524,11 @@ pub struct Progress {
 }
 
 impl Progress {
+    /// Create Progress.
+    pub fn new(current: u64, total: u64) -> Self {
+        Self { current, total }
```
johannesvollmer, line 529: Needed to add this function to actually create a progress object - Alternatively could make the fields public and use a struct constructor.

src/io/free_functions.rs
```diff
@@ -3,32 +3,7 @@ use std::io::{BufRead, BufReader, BufWriter, Seek};
...
-use crate::codecs::dds;
-#[cfg(feature = "tiff")]
-use crate::codecs::tiff;
-#[cfg(feature = "webp")]
-use crate::codecs::webp;
-#[cfg(feature = "farbfeld")]
-use crate::codecs::farbfeld;
-#[cfg(any(feature = "avif-encoder", feature = "avif-decoder"))]
-use crate::codecs::avif;
+use crate::codecs::*;
```
johannesvollmer, line 7: Simplified this huge list of imports - should yield an exactly equivalent import. Instead of checking for each codec and importing it, import all active codecs

src/io/free_functions.rs
```diff
@@ -80,9 +58,11 @@ pub fn load<R: BufRead + Seek>(r: R, format: ImageFormat) -> ImageResult<Dynamic
         #[cfg(feature = "ico")]
         image::ImageFormat::Ico => DynamicImage::from_decoder(ico::IcoDecoder::new(r)?),
         #[cfg(feature = "hdr")]
-        image::ImageFormat::Hdr => DynamicImage::from_decoder(hdr::HdrAdapter::new(BufReader::new(r))?),
+        image::ImageFormat::Hdr => DynamicImage::from_decoder(hdr::HdrAdapter::new(r)?),
+        #[cfg(feature = "openexr")]
+        image::ImageFormat::OpenExr => DynamicImage::from_decoder(openexr::ExrDecoder::read(r, None)?),
         #[cfg(feature = "pnm")]
-        image::ImageFormat::Pnm => DynamicImage::from_decoder(pnm::PnmDecoder::new(BufReader::new(r))?),
```
johannesvollmer, line 85: Made this radical change while I was at it. In the rare case an unbuffered reader is used, these two image formats now might have worse performance, but now it's at least equal for all image formats. In the image crate, this reader is always already be buffered. I also clarified this in the documentation.

johannesvollmer: See the changes in `io::reader.rs` below

src/codecs/openexr.rs
```diff
@@ -0,0 +1,433 @@
...
+
+            exr::prelude::Image
```
johannesvollmer, line 284: OpenEXR supports many different compression methods, each with a different a trade-off between speed and compression ratio. By default, a fast but rather weak compression is used (plain run-length encoding). This could also use deflate compression, which is slower, but yields small images. What do you think?

src/buffer.rs
HeroicKatora: We hadn't yet decided if they were ready. I think we don't have any test coverage on the conversion to/from the other color types for example. As a pure container though they should be fine.

src/io/free_functions.rs
fintelia: Seems like a reasonable change, but I'd feel better if it could go in its own quick PR.

src/codecs/openexr.rs
HeroicKatora: This increases the surface quite a bit. Note that for (2) specifically there is the [`image::io::Reader`](https://docs.rs/image/0.23.14/image/io/index.html) which allows a composition of the different modes without a multiplicative effect on the number of methods. The raw methods are slated for deprecation once the equivalent writer exists.

src/image.rs
HeroicKatora: Can you mark this `pub(crate)`? Currently there is one potential invariant (`current <= total`) that we may want to assert for instances of this type.

src/buffer.rs
johannesvollmer: I only thought about is as a container. I'll make them public :)

src/codecs/openexr.rs
johannesvollmer: The exr library currently expects strongly typed data, meaning actual `f32` values. It supports raw bytes, but then we can't use the existing abstractions from the exr library. 

I think it would be undefined behaviour if we tried to interpret unaligned data as `&[f32]`. In this sense, the exr library currently doesn't support unaligned data. My favourite option would be to align a new buffer in case of misalignment, but the error would definitely be quicker to implement.

src/image.rs
johannesvollmer: Made it `pub(crate)`

src/codecs/openexr.rs
johannesvollmer: Went with constructing f32 from bytes on the fly for now, as it was the quickest change supporting unaligned access, is this alright?

src/io/free_functions.rs
johannesvollmer: Reverted the change, but kept the documentation

src/codecs/openexr.rs
johannesvollmer: Sorry if I didn't understand that, english is not my mother tongue. Did I get that right that you suggest I keep these methods and later mark them deprecated once `Rgb<f32>` implements `Pixel` and dynamic image supports `Rgba32F`?

johannesvollmer: @HeroicKatora So, do we keep these?

src/dynimage.rs
```diff
@@ -882,7 +882,11 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::rotate270(p))
     }
 
-    /// Encode this image and write it to ```w```
+    /// Encode this image and write it to ```w```.
+
```
fintelia, line 886: nit: should be a `///` on this line.

src/dynimage.rs
```diff
@@ -1309,6 +1313,9 @@ where
 /// See [`ImageOutputFormat`](../enum.ImageOutputFormat.html) for
 /// supported types.
 ///
+/// Assumes the writer is buffered. In most cases,
+/// you should wrap your writer in a `BufWriter` for best performance.
+    ///
```
fintelia, line 1318: nit: bad indentation

src/codecs/openexr.rs
fintelia: Could you replace these with new `Reader::{decode_rgb32f, decode_rgba32f}` methods?

src/codecs/openexr.rs
```diff
@@ -0,0 +1,489 @@
...
+        .ok_or_else(|| ImageError::Limits(LimitError::from_kind(LimitErrorKind::InsufficientMemory)))
+}
+
+
+// TODO this could be a generic function that works for any format
+/// Write an `Rgb32FImage` to the specified file path, replacing any existing file.
+// TODO progress? load rect?
+pub fn write_rgb_image_to_file(path: impl AsRef<Path>, image: &Rgb32FImage) -> ImageResult<()> {
+    write_rgb_image(BufWriter::new(std::fs::File::create(path)?), image)
+}
```
fintelia, line 85: Let's skip adding this one line function for now. Shouldn't be too hard for people to write it themselves.

src/codecs/openexr.rs
```diff
@@ -0,0 +1,489 @@
...
+        bytemuck::cast_slice(image.as_raw().as_slice()),
+        image.width(), image.height(),
+        ColorType::Rgb32F
+    )
+}
+
+// TODO this could be a generic function that works for any format
+/// Write an `Rgba32FImage` to the specified file path, replacing any existing file.
+// TODO progress? load rect?
+pub fn write_rgba_image_to_file(path: impl AsRef<Path>, image: &Rgba32FImage) -> ImageResult<()> {
```
fintelia, line 104: Same as above

src/codecs/openexr.rs
```diff
@@ -0,0 +1,489 @@
...
+                "image does not contain non-deep rgb channels"
+            )))?;
+
+        // TODO binary search, channels are sorted
+        let has_alpha = exr_reader.headers()[header_index]
+            .channels.list
+            //.binary_search_by_key(b"A", |chan| chan.name.bytes())
+            .iter()
+            .find(|chan| chan.name.eq("A")) // TODO eq_lowercase only if exrs supports it
+            .is_some();
```
fintelia, line 175: nit: `.find(...).is_some()` is equivalent to `.any(...)`

src/codecs/openexr.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    buffered_write.write_all(seekable_write.into_inner().as_slice())?;
+    Ok(())
+}
+
+
+// TODO is this struct and trait actually used anywhere?
+/// A thin wrapper that implements `ImageEncoder` for OpenEXR images. Will behave like `image::codecs::openexr::write_buffer`.
+#[derive(Debug)]
+pub struct Encoder<W> (W);
```
fintelia, line 352: Please call this `ExrEncoder`

src/codecs/openexr.rs
```diff
@@ -0,0 +1,489 @@
...
+
+
+
+/// Write a raw byte buffer of pixels,
+/// panicking if the buffer is not aligned to `f32`
+/// or if it has an invalid length.
+///
+/// Assumes the writer is buffered. In most cases,
+/// you should wrap your writer in a `BufWriter` for best performance.
+pub fn write_buffer(
```
fintelia, line 283: Could you either make this private or move the contents into `ExrEncoder::write_image`?

src/codecs/openexr.rs
johannesvollmer: Sure, sounds awesome, where can I find those methods? Are they in the master branch? Couldn't find them, not even with global search in the whole `image` repository lol

fintelia: They don't exist now. I was suggesting you add them to the [`io::Reader`](https://docs.rs/image/0.23.14/image/io/struct.Reader.html) struct.

johannesvollmer: Oh, haha :D Yes I'd love to do that, now I got it

johannesvollmer: Are we going to add the other variants later? Like `Rgb8` and others?

src/dynimage.rs
johannesvollmer: This already is a triple slash, what do you mean?

fintelia: There's no triple slash on line 886

src/codecs/openexr.rs
johannesvollmer: Update: Moving these methods turns out to be a difficult refactor. It seems to be too specific for the general reader interface. I'm stuck with the following code:

```rust

    fn read_as_buffer<Pixel, PixelComponent, PixelStorage>(self) -> ImageResult<ImageBuffer<Pixel, PixelStorage>> {
        let decoder: Box<dyn ImageDecoder> = ???;
        let buffer: Vec<PixelComponent> = decoder_to_vec(decoder)?;
        ImageBuffer::from_raw(width, height, buffer)
    }
```

Maybe this should rather be a function with the following type:

```rust
trait Decoder {
    // ...

    fn decode_as_buffer <Px: Pixel> (self) -> ImageBuffer<Px, Vec<Px::SubPixel>> {
        assert_eq!(Px::COLOR_TYPE, self.color_type());
        let (width, height) = self.dimensions();
        let buffer: Vec<Px::SubPixel> = decoder_to_vec(self)?;
        ImageBuffer::from_raw(width, height, buffer)
    }
}
```

HeroicKatora: I don't quite follow, where does the necessity of all the type parameters come from? In the implementation here you just had `ImageResult<Rgba32FImage>` which should be fine. In particular there is no need to generalize to _all_ buffer types and to _all_ `impl Pixel` types. Rather, and this might solve your confusion on `Box<dyn ImageDecoder>`, I wouldn't generalize this further than for pixel types whose components implement `bytemuck::Pod`. This covers all integers and floats which should be more then fine.

This allows you to decode to a vector that you're allocating yourself by using something like: `ImageDecoder::read_image(bytemuck::cast_slice_mut(&mut the_sample_vec[..]))`

src/dynimage.rs
johannesvollmer: Oh right

src/codecs/openexr.rs
johannesvollmer: > This allows you to decode to a vector that you're allocating yourself by using something like: ImageDecoder::read_image(bytemuck::cast_slice_mut(&mut the_sample_vec[..]))

The only point of these helper methods is to allocate a buffer, such that I don't have to repeat the boilerplate allocation code twice in every test. This is less error prone and the intent is immediately obvious, resulting in cleaner code. Maybe this simpler method could even be used in other places in the existing `image` crate code.

---
 
> I wouldn't generalize this further than for pixel types whose components implement bytemuck::Pod. This covers all integers and floats which should be more then fine.

Yes, that's what I did:

```rust

trait ImageDecoder {
    // ...

    /// Fills an `ImageBuffer` with the image pixels.
    /// You must provide a pixel type that exactly matches the image color type,
    /// no conversion is done. For incompatible pixel types, an error is returned.
    fn decode_into_pixels<Px: Pixel> (self) -> ImageResult<ImageBuffer<Px, Vec<Px::Subpixel>>>
        where Px::Subpixel: bytemuck::Pod,
              Px: 'static,
    {
        if Px::COLOR_TYPE != self.color_type() {
            return Err(ImageError::Decoding(DecodingError::new(
                ImageFormatHint::Unknown, // TODO we may know this
                "The color type of the provided Pixel must match the color type of the ImageDecoder"
            )));
        }

        let (width, height) = self.dimensions();
        let buffer: Vec<Px::Subpixel> = decoder_to_vec(self)?;

        ImageBuffer::from_raw(width, height, buffer)

            // this should be the only reason for the "from raw" call to fail,
            // even though such a large allocation would probably cause an error much earlier
            .ok_or_else(|| ImageError::Limits(LimitError::from_kind(LimitErrorKind::InsufficientMemory)))
    }
}


fn usage(){
    OpenExrDecoder::read(read, Some(true))?.decode_into_pixels::<Rgba<f32>>()
}
```


I got this working, but the Pixel trait is implemented incorrectly for `Rgb<f32>`. The `Rgb<f32>::SubPixel` type will not resolve to an `f32` color type. The Pixel trait is only really implemented for `Rgb8` and `Rgb16`, not `f32`. This causes the error check in the first line to always fail for `f32` colors.


Instead of this generic implementation, we could copy & paste all the implementations manually for all the color types.... Is this really the way to go? The pixel type must be fixed sooner or later anyways, so maybe we should probably just do that instead of manually copying the code.

src/codecs/openexr.rs
```diff
@@ -0,0 +1,489 @@
...
+
+impl<R: BufRead + Seek> ExrDecoder<R> {
+
+    /// Create a decoder. Consumes the first few bytes of the source to extract image dimensions.
+    /// Assumes the reader is buffered. In most cases,
+    /// you should wrap your reader in a `BufReader` for best performance.
+    /// If alpha preference is specified, an alpha channel will
+    /// always be present or always be not present in the returned image.
+    /// If alpha preference is none, the alpha channel will only be returned if it is found in the file.
+    pub fn read(source: R, alpha_preference: Option<bool>) -> ImageResult<Self> {
```
fintelia, line 147: Could you rename this `ExrDecoder::with_alpha_preference` and add a `ExrDecoder::new` method which assumes alpha_preference=None? That would be more consistent with other decoders

fintelia: Honestly, at some point we should really make the `new` method part of the ImageDecoder trait but there's some complications involved with that.

src/codecs/openexr.rs
HeroicKatora: > The only point of these helper methods is to allocate a buffer, such that I don't have to repeat the boilerplate allocation code twice in every test. 
> The Rgb<f32>::SubPixel type will not resolve to an f32 color type. The Pixel trait is only really implemented for Rgb8 and Rgb16, not f32

This helper method as sketched would then be better suited for `impl ImageBuffer`. It is similar to [`DynamicImage::from_decoder`](https://docs.rs/image/0.23.14/image/enum.DynamicImage.html#method.from_decoder). I'm afraid the `Pixel` issue is something we can't easily resolve in this major version of `image`. We could, however, already add an improved trait that is _just_ used for these method. Something like:

```
#[repr(u8)]
enum SampleType {
    U8 = 0,
    U16 = 1,
…
    F32 = 2,
}

trait Sample: Primitive {
    const TYPE: SampleType;
}

mod hidden {
    pub trait PixelV2 { …  }
    impl<T: Sample> PixelV2 for Rgb<T> { … }
}

impl<Px> ImageBuffer<Px, Vec<Px::Subpixel>>
where
    Px: hidden::PixelV2,
{
    fn from_decoder(mut decoder: impl ImageDecoder) -> ImageResult<Self> {
        if Px::COLOR_TYPE != decoder.color_type() { … }
    }
}
```

In any case, I would much rather discuss these helpers in a separate PR _after_ the core EXR functionality has been added together with the new `ColorType` variant :)

src/codecs/openexr.rs
johannesvollmer: > we should really make the new method part of the ImageDecoder

I thought so too some time before haha :)

HeroicKatora: The embedded hal crates very much avoid having a constructor as part of the trait and I think this style is, for the most part, pretty clean. An alternative would be a method to 'suggest' a color type to an `ImageDecoder` so that you can modify the decoding later. Afterall, the choice need not be final after having read the header of an image.

src/codecs/openexr.rs
johannesvollmer: Interesting idea, I like it. I propose we keep the helper methods, but private, then we can delay the discussion without duplicating code in the tests. Okay?

johannesvollmer:
Got a build error, out of nowhere: `error: failed to run custom build command for "rav1e v0.5.0-alpha"`. Deleting the `target` folder and updating cargo does not help

src/codecs/openexr.rs
```diff
@@ -0,0 +1,481 @@
...
+// private, access via `OpenExrEncoder`
+fn write_buffer(
+    mut buffered_write: impl Write/* + Seek*/, unaligned_bytes: &[u8],
+    width: u32, height: u32, color_type: ColorType
+) -> ImageResult<()>
+{
+    let width = width as usize;
+    let height = height as usize;
+
+    if unaligned_bytes.len() < width * height * color_type.bytes_per_pixel() as usize {
```
fintelia, line 211: This multiplication can overflow. It would be good to check for that case

src/codecs/openexr.rs
HeroicKatora: I'm fine with private helpers :+1:

src/codecs/openexr.rs
johannesvollmer: Like this?
```rust
    // if the width and height does not match the length of the bytes, return error
    let has_invalid_size = if let Some(expected_byte_count) = expected_byte_count {
        unaligned_bytes.len() < expected_byte_count
    }
    // otherwise, size calculation overflowed, is bigger than memory,
    // therefore data is too small, so it is invalid.
    else { true };

    if has_invalid_size {
        return Err(ImageError::Encoding(EncodingError::new(
            ImageFormatHint::Exact(ImageFormat::OpenExr),
            "byte buffer not large enough for the specified dimensions and f32 pixels"
        )));
    }
```

fintelia: Yeah, though you could probably do something like `if expected_byte_count.map(...).unwrap_or(true) { ... }` to simplify the code

johannesvollmer: Yes! Probably like this:

```rust
    let expected_byte_count = width.checked_mul(height)
        .and_then(|size| size.checked_mul(color_type.bytes_per_pixel() as usize));

    // if the width and height does not match the length of the bytes, the arguments are invalid
    let has_invalid_size_or_overflowed = expected_byte_count
        .map(|expected_byte_count | unaligned_bytes.len() < expected_byte_count)

        // otherwise, size calculation overflowed, is bigger than memory,
        // therefore data is too small, so it is invalid.
        .unwrap_or(true);

    if has_invalid_size_or_overflowed {
        return Err(ImageError::Encoding(EncodingError::new(
            ImageFormatHint::Exact(ImageFormat::OpenExr),
            "byte buffer not large enough for the specified dimensions and f32 pixels"
        )));
    }
```

johannesvollmer:
Hey! I've finally had the time to press the `push commits` button in my IDE! Sorry for the delay. 

So, from my perspective, we could merge this soon, as a minimal support. 

There's of course a lot of other things left to optimize, including the exr library itself. But for now, I estimate that 95% of uses cases are already covered, which is RGB images with F32 and F16 precision. I included an extensive list of limitations in the module documentation. Here's the contents:

- only color type `Rgba32F` and `Rgba16F` are supported
- only non-deep rgb/rgba files supported, no conversion from/to YCbCr or similar
- only the first non-deep rgb layer is used
- only the largest mip map level is used
- pixels outside display window are lost
- meta data is lost
- dwaa/dwab compressed images not supported yet by the exr library
- (chroma) subsampling not supported yet by the exr library

Some of these limitations are simplifications, which may possibly never change: Only the first layer and largest mip map level. However, conversion between color models, meta data, and subsampling might be something we could add later.

_Note: The YCbCr conversion is specified in the OpenEXR standard, so it should be easy to implement, as it is fully defined._

HeroicKatora (approved):
Very nice iteration overall. LGTM.

Regarding linear RGB vs. sRGB I hope we can solve it with the meta data PR and `f32` in the next major release.

johannesvollmer:
That would be great!

src/codecs/openexr.rs
```diff
@@ -0,0 +1,495 @@
...
+                    // only keep pixels inside the data window
+                    // TODO filter chunks based on this
+                    if index_in_display_window.x() >= 0 && index_in_display_window.y() >= 0
+                        && index_in_display_window.x() < display_window.size.width() as i32
+                        && index_in_display_window.y() < display_window.size.height() as i32
+                    {
+                        let index_in_display_window = index_in_display_window.to_usize("index bug").unwrap();
+                        let first_f32_index = index_in_display_window.flat_index_for_size(display_window.size);
+
+                        buffer[first_f32_index * channel_count .. (first_f32_index + 1) * channel_count]
```
johannesvollmer, line 170: In the decoder, this indexing operation will panic when the buffer has an invalid size. The size should probably be checked and return an error in case it is invalid, right?

src/codecs/openexr.rs
```diff
@@ -0,0 +1,495 @@
...
+                progress_callback(Progress::new(
+                    (progress*blocks_in_header as f64) as u64, blocks_in_header) // TODO precision errors?
+                );
+            })
+            .from_chunks(self.exr_reader).map_err(to_image_err)?;
+
+        // TODO this copy is strictly not necessary, but the exr api is a little too simple for reading into a borrowed target slice
+
+        // this cast is safe and works with any alignment, as bytes are copied, and not f32 values.
+        unaligned_bytes.copy_from_slice(bytemuck::cast_slice(result.layer_data.channel_data.pixels.as_slice()));
```
johannesvollmer, line 189: This copy operation will panic if the caller passes a buffer with invalid length. We should probably check this and then return an error bofore calling `copy_from_slice`, right?

johannesvollmer:
No wait.... the decoder should panic instead? Whoopsie :D Changed it to a panic. 

Would it make sense to unify this behaviour to be the same across decoding and encoding?

johannesvollmer:
There seems to be a problem with the CI runner:

```
info: syncing channel updates for 'beta-x86_64-unknown-linux-gnu'
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-beta.toml.sha256' to '/home/runner/.rustup/tmp/vnafuiemrcd4vsc4_file': failed to make network request: error sending request for url (https://static.rust-lang.org/dist/channel-rust-beta.toml.sha256): error trying to connect: Connection reset by peer (os error 104): error trying to connect: Connection reset by peer (os error 104): Connection reset by peer (os error 104): Connection reset by peer (os error 104)
```

johannesvollmer:
re-triggering the CI worked, now I'm ready to go :)

johannesvollmer:
Thanks for your support <3

jzarzeckis:
Hey. as a newbie I may get it all wrong, but I'm trying this out with the image in the tests folder, and I called:
```rs
let im = open("assets/cropping - uncropped original.exr").unwrap();
```

And got the following error:
```
Unsupported(UnsupportedError { format: Unknown, kind: Color(Rgba32F) })
 ```

I wonder why the format is `Unknown`. Did it fail to parse the extension? Just tested with a PNG, and it did work fine.

FYI - this is what I have in my `cargo.toml`:
```toml
image = { git = "https://github.com/image-rs/image", rev="3197e3d964f5d214cbdec98206fbc895723e5c72" }
```

johannesvollmer:
Hi! Thanks for taking the time to post this. Next time, don't be afraid to open a full-blown issue and tag me in there :)

Summary: While the exr file format can be decoded and encoded, the rest of the image library needs more time to properly implement f32 support everywhere. 

In detail: The support is pretty basic and currently does not support being loaded as a `DynamicImage`. This is because the dynamic image does not support F32 data yet - regardless of the file format. As a workaround, you can still load an exr image to an f32 buffer by using the `OpenEXRDecoder` directly. We hope that in the next major release, this won't be necessary.

jzarzeckis:
Thanks for the clarification. Yeah - for now I'm playing around with the decoder directly.

johannesvollmer:
You're welcome. If you notice anything else, let me know :)

PR #1477, opened 2021-05-19
Require Seekable Writer

johannesvollmer:
This PR changes the API, now requiring all writers to be seekable. 

Non-Seekable writers are pretty rare, and can still be used by allocating a temporary seekable byte vector. This allows us to allocate less in the common case, which is writing to an already seekable destination. At this moment, no encoder utilizes this advantage yet (TiffEncoder only writes to files?). The individual encoders could be revisited to see whether they can already profit from this new requirement.

@fintelia @HeroicKatora  I looked at the avif encoder. The `fallback` buffer is required for reasons other than seeking. It is necessary because of the design of the avif library, which allocates a vector in any case. I was not yet able to remove that temporary buffer even when requiring `Seek` from the writer.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,13 +16,12 @@
 //! Load images using [`io::Reader`]:
 //!
 //! ```rust,no_run
-//! # use std::io::Cursor;
 //! use image::io::Reader as ImageReader;
 //! # fn main() -> Result<(), image::ImageError> {
 //! # let bytes = vec![0u8];
 //!
 //! let img = ImageReader::open("myimage.png")?.decode()?;
-//! let img2 = ImageReader::new(Cursor::new(bytes)).decode()?;
+//! let img2 = ImageReader::new(std::io::Cursor::new(bytes)).decode()?;
 //! # Ok(())
 //! # }
 //! ```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,7 +30,7 @@
 //! And save them using [`save`] or [`write_to`] methods:
 //!
 //! ```rust,no_run
-//! # use std::io::Write;
+//! # use std::io::{Write, Cursor};
 //! # use image::ImageOutputFormat;
 //! # use image::DynamicImage;
 //! # #[cfg(feature = "png")]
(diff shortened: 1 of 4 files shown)


johannesvollmer:
Also: Why do some `write` functions accept `ImageFormat`? I would have assumed that all `write` functions use an `ImageOutputFormat`?

johannesvollmer:
I sketched out some radical changes (that you probably already have implemented somewhere?) to show you what I mean:

https://github.com/image-rs/image/compare/next...johannesvollmer:image-output-format-reuse

fintelia (approved):

HeroicKatora (changes requested):
A small number of tests will likely need to be changed from `Vec<u8>` to `Cursor<Vec<u8>>` or similar.
See [this CI run](https://github.com/image-rs/image/pull/1477/checks?check_run_id=2646568186)

```


---- src/lib.rs - (line 32) stdout ----
error[E0277]: the trait bound `Vec<u8>: Seek` is not satisfied
  --> src/lib.rs:43:6
   |
13 | img2.write_to(&mut bytes, image::ImageOutputFormat::Png)?;
   |      ^^^^^^^^ the trait `Seek` is not implemented for `Vec<u8>`
```

src/lib.rs
```diff
@@ -40,7 +40,7 @@
 //! img.save("empty.jpg")?;
 //!
 //! let mut bytes: Vec<u8> = Vec::new();
-//! img2.write_to(&mut bytes, image::ImageOutputFormat::Png)?;
+//! img2.write_to(&mut Cursor::new(&mut bytes), image::ImageOutputFormat::Png)?;
```
johannesvollmer, line 43: Do you think the use of the Cursor should have an explanation, or do we assume this is common knowledge?

HeroicKatora: I'd say it is common enough. We might give readers a small nudge by making the `use` of `Cursor` visible.

johannesvollmer: Sounds good. Any more changes before merging?

HeroicKatora (approved):

johannesvollmer:
Thanks :)

PR #1478, opened 2021-05-19
Add minimal F32 support to ColorType

johannesvollmer:
Add F32 RGB/RGBA variants to both color type enumerations. 

### HDR Image format
Internally requires only 4 bytes of storage, but looking up a pixel should yield an `Rgba<f32>` color. It should not be necessary to allocate a complete `RGBA32F` buffer just to to offer the `Rgba<f32>` value, but it would be possible now. I can probably implement that change in this PR if desired.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -30,6 +30,11 @@ pub enum ColorType {
     /// Pixel is 8-bit BGR with an alpha channel
     Bgra8,
 
+    /// Pixel is 32-bit float RGB
+    Rgb32F,
+    /// Pixel is 32-bit float RGBA
+    Rgba32F,
+
     #[doc(hidden)]
     __NonExhaustive(crate::utils::NonExhaustiveMarker),
 }
@@ -44,6 +49,8 @@ impl ColorType {
             ColorType::Rgba8 | ColorType::Bgra8 | ColorType::La16 => 4,
             ColorType::Rgb16 => 6,
             ColorType::Rgba16 => 8,
+            ColorType::Rgb32F => 3 * 4,
+            ColorType::Rgba32F => 4 * 4,
             ColorType::__NonExhaustive(marker) => match marker._private {},
         }
     }
@@ -52,8 +59,8 @@ impl ColorType {
     pub fn has_alpha(self) -> bool {
         use ColorType::*;
         match self {
-            L8 | L16 | Rgb8 | Bgr8 | Rgb16 => false,
-            La8 | Rgba8 | Bgra8 | La16 | Rgba16 => true,
+            L8 | L16 | Rgb8 | Bgr8 | Rgb16 | Rgb32F => false,
+            La8 | Rgba8 | Bgra8 | La16 | Rgba16 | Rgba32F => true,
             __NonExhaustive(marker) => match marker._private {},
         }
     }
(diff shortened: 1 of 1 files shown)


johannesvollmer:
What about the DynamicImage enum? How about a `Rgb32F, Rgba32F` variant and some `Rgba32FImage, Rgba<f32>` buffers and pixels? 

How much is this minimal f32 support worth if the dynamic image cannot contain the float pixels?

fintelia:
I think this change is valuable even without those other items.

The way I think about the library is that there are multiple "levels" of API. The low-level API involves `ColorType` and the `ImageDecoder` and operates on raw byte slices. This PR would enable full floating point support for that level.

Built on top of that is a high-level API which involving `DynamicImage`, `ImageBuffer`, and `GenericImage`. Full floating point support could be added here in a followup.

johannesvollmer:
Ok, makes sense. What do you think about changing the HDR image encoder to use F32? Would this need to go into the `next` branch?

fintelia:
`HdrEncoder::encode` already takes a `&[Rgb<f32>]` so that shouldn't need to change (`Rgb<f32>` is just a newtype wrapper around a `[f32; 3]`).

johannesvollmer:
Ooops sorry, I mean the decoder. It will allocate an Rgbe8 storage, right? 

Maybe I'll need some pointers to understand the architecture of the library. Where should I start? Is there an overview about the image classes?

fintelia:
The main reading I'd recommend is overview text on https://docs.rs/image/0.23.14/image/, but it is also worth noting that the library has evolved over many years and though we try to increase API consistency as we go, there are a lot of rough edges. I hadn't looked at `HdrDecoder` for a while, but I notice that unlike our other decoders it doesn't implement `ImageDecoder. That could probably be fixed once ColorType::Rgb32F is added to describe the output format.

Narann:
I don't want to twist the thread, but I think it's important to keep the idea that, on the (very) long run, a `Rgb<f16>`/`Rgba<f16>`/`ColorType::Rgb16F`/`ColorType::Rgba16F` could make sense.

Keep the good work! :)

johannesvollmer:
> That could probably be fixed once ColorType::Rgb32F is added to describe the output format

That's what I meant. As it probably is a breaking change, that would be adequate for the `next` branch, and not master, right?

> Rgb<f16>/Rgba<f16>/ColorType::Rgb16F/ColorType::Rgba16F could make sense

Yes, we discussed that. We won't want to rule it out as a storage format, but for now it's only F32.

fintelia:
> > That could probably be fixed once ColorType::Rgb32F is added to describe the output format
> 
> That's what I meant. As it probably is a breaking change, that would be adequate for the next branch, and not master, right?

ColorType is actually marked non-exhaustive, so adding variants isn't a breaking change.

johannesvollmer:
Yes, but removing the HDR adapter probably is?:D

johannesvollmer:
Okay, then I think we could merge this. I'll open an Issue for the HDR adapter so we don't forget to have a look at it later.

fintelia (approved):

johannesvollmer:
Cool thanks <3

PR #1482, opened 2021-05-22
Remove Bgr and Bgra color types

fintelia:
Remove all handling of BGR and BGRA images, except for in ExtendedColorType. Users can trivially convert images just by swapping the red and blue components, but it is unclear whether they're even being used at all now.

Started as an experiment to see what it would take to remove ColorType::{Bgr8, Bgra8}, but it turns out that that the pixel structs are sufficiently entangled that you can't really remove one without the other.

### DIFF
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -80,7 +80,7 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for AvifDecoder<R> {
     }
 
     fn color_type(&self) -> ColorType {
-        ColorType::Bgra8
+        ColorType::Rgba8
     }
 
     fn into_reader(self) -> ImageResult<Self::Reader> {
@@ -156,6 +156,11 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for AvifDecoder<R> {
                     buf[3 + i * 4] = slice[i];
                 }
             }
+        }
+
+        // Convert Bgra to Rgba
+        for chunk in buf.chunks_exact_mut(4) {
+            chunk.swap(0, 2);
         }
 
         Ok(())
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ All image processing functions provided operate on types that implement the `Gen
 | AVIF   | Only 8-bit | Lossy |
 | PNM    | PBM, PGM, PPM, standard PAM | Yes |
 | DDS    | DXT1, DXT3, DXT5 | No |
-| TGA    | Yes | RGB(8), RGBA(8), BGR(8), BGRA(8), Gray(8), GrayA(8) |
+| TGA    | Yes | RGB(8), RGBA(8), Gray(8), GrayA(8) |
 | farbfeld | Yes | Yes |
 
 ### The [`ImageDecoder`](https://docs.rs/image/*/image/trait.ImageDecoder.html) and [`ImageDecoderExt`](https://docs.rs/image/*/image/trait.ImageDecoderExt.html) Traits
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -6,7 +6,7 @@ use std::ops::{Deref, DerefMut, Index, IndexMut, Range};
 use std::path::Path;
 use std::slice::{ChunksExact, ChunksExactMut};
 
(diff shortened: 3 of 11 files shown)


src/codecs/avif/decoder.rs
```diff
@@ -158,6 +158,11 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for AvifDecoder<R> {
             }
         }
 
+        // Convert Bgra to Rgba
+        for chunk in buf.chunks_exact_mut(4) {
+            chunk.swap(0, 2);
+        }
```
HeroicKatora, line 164: I would like to see this handled in `DynamicImage` instead, if possible. The decision to remove `Bgra` would then be local to our buffer types while more complex third-parties can make use of the original byte order.

fintelia: I'd say that the correct thing would be to report `Bgra` as the `original_color_type()`, and then a forthcoming API addition could expose the ability for advanced users to decode images with the original color type instead of the more constrained `color_type()`. Keeping `ColorType::Bgra` means that _every_ user of ImageDecoder has to at least in theory handle BGRA images.

The library actually isn't even currently consistent on this area. The TGA decoder actually [already swaps BGR -> RGB](https://github.com/image-rs/image/blob/7d280f5745fcbd1c8af819b265dc81b63cacd6b4/src/codecs/tga/decoder.rs#L318).

fintelia:
If there are no objections, I'll go ahead and merge this into `next`. We can revert later if this ends up causing issues

HeroicKatora (approved):

gyk:
Hi, I think Bgr and Bgra are still relevant if you have to make intensive use of system APIs. Here's some snippet from our code
base (a desktop application):

```rs
// Windows: Get color type from BITMAPINFOHEADER
let color = if bmih.biBitCount == 24 {
    ColorFormat::Bgr
} else if bmih.biBitCount == 32 {
    ColorFormat::Bgra
} else {
    // ...
};


// macOS: Get color type from CGImage
let info = unsafe { CGImageGetAlphaInfo(image.as_ptr()) };

let color = match info {
    | CGImageAlphaInfo::CGImageAlphaFirst
    | CGImageAlphaInfo::CGImageAlphaPremultipliedFirst => ColorFormat::Bgra,

    // ...
};
```

Although this can be done by swapping raw bytes, keeping BGR(A) type would make it more convenient to use, as stated in #451.

fintelia:
@gyk Unfortunately having more color types adds complexity throughout the crate and to our public API. If you have a specific use case and want to talk through possible API changes to accommodate, please feel free to open a new issue

HeroicKatora:
@gyk Also note: we've removed them from our _internal_ buffer representation as it vastly simplified the algorithmic portions. The expectation was that one could perform a conversion at the interface, where this has always been necessary in many cases regardless. For example, expanding bit-maps to bytes due to Rust's aliasing unit being a full byte. Or to merge all channels to a single plane. Similar issues had been internally where the duplication of channel order meant some inconsistencies in supported operations. I hope that this motivates better converters rather than undoing the buffer simplifications?

gyk:
@fintelia @HeroicKatora Ok, that sounds reasonable. Thanks!

WilliamVenner:
Why would you remove BGR/BGRA? It is a very common format for system APIs and buffers.

What's worse is that this PR didn't even come with a major version bump. This is a breaking change.

fintelia:
@WilliamVenner some of the motivation is described earlier in this thread. And we *did* do a major version bump when making this change: `0.23.x` -> `0.24.0`

zaynetro:
> @fintelia macos CVPixelBufferCreate uses kCVPixelFormatType_32BGRA and it doesn't support kCVPixelFormatType_32RGBA

Example of converting from BGRA8888 to RGBA:

```
fn convert_bgra(width: u32, height: u32, mut bgra: Vec<u8>) -> Option<RgbaImage> {
    for src in bgra.chunks_exact_mut(4) {
        let (blue, green, red, alpha) = (src[0], src[1], src[2], src[3]);
        src[0] = red;
        src[1] = green;
        src[2] = blue;
        src[3] = alpha;
    }
    RgbaImage::from_raw(width, height, bgra)
}
```

You can find more alternatives in this thread: https://users.rust-lang.org/t/converting-a-bgra-u8-to-rgb-u8-n-for-images/67938

PR #1485, opened 2021-05-29
Implement From on DynamicImage

Palladinium:
This adds some convenience `From` implementations from each `DynamicImage` variant to the enum itself.
You can already convert any variant to the enum by just constructing the variant itself, but this can help cut down repetition and allow for some generic bounds such as `T: Into<DynamicImage>`.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -983,6 +983,66 @@ impl DynamicImage {
     }
 }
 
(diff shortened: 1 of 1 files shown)


fintelia (approved):

HeroicKatora:
Thanks, that is quite useful.

PR #1487, opened 2021-05-31
[WIP] Silicon Graphics Image support

Moggers:
This is a naive implementation of *reading* SGI images, it lacks support for BPC=2, and I haven't had a chance to test the dimensions=2 or dimensions=1.
It works, but still needs a bit of cleanup and tests.


fintelia:
Thanks for sharing this! Do you have a sense of how widely used SGI files are? I hadn't heard of them before so I don't have a good sense of whether this is a format that other people are going to want to work with using `image`.

Moggers:
>Thanks for sharing this! Do you have a sense of how widely used SGI files are? I hadn't heard of them before so I don't have a good sense of whether this is a format that other people are likely going to want to work with using image.

if there is a god and he is merciful, I doubt very many people at all will be using this format. This usage is the only time I've ever run into the format, and the software which outputs and ingests them is quite old. On one hand the number of people who will have to use this is going to be vanishingly low.. On the other hand if people do need to work with SGI images they're not going to have much choice aside from implementing it themselves (at the moment).

I've no idea what extent of the work involved to maintain each format is, so I'd understand if you don't want to take on the burden of the additional implementation.

fintelia:
Maintaining format decoders isn't a huge amount of work, but it isn't zero either. If someone else comes along and says this is super important we can revive it or see about spinning it off into another crate, but for now I'm going to go ahead and close this.

HeroicKatora:
@Moggers I'm agreeing with @fintelia here. You're encouraged to publish this as a standalone crate, similar to `png`. It's quite some overhead to handle potential bug reports and feature requests on a format where none of the usual maintainers have any experience with. However, note that adding bindings to a separate crate within image is another matter. A simple 'translation layer' is easier to maintain and I think we can be open to discussing the complexity of such a thing separately :)

NWPlayer123:
Following up - I could use support for this. I'm in the process of writing a Bevy loader for Panda3D models/Toontown, which has native support for `.rgb`. Toontown mostly uses it for storing alpha channels (dimensions=2, RLE) alongside a jpeg file. I ask because the easiest way to get "official" support would be to support it here as bevy_image uses this crate for all its texture loading. For now, I'll probably just add a separate AssetLoader on my side, but let it be known.

Issue #1243, opened 2020-05-29
Discuss name of `GenericImage::opt_pixel`


HeroicKatora:
`try_get_pixel` is not perfect (it may indeed suggest that the return type is a `Result`), but I think it conveys the meaning of what the function does better than `opt_get_pixel`. I wouldn't have guessed what the function behavior is just by its name.

_Originally posted by @lovasoa in https://github.com/image-rs/image/pull/1229#issuecomment-636150657_

HeroicKatora:
I'll quote my original intention for choosing these names:

> I don't think `try` is appropriate, the return type [is not an Result]. What I would really want is `get_` as there is prior art for that in `slice::get` itself but alas the name `get_pixel` is taken and `get` seemed too risky to add. So this models the `_ref`, `_mut` suffixes instead, but as a prefix since it is more important and to parallel the `get_pixel` name. I intend to rename this as part of moving the default impl in `0.24`.
> 
> If the justification doesn't sound right, open an issue and we'll get it changed or even reverted for now to avoid a commitment to the name with the release which I'd like to land this WE.

fintelia:
Does it have to return an option? We could also return say a `Result<.., OutOfBoundsError>`. That would also clarify that the method should only be returning an Err / None if the provided coordinates were out of bounds.

HeroicKatora:
Similar story, I think it should have the same signature as `slice::get` because it is basically that. Adding a ZST error type only to justify the name appears seem like putting the cart before the horse.

okaneco:
I experimented with changing `GenericImageView::get_pixel` to return `Option<Self::Pixel>` and `GenericImage::get_pixel_mut` to return `Option<&mut P>` in this branch https://github.com/okaneco/image/commit/710d0441ecbf5acd454b25887fc6be5da5e76e20 and it passes tests locally. The relevant library changes occupy the first 55% of the page, below that is mainly `unwrap` boilerplate in `imageops` to maintain current behavior. I initially tried to return `Option<&Self::Pixel>` but ran into lifetime issues eventually with `SubImage` and the `imageops` functions/`DynamicImage`.

This avoids panics in `get_pixel`, `get_pixel_mut`, and `put_pixel` (should `put_pixel` return Err if the pixel doesn't exist?). The changes seem to improve some benchmarks. Using `cargo +nightly bench --features=benchmarks` with `rustc 1.52.0-nightly (107896c32 2021-03-15)`.
```
// branch
[...]

PR #1500, opened 2021-06-26
Add non-panicking get_pixel/mut methods to ImageBuffer

okaneco:
This relands part of #1229 which added methods to `ImageBuffer` that return
`Option<&P>` and `Option<&mut P>` from `get_pixel[_mut]`-like methods. It does
not include adding methods to the `GenericImage` and `GenericImageView` traits.

Naming is still a concern #1243.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -726,6 +726,14 @@ where
         }
     }
 
+    /// Gets a reference to the pixel at location `(x, y)` or returns `None` if
+    /// the index is out of the bounds `(width, height)`.
+    pub fn get_pixel_checked(&self, x: u32, y: u32) -> Option<&P> {
+        self.pixel_indices(x, y)
+            .and_then(|pixel_indices| self.data.get(pixel_indices))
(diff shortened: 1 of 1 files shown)


fintelia:
If this is purely about getting the optimizer to produce better code, I wonder if we should instead do something like:
```rust
if x >= self.width { 
    return None;
}
let num_channels = <P as Pixel>::CHANNEL_COUNT as usize;
let i = (y as usize).saturating_mul(self.height as usize).saturating_add(x as usize);

self.data.get(i..i+num_channels)
    .map(|pixel_indices| <P as Pixel>::from_slice(pixel_indices))
```

This way we incorporate the invariant that width * height = data.len(), and avoid an extra check.

okaneco:
Thanks for that suggestion, I've incorporated that change.

I'm interested in better code generation but more so in providing methods that return Option or Result to users where possible instead of panicking.

fintelia (approved):

PR #1510, opened 2021-07-15
Better handling of image size overflows

fintelia:
Avoids a possible panic from allocating more than isize::MAX bytes for a single Vec, but otherwise doesn't prevent attempts to do unreasonably large allocations.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -387,7 +387,7 @@ pub(crate) fn load_rect<'a, D, F, F1, F2, E>(x: u32, y: u32, width: u32, height:
     let scanline_bytes = decoder.scanline_bytes();
     let total_bytes = width * height * bytes_per_pixel;
 
-    if buf.len() < usize::try_from(total_bytes).unwrap_or(usize::MAX) {
+    if buf.len() < usize::try_from(total_bytes).unwrap_or(usize::max_value()) {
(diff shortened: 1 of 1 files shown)


src/image.rs
```diff
@@ -555,10 +562,10 @@ pub trait ImageDecoder<'a>: Sized {
     /// This is the size of the buffer that must be passed to `read_image` or
     /// `read_image_with_progress`. The returned value may exceed usize::MAX, in
     /// which case it isn't actually possible to construct a buffer to decode all the image data
-    /// into.
+    /// into. If, however, the size does not fit in a u64 then u64::MAX is returned.
     fn total_bytes(&self) -> u64 {
         let dimensions = self.dimensions();
-        u64::from(dimensions.0) * u64::from(dimensions.1) * u64::from(self.color_type().bytes_per_pixel())
+        (u64::from(dimensions.0) * u64::from(dimensions.1)).saturating_mul(u64::from(self.color_type().bytes_per_pixel()))
```
HeroicKatora, line 568: This suffers from small readability problem now, rather split up in two lines:
```suggestion
        let pixel_count = u64::from(dimensions.0) *u64::from(dimensions.1);
        let bpp = u64::from(self.color_type().bytes_per_pixel());
        pixel_count.saturating_mul(bpp)
```

HeroicKatora (commented):
:+1: , after readability nit

PR #1516, opened 2021-07-19
Add Rgb(a)32F DynamicImage Variant

johannesvollmer:
_Work in progress | Depends on other Issues to be resolved first_

Adds F32 support to the DynamicImage.

Points to be discussed:
- [ ] `grayscale(image)` returns `Luma<T>`, but `Luma<f32>` isn't a supported color type, so... `f32` images will panic?
- [ ] `impl GenericImage for DynamicImage { type SubPixel = u8; }` is difficult for float images, as it is unclear how `f32` colors should be converted to `u8` colors. Maybe the default subpixel type should be `f32` for best quality, instead of `u8`?

(couldn't resist to add other changes as well, they are of course to be discussed)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ name = "image"
 path = "./src/lib.rs"
 
 [dependencies]
-bytemuck = "1"
+bytemuck = { version = "1.7.0", features = ["extern_crate_alloc"] } # includes cast_vec
 byteorder = "1.3.2"
 num-iter = "0.1.32"
 num-rational = { version = "0.3", default-features = false }
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -41,6 +41,7 @@ use crate::traits::Pixel;
 /// A Dynamic Image
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub enum DynamicImage {
+
     /// Each pixel in this image is 8-bit Luma
     ImageLuma8(GrayImage),
 
@@ -64,63 +65,49 @@ pub enum DynamicImage {
 
     /// Each pixel in this image is 16-bit Rgb with alpha
     ImageRgba16(Rgba16Image),
+
+    /// Each pixel in this image is 32-bit float Rgb
+    ImageRgb32F(Rgb32FImage),
+
+    /// Each pixel in this image is 32-bit float Rgb with alpha
+    ImageRgba32F(Rgba32FImage),
 }
 
 macro_rules! dynamic_map(
-        ($dynimage: expr, ref $image: ident => $action: expr) => (
-                match $dynimage {
-                        DynamicImage::ImageLuma8(ref $image) => DynamicImage::ImageLuma8($action),
-                        DynamicImage::ImageLumaA8(ref $image) => DynamicImage::ImageLumaA8($action),
-                        DynamicImage::ImageRgb8(ref $image) => DynamicImage::ImageRgb8($action),
-                        DynamicImage::ImageRgba8(ref $image) => DynamicImage::ImageRgba8($action),
-                        DynamicImage::ImageLuma16(ref $image) => DynamicImage::ImageLuma16($action),
-                        DynamicImage::ImageLumaA16(ref $image) => DynamicImage::ImageLumaA16($action),
-                        DynamicImage::ImageRgb16(ref $image) => DynamicImage::ImageRgb16($action),
-                        DynamicImage::ImageRgba16(ref $image) => DynamicImage::ImageRgba16($action),
-                }
-        );
+        ($dynimage: expr, $image: pat => $action: expr) => ({
+            use DynamicImage::*;
+            match $dynimage {
+                ImageLuma8($image) => ImageLuma8($action),
+                ImageLumaA8($image) => ImageLumaA8($action),
+                ImageRgb8($image) => ImageRgb8($action),
+                ImageRgba8($image) => ImageRgba8($action),
+                ImageLuma16($image) => ImageLuma16($action),
+                ImageLumaA16($image) => ImageLumaA16($action),
+                ImageRgb16($image) => ImageRgb16($action),
+                ImageRgba16($image) => ImageRgba16($action),
+                ImageRgb32F($image) => ImageRgb32F($action),
+                ImageRgba32F($image) => ImageRgba32F($action),
+            }
+        });
 
-        ($dynimage: expr, ref mut $image: ident => $action: expr) => (
-                match $dynimage {
-                        DynamicImage::ImageLuma8(ref mut $image) => DynamicImage::ImageLuma8($action),
-                        DynamicImage::ImageLumaA8(ref mut $image) => DynamicImage::ImageLumaA8($action),
-                        DynamicImage::ImageRgb8(ref mut $image) => DynamicImage::ImageRgb8($action),
-                        DynamicImage::ImageRgba8(ref mut $image) => DynamicImage::ImageRgba8($action),
-                        DynamicImage::ImageLuma16(ref mut $image) => DynamicImage::ImageLuma16($action),
-                        DynamicImage::ImageLumaA16(ref mut $image) => DynamicImage::ImageLumaA16($action),
-                        DynamicImage::ImageRgb16(ref mut $image) => DynamicImage::ImageRgb16($action),
-                        DynamicImage::ImageRgba16(ref mut $image) => DynamicImage::ImageRgba16($action),
-                }
-        );
-
-        ($dynimage: expr, ref $image: ident -> $action: expr) => (
-                match $dynimage {
-                        DynamicImage::ImageLuma8(ref $image) => $action,
-                        DynamicImage::ImageLumaA8(ref $image) => $action,
-                        DynamicImage::ImageRgb8(ref $image) => $action,
-                        DynamicImage::ImageRgba8(ref $image) => $action,
-                        DynamicImage::ImageLuma16(ref $image) => $action,
-                        DynamicImage::ImageLumaA16(ref $image) => $action,
-                        DynamicImage::ImageRgb16(ref $image) => $action,
-                        DynamicImage::ImageRgba16(ref $image) => $action,
-                }
-        );
-
-        ($dynimage: expr, ref mut $image: ident -> $action: expr) => (
-                match $dynimage {
-                        DynamicImage::ImageLuma8(ref mut $image) => $action,
-                        DynamicImage::ImageLumaA8(ref mut $image) => $action,
-                        DynamicImage::ImageRgb8(ref mut $image) => $action,
-                        DynamicImage::ImageRgba8(ref mut $image) => $action,
-                        DynamicImage::ImageLuma16(ref mut $image) => $action,
-                        DynamicImage::ImageLumaA16(ref mut $image) => $action,
-                        DynamicImage::ImageRgb16(ref mut $image) => $action,
-                        DynamicImage::ImageRgba16(ref mut $image) => $action,
-                }
+        ($dynimage: expr, |$image: pat| $action: expr) => (
+            match $dynimage {
+                DynamicImage::ImageLuma8($image) => $action,
+                DynamicImage::ImageLumaA8($image) => $action,
+                DynamicImage::ImageRgb8($image) => $action,
+                DynamicImage::ImageRgba8($image) => $action,
+                DynamicImage::ImageLuma16($image) => $action,
+                DynamicImage::ImageLumaA16($image) => $action,
+                DynamicImage::ImageRgb16($image) => $action,
+                DynamicImage::ImageRgba16($image) => $action,
+                DynamicImage::ImageRgb32F($image) => $action,
+                DynamicImage::ImageRgba32F($image) => $action,
+            }
         );
 );
 
 impl DynamicImage {
+
     /// Creates a dynamic image backed by a buffer of grey pixels.
     pub fn new_luma8(w: u32, h: u32) -> DynamicImage {
         DynamicImage::ImageLuma8(ImageBuffer::new(w, h))
@@ -163,6 +150,16 @@ impl DynamicImage {
         DynamicImage::ImageRgba16(ImageBuffer::new(w, h))
     }
 
+    /// Creates a dynamic image backed by a buffer of RGB pixels.
+    pub fn new_rgb32f(w: u32, h: u32) -> DynamicImage {
+        DynamicImage::ImageRgb32F(ImageBuffer::new(w, h))
+    }
+
+    /// Creates a dynamic image backed by a buffer of RGBA pixels.
+    pub fn new_rgba32f(w: u32, h: u32) -> DynamicImage {
+        DynamicImage::ImageRgba32F(ImageBuffer::new(w, h))
+    }
+
     /// Decodes an encoded image into a dynamic image.
     pub fn from_decoder<'a>(decoder: impl ImageDecoder<'a>)
         -> ImageResult<Self>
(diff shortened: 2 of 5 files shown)


johannesvollmer:
I think we can't give up `impl<T> Pixel for Rgba<T>` in favor of `for Rgba<f32> & for Rgba<u16> & for Rgba<u8>`: This results in problems when converting from any color to luma in `fn grayscale`.

It does not compile, because the architecture does not encode that any `Rgb<T>` can be converted to any `Luma<T>`, with the result still being `Pixel` (because an arbitrary `Rgb<u128>` may implement Pixel, but `Luma<u128>` may not).

johannesvollmer:
@fintelia @HeroicKatora 

Here's a first draft. Would you like to discuss?

These are the most important limitations of this approach concerning `DynamicImage`:
- `DynamicImage` cannot implement `Eq` and `Hash` anymore because `f32` doesn't want to implement that
- `Luma<f32>` and `LumaA<f32>` do not implement `Pixel`. Therefore, `fn grayscale` panics. Could also return an error


These are the most important limitations of this approach concerning `Pixel`:
- Assumes white means `1.0` for `f32`, otherwise all the color traits cannot be implemented for `f32`
- `ColorType` is incorrect for `Luma&f32` and for `usize` and `i32` (but this was the case for f32 before, so apparently this is not worse than before?)
- Does not scale very well in case we want to add more colors, because each color needs a `const ColorType` in `trait Sample`

Also, I could not resist doing some deprecation progress (boyscout rule).

src/traits.rs
```diff
@@ -116,15 +116,64 @@ impl Lerp for f32 {
     }
 }
 
-use crate::color::{LSample, LaSample, RgbSample, RgbaSample};
+pub trait Sample: Primitive + 'static {
+    const MAX_SAMPLE_VALUE: Self;
+    const RGB_COLOR_TYPE: ColorType;
+    const RGBA_COLOR_TYPE: ColorType;
+    const L_COLOR_TYPE: ColorType;
+    const LA_COLOR_TYPE: ColorType;
```
fintelia, line 124: Perhaps these should be `Option<ColorType>`?

johannesvollmer: I tried that, but I wasn't brave enough to also make the `trait Pixel::COLOR_TYPE` optional. That would be a lot cleaner though. And it would result in runtime errors instead of subtly incorrect behavior.

johannesvollmer: I can do that. Alternatively, we could also make a separate trait only for the color type. Then we could rule out these errors at compile time, instead of sometimes throwing a runtime error.

```rust
trait PixelWithColorType {
    const COLOR_TYPE: ColorType;
}
```

And then only some pixels will implement that, and the methods that use this constants could require this trait in addition to `Pixel`. What do you think?

johannesvollmer: I'd like to do this:
```rust
impl Sample for f32 {
    const RGBA_COLOR_TYPE: ImageResult<ColorType> = Ok(ColorType::Rgba32F);
    const L_COLOR_TYPE: ImageResult<ColorType> = Err(UnsupportedPixelError);
}

/*const UnsupportedPixelError: ImageError = ImageError::Unsupported(UnsupportedError::from_format_and_kind(
    ImageFormatHint::Unknown, UnsupportedErrorKind::Color(ExtendedColorType::Unknown(255))
));*/

/*const UnsupportedPixelError: ImageError = ImageError::Parameter(ParameterError::from_kind(
    ParameterErrorKind::FailedAlready
));*/
```

But I can't find an error variant which would make sense

fintelia: I'm nervous that fracturing Pixel like that. Downstream code might end up having to carry around a whole bunch of different trait bounds. Worse adding any new bounds would be API breaking.

Making Pixel::COLOR_TYPE optional would be an improvement. Though I wonder whether it would be feasible to simply only implement Pixel for color types. That might entail converting `Rgb<T>` -> `Rgb8` + `Rgb16` + `Rgb32F`, etc.

johannesvollmer: > Worse adding any new bounds would be API breaking

How is that worse than changing the associated constant in a trait? 

In general, I would not say that a monolithic trait has advantages. The pixel trait has multiple responsibilities, two of which are (a) data storage for image processing and (b) color format inference for image file saving. Having both of those responsibilities in one trait is problematic, because you will end up in one of the following drawbacks: Either you force a color format on all pixels (even if you only need and intermediate `Luma<f32>` image), which now is a runtime error. Or you unnecessarily limit color processing algorithms and temporary storage to the few select pixel types which can be saved to a file, which could otherwise also work with `Luma<f32>`.

johannesvollmer: I got a little carried away, thinking of an ideal solution. To cut this short, I think that splitting up the trait may be a larger change, for another time. Maybe we should focus on going `f32` with the smallest amount of changes for now :)

johannesvollmer:
So, can you do another review?
@fintelia @HeroicKatora 

Something blocking this?

src/traits.rs
```diff
@@ -99,20 +110,84 @@ impl Lerp for u32 {
...
+    }
+}
+
+
+pub trait Sample: Primitive + 'static {
+    const MAX_SAMPLE_VALUE: Self;
+    const RGB_COLOR_TYPE: ColorTypeOrErr;
+    const RGBA_COLOR_TYPE: ColorTypeOrErr;
+    const L_COLOR_TYPE: ColorTypeOrErr;
+    const LA_COLOR_TYPE: ColorTypeOrErr;
```
HeroicKatora, line 127: Usage of an associated constant brings up an interesting issue that wasn't obvious to me from the start, related to forward compatibility. We can't really change these constants in a minor version without risk of breakage. At least we don't have a guarantee here. This is because downstream users might have code such as this: 

```rust
<u32 as Sample>::L_COLOR_TYPE.is_err()
```

[Playground for trying it out](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4d812535bcbac5b0446022ae375db23c)

That's quite unfortunate, isn't it? The only real remedy would be _hiding_ the existance of these constants by putting them in a public-private trait.

But of course the same would apply to the constants in `Luma<_> as Pixel`. Of course there we could change them to not be constants afterall, but just have static accessor methods. But what's the cost associated with that?

HeroicKatora (commented):
> Something blocking this?

Yeah, finding time for careful deliberation on the maintainability of the interface which I finally did. I'm pretty content with most of these changes. There's maybe two things that I find slightly annoying from an user perspective:

* Is the `Primitive` trait useful any more? We do control both of these traits after all. Due to the changes to `Pixel` everyone needs to switch to `Sample` anyways. The churn could be far less if we'd merge those two traits and just change `Primitive`.

johannesvollmer:
Glad that you took the time :) 

> Is the Primitive trait useful any more? We do control both of these traits after all. Due to the changes to Pixel everyone needs to switch to Sample anyways. The churn could be far less if we'd merge those two traits and just change Primitive

Yes. However, I think `Sample` would be the best name for the merged trait. It could be confusing for downstream users to find out that the generic sounding `Primitive` trait contains highly color-related content. I'd favor merging the traits, given that the name of the merged trait is not confusing.

src/traits.rs
johannesvollmer: I think the solution with the associated constant has a heap of problems. It's not a clean design, it's a rigid architecture that makes future changes harder. The clean solution would be to change the `Pixel` trait.

The example you linked in the playground shows one of the symptoms. However, if we cannot change the pixel trait, this is an acceptable trade-off. If we need to change it, let it be a breaking change. That's the sacrifice we need to make by not cleaning up the pixel trait

src/traits.rs
```diff
@@ -99,20 +110,84 @@ impl Lerp for u32 {
...
+impl Lerp for f32 {
+    type Ratio = f32;
+
+    fn lerp(a: Self, b: Self, ratio: Self::Ratio) -> Self {
+        a + (b - a) * ratio
+    }
+}
+
+
+pub trait Sample: Primitive + 'static {
```
HeroicKatora, line 122: Opening this to deliberate on the name of the trait (whether or not we merge it with the current `Primitive`). My problem with the name `Sample` is that in GPU is refers to one computed color instance of a pixel in a fragment—of which there could be multiple in multisampled graphics pipelines.

Another alternative would be `Channel` but this one has the added meaning of an _interpretation_ of the number. For example, red channel, green channel of an RGB-ish pixel make sense but 'the u8 channel' doesn't.

The last name that comes to mind is `Component` which, at least in my intuitive interpretation, is very closely related to the bit-encoding. That could be fitting here. We currently only provide color representations with homogeneous component depths. So the description of this trait could be summed up similar to:

> A trivial type that can be used for any channel value in a color representation.

src/traits.rs
```diff
@@ -99,20 +110,84 @@ impl Lerp for u32 {
...
+    const MAX_SAMPLE_VALUE: Self;
```
HeroicKatora, line 123: Together with the above this is a bit odd. If it is an inherent property of `Component` then it must be the same for all usages in `Pixel`—including in alpha and non-alpha channels. This may be okay, or not? I don't have strong evidence either way.

src/traits.rs
HeroicKatora: Pretty much exactly my throughts. However, we are in the fortunate situation of being able to break the pixel trait. We should only be careful and keep the changes as small as possible.

src/traits.rs
johannesvollmer: Agreed :)

I'd favor `Component`, but it is an overused word, maybe `PixelComponent`? Or, instead of `Channel`, we could use `ChannelType` or `ChannelPrimitive`?

src/traits.rs
johannesvollmer: Good point. 

There's a few good places to put this value, but they are all more complex than this constant. I would define `0..1` as the default float range for all channels, and would only make it more complex once someone actually has a use case for it. Keeping it simple for now. 

Rename the constant to `DEFAULT_MAX_COMPONENT_VALUE` or even `DEFAULT_MAX_COMPONENT_VALUE_FOR_ALL_CHANNELS` would keep the door open for adding some kind of override later on (maybe in a color space definition, or even in the `invert(max_rgb: 1.0, max_alpha: 1.0)` function call).

Keep in mind that merging this pr will not set this decision in stone. It could still be added to the roadmap for the next release.

src/traits.rs
johannesvollmer: I think it would be nice to also rename `trait Pixel { type Subpixel: PixelComponent; }` to `.. { type Component: PixelComponent }`. What do you think?

johannesvollmer: Should we rename the `sample` in all places in the project, for example also  `Buffer::into_flat_samples` to  `Buffer::into_flat_components`?

src/traits.rs
johannesvollmer: The last time I tried to come up with a clean solution, I ended up completely reworking the pixel trait. I don't know whether I can come up with a smaller refactor, but I can try

src/traits.rs
HeroicKatora: > I think it would be nice to also rename trait Pixel { type Subpixel: PixelComponent; ... } to trait Pixel { type Component: PixelComponent ... }. What do you think?

> Should we rename the sample in all places in the project, for example also Buffer::into_flat_samples to Buffer::into_flat_components?

Maybe but it's examples of those 'nice but high churn from breakage' changes. For the methods I would deprecate/hide the existing implementations and then add the new naming on top (and remove in 0.25). I wouldn't rename the associated type as there is no compatibility/deprecation on it at the moment. The trait rename is acceptable in so far as the trait itself will undergo a major change.

src/traits.rs
HeroicKatora: What if we move the associated const `COLOR_TYPE` into a subtrait, something like `ColoredPixel`.

HeroicKatora: Because I'm decently sure the constant isn't actually being used in the _type system_ (how could it?) so it should be generally feasible to add that extra bound to usage sites.

src/traits.rs
fintelia: Also, please only do that sort of widespread change in a standalone PR! If you are purely changing names it can just be a matter of skimming everything, but if there are functional changes as well the reviewing is much more time consuming

fintelia: For a merged PixelComponent trait, I would also suggest making it sealed so that we can guarantee there are no downstream implementations.

johannesvollmer: Alright, I'll rename only the types that were introduced in this pr and make the trait sealed :)

src/traits.rs
johannesvollmer: The `COLOR_TYPE` constant is used in the `impl trait Pixel for $T` macro. It's necessary because the Pixel trait defines the constant `COLOR_TYPE: ColorTypeOrErr`.

I'll try separating them

src/traits.rs
HeroicKatora: To be frank I think sealing the trait in this one makes it _worse_. We don't gain any additional ability to change the associated constants or to resolve the color space tension. We can't rely on the sealing for any effect as most of our implementations are generic over `<T: Primitive>` anyways.

But now no one downstream can change our defaults by, e.g. implementing a `Component` that has its maximum value at a different constant or one for their own half-floats, for non-NaN floats etc.

johannesvollmer:
Considering all the discussions about renaming and splitting, what do you think of the following `PixelComponent` definition?

```rust
/// The type of each channel in a pixel. For example, this can be `u8`, `u16`, `f32`.
// TODO rename to `PixelComponent`
pub trait Primitive: Copy + NumCast + Num + PartialOrd<Self> + Clone + Bounded {
    const DEFAULT_MAX_COMPONENT_VALUE: Self = usize::MAX;
}

/* ... */

/// The type of each channel in a pixel, with an associated color type.
pub trait ComponentWithColorType: Primitive + 'static + private::Sealed {
    const RGB_COLOR_TYPE: ColorTypeOrErr;
    const RGBA_COLOR_TYPE: ColorTypeOrErr;
    const L_COLOR_TYPE: ColorTypeOrErr;
    const LA_COLOR_TYPE: ColorTypeOrErr;
}
```

HeroicKatora:
The main problem seems to be the associated `ColorType` constant, in combination with the impls on `Luma*`. The color type just is not a property of the component, what's stopping us from getting rid of it? I would move that to an extension trait that we _selectively_ implement in `Luma*`, `Rgb*` etc.

src/traits.rs
HeroicKatora: Yes, I know that's what Pixel is doing but it's not a functionality. That's just the same question again on a different trait. But to what effect?

HeroicKatora:
So more concretely I would suppose something like this:

```rust
/// The type of each channel in a pixel. For example, this can be `u8`, `u16`, `f32`.
// TODO rename to `PixelComponent`
pub trait Primitive: Copy + NumCast + Num + PartialOrd<Self> + Clone + Bounded {
    const DEFAULT_MAX_COMPONENT_VALUE: Self = usize::MAX;
}

pub trait Pixel {
}

pub trait ColoredPixel: Pixel {
    const COLOR_TYPE: ColorType;
}

```

johannesvollmer:
I proposed such a solution once. It's one of the cleanest solutions I had come up with. I don't recall why we decided against it at that time, but I think it's a great solution. I'll go on and push some changes with this strategy next

fintelia:
I think we're at the stage were we should probably be merging `next` into `master`, potentially with one final 0.23.x release first. @HeroicKatora do you have thoughts?

fintelia:
:+1: In that case, seems to just be a matter of getting all the merges to go through so everything ends up on `master`

PR #1517, opened 2021-07-20
Remove deprecated items

### LINKED ISSUE #1434: Remove `math::{nq, utils}`
All the items in these two submodules are deprecated, so at the next major release we should remove them.

JulianKnodt:
This removes deprecated items for the next release.

Fixes #1434


fintelia:
I just tried to switch the target branch to `next`, but it doesn't seem to have gone cleanly. Could you try making a PR against that branch instead? We'll be using it for the next major release and it might already contain some of these changes

PR #1521, opened 2021-07-26
Use seekable writer

### LINKED ISSUE #1489: Optimize Write for Seeking
This issue is a reminder to optimize encoders based on the change that writers are now seekable in the API. For example, the exr encoder writes to a seekable intermediate buffer, which can be removed now that the writers are seekable.

johannesvollmer:
Use the seekable writer to improve performance and to simplify code. Also, cleaned up some imports in the openexr module.

__The writers were not seekable, even though we had already merged seekable writers into `next`. I had to re-add seekable trait bounds to all methods.__ Maybe those changes got lost?

Modules that benefit from seekable writers, because they had to write to an intermediate `Cursor<Vec<u8>>` previously:
- [x] openexr
- [x] tiff

Closes #1489


johannesvollmer:
@fintelia @HeroicKatora I think this could be ready for merging

fintelia (approved):

johannesvollmer:
Thanks :)

PR #1522, opened 2021-07-26
Lets talk about buffering

johannesvollmer:
While scrolling through the code, I noticed that `BufReader` and `BufWriter` are used more than expected. 

1. Sometimes, an already buffered reader or writer is buffered again. Where this is desired, we should add documentation that explains why this is useful. If not, we should remove the unnecessary buffering.
2. `BufRead` is an overly strict trait bound. Most of the time, we should use `Read` instead, with proper documentation. The `BufRead` trait has super specialized methods like `fn fill_buf(&mut self) -> Result<&[u8]>;`. Where these methods are not used, the trait bound is not required.
3. Whether a reader should be buffered or not is sometimes not documented

- [x] farbfeld: removed unnecessary buffering in writer and reader
- [x] use plain read instead of bufread where possible (not possible in pnm and hdr codecs)
- [x] rename identifiers and add documentation to clarify whether a reader or writer is expected to be buffered
- [x] do not buffer only hdr and pmn again in `fn load`

### DIFF
--- a/src/io/free_functions.rs
+++ b/src/io/free_functions.rs
@@ -1,5 +1,5 @@
 use std::fs::File;
-use std::io::{BufRead, BufReader, BufWriter, Seek};
+use std::io::{BufReader, BufWriter, Seek, Read, BufRead};
 use std::path::Path;
 use std::u32;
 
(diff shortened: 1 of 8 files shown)


src/io/free_functions.rs
```diff
@@ -14,107 +14,100 @@ use crate::image::ImageFormat;
...
-        image::ImageFormat::Ico => DynamicImage::from_decoder(ico::IcoDecoder::new(r)?),
+        image::ImageFormat::Ico => DynamicImage::from_decoder(ico::IcoDecoder::new(buffered_reader)?),
         #[cfg(feature = "hdr")]
-        image::ImageFormat::Hdr => DynamicImage::from_decoder(hdr::HdrAdapter::new(BufReader::new(r))?),
+        image::ImageFormat::Hdr => DynamicImage::from_decoder(hdr::HdrAdapter::new(buffered_reader)?),
         #[cfg(feature = "openexr")]
-        image::ImageFormat::OpenExr => DynamicImage::from_decoder(openexr::OpenExrDecoder::new(r)?),
+        image::ImageFormat::OpenExr => DynamicImage::from_decoder(openexr::OpenExrDecoder::new(buffered_reader)?),
         #[cfg(feature = "pnm")]
-        image::ImageFormat::Pnm => DynamicImage::from_decoder(pnm::PnmDecoder::new(BufReader::new(r))?),
```
johannesvollmer, line 65: HDR and PNM were probably only wrapping the reader inside another `BufReader` to satisfy `BufRead` for the decoder. This is no longer necessary because this `load` function accepts a `BufRead` for a long time already. Maybe this wrapping should have been removed earlier, when changing the `load` function signature to accept `BufRead`.

src/io/free_functions.rs
```diff
@@ -129,7 +122,6 @@ pub(crate) fn save_buffer_impl(
     color: color::ColorType,
 ) -> ImageResult<()> {
     let format =  ImageFormat::from_path(path)?;
-    let fout = &mut BufWriter::new(File::create(path)?);
```
johannesvollmer, line 132: this unused variable was still here because `#[allow(unused_variables)]`

HeroicKatora (approved):
LGTM. One thing that this whole `BufRead` vs. `Read` shows is a lack of the standard library, an _optional_ buffered reader where we can discover at runtime if buffering is available or not. I was debating a possible approach here <https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ee7d41e26bb6eb2952e75ad65ccaf5a7> but it's pretty cursed so I didn't want to publish that as a crate ^^.

johannesvollmer:
Yeah, to show buffered/unbuffered, I'd rather use a real type instead of pain documentation, but that's just how it is now haha

PR #1525, opened 2021-08-03
optimise resize

JAicewizard:
A lot of time in resizing is spend on the float conversions. This is partially because `.round()` is slow but also because the type used for the rounding caused unnecessary overhead. This simplifies this, and adds some other performance improvements.

The first commit is a format (oops, I left one comment in. ), I dont know what your policy is on this in PRs. I kept this as a separate commit to make sure it would be easy to review.

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -110,14 +110,28 @@ struct FloatNearest(f32);
 // to_i64, to_u64, and to_f64 implicitly affect all other lower conversions.
 // Note that to_f64 by default calls to_i64 and thus needs to be overridden.
 impl ToPrimitive for FloatNearest {
+    // to_{i,u}64 is required, to_{i,u}{8,16} are usefull.
+    // If a usecase for full 32 bits is found its trivial to add
+    fn to_i8(&self) -> Option<i8> {
+        self.0.round().to_i8()
+    }
+    fn to_i16(&self) -> Option<i16> {
+        self.0.round().to_i16()
+    }
     fn to_i64(&self) -> Option<i64> {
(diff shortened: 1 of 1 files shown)


fintelia:
I don't think that your change from `FloatNearest` -> `.round()` is correct for floating point images, since those shouldn't be rounded at all.

JAicewizard:
Thats a bit of a bummer, thats where the the majority of the performance gain came from.

Without that it is still 17% less time spend, but thats nowhere near the 37% with this.

JAicewizard:
I managed to fix the real cause of the slowdown, the `f32`->`{u,i}64`->`{u,i}{8,16,32,64,128}` chain. This was just one abstraction to many for rustc. Adding dedicated methods for `{u,i}{8,16}` completely resolves this.

JAicewizard:
I noticed that `filter3x3` doesnt round the values. Is there a reason why?

fintelia:
> I noticed that filter3x3 doesnt round the values. Is there a reason why?

Nope, that is a bug as far as I can tell. Must have been missed when resize was converted to do rounding in #1408

fintelia:
So I think the changes to FloatNearest are good to merge, but I'd rather not do the reformatting at the same time. Could you isolate those changes?

JAicewizard:
Yup, that's why I made that a separate commit. Easy delete, its a force push, so if you pulled it make sure to force pull as well.

fintelia (approved):

fintelia:
Could you rebase this on `master`? That should fix the CI issues

JAicewizard:
Done

PR #1536, opened 2021-08-11
Chroma decoding for webp images.

tristanphease:
Adds chroma decoding for webp images. Note loop filtering still isn't implemented so it is still a little blocky.


src/codecs/webp/decoder.rs
```diff
@@ -169,16 +168,16 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for WebPDecoder<R> {
...
 
     fn into_reader(self) -> ImageResult<Self::Reader> {
-        Ok(WebpReader(Cursor::new(self.frame.ybuf), PhantomData))
+        Ok(WebpReader(Cursor::new(self.frame.to_rgb_vec()), PhantomData))
     }
 
     fn read_image(self, buf: &mut [u8]) -> ImageResult<()> {
         assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
-        buf.copy_from_slice(&self.frame.ybuf);
+        buf.copy_from_slice(&self.frame.to_rgb_vec());
```
fintelia, line 180: Could we avoid this copy?

tristanphease: How would we do that, could we return an iterator instead of a vector to convert from YUV to RGB?

HeroicKatora: I think what he meant was that we are performing two copies: Once to construct the return value of `to_rgb_vec` which also takes an additional allocation, an then again when we copy that data to the output slice. Note the implementation of `to_rgb_vec` just requires access to a slice after it has been allocated. So instead there could be a helper function `fill_rgb(&self, _: &mut [u8])` which is called by both this and `to_rgb_vec`.

src/codecs/webp/vp8.rs
```diff
@@ -865,6 +871,50 @@ pub struct Frame {
...
+    fn chroma_width(&self) -> u16 {
+        (self.width + 1) / 2
+    }
+
+    fn chroma_height(&self) -> u16 {
+        (self.height + 1) / 2
+    }
+
+    /// Converts the 4:2:0 YUV vectors to an rgb vector
+    pub fn to_rgb_vec(&self) -> Vec<u8> {
```
fintelia, line 885: You could probably just fold this into the `into_reader` function, since that's the only place this is called. Could even be a simple as:
```rust
fn into_reader(self) -> ImageResult<Self::Reader> {
    let mut data = vec![0; self.ybuf.len() * 3];
    self.fill_rgb(&data);
    Ok(WebpReader(Cursor::new(data), PhantomData))
}
```

tristanphease: Seems like a good idea.

src/codecs/webp/vp8.rs
```diff
@@ -1382,6 +1424,113 @@ impl<R: Read> Vp8Decoder<R> {
...
+                urb.copy_from_slice(&resdata[16 * 16 + i * 16..16 * 16 + i * 16 + 16]);
+
+                let y0 = 1 + y * 4;
+                let x0 = 1 + x * 4;
+                add_residue(&mut uws, &urb, y0, x0, stride);
+
+                let mut vrb = [0i32; 16];
+                vrb.copy_from_slice(&resdata[20 * 16 + i * 16..20 * 16 + i * 16 + 16]);
+
+                add_residue(&mut vws, &vrb, y0, x0, stride);
```
paolobarbolini, line 1522:
```suggestion
                let urb: &[i32; 16] = resdata[16 * 16 + i * 16..][..16].try_into().unwrap();

                let y0 = 1 + y * 4;
                let x0 = 1 + x * 4;
                add_residue(&mut uws, urb, y0, x0, stride);

                let vrb: &[i32; 16] = resdata[20 * 16 + i * 16..][..16].try_into().unwrap();

                add_residue(&mut vws, vrb, y0, x0, stride);
```

This removes two mem copies. I know the `.try_into()` looks bad but it can never fail and the compiler optimizes it away.

You'll also have to add `use std::convert::TryInto;` at the beginning of the file. I can't seem to leave a code suggestion for it.

tristanphease: Will also do this for the luma one then too.

fintelia (approved):
Could you rebase this on master? That should fix the CI issues

PR #1539, opened 2021-08-14
Don't run CI on rustc older than 1.46

fintelia:
This should prevent CI failures and allow us to merge other pull requests


HeroicKatora:
@fintelia Given that this breaks the `0.23` MSRV idea, should we just promote `next` to master and begin work on `0.24` on that branch?

HeroicKatora:
Also, Debian stable (`bullseye`) has been released with `1.48` so we might as well bump it forward to that.

PR #1541, opened 2021-08-17
Bump png to 0.17.0

ldm0:
Ref: https://github.com/image-rs/image-png/issues/272

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ num-rational = { version = "0.4", default-features = false }
 num-traits = "0.2.0"
 gif = { version = "0.11.1", optional = true }
 jpeg = { package = "jpeg-decoder", version = "0.1.22", default-features = false, optional = true }
-png = { version = "0.16.5", optional = true }
+png = { version = "0.17.0", optional = true }
 scoped_threadpool = { version = "0.1", optional = true }
 tiff = { version = "0.6.0", optional = true }
 ravif = { version = "0.7.0", optional = true }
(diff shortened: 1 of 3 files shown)


fintelia:
Could you look into why the one CI job is failing?

ldm0:
> Could you look into why the one CI job is failing?

I forget to update png in the `Cargo.toml.public-private-dependencies`.

Change is done.

HeroicKatora (approved):

ldm0:
Ping :D

fintelia:
The PNG crate actually has a new Minimum Supported Rust Version, so this would have to happen as a major release

ldm0:
> The PNG crate actually has a new Minimum Supported Rust Version, so this would have to happen as a major release

So move this PR's target to the `next` branch?

HeroicKatora:
> The PNG crate actually has a new Minimum Supported Rust Version, so this would have to happen as a major release

Probably yes. I neverless want to interject: with regards to this, the people at `wgpu`/`naga` had the exact same issue with `bitflags`. The have published a version requiring `~1.2.0`. I informed them of the consequences and downstream people reported the necessity to downgrade to `1.2.0`. Now they're thinking about upholding the MSRV promise when you manually do that but still declaring the dependency as `>=1.2`.

In any case, the changes at png and here are more than that dependency and the MSRV will have to be bumped regardless. As mentioned in the other comments, `master` already contains the MSRV bump so we could just merge `next` and develop the next major version there.

fintelia:
@HeroicKatora the `master` branch -- like the rest of the 0.23.x series -- should still compile with 1.34.2 provided that something else in the dependency tree pins bitflags appropriately. I opened #1544 to confirm that.

How about we make more one release on the 0.23.x series with the current contents of master and the old MSRV, and then do a version bump which pulls in all the changes from the next branch?

ldm0:
Close this since #1543 was landed.

PR #1537, opened 2021-08-13
Improve usage of `Vec::extend`

paolobarbolini:
Just a few random improvements I noticed could be made.

I license past and future contributions under the dual MIT/Apache-2.0 license,
allowing licensees to chose either at their option.

PR #1542, opened 2021-08-17
Reduce allocations and `memcpy`s

paolobarbolini:
In the same style as #1537 - a few random improvements I noticed could be made.

I haven't run a benchmark on these but:

* `encode_dxt_colors` gets called in a for loop so it's probably worth it to remove the `.to_vec()`
* The `Vec::split_to` optimization removes 2 allocations and 2 `memcpy`s
* [comparing](https://rust.godbolt.org/z/37aj341dn) `crate::copy_memory` with `slice::copy_from_slice` shows that they generate the same assembly

---

### DIFF
--- a/src/codecs/dxt.rs
+++ b/src/codecs/dxt.rs
@@ -534,7 +534,7 @@ fn encode_dxt_colors(source: &[u8], dest: &mut [u8], is_dxt1: bool) {
     let mut colorspace_ = [[0u8; 3]; 16];
     let mut colorspace_len = 0;
     for color in &targets {
-        if !colorspace_.contains(color) {
+        if !colorspace_[..colorspace_len].contains(color) {
             colorspace_[colorspace_len] = *color;
             colorspace_len += 1;
         }
--- a/src/codecs/dxt.rs
+++ b/src/codecs/dxt.rs
@@ -525,16 +525,21 @@ fn encode_dxt_colors(source: &[u8], dest: &mut [u8], is_dxt1: bool) {
         *d = [s[0], s[1], s[2]];
     }
 
-    // and a set of colors to pick from.
-    let mut colorspace = targets.to_vec();
-
     // roundtrip all colors through the r5g6b5 encoding
-    for rgb in &mut colorspace {
+    for rgb in &mut targets {
         *rgb = enc565_decode(enc565_encode(*rgb));
     }
 
     // and deduplicate the set of colors to choose from as the algorithm is O(N^2) in this
-    colorspace.dedup();
+    let mut colorspace_ = [[0u8; 3]; 16];
+    let mut colorspace_len = 0;
+    for color in &targets {
+        if !colorspace_.contains(color) {
+            colorspace_[colorspace_len] = *color;
+            colorspace_len += 1;
+        }
+    }
+    let mut colorspace = &colorspace_[..colorspace_len];
 
     // in case of slight gradients it can happen that there's only one entry left in the color table.
     // as the resulting banding can be quite bad if we would just left the block at the closest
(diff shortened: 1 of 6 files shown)


src/codecs/tga/decoder.rs
```diff
@@ -289,10 +290,13 @@ impl<R: Read + Seek> TgaDecoder<R> {
         let remain_len = self.line_remain_buff.len();
 
         if remain_len >= line_num_bytes {
-            let remain_buf = self.line_remain_buff.clone();
+            // `Vec::split_to` if std had it
+            let bytes = {
+                let bytes_after = self.line_remain_buff.split_off(line_num_bytes);
+                mem::replace(&mut self.line_remain_buff, bytes_after)
+            };
```
HeroicKatora, line 297: I wonder hower `splice(..line_num_bytes, []).collect()` performs?

src/codecs/tga/decoder.rs
```diff
@@ -304,7 +308,8 @@ impl<R: Read + Seek> TgaDecoder<R> {
         pixel_data.extend_from_slice(&line_data[..num_bytes]);
 
         // put the remain data to line_remain_buff
-        self.line_remain_buff = line_data[num_bytes..].to_vec();
+        debug_assert!(self.line_remain_buff.is_empty());
```
HeroicKatora, line 311: Please add a comment that this is due to `Vec::append` above.

src/codecs/dxt.rs
```diff
@@ -525,16 +525,21 @@ fn encode_dxt_colors(source: &[u8], dest: &mut [u8], is_dxt1: bool) {
...
+    for rgb in &mut targets {
         *rgb = enc565_decode(enc565_encode(*rgb));
     }
 
     // and deduplicate the set of colors to choose from as the algorithm is O(N^2) in this
-    colorspace.dedup();
+    let mut colorspace_ = [[0u8; 3]; 16];
+    let mut colorspace_len = 0;
+    for color in &targets {
+        if !colorspace_.contains(color) {
```
HeroicKatora, line 537: `dedup` is not the same as keeping only unique colors and this will fail to include `[0, 0, 0]` if it ever occurs. I don't know enough about `tga` to know for certain if this behavioral difference affects the actual encoding correctness.

src/codecs/tga/decoder.rs
paolobarbolini: I neven used `Vec::splice` and seeing how [bad the generated code is](https://rust.godbolt.org/z/YqfM973jE) I'll probably never use it :laughing:.
`Vec::split_off` is much much simpler, it just does [an allocation and a memcpy](https://doc.rust-lang.org/1.54.0/src/alloc/vec/mod.rs.html#1868-1901).

src/codecs/webp/vp8.rs
```diff
@@ -1903,11 +1903,8 @@ fn predict_4x4(ws: &mut [u8], stride: usize, modes: &[IntraMode], resdata: &[i32
                 IntraMode::HU => predict_bhupred(ws, x0, y0, stride),
             }
 
-            // Create a [i32; 16] array for add_residue by copying the
-            // slice from resdata into rb (slices do not work).
-            let mut rb = [0i32; 16];
-            rb.copy_from_slice(&resdata[i * 16..i * 16 + 16]);
-            add_residue(ws, &rb, y0, x0, stride);
+            let rb: &[i32; 16] = resdata[i * 16..][..16].try_into().unwrap();
+            add_residue(ws, rb, y0, x0, stride);
```
fintelia, line 1907: Do we know whether this is faster? The entire array fits in a single cache line, so copying it is extremely fast and this may not be a net win

paolobarbolini: Comparing the previous and newer version when using `opt-level=3` generates the same number of instructions
1.54: https://rust.godbolt.org/z/hGf7GYn4r
1.34: https://rust.godbolt.org/z/Y669fv3h4

While compiling with a lower `opt-level` generates less instructions for the `.try_into()` version
1.54: https://rust.godbolt.org/z/fda9r8rKa
1.34: https://rust.godbolt.org/z/j6joTovj5

so to me this seems worth it.

HeroicKatora (approved):

PR #1544, opened 2021-08-22
Verify that master still works on 1.34.2

fintelia:
Not intended for merging... just to trigger a CI run


fintelia:
Starting to think this might not be worth the effort...

fintelia:
Conclusion: getting the current master branch to compile on 1.34.2 requires patching not just bitflags but also a bunch of other dependencies, disabling a few features, and even then some of the tests don't work

PR #1546, opened 2021-08-24
Perf(tga):  rle reduce allocations.

odriansu:
This pull request attempts to improve decoding perfomance by moving an allocation out of the while loop and replace the for loop by an iterator.

benchmark shows a slightly more than wice as fast decoding speed.

### DIFF
--- a/src/codecs/tga/decoder.rs
+++ b/src/codecs/tga/decoder.rs
@@ -238,6 +238,7 @@ impl<R: Read + Seek> TgaDecoder<R> {
     /// Reads a run length encoded data for given number of bytes
     fn read_encoded_data(&mut self, num_bytes: usize) -> io::Result<Vec<u8>> {
         let mut pixel_data = Vec::with_capacity(num_bytes);
+        let mut pixel = Vec::with_capacity(self.bytes_per_pixel);
 
         while pixel_data.len() < num_bytes {
             let run_packet = self.r.read_u8()?;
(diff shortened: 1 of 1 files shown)


fintelia (approved):

src/codecs/tga/decoder.rs
```diff
@@ -247,15 +248,14 @@ impl<R: Read + Seek> TgaDecoder<R> {
...
+                let repeat_count = ((run_packet & !0x80) + 1) as usize * self.bytes_per_pixel;
                 self.r
                     .by_ref()
                     .take(self.bytes_per_pixel as u64)
-                    .read_to_end(&mut data)?;
-                for _ in 0usize..repeat_count {
-                    pixel_data.extend_from_slice(&data);
-                }
+                    .read_to_end(&mut pixel)?;
+                let data = pixel.iter().cycle().take(repeat_count);
```
HeroicKatora, line 256: This is just slightly obscure, I would appreciate a comment. And maybe move the multiplication with `self.bytes_per_pixel` into the argument here because `repeat_count` is a term that comes from tga itself and this way we preserve its original meaning and value.

src/codecs/tga/decoder.rs
```diff
@@ -238,6 +238,7 @@ impl<R: Read + Seek> TgaDecoder<R> {
     /// Reads a run length encoded data for given number of bytes
     fn read_encoded_data(&mut self, num_bytes: usize) -> io::Result<Vec<u8>> {
         let mut pixel_data = Vec::with_capacity(num_bytes);
+        let mut pixel = Vec::with_capacity(self.bytes_per_pixel);
```
HeroicKatora, line 241: Can you rename this to `repeat_buf` to clarify its use and avoid confusion with `pixel_data`.

HeroicKatora (approved):

PR #1547, opened 2021-08-26
Add adaptive filter option to png encoder

okaneco:
- Add `Adaptive` to `codecs::png::FilterType` enum
- Make `Adaptive` the default `FilterType`
- Use `CompressionType::default` and `FilterType::default` in `codecs::png::PngEncoder::new`

png 0.17 includes the Adaptive filtering method for encoding which helps reduce file size.

I've updated the doc comments in `codecs::png::FilterType` and `codecs::png::PngEncoder::new_with_quality` which mentioned changing the filter type to adaptive filtering when it became available.

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -457,8 +457,7 @@ pub enum CompressionType {
 
 /// Filter algorithms used to process image data to improve compression.
 ///
-/// The default filter is `Sub` though this default may change in the future, most notable if an
-/// adaptive encoding option is implemented.
+/// The default filter is `Adaptive`.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


okaneco:
The `png` feature tests pass on my machine using a local version of `image-png` with this commit https://github.com/image-rs/image-png/commit/64b34e401b57bb9bccd3c68c06f8810b0df29ace to build `image`.

This should be unblocked after the next `image-png` release.

fintelia:
@okaneco png version 0.17.1 is published containing the fix.

okaneco:
Thanks for getting that release out so quickly. I've re-pushed and tests look good now.

fintelia:
Thanks!

PR #1549, opened 2021-09-02
Loop filtering for WebP images

tristanphease:
Adds loop filtering for WebP images, so that decoding for Rgb images is almost perfect.
I believe there's still some discrepancy somewhere in decoding but I think it's not related to loop filtering.
The code is quite messy and definitely could use some improvement.
Additionally all the reference images I am using to test use the simple loop filtering only so I haven't been able to test
 whether the normal filtering works.


HeroicKatora (approved):
High level review the code looks good albeit there's surely room for optimization due to translation of some C-like structure to rust idioms. I can't really say about the format specific implementation details. But the results look good so probably :smile:

PR #1557, opened 2021-09-23
External test data

197g:
This is a draft for a system that would allow _external_ packagers to run our test suite based on the published packages, without requiring us to add that test data into the packages. This is based on a recent idea of mine which I published as [`xtest-data`](https://crates.io/crates/xtest-data). Comment of all sorts welcome on the approach.

I'm hoping to quickly arrive at a stable version—maybe even `1.0`—of the linked crate by prototyping it with `image` or its sub-crates. These should be a decent reflection of real world testing and we certainly have a relatively heavy test suite in terms of file sizes. One particular pain point which I haven't been able to integrate yet would be to move data out-of-tree altogether but I imagine this to be feasible with submodules—indeed, submodules should already work without requiring everyone to download all data—but more interesting would be `git lfs` etc.

## How to run the tests

* Manually package the crate: `cargo package --no-verify`
* Unpack the crate at some location:
  ```bash
  gunzip -c target/package/image-0.23.14.crate | tar -C /tmp --extract --file -
  cd /tmp/image-0.23.14
  ```
* Do the dance for the crate (note that only `reference_images` works atm):
  ```bash
  # Note the error output here, no network connection yet.
  cargo test --tests check_references -- --nocapture
  # Instruct the crate to auto-setup the directories
  CARGO_XTEST_DATA_FETCH=1 cargo test --tests check_references -- --nocapture
  ```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,8 @@ exclude = [
     "tests/*",
 ]
 include = [
-    "tests/reference_images.rs"
+    "tests/reference_images.rs",
+    "benches"
 ]
 
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ num-complex = "0.3"
 glob = "0.3"
 quickcheck = "0.9"
 criterion = "0.3"
-xtest-data = "0.0.2"
+xtest-data = "0.0.3"
 
 [features]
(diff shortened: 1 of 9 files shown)


197g:
This was more of a sketch for developing that crate anyways. Maybe Debian packaging with glycin can re-ignite interest in the underlying issue and improve those, then re-opening it.

PR #1572, opened 2021-10-07
support BMP encode with palette (indexed color) and decode to image and a separate palette

### LINKED ISSUE #1561: Support palette for BMP files
I would like to be able to support palette for bmp files.

My specific use case for this functionality is: I want to generate a palette BMP file, you know, just like a png file with palette.

This is more generally applicable to anyone who wants a palette bmp file ;)

## Draft

Sounds like this lib already supports palette (partially):

https://github.com/image-rs/image/blob/master/src/codecs/bmp/encoder.rs#L192-L195

So the missing piece is quite small: instead of blindly write 256 colors to the palette, we can let the user provide a palette and write to it.

fzyzcjy:
Closes #1561

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -544,7 +544,7 @@ pub struct BmpDecoder<R> {
     reader: R,
 
     bmp_header_type: BMPHeaderType,
-    skip_palette: bool,
+    indexed_color: bool,
 
     width: i32,
     height: i32,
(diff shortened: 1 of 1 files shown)


fzyzcjy:
Done. now decoder also works. Here is my sample code (I need to access both indices themselves and the palette)

```
pub struct IndexedImage {
    pub mat: Mat, // the index
    pub palette: Palette,
}

pub fn imdecode(buf: &[u8]) -> Result<IndexedImage> {
    let mut cursor = Cursor::new(buf);
    let mut decoder = BmpDecoder::new_skip_palette(&mut cursor, true)?;

    let palette = decoder
        .get_palette()
        .ok_or_else(|| anyhow!("no palette exists"))?
        .to_owned();

    let mut mat = Mat::zeros(height as i32, width as i32, CV_8UC1)?.to_mat()?;
    decoder.read_image(mat.data_bytes_mut()?)?;

    Ok(IndexedImage::new(mat, palette))
}

pub fn imencode(im: IndexedImage) -> Result<Vec<u8>> {
    let size = im.mat.size()?;

    let mut buf = Vec::new();
    let mut encoder = BmpEncoder::new(&mut buf);
    encoder.encode_with_palette(
        im.mat.data_bytes()?,
        size.width as u32,
        size.height as u32,
        ColorType::L8,
        Some(&im.palette),
    );

    Ok(buf)
}
```

fzyzcjy:
🎉

Issue #1591, opened 2021-10-29
FromPrimitive rounding modes


fintelia:
Before our next major release, we should make sure that conversion between primitives rounds how we want it to. This has become considerably more noticeable now that f32 component types are being added.

_Originally discussed this thread: https://github.com/image-rs/image/pull/1585#discussion_r735921529_

PR #1593, opened 2021-10-30
Color conversion with rounding

HeroicKatora:
Make sure that color conversion consistently happens with _rounding_ instead of truncation when loosing precision. Updates the reference test suite to keep preverse the images original color type. It previously always stored them as `rgba8` but no longer invents alpha channels and is more or less lossless. This makes the test suite more sensitive to changes in decoded color types.

Closes: #1591


src/traits.rs
```diff
@@ -152,7 +152,7 @@ impl PixelWithColorType for Rgb<u16> { const COLOR_TYPE: ColorType = ColorType::
 impl PixelWithColorType for Rgb<f32> { const COLOR_TYPE: ColorType = ColorType::Rgb32F; }
 
 impl PixelWithColorType for Rgba<u8> { const COLOR_TYPE: ColorType = ColorType::Rgba8; }
-impl PixelWithColorType for Rgba<u16> { const COLOR_TYPE: ColorType = ColorType::Rgba8; }
+impl PixelWithColorType for Rgba<u16> { const COLOR_TYPE: ColorType = ColorType::Rgba16; }
```
HeroicKatora, line 155: Note this fix! It should probably go through, no matter what we decide on with regards to conversion.

fintelia (approved):

Issue #1466, opened 2021-04-22
DynamicImage::grayscale does not maintain alpha channel


Jacherr:
When attempting to convert a RGBA image to grayscale, the alpha channel gets deleted/not respected, which results in any transparent or semi-transparent parts of the image to become completely black in the output DynamicImage.

I'm not sure if this is intended or not, but it would be ideal if the alpha channel was respected when converting an image with an alpha channel to grayscale.

drahnr:
I have unit test that used to work (image `0.23.11` -> `0.23.14` )

This unit test now fails:
https://github.com/spearow/juice/blob/master/greenglas/tests/image_spec.rs#L53-L60

PR #1596, opened 2021-10-31
Issue 1466

ForgottenMaster:
Fix for DynamicImage::grayscale not preserving the alpha channel for input images that have an alpha channel. Fixes issue #1466 

[src/imageops/colorops.rs]
Made new functions grayscale_alpha and grayscale_with_type_alpha which do the same as their non-alpha counterparts except they preserve the alpha channel
Uses LumaA rather than Luma for the pixel type
Uses .to_luma_alpha() rather than .to_luma()

[src/imageops/mod.rs]
Expose the new functions from the imageops module so we can access them in src/dynimage.rs

[src/dynimage.rs]
Updated DynamicImage::grayscale method to use grayscale_alpha/grayscale_with_type_alpha instead of the non-alpha counterparts when the input format has an alpha channel
Added tests to ensure that the alpha channel is discarded/preserved appropriately based on the input format

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -13,18 +13,12 @@ type Subpixel<I> = <<I as GenericImageView>::Pixel as Pixel>::Subpixel;
 
 /// Convert the supplied image to grayscale. Alpha channel is discarded.
 pub fn grayscale<I: GenericImageView>(image: &I) -> ImageBuffer<Luma<Subpixel<I>>, Vec<Subpixel<I>>>
-    where
-        Subpixel<I>: 'static,
-        <Subpixel<I> as Num>::FromStrRadixErr: 'static,
 {
     grayscale_with_type(image)
 }
 
 /// Convert the supplied image to grayscale. Alpha channel is preserved.
 pub fn grayscale_alpha<I: GenericImageView>(image: &I) -> ImageBuffer<LumaA<Subpixel<I>>, Vec<Subpixel<I>>>
-    where
-        Subpixel<I>: 'static,
-        <Subpixel<I> as Num>::FromStrRadixErr: 'static,
 {
     grayscale_with_type_alpha(image)
 }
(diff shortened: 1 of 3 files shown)


src/imageops/colorops.rs
```diff
@@ -42,6 +51,28 @@ where
     out
 }
 
+/// Convert the supplied image to a grayscale image with the specified pixel type. Alpha channel is preserved.
+pub fn grayscale_with_type_alpha<NewPixel, I: GenericImageView>(image: &I) -> ImageBuffer<NewPixel, Vec<NewPixel::Subpixel>>
+where
+    Subpixel<I>: 'static,
+    <Subpixel<I> as Num>::FromStrRadixErr: 'static,
```
HeroicKatora, line 58: Where does this bound come from? I'm a bit confused where this type even comes into play..

ForgottenMaster: Hey, thanks for replying.  I copied the grayscale_with_type method here and changed the to_luma call to a to_luma_alpha

Being as it was copied/modified from the non alpha preserving one it came with the same bounds as grayscale_with_type

I will look into this some more, it might be that we can remove this bound here, and also from the grayscale_with_type function

HeroicKatora: Just want to add, I'm mostly alright with approving the PR in the current state. I'm just very curious. If it turns out to be some convoluted thing it may be better to make an issue for this instead, removing the bound is SemVer compatible.

HeroicKatora: So it seems that this bound was initially introduced in 366b4f57200acc7144914d909484beb7c6985af5, 7 years ago. I imagine this is all related to the rather unfortunate requirements of `Pixel: 'static`. Due to how trait work, this cascades as requirements to all of its associated types and through the `Subpixel: Primitive` and `Primitive: Num` this might be why it was necessary.

Fortunately, the requirement `Pixel: 'static` was recently removed in #1597. It would be interesting to simply remove all these `'static` bounds and see if that compiles.

ForgottenMaster: Yeah, it looks like that is precisely the case and it was required to specify the lifetime bounds as a result of those bounds being present on the trait

Since the removal of the `Pixel: 'static` bound in #1597, I was able to remove the lifetime bounds on these functions also

HeroicKatora (approved):

fintelia (approved):

HeroicKatora:
Thank you!

PR #1598, opened 2021-11-01
Make imageops::{overlay, replace} take coordinates as i64's

### LINKED ISSUE #1575: Overlay and replace with negative position
`overlay()` and `replace()` currently take `u32` for `x` and `y`:

https://github.com/image-rs/image/blob/8ab2d70c6bd725da9b6324316be78c22181cb99a/src/imageops/mod.rs#L261-L262

As such `x` and `y` cannot be negative. This means in effect that the overlaid image can only cross the right and bottom edge of the destination image but not the left and top edge. This is pretty surprising. I think it should either not be allowed to go out of bounds at all or - my much preferred solution - allow the offset to be negative, so I can place the overlaid image wherever I want, even if it sticks out at the left or top.

fintelia:
Fixes #1575

### DIFF
--- a/src/imageops/mod.rs
+++ b/src/imageops/mod.rs
@@ -1,8 +1,6 @@
 //! Image Processing Functions
 use std::cmp;
 
-use num_traits::{NumCast};
-
 use crate::image::{GenericImage, GenericImageView, SubImage};
 use crate::traits::{Lerp, Primitive, Pixel};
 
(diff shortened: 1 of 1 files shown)


src/imageops/mod.rs
```diff
@@ -1,8 +1,6 @@
 //! Image Processing Functions
 use std::cmp;
 
-use num_traits::{NumCast};
```
fintelia, line 4: `NumCast` provides a `i64::from` method, which conflicts with the implementation provided by the `From` trait. I just switched the couple places that refer to NumCast to reference it by its full name.

src/imageops/mod.rs
```diff
@@ -81,72 +79,65 @@ fn crop_dimms<I: GenericImageView>(
...
-    let x_range = top_width.saturating_add(x) // Calculate max coordinate
-        .min(bottom_width) // Restrict to lower width
-        .saturating_sub(x); // Determinate length from start `x`
-    let y_range = top_height.saturating_add(y)
-        .min(bottom_height)
-        .saturating_sub(y);
-    (x_range, y_range)
+    x: i64,
+    y: i64,
+) -> (u32, u32, u32, u32, u32, u32) {
```
HeroicKatora, line 96: That's decently complex. How about having a simple struct for this with public members and methods to construct the x-y-ranges (`core::ops::Range<u32>`) for the bot and top image. I wouldn't mind as much if it weren't part of the public API.

fintelia: How important is it to keep the improved overlay_bounds function as part of the public API?

fintelia: Ok, this PR now leaves `overlay_bounds` unchanged and converts internal uses to a private overlay_bounds_ext function instead

HeroicKatora (approved):

PR #1599, opened 2021-11-01
HDR Decoder: float by default

johannesvollmer:
I suggest we replace the default color type of the hdr format. It was rgb8, but instead we should consider making it f32, to avoid loosing information.

 I also thought about whether it was possible to implement `Pixel` for `Rgbe8`, which looks like the optimal solution, since it saves memory but retains the float precision and infinite range. However, there is no color type associated with that, and we cannot directly return a pixel buffer, so that seems impossible.

Also adds some tests for round-trip checks and (soon) pixel checks ns across image formats.

- [x] decode from file to f32 buffers
- [x] fix tests


src/codecs/hdr/decoder.rs
```diff
@@ -143,14 +144,25 @@ impl<R: BufRead> HdrAdapter<R> {
     }
 
     /// Read the actual data of the image, and store it in Self::data.
-    fn read_image_data(&mut self, buf: &mut [u8]) -> ImageResult<()> {
-        assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
+    fn read_image_data_f32(&mut self, image_bytes: &mut [u8]) -> ImageResult<()> {
+        panic!("dsfowsidjw");
```
johannesvollmer, line 148: running the tests currently does not trigger this panic. some ideas why this is not triggered?

src/codecs/hdr/decoder.rs
```diff
@@ -414,70 +428,56 @@ impl<R: BufRead> HdrDecoder<R> {
...
 
-        let chunks_iter = output_slice.chunks_mut(self.width as usize);
-        let mut pool = Pool::new(8); //
-
-        (pool.scoped(|scope| {
-            for chunk in chunks_iter {
-                let mut buf = vec![Default::default(); self.width as usize];
-                read_scanline(&mut self.r, &mut buf[..])?;
-                let f = &f;
-                scope.execute(move || {
```
johannesvollmer, line 446: this code uses multi threading, but the compression is happening on one thread, while only copying the decompressed pixels is done on multiple threads. this seems like an accident. are there benchmarks that show us whether this multi threading speeds things up?

src/codecs/hdr/decoder.rs
johannesvollmer: the benchmarks trigger this panic

src/codecs/hdr/decoder.rs
johannesvollmer: without the threadpool:
```
load-Hdr/image1.hdr     time:   [356.12 us 361.68 us 367.43 us]
  7 (7.00%) high mild
  3 (3.00%) high severe

load-Hdr/rgbr4x4.hdr    time:   [1.7318 us 1.7542 us 1.7803 us]
  8 (8.00%) high mild
  6 (6.00%) high severe
```

fintelia:
Anyone remember why this stalled? The changes look reasonable, modulo figuring out whether they include any API-breaking changes.

johannesvollmer:
I believe the following happened:
1. I changed the code such that I was pretty confident it might work
2. Add some tests for further confidence
3. Notice that an unconditional panic never is detected by the tests

johannesvollmer:
Note: This only handles decoding, not encoding yet

johannesvollmer:
Anyways, do we consider changing the returned `ColorType` to be a breaking change?

Some users might expect the returned type to be `Rgb8`, which would mean that with the new version, their code no longer works, failing at runtime.

HeroicKatora:
The short answer: not really, but we don't 100% know.

Downstream could rely on matching `DynamicImage` to get results and panic on mismatching code paths. Code would break even though to the type system it isn't a SemVer major change. Note also, `DynamicImage::from_decoder(…).into_rgba8(self)` would continue to work fine.

Some changes to color types had been made previously but most from rgba to rgb and _not_ to the sample type. Potentially, there could be some code that requires opt-in to th changes in `image::io::Reader` (by specifying some 'support-level') but making a promise to keep color types stable is a significant maintenance burden.

I'd say it's okay, although we may seek to clarify this through the `Reader` or decoder at a future point. With jpeg (CMYK) and avif (Yuv) there are certainly potentially going to be even more drastic color type changes happening in the future.

src/codecs/hdr/decoder.rs
```diff
@@ -414,70 +425,56 @@ impl<R: BufRead> HdrDecoder<R> {
...
-                    for (dst, &pix) in chunk.iter_mut().zip(buf.iter()) {
-                        *dst = f(pix);
-                    }
-                });
+        let mut rgbe_pixel_line_tmp_buffer = vec![Default::default(); self.width as usize];
+
+        for y_index in 0 .. self.height as usize {
```
johannesvollmer, line 444: Will this treat the image being upside down?

src/codecs/hdr/decoder.rs
johannesvollmer: has been resolved in a different pr already

src/codecs/hdr/decoder.rs
johannesvollmer: no, it's correct as is, verified by hand (comparing with the reference png)

johannesvollmer:
revisiting this:
- reduced pr scope to exclude the hdr encoder, do that in another pr, this pr is only the decoder
- removed some roundtrip testing because it seemed of little value

now, the reference tests fail. this is because the png format doesn't support float. added the workaround that all hdr images are converted from rgb32f to rgb8 before comparing with png. 

current state: it seems to be on the right way, but some channels are apparently not handled correctly. you can see this when comparing the rendered png reference image with the hdr image, it doesn't match.

johannesvollmer:
grafik

left: the original hdr image, as displayed in affinity photo

right: result of decoding the hdr image using the new decoder, converting that to rgb8, and writing it out as png

maybe the problem is that we must do gamma correction for the output to match?

johannesvollmer:
added one commit that implements the gamma correction. but, as i said, i think there is a better way: output linear values and update the tested reference images.

fintelia:
Could you take a look at https://github.com/image-rs/image/pull/1937?

johannesvollmer:
Yes indeed. That pr achieves the same with less code changes, but

- the tests shouldn't work? I can run the tests for that pr to check it
- this pr does more: it removes some of the buffer copying


oh, the debug statements should be removed from this pr

johannesvollmer:
the tests need to be adjusted either pr, because changing non-linear u8 to linear f32 should result in failing tests. 

which is another breaking change. we should probably provide functionality to obtain the old behavior without much of a hassle

johannesvollmer:
note: i checked out that PR and indeed the tests fail, as they should.

fintelia:
Superseded by #2150

Issue #1600, opened 2021-11-04
Resizing broken in 0.24


Gelbpunkt:
Hi there, seems like the recent changes to 0.24 broke `resize`, at least for my image files. Only `FilterType::Nearest` works, all others produce a vec of 0s. The input image is a `DynamicImage::ImageRgba8`.

## Expected

Resizing produces a non-zero vec

## Actual behaviour

I get a vec of 0s.

## Reproduction steps

Image: <https://git.travitia.xyz/Adrian/okapi-rewrite/-/raw/current/assets/images/casts/ranger.png>

Code:
```rs
let buf: &[u8] = include_bytes!("ranger.png");
let image = load_from_memory(buf).unwrap().to_rgba8();
let result = resize(
    &image,
    22,
    22,
    FilterType::Lanczos3,
);
// result only contains 0s unless using Nearest
```

PR #1602, opened 2021-11-05
Fix issue #1600

fintelia:
Seems to be caused by a bad merge between https://github.com/image-rs/image/commit/d46578d6d171309143f820c9e9c8cabc6fce0ba4 and https://github.com/image-rs/image/commit/4eed468f852c80eca718abe679877be377488fb9

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -292,14 +292,12 @@ where
                 t.3 += vec.3 * w;
             }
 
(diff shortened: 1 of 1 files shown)


HeroicKatora (approved):
Bad merge indeed, I vaguely remember doing it. Thanks for the regression test, it's slightly uneasing that this wasn't caught before.

PR #1604, opened 2021-11-05
Use a 4-channel f32 pixel intermediate type for resizing.

### LINKED ISSUE #1603: Unnecessary loss of precision when resizing
In imageops::resize resizing is done in two passes, vertical and horizontal. This isn't a problem for scaling mathematically, but right now the temporary ImageBuffer uses the same pixel format as the input/output pixels, causing unnecessary rounding and clamping and making the results less accurate. This is actually a significant portion of the complexity of scaling right now.

## Expected

Precision for internal calculations should be maintained and rounding/clamping should only occur when necessary.

awused:
This maintains precision and avoids a lot of unnecessary rounding to integers for intermediate values. This improves precision and performance.

On a 2950x:
Before: test imageops::sample::tests::bench_resize                     ... bench:   3,179,804 ns/iter (+/- 129,332) = 641 MB/s
After: test imageops::sample::tests::bench_resize                     ... bench:   1,563,759 ns/iter (+/- 87,143) = 1304 MB/s

Since there's no real generic [f32; 4] pixel type I just used Rgba32FImage as the intermediate type. The image is never converted to Rgba and the channel order is passed through, which is a bit weird but should be sound.

Fixes #1603

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -7,7 +7,7 @@ use std::f32;
 
 use num_traits::{NumCast, ToPrimitive, Zero};
 
-use crate::ImageBuffer;
+use crate::{ImageBuffer, Rgba32FImage};
 use crate::image::GenericImageView;
 use crate::utils::clamp;
 use crate::traits::{Enlargeable, Pixel, Primitive};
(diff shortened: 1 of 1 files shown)


HeroicKatora (commented):
There was a correctness issue (#1600) that conflicted with this one. Could you please rebase, the approach is still the same. Also, I fully agree on weird but sound. Using intermediate f32-images makes this implementation looks quite like an actual shader pipeline :smile:

awused:
It's been rebased and the benchmarks have been redone.

HeroicKatora (approved):
Thanks!

PR #1606, opened 2021-11-05
Deprecate PngEncoder::encode and IcoEncoder::encode

fintelia:
It seems to be the only function that expects input in big endian, so getting
rid of it will make the rest of the library more consistent.

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -506,6 +506,9 @@ impl<W: Write> PngEncoder<W> {
     }
 
     /// Encodes the image `data` that has dimensions `width` and `height` and `ColorType` `c`.
+    ///
+    /// Expects data in big endian.
+    #[deprecated = "Use `PngEncoder::write_image` instead. Beware that `write_image` has a different endianness convention"]
     pub fn encode(self, data: &[u8], width: u32, height: u32, color: ColorType) -> ImageResult<()> {
         let (ct, bits) = match color {
             ColorType::L8 => (png::ColorType::Grayscale, png::BitDepth::Eight),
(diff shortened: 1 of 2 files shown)


src/codecs/png.rs
```diff
@@ -575,6 +578,11 @@ impl<W: Write> PngEncoder<W> {
 }
 
 impl<W: Write> ImageEncoder for PngEncoder<W> {
+    /// Write a PNG image with the specified width, height, and color type.
+    ///
+    /// For color types with 16-bit per channel or larger, the contents of `buf` should be in
+    /// native endian. PngEncoder will automatically convert to big endian as required by the
+    /// underlying PNG format.
```
fintelia, line 585: Adding this doc-comment makes rustdoc expand ImageEncoder::write_image by default.

src/codecs/png.rs
```diff
@@ -532,6 +532,9 @@ impl<W: Write> PngEncoder<W> {
     }
 
     /// Encodes the image `data` that has dimensions `width` and `height` and `ColorType` `c`.
+    ///
+    /// Expects data in big endian.
+    #[deprecated = "Use `PngEncoder::write_image` instead. Beware that `write_image` has a different endianness convention"]
     pub fn encode(self, data: &[u8], width: u32, height: u32, color: ColorType) -> ImageResult<()> {
```
HeroicKatora, line 538: Maybe add `#[doc(hidden)]`?

fintelia: Hm, if we do that then users won't be able to see that the function expected big endian. (Older releases didn't mention it)

Your call

HeroicKatora: I guess we can have it deprecated for a couple of versions before hiding it :+1:

HeroicKatora (approved):

Issue #1046, opened 2019-09-29
`ImageDecoderExt` is named incorrectly


HeroicKatora:
Lifting this from [this review](https://github.com/image-rs/image/pull/1044#discussion_r329183955) which discussed some changes to the `ImageDecoder`. One thing that came up was that *efficient* decoding of parts of an image is not particularly ergonomic to use nor to implement. The name of the trait `ImageDecoderExt` also obscures the fact that it is used for this purpose.

> > The typical *Ext naming is for traits that purely extend an interface and are implementation for all T: BaseTrait, so this is also a confusing naming.
>
> This is a very good point. Any ideas for alternate names? Perhaps ImageDecoderSeekable or something?

`ImageSeek` may be a bit terse but not inaccurate.

PR #1611, opened 2021-11-13
Rename ImageDecoderExt

HeroicKatora:
Closes: #1046 

Since there is no way to introduce aliases for traits we must do this in a major change, and can't have a backwards compatible, hidden way for migration.


fintelia:
If we were willing to make all `ImageDecoder`'s require a `Seek` bound from their inner readers, then we could merge the two traits. Otherwise :+1: on renaming ImageDecodeExt -> ImageDecoderRect in the 0.24 release

HeroicKatora:
Even with the `Seek` bound it would still be additional functionality to implement. I suppose one could make a somewhat complex system to make it optional, however this might be convoluted.

Issue #1616, opened 2021-11-20
Attempt to multiply with overflow in PNM decoder


saethlin:
Running the PNM fuzzing target, this input produces a crash due to an attempt to multiply with overflow:
```
[80, 54, 10, 52, 50, 57, 52, 56, 50, 57, 52, 56, 35, 56, 10, 52, 10, 48, 10, 12, 12, 56]
```
It looks like the offending multiply is this: https://github.com/image-rs/image/blob/0f1c38a97ec809f399a2a5edfb85c7e17fcd46ec/src/codecs/pnm/decoder.rs#L608

I suspect this should be a `checked_mul`

fintelia:
Yeah, I think the right fix is switching the type for the second argument from u32 -> usize while also replacing the multiplication with a `checked_mul`.

PR #1619, opened 2021-11-21
Added overflow handling to multiplication of width and components

shikharvashistha:
Greetings of the day @fintelia,

I've fixed the issue as suggested, kindly review the pull and suggest changes if any.

Thanks & Regards
@shikharvashistha
Fixes : #1616

### DIFF
--- a/src/codecs/pnm/decoder.rs
+++ b/src/codecs/pnm/decoder.rs
@@ -1,4 +1,5 @@
 use std::convert::TryFrom;
+use std::convert::TryInto;
 use std::error;
 use std::fmt::{self, Display};
 use std::io::{self, BufRead, Cursor, Read};
@@ -228,7 +229,7 @@ enum TupleType {
 
 trait Sample {
     fn bytelen(width: u32, height: u32, samples: u32) -> ImageResult<usize>;
-    fn from_bytes(bytes: &[u8], row_size: u32, output_buf: &mut [u8]) -> ImageResult<()>;
+    fn from_bytes(bytes: &[u8], row_size: usize, output_buf: &mut [u8]) -> ImageResult<()>;
     fn from_ascii(reader: &mut dyn Read, output_buf: &mut [u8]) -> ImageResult<()>;
 }
 
@@ -629,14 +630,13 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for PnmDecoder<R> {
 }
 
 impl<R: Read> PnmDecoder<R> {
-    fn read_samples<S: Sample>(&mut self, components: usize, buf: &mut [u8]) -> ImageResult<()> {
+    fn read_samples<S: Sample>(&mut self, components: u32, buf: &mut [u8]) -> ImageResult<()> {
         match self.subtype().sample_encoding() {
             SampleEncoding::Binary => {
-                let width = self.header.width();
+                let width = TryInto::<usize>::try_into(self.header.width()).unwrap();
                 let height = self.header.height();
-                let bytecount = S::bytelen(width, height, components)?;
+                let bytecount = S::bytelen(width.try_into().unwrap(), height, components)?;
                 let mut bytes = vec![];
-
                 self.reader
                     .by_ref()
                     // This conversion is potentially lossy but unlikely and in that case we error
@@ -647,8 +647,11 @@ impl<R: Read> PnmDecoder<R> {
                 if bytes.len() != bytecount {
                     return Err(DecoderError::InputTooShort.into());
                 }
-
-                S::from_bytes(&bytes, usize::checked_mul(width, components), buf)
+                let row_size = match width.checked_mul(components.try_into().unwrap()) {
+                    Some(n) => n,
+                    None => return Err(ImageError::IoError(io::ErrorKind::UnexpectedEof.into())),
+                };
+                S::from_bytes(&bytes, row_size, buf)          
             }
             SampleEncoding::Ascii => self.read_ascii::<S>(buf),
         }
@@ -694,7 +697,11 @@ impl Sample for U8 {
         Ok((width * height * samples) as usize)
     }
 
-    fn from_bytes(bytes: &[u8], _row_size: u32, output_buf: &mut [u8]) -> ImageResult<()> {
+    fn from_bytes(
+        bytes: &[u8],
+        _row_size: usize,
+        output_buf: &mut [u8],
+    ) -> ImageResult<()> {
         output_buf.copy_from_slice(bytes);
         Ok(())
     }
@@ -712,7 +719,11 @@ impl Sample for U16 {
         Ok((width * height * samples * 2) as usize)
     }
 
-    fn from_bytes(bytes: &[u8], _row_size: u32, output_buf: &mut [u8]) -> ImageResult<()> {
+    fn from_bytes(
+        bytes: &[u8],
+        _row_size: usize,
+        output_buf: &mut [u8],
+    ) -> ImageResult<()> {
         output_buf.copy_from_slice(bytes);
         for chunk in output_buf.chunks_exact_mut(2) {
             let v = BigEndian::read_u16(chunk);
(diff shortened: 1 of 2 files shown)


5225225:
`checked_mul` returns a `Option<usize>`, which means it can fail.

There are multiple ways to solve this, but what I would suggest doing is something like

```rust
let row_size = match width.checked_mul(components) {
    Some(n) => n,
    None => return Err(<some error>),
};
```

Also, it's not entirely clear that using `usize` here is correct. It would be *possible* to change it to be `usize` everywhere (maybe, this doesn't *look* like it would affect the public API), but we use `u32` everywhere else.

I'd keep it as a `u32`, and then you just need to deal with the error.

For the error... there doesn't look to be a good error to return already. So you'll want to make a new one.

To do that, you'll need to add a new enum variant to the `DecoderError` (it's at the top of `pnm/decoder.rs`). I'd do something like

```rust
/// The row was too wide and overflowed in multiplication
RowTooWide {
    width: u32,
    components: u32,
},
```

And then you'll need to add a human readable error for this. I'd say what the width and the components were, like

```rust
DecoderError::RowTooWide { width, components } =>
    f.write_fmt(format_args!("Row too wide: have width={}, components={}", width, components)),
```

And finally, I'd add a test. You can add the test at the bottom of `pnm/decoder.rs`, I'd look at how `dimension_overflow` works. Try and feed it the error file, and assert that it fails with your RowTooWide error.

---

Sorry about the wall of text! If anything's unclear, feel free to ask.

Also pinging @fintelia since i'm not sure if my suggestion to keep it as a `u32` is right. But we *seem* to use `u32` reasonably consistently for image dimensions elsewhere.

fintelia:
We use `u32` to store image dimensions quite frequently, but what concerns me here is that argument actually specifies the width of the image times the number of channels. I suggested changing to a `usize` in https://github.com/image-rs/image/issues/1616#issuecomment-974673631 because we know that number will be at most equal to the size of `buf`, which by definition must fit into a `usize`.

fintelia:
@shikharvashistha thanks for this PR! It isn't quite ready to merge yet, but we can work through some changes to get everything working.

First off, could you make sure you able to run `cargo check` and `cargo test` locally? Based on the Continious Integration tests, it seems that your changes don't compile, which is easiest to resolve by running the compiler locally and adjusting your code accordingly.

Next, I should perhaps provide more detailed directions. When I suggested changing the argument type, I mean changing line 201 so that from_bytes's second argument was `row_size: usize` on line 201: 
https://github.com/image-rs/image/blob/0f1c38a97ec809f399a2a5edfb85c7e17fcd46ec/src/codecs/pnm/decoder.rs#L201
Once you make that change, you'll see a couple compiler errors generated. These indicate places where other code has to be updated to get the types to match. Most of these should be straightforward to fix, but please ask questions if you get stuck.

shikharvashistha:
cargo check && cargo test seems to be working for me now

src/codecs/pnm/decoder.rs
```diff
@@ -589,11 +633,10 @@ impl<R: Read> PnmDecoder<R> {
     fn read_samples<S: Sample>(&mut self, components: u32, buf: &mut [u8]) -> ImageResult<()> {
         match self.subtype().sample_encoding() {
             SampleEncoding::Binary => {
-                let width = self.header.width();
+                let width = TryInto::<usize>::try_into(self.header.width()).unwrap();
                 let height = self.header.height();
-                let bytecount = S::bytelen(width, height, components)?;
+                let bytecount = S::bytelen(width.try_into().unwrap(), height, components)?;
```
HeroicKatora, line 638: Are these not undoing each other? In that case I would delay the conversion to `usize` and reassign it later.

src/codecs/pnm/decoder.rs
```diff
@@ -697,8 +755,8 @@ impl Sample for PbmBit {
         Ok((linelen * height) as usize)
     }
 
-    fn from_bytes(bytes: &[u8], row_size: u32, output_buf: &mut [u8]) -> ImageResult<()> {
-        let mut expanded = utils::expand_bits(1, row_size, bytes);
+    fn from_bytes(bytes: &[u8], row_size: usize, output_buf: &mut [u8]) -> ImageResult<()> {
+        let mut expanded = utils::expand_bits(1, row_size.try_into().unwrap(), bytes);
```
HeroicKatora, line 759: This `unwrap` was added due to the type change. It should be explained, in a comment or by using `expect`, why this can not panic—otherwise this is seems to make stability worse than the panic-in-debug from multiplication. An alternative would be to return an error here, the result type is already available for this.

src/codecs/pnm/decoder.rs
```diff
@@ -604,12 +647,17 @@ impl<R: Read> PnmDecoder<R> {
...
                 }
-
-                S::from_bytes(&bytes, width * components, buf)
-            }
-            SampleEncoding::Ascii => {
-                self.read_ascii::<S>(buf)
+                let row_size = match width.checked_mul(components.try_into().unwrap()) {
+                    Some(n) => n,
+                    None => return Err(ImageError::IoError(io::ErrorKind::UnexpectedEof.into())),
+                };
```
HeroicKatora, line 653: Combining `unwrap` and `checked_mul` seems inconsistent. Note that not all platforms _guarantee_ that `usize` can be converted to `u32` and vice-versa. Rather the failure to conversion could also result in an error return.

src/codecs/pnm/decoder.rs
```diff
@@ -604,12 +647,17 @@ impl<R: Read> PnmDecoder<R> {
...
                     return Err(DecoderError::InputTooShort.into());
                 }
-
-                S::from_bytes(&bytes, width * components, buf)
-            }
-            SampleEncoding::Ascii => {
-                self.read_ascii::<S>(buf)
+                let row_size = match width.checked_mul(components.try_into().unwrap()) {
+                    Some(n) => n,
+                    None => return Err(ImageError::IoError(io::ErrorKind::UnexpectedEof.into())),
```
HeroicKatora, line 652: This error kind doesn't really fit the failure. It's not caused by the environment, so it should be a `DecodingError` of some sort. Feel free to add one to variant to `DecoderError` in this file if none of them apply.

src/codecs/pnm/decoder.rs
```diff
@@ -1179,7 +1253,7 @@ WIDTH 4294967295
 HEIGHT 4294967295
 ENDHDR
 \xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef";
-        
+
```
HeroicKatora, line 1256: Add the test case from the original issue here, please.

HeroicKatora (commented):
It seem like there is _more_ panic after the changes, please review the code and see if the comments can help fix that.

src/codecs/pnm/decoder.rs
```diff
@@ -1262,3 +1266,17 @@ ENDHDR
...
+fn read_samples_test() {
+    let data = vec![80, 54, 10, 52, 50, 57, 52, 56, 50, 57, 52, 56, 35, 56, 10, 52, 10, 48, 10, 12, 12, 56];
+    let mut decoder = PnmDecoder::new(&data[..]).unwrap();
+    assert_eq!(decoder.color_type(), ColorType::L8);
+    assert_eq!(decoder.dimensions(), (16, 16));
+    assert_eq!(decoder.subtype(), PnmSubtype::Graymap(SampleEncoding::Binary));
+    let mut image = vec![0; decoder.total_bytes() as usize];
+    decoder.read_image(&mut image).unwrap();
+    assert_eq!(image, vec![80, 54, 10, 52, 50, 57, 52, 56, 50, 57, 52, 56, 35, 56, 10, 52, 10, 48, 10, 12, 12, 56]);
+
```
HeroicKatora, line 1280: The asserts do not hold for that image. We're only interested in not crashing and having a valid eader. They could just be removed (please verify on your local machine that this works):

```suggestion
fn read_samples_test() {
    let data = vec![80, 54, 10, 52, 50, 57, 52, 56, 50, 57, 52, 56, 35, 56, 10, 52, 10, 48, 10, 12, 12, 56];
    let mut decoder = PnmDecoder::new(&data[..]).unwrap();
    let mut image = vec![0; decoder.total_bytes() as usize];
    let _ = decoder.read_image(&mut image);
}
```

shikharvashistha:
Apologies, it's difficult for me to complete it at this time due to other time bound commitments, other's who are interested can take up the challenge.

HeroicKatora:
Thanks for the notification, if you keep the branch and repository up then we (meaning anyone with write-access to this repository) can push further work into it. No problems at all, thanks for everything done already :)

Issue #1406, opened 2021-01-23
Make `GenericImageView` & `GenericImage` traits object safe


garlic0708:
I'm trying to implement a function that receives an `ImageBuffer`, and may either return the `ImageBuffer` itself or a `SubImage` which is a view into the `ImageBuffer` parameter, so I try to define the return type of the function as `Box<dyn GenericImageView>`, as both `ImageBuffer` and `SubImage` are implementors of the trait.

However, this gives a compile error. It turns out that the `pixels` method makes the trait object unsafe, thus preventing the making a dynamic object of the trait:
```
error[E0038]: the trait `GenericImageView` cannot be made into an object
   --> src/get_dominant_color/kmeans.rs:6:75
    |
6   | ...) -> Box<dyn GenericImageView<Pixel=I::Pixel, InnerImageView=I::InnerImageView>> {
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `GenericImageView` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> /Users/garlic0708/.cargo/registry/src/mirrors.ustc.edu.cn-61ef6e0cd06fb9b8/image-0.23.12/./src/image.rs:713:8
    |
713 |     fn pixels(&self) -> Pixels<Self> {
    |        ^^^^^^ the trait cannot be made into an object because method `pixels` references the `Self` type in its return type
```

I was not opting for `ImageBuffer` as the return type, because I noticed that the `to_image` method on `SubImage` performs a copy of the cropped image, which is potentially time consuming, but it seems that it is the only option as of now.

## Draft

Can we define a separate trait for the `pixels` method, leaving the `GenericImageView` & `GenericImage` traits object safe, thus enabling the use of a dynamic object of the object safe part of the image access & manipulation methods?

HeroicKatora:
[...]

Issue #1118, opened 2020-01-24
RFC: Rethink Pixel Types


fintelia:
Right now, most of the image algorithms are implemented for the _GenericImage_ trait. I see a couple problems with this design:

1. Without being able to assume contigious memory layout it is hard to get good cache performance
2. It makes doing dynamic dispatch for something like DynamicImage rather awkward, requiring a match statement to statically dispatch for each possible color type. 
3. Writing generic code frequently requires adding methods to the Pixel trait (like [invert](https://docs.rs/image/0.23.0-preview.0/image/trait.Pixel.html#tymethod.invert) or [to_rgba](https://docs.rs/image/0.23.0-preview.0/image/trait.Pixel.html#tymethod.to_rgba)). This isn't viable for downstream users who can't add methods themselves.
4. We rely heavily on inlining/optimization to counter the overhead of doing a function call per pixel in many algorithms.

I propose we instead implement the various algorithms directly within a single trait that both statically and dynamically typed images can implement. We'll lose the ability for downstream users to add new pixel types (though that capability was dubious in practice since we really don't let users actually describe the meaning of pixel types), but should end up with a design that is both cleaner and hopefully more correct.

```rust
/// Encodes valid pairs of color type + space. `ColorSpace` can be 
/// crate-private to start, so we don't have to decide how it works 
/// right away.
pub struct PixelType(ColorType, ColorSpace);
impl PixelType {
    pub fn linear(ty: ColorType) -> ImageResult<Self> { ... }
    pub fn color_type(&self) -> ColorType { self.0 }
}

pub trait ContiguousImage {
    fn dimensions(&self) -> (u32; u32);
    fn pixel_type(&self) -> PixelType;
    fn data(&self) -> &[u64]; // u64 lets us transmute to other primitive types
    fn data_mut(&mut self) -> &mut [u64];

    // Provided methods...
    fn as_bytes(&self) -> &[u8] { bytemuck::bytes_of(self.data()) }
    fn as_bytes_mut(&mut self) -> &mut [u8] { bytemuck::bytes_of_mut(self.data_mut()) }
    fn convert_to(&self, new: PixelType) -> DynamicImage { ... }
    fn resize(&self, ...) -> DynamicImage { ... }
    fn blur(&mut self, radius: f32) { ... }
    fn rotate(&mut self, angle: f32) { ... }
    fn horizontal_convolution(&mut self, filter: &[f32]) { ... }
[...]

PR #1620, opened 2021-11-21
Dyn generic image

HeroicKatora:
Splits `view` and `sub_image` from the main image traits to a different set of traits, removing the need for `InnerImage` associated types. Also bounds the relevant method that prevented the current implementation from being used as a `dyn` trait object.

Related: #1118, #1406

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,6 @@ While some of the methods for `GenericImage` are...
 An image parameterised by its Pixel types, represented by a width and height and a vector of pixels. It provides direct access to its pixels and implements the `GenericImageView` and `GenericImage` traits.
 
 ```rust
-extern crate image;
-
 use image::{GenericImage, GenericImageView, ImageBuffer, RgbImage};
 
 // Construct a new RGB ImageBuffer with the specified width and height.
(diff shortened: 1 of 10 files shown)


fintelia:
Really like removing some of the complexity from GenericImage / GenericImageView!

How important is the transparent handling of stacked sub-images? Are there any algorithms that rely on that specifically? An alternative strategy would be to remove all the related traits and just have some methods on SubImage:
```rust
impl<I: GenericImageView> SubImage<I> {
    fn new(img: I, ...) -> SubImage<I> { ... }
    fn view(&self, ...) -> SubImage<I> { ... }
}
impl<I: GenericImage> SubImage<I> {
    fn new_mut(img: I, ...) -> SubImage<I> { ... }
    fn view_mut(&mut self, ...) -> SubImage<I> { ... }
}
```

HeroicKatora:
Regardin this concern and the general desire to release `0.24` somewhat soon, I would remove the traits and impls for now in one separate commit and replace them with inherent methods. If there is a negative reaction and strong desire to bring them back it's easier and SemVer compatible to bring them back now :)

fintelia (approved):

PR #1624, opened 2021-11-24
Lossless WebP image decoding

tristanphease:
Does lossless image decoding for WebP images.
It worked on all the lossless images I tried, so as best as I can tell it works.
Definitely scope for optimisations and there's probably too much type conversion going on but the lossless does a lot of indexing with random numbers.
Feedback would be appreciated.


src/codecs/webp/decoder.rs
```diff
@@ -48,65 +47,80 @@ impl From<DecoderError> for ImageError {
...
-
-        let mut decoder = WebPDecoder {
-            r,
-            have_frame: false,
-            frame: f,
+    pub fn new(mut r: R) -> ImageResult<WebPDecoder<R>> {
+        
+        let _size = WebPDecoder::read_file_header(&mut r)?;
+
+        let inner_decoder = WebPDecoder::read_frame(r)?;
```
HeroicKatora, line 67: This is probably the most contentious line, the rest of the PR seems fine. However, moving the point of error for a bad image, as well as the costs associated with reading it, into the constructor is quite a different. In particular we no longer have a fast path into reading merely the header and metadata.

src/codecs/webp/lossless.rs
```diff
@@ -0,0 +1,725 @@
...
+const ALPHABET_SIZE: [u16; HUFFMAN_CODES_PER_META_CODE] = [
+    256 + 24,
+    256, 256, 256,
+    40
+];
+
+#[inline]
+pub(crate) fn div_round_up(num: u16, den: u16) -> u16 {
+    (num + den - 1) / den 
+}
```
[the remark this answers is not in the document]
tristanphease: Yeah I just used the div_round_up from the specification, but in libwebp they have
```
static WEBP_INLINE uint32_t VP8LSubSampleSize(uint32_t size,
                                              uint32_t sampling_bits) {
  return (size + (1 << sampling_bits) - 1) >> sampling_bits;
}
```
Probably inlining it makes sense, not sure about overflow.

src/codecs/webp/decoder.rs
tristanphease: I did this at the start while hacking it together to work initially.
I did it this way because the original code had new() call decoder.read_metadata() which called self.read_frame() so it didn't seem to make much difference.
I could try moving it around.

Do we need this behaviour? :
```
let mut framedata = Vec::new();
self.r.by_ref().take(len as u64).read_to_end(&mut framedata)?;
let m = io::Cursor::new(framedata);
```

src/codecs/webp/lossless.rs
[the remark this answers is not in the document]
HeroicKatora: Then I suppose this is not a problem. Inlining should make most differences. go away.

src/codecs/webp/decoder.rs
HeroicKatora: I did not realize that the old version was following that pattern as well. The place is presumably fine then. The usage of `read_to_end` ensures that the full chunks has been consumed from the underlying reader in all cases, and `io::Cursor` just enables reading from that chunk.

It would be possible to restore that behavior by moving the read of `_len` from `LosslessDecoder::new` to `read_frame` and adding similar code there.

tristanphease: The issue with that is then we have to change the inner decoder signature to ```InnerDecoder<Cursor<Vec<u8>>>``` which seems a little cumbersome.  
Not sure what the most idiomatic thing to do is, maybe instead of storing the decoder in the ```WebPDecoder```, we could copy their frames to the ```WebPDecoder``` and use them like the old decoder.rs does?

tristanphease: Oh the frame copy is done so the vp8decoder is dropped after the frame is read; I guess that's behaviour we want to preserve?

tristanphease: Ok, I made changes so the decoder acts like the old version, just with the enum for different frame types.

HeroicKatora: Perfect, this makes it easier to keep track of all the changes introduced by the PR. Looks good to me then, if you want to change this then feel free to provide motivation and a separate PR :+1:

HeroicKatora (approved):

PR #1627, opened 2021-11-27
Bump gif dependency to 0.11.3

simon-frankau:

"image" currently uses gif 0.11.1, which is a little over a year old. I suggest bumping onto the latest version (that just happens to include a bug fix I care about. ;).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ byteorder = "1.3.2"
 num-iter = "0.1.32"
 num-rational = { version = "0.4", default-features = false }
 num-traits = "0.2.0"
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Is the current library incompatible with `gif = 0.11.1`? If not, then this doesn't really do anything. `cargo` will choose the newest compatible version according to the dependency specification when creating a lockfile, in particular it already uses `0.11.3` which is compatible with `0.11.1`.

Locally anyone can reperform version selection with `cargo update`.

simon-frankau:
Ah, thanks for taking the time to explain. I'll go away and read the Cargo docs!

Issue #1433, opened 2021-02-13
What to do in 0.24


HeroicKatora:
We have collected some amount of cruft that could be fixed in a new major version. There is no single issue that is a pressing concern, some have workaround, but I'd want to arrive at somewhat of a consensus or plan nevertheless as in total it is awkward. In particular the Rust compiler has moved forward quite a bit. I'd make a `next` branch like last time, which I believe worked out quite well.

The main issues that could be addressed:

#### Pure renames and restructing
- [x] remove the deprecated top-level decoders, fixing the high-level structure
- [x] remove scream case names, including the ones we missed last time
- [ ] minor renames of functions in `imageproc` that are now more flexible (#1201)
- [x] Remove empty module `math` (#1434)

#### Design issues
- [x] floating point images in `DynamicImage`, or a rework that would allow addressing this in a future minor change
  - [x] figure out consistent rounding (#1591)
-  potentially some improvements with const-generics
- [ ] prepare for potential no_std support, with `extern crate alloc;`
- [ ] `async` (#1397) but we'd need a sketch first
- [ ] changes to the `GenericImage` trait that makes it object safe (#784, #1406, #1118)

#### Implementation is breaking
- [x] constify some things such as pre-computed tables (#1326)
- [x] fix color accuracy in luma conversion (#1651) (#1215)
- ~~simplify the pixel trait (#1099,  #956)~~
- [ ] fix the API concerns on `GenericImage` (#1340, 
- [x] stream decoders: (#1046, #1219) (#1611)
- [x] troubles with usage of the `Seek` trait on readers: (#1183)
- [x] fix some clarity issues that require newer compiler (#1333)
- [x] review `ImageOutputFormat::Unknown` variant (https://github.com/image-rs/image/pull/1453#discussion_r593829446)
- [x] remove `'static` bound on `Pixel` (#1597 )
- [ ] optional `get_pixel` methods (and more clarity on assert vs. Result)
- [x] [This comment in `buffer`](https://github.com/image-rs/image/blob/d2f9fe8026db226f3fa6ff4e2473aca5977e6d26/src/buffer.rs#L1152-L1154) about `slice::copy_within` usage
- [x] Think about `replace` and `overlay` signature (#1575)
- [x] Inconsistent signature in `*Encoder::new` (#1605)
- [x] `DynamicImage` as `non_exhaustive` would be more future-proof

## Unanswered questions

* When to do the major version, how to communicate this.
[...]

PR #1632, opened 2021-12-02
Make DynamicImage non-exhaustive

HeroicKatora:
As suggested by [@5225225](https://github.com/image-rs/image/issues/1433#issuecomment-963592987). Downstream code exhaustive matching on the current variants should use or petition for methods that extract into one types instead. Of particular note and example I'd like to point out [`as_bytes`] which essentially promises that all future variants will only contain matrices of trivial and padding-free channel types.

Overall, it's expected that this allows `0.24` to be more stable while we may discuss how to introduce half-float, or non-rgb combinations.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -29,9 +29,31 @@ use crate::traits::Pixel;
 use crate::{Rgb32FImage, Rgba32FImage};
 
 /// A Dynamic Image
+///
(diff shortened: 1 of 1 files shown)


fintelia:
> Overall, it's expected that this allows 0.24 to be more stable while we may discuss how to introduce half-float, or non-rgb combinations.

And of course whether we _should_ introduce half-float or non-rgb combinations :smile:

fintelia (approved):

PR #1635, opened 2021-12-04
Fix image-rs#1618: bmp::decoder::extend_buffer panic

alichay:
`bmp::decoder::extend_buffer` had a working strategy for cases where the buffer was being doubled or more, but if the buffer wasn't being doubled, the part of the function that copied the old data to the end of the buffer absolutely blew up sometimes.

This implementation is simpler than the old one, and still really fast according to some quick profiles.

I already put the copyright thing in another PR, but I'm just stating it again because that PR hasn't landed :)


HeroicKatora (approved):

src/codecs/bmp/decoder.rs
```diff
@@ -295,17 +295,10 @@ fn extend_buffer(buffer: &mut Vec<u8>, full_size: usize, blank: bool) -> &mut [u
         old.copy_from_slice(&new[..old_size]);
         new
     } else {
```
fintelia, line 297: I think we can get rid of this part now, and just use `copy_within` unconditionally?

alichay: I kept them separate because the >=2len path's `copy_from_slice` uses memcpy rather than memmove like `copy_within`, but it seems most libc impls will optimize non-overlapping memmove to a memcpy anyway so that seems like a decent simplification

fintelia (approved):

Issue #884, opened 2019-03-19
How to encode multiple pictures in a single .ico file? - Ico Encoding


GarkGarcia:
I'm trying to create a function which takes a `&Vec<RgbaImage>` and saves them as a single `.ico` file in a given `path: &String`. I can't figure out how to cram multiple files pictures in a single `ico`... Here's what I've got so far:

```rs
fn ico(bufs: &Vec<RgbaImage>, path: &String) -> Result<String, io::Error> {
    match fs::File::create(path) {
        Ok(file) => {
            let mut output = ico::ICOEncoder::new(file);

            for buf in bufs {
                let (w, h) = buf.dimensions();

                if let Err(err) = output.encode(&buf.clone().into_raw(), w, h, ColorType::RGBA(8)) {
                    return Err(err);
                }
            }

            Ok(path.clone())
        },
        Err(err) => Err(err)
    }
}
```

Is there a way to do it?

fintelia:
I don't think that is currently possible. We'd be open to a pull request to add the feature though

paolobarbolini:
I ran into this same issue today and I ended up copy and pasting the encoder into my project and making it work for multiple images. I'll try to send a PR later, though I also think it'd be cool if the encoder took an externally encoded png, just so it could first be run through something like oxipng. Maybe it's just me having too much fun with this stuff, but [having a <1kb favicon](https://static.m4ss.net/favicon/favicon.ico) with 16x16, 32x32, 192x192 and 256x256 variants when most 16x16 and 32x32 favicons are 4kb is awesome :smile:

RunDevelopment:
This has been possible since v0.24.0 (released 4 years ago) with [`IcoEncoder::encode_images`](https://docs.rs/image/0.24.0/image/codecs/ico/struct.IcoEncoder.html#method.encode_images).

PR #1637, opened 2021-12-06
Make it possible to provide a pre-encoded image to the .ico encoder

paolobarbolini:
As mentioned in the second step of https://github.com/image-rs/image/issues/884#issuecomment-985755468.

Makes it possible to use a pre-encoded .png or .bmp file without having to first decode it. Also opens the ability to use something like oxipng to those who want to have fun with extremely small favicons :smiley: 

I haven't added any tests yet.

---

### DIFF
--- a/src/codecs/ico/encoder.rs
+++ b/src/codecs/ico/encoder.rs
@@ -1,11 +1,12 @@
 use byteorder::{LittleEndian, WriteBytesExt};
-use std::io::{self, Write};
+use std::io::{self, Cursor, Write};
 
 use crate::color::ColorType;
 use crate::error::ImageResult;
-use crate::image::ImageEncoder;
+use crate::image::{ImageDecoder, ImageEncoder};
 
-use crate::codecs::png::PngEncoder;
+use crate::codecs::bmp::BmpDecoder;
+use crate::codecs::png::{PngDecoder, PngEncoder, PNG_SIGNATURE};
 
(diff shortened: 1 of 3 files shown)


fintelia:
Should the interface allow providing multiple images to be included in the same ico? Or would that add too much complexity?

paolobarbolini:
> Should the interface allow providing multiple images to be included in the same ico? Or would that add too much complexity?

I think it should. I was planning to do that in a followup PR

src/codecs/ico/encoder.rs
```diff
@@ -54,6 +55,48 @@ impl<W: Write> IcoEncoder<W> {
...
+            let (width, height) = decoder.dimensions();
+            let color_type = decoder.color_type();
+            (width, height, color_type)
+        } else {
+            let decoder = BmpDecoder::new(Cursor::new(encoded_image))?;
+
+            let (width, height) = decoder.dimensions();
+            let color_type = decoder.color_type();
+            (width, height, color_type)
+        };
```
HeroicKatora, line 76: This puts it in a weird spot where we decode _some_ of the data but do not fully validate it either. Could we not expose the `pre_encoded_image_impl` method directly?

paolobarbolini: Sure, this is here only to make it harder to declare the wrong image dimensions or color type.

fintelia (approved):

PR #1639, opened 2021-12-17
Make it possible to supply multiple images to the .ico encoder

paolobarbolini:
Continuation of #1637

### DIFF
--- a/src/codecs/ico/encoder.rs
+++ b/src/codecs/ico/encoder.rs
@@ -2,7 +2,7 @@ use byteorder::{LittleEndian, WriteBytesExt};
 use std::io::{self, Write};
 
 use crate::color::ColorType;
-use crate::error::ImageResult;
+use crate::error::{ImageError, ImageResult, ParameterError, ParameterErrorKind};
 use crate::image::ImageEncoder;
 
 use crate::codecs::png::PngEncoder;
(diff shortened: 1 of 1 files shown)


src/codecs/ico/encoder.rs
```diff
@@ -55,30 +55,52 @@ impl<W: Write> IcoEncoder<W> {
...
     /// an error will be returned.
-    pub fn write_pre_encoded_image(
+    /// `images` must have a lenght between 1 and 65535 (inclusive).
+    pub fn write_pre_encoded_images(
         mut self,
-        encoded_image: &[u8],
-        width: u32,
-        height: u32,
-        color_type: ColorType,
+        images: &[(&[u8], u32, u32, ColorType)],
```
fintelia, line 73: Maybe replace the tuple with an `IcoFrame` struct?

src/codecs/ico/encoder.rs
```diff
@@ -19,6 +20,72 @@ pub struct IcoEncoder<W: Write> {
...
+    // Stored as `0 => 256, n => n`
+    height: u8,
+    color_type: ColorType,
+}
+
+impl<'a> IcoFrame<'a> {
+    /// Construct a new `IcoFrame` using a pre-encoded PNG or BMP
+    ///
+    /// The `width` and `height` must be between 1 and 256 (inclusive).
+    pub fn new(
```
paolobarbolini, line 38: I'm thinking about the naming and this probably isn't misuse resistant enough :thinking:

HeroicKatora: It's either `new` or a variant of `with_`. So what about `with_encoded`?

src/codecs/ico/encoder.rs
```diff
@@ -19,6 +20,72 @@ pub struct IcoEncoder<W: Write> {
     w: W,
 }
 
+/// An ICO image entry
+pub struct IcoFrame<'a> {
+    // Pre-encoded PNG or BMP
+    encoded_image: Cow<'a, [u8]>,
+    // Stored as `0 => 256, n => n`
+    width: u8,
```
fintelia, line 28: Wait, why `u8`?

paolobarbolini: Because ICOs have a 8bit width and height, so this pre-encodes that instead of doing it late into the encoding stage

HeroicKatora (approved):

Issue #1647, opened 2022-01-12
Avif and webp recognition is bugged


installgentoo:
image v0.23.14

## Expected

Avif/webp should work

## Actual behaviour

Format not recognised on images of odd size

## Reproduction steps

Convert image to avif with imagemagick, use libaom backbend for libheif. You get an apparently valid file which begins with 0x00 0x00 0x00 0x1C 0x66 0x74 0x79 0x70 0x61 0x76 0x69 0x66 
Magic bytes for avif are (b"\0\0\0 ftypavif", ImageFormat::Avif), and space is 0x20
You can already see the issue. We need another variant of avif with fileseparator in palce of space.

installgentoo:
Moreover mp4parse fails at parsing valid avif

an example file https://github.com/installgentoo/assorted_curiosities/blob/master/trash/ortodox_grafix.avif
properly displayed by firefox, imagemagick and geeqie.

installgentoo:
also dav1d_get_picture() returns 0 on that image, even though content is parsed properly.

installgentoo:
// Cross-correlate pixel format with planes and alignment.
    // wrapping_sub is wanted. If num_planes is 0, this turns in a very big number that
    // still represents an invalid number of planes.
    let last_src_plane = src_format.num_planes.wrapping_sub(1);
    if !pixel_format::is_compatible(src_pixel_format, width, height, last_src_plane) {
        return Err(ErrorKind::InvalidValue);
    }

specifically fails on this

installgentoo:
Okay i've debugged a bit more, and removing that check at dcv-color-primitives-0.1.16/src/lib.rs:827 makes my avifs load correctly.

Could some actual developer help me figure out where the last_src_plane(i suspect?) is set incorrectly? This is something in image crate methinks.

installgentoo:
On top of that dcv issue dav1d 1.0.0 seems to just break avif support. Apparently you have to flush something somewhere now.

HeroicKatora:
> On top of that dcv issue dav1d 1.0.0 seems to just break avif support

Do you happen to have a reproduction? I suspect we should do a call to `primary_decoder.flush()?;` here:
https://github.com/image-rs/image/blob/master/src/codecs/avif/decoder.rs#L40-L41

installgentoo:
I'm actively trying to debug.

You can grab the test.avif image from my second post and try opening it for yourself.

installgentoo:
[...]

PR #1649, opened 2022-01-12
Add new magic byte

installgentoo:
Add compatibility with .avif files created via libheif


installgentoo:
see https://github.com/image-rs/image/issues/1647

HeroicKatora (approved):

PR #1651, opened 2022-01-12
Use enlargeable types in to_luma functions.

### LINKED ISSUE #1214: to_luma methods can have floating point issues

HeroicKatora:
This is an attempt to fix #1214, using the `Enlargeable` trait to
select a suitable type for the calculation in `rgb_to_luma` and
`bgr_to_luma`.

Integer pixel types will use suitable integer pixel types for the
calculation, while floating point pixel types will use floating point
types.

This PR also provides the `Enlargeable` trait for all `Primitive`
types (luckily, `Primitive` is not implemented for 128 bit numeric
types).

The `Color` implementation of all library-defined structs is then
restricted to a subset of Primitive channel types. This is necessary to
allow its conversion operations to utilize the private `Enlargeable`.
Notably however we _relax_ the type itself to not utilize any trait
bound. It's a simple wrapper around an array now.

Rebase of #1215 which was auto-closed due to deleted base branch..

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -75,7 +75,7 @@ pub trait Enlargeable: Sized + Bounded + NumCast {
     fn clamp_from(n: Self::Larger) -> Self {
         if n > Self::max_value().to_larger() {
             Self::max_value()
-        } else if n < Self::min_value().to_larger(){
+        } else if n < Self::min_value().to_larger() {
             Self::min_value()
         } else {
             NumCast::from(n).unwrap()
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -70,15 +70,20 @@ declare_primitive!(f64: (0.0)..1.0);
 /// An Enlargable::Larger value should be enough to calculate
 /// the sum (average) of a few hundred or thousand Enlargeable values.
 pub trait Enlargeable: Sized + Bounded + NumCast {
-    type Larger: Primitive + AddAssign;
+    type Larger: Copy + NumCast + Num + PartialOrd<Self::Larger> + Clone + Bounded + AddAssign;
 
     fn clamp_from(n: Self::Larger) -> Self {
-        // Note: Only unsigned value types supported.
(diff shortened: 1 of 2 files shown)


src/traits.rs
```diff
@@ -70,16 +70,21 @@ declare_primitive!(f64: (0.0)..1.0);
 /// An Enlargable::Larger value should be enough to calculate
 /// the sum (average) of a few hundred or thousand Enlargeable values.
 pub trait Enlargeable: Sized + Bounded + NumCast {
```
fintelia, line 72: Could you make this `pub(crate)` for clarity?

HeroicKatora: I'm afraid, no. It's being used in a public interface even though it is not accessible, pretty much like a sealed trait.

HeroicKatora:
@fintelia Other concerns with the public change of this PR? It overlaps with reworking the `Pixel` trait in that if we removed `invert` and color conversion as originally sketched in #1099 then it might be possible to relax this to `Primitive` again.

HeroicKatora:
These are the two API breaking issues still open from `0.24`. We could also do a release just to get it out and then work hard on _actually_ fixing color, with conversion and better buffer API, in a `0.25`.

fintelia:
I think maybe merge this and then roll out a `0.24`? A broader Pixel rework for `0.25` seems reasonable to me.

fintelia (approved):

Issue #1623, opened 2021-11-22
Deprecate codecs::dxt


fintelia:
Quoting myself in https://github.com/image-rs/image/pull/1495#issuecomment-863390561:

> Right now our GPU compressed texture handling is really not all that great. For decoding we don't support any of the newer formats like BC6, BC7 or ASTC, And for encoding, our code is slow, low quality, and doesn't support any of the newer formats.

As a result, pretty much any use of the module would be better served by [squish-rs](https://crates.io/crates/squish/2.0.0-beta1) which implements faster and higher quality encoding (though be sure to use `2.0.0-beta1` or later to get the recent bugfixes).

Additionally, unlike the other codecs, DXT is not actually an image file format but rather a compression format. This distinction matters for instance because it means that the DxtDecoder::new() has to be passed the image metadata unlike all our other decoders which don't need that. My suggestion is to just completely deprecate (and eventually remove) the codecs::dxt module. We can still keep the codecs::dds by porting it to rely on squish-rs instead.

coderedart:
reached here while checking for basis universal support.
It seems squish is archived now.

work continuing at https://github.com/jansol/texpresso

Shadlock0133:
If it's useful for anyone, I've implemented BC7 decoder without any dependencies here: https://github.com/Shadlock0133/bigblob-decoder/blob/main/src/bc7/decode.rs

CensoredUsername:
As the original author of the DXT codec, that's fine with me. I wrote it for another project (which didn't need full DDS support) when there wasn't any other rust impl of the algorithm yet, and figured it'd be useful for other people. It's pretty ad hoc and just brute-forces some of the harder parts of the algorithm so it's indeed fairly slow. If a better implementation exists, by all means use it instead.

PR #1654, opened 2022-01-16
Deprecate DXT support

fintelia:
https://github.com/image-rs/image/issues/1623 describes the motivation:

> > Right now our GPU compressed texture handling is really not all that great. For decoding we don't support any of the newer formats like BC6, BC7 or ASTC, And for encoding, our code is slow, low quality, and doesn't support any of the newer formats.
> 
> As a result, pretty much any use of the module would be better served by squish-rs which implements faster and higher quality encoding (though be sure to use 2.0.0-beta1 or later to get the recent bugfixes).
> 
> Additionally, unlike the other codecs, DXT is not actually an image file format but rather a compression format. This distinction matters for instance because it means that the DxtDecoder::new() has to be passed the image metadata unlike all our other decoders which don't need that. My suggestion is to just completely deprecate (and eventually remove) the codecs::dxt module. We can still keep the codecs::dds by porting it to rely on squish-rs instead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -222,6 +222,7 @@ pub mod codecs {
     #[cfg(feature = "dds")]
     pub mod dds;
     #[cfg(feature = "dxt")]
+    #[deprecated = "DXT support will be removed in a future version. See https://github.com/image-rs/image/issues/1623"]
     pub mod dxt;
     #[cfg(feature = "farbfeld")]
     pub mod farbfeld;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -222,7 +222,7 @@ pub mod codecs {
     #[cfg(feature = "dds")]
     pub mod dds;
     #[cfg(feature = "dxt")]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -222,6 +222,7 @@ pub mod codecs {
     #[cfg(feature = "dds")]
     pub mod dds;
     #[cfg(feature = "dxt")]
+    #[deprecated = "DXT support will be removed in a future version. See https://github.com/image-rs/image/issues/1623"]
```
HeroicKatora, line 225:
```suggestion
    #[deprecated = "DXT support will be removed or reworked in a future version. Prefer the `squish` crate instead. See https://github.com/image-rs/image/issues/1623"]
```

HeroicKatora (commented):
Maybe we should propose the known alternative directly in the notice as well.

HeroicKatora (approved):

PR #1657, opened 2022-01-24
Export Enlargeable

Gelbpunkt:
This is required for e.g. imageproc.

Previously:
```rust
impl<T, U> WithChannel<U> for Rgb<T>
where
    T: Primitive + 'static,
    U: Primitive + 'static,
{
    type Pixel = Rgb<U>;
}
```

Now:
```rust
impl<T, U> WithChannel<U> for Rgb<T>
where
    T: Primitive + Enlargeable + 'static,
    U: Primitive + Enlargeable + 'static,
{
    type Pixel = Rgb<U>;
}
```

The trait is `pub`, but not exported anywhere.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,7 +136,7 @@ pub use crate::buffer_::{
 pub use crate::flat::FlatSamples;
 
 // Traits
-pub use crate::traits::{EncodableLayout, Pixel, PixelWithColorType, Primitive};
+pub use crate::traits::{EncodableLayout, Enlargeable, Pixel, PixelWithColorType, Primitive};
 
 // Opening and loading images
 pub use crate::dynimage::{


HeroicKatora:
The trait was introduced as sealed/private, to be on the safe side of being able to modify it. For example, adding additional methods that clarify how exactly we need it to be 'enlargeable'. What about this alternative way to write the trait impl in `imageproc` which does not touch any of the traits?


```
impl<T, U> WithChannel<U> for Rgb<T>
where
    Rgb<T>: Pixel,
    Rgb<U>: Pixel,
{
    type Pixel = Rgb<U>;
}
```

Gelbpunkt:
Right, this works (not sure why it does not without specifying subpixel explicitly):

```rust
impl<T, U> WithChannel<U> for Rgb<T>
where
    Rgb<T>: Pixel<Subpixel = T>,
    Rgb<U>: Pixel<Subpixel = U>,
    U: Primitive + 'static,
{
    type Pixel = Rgb<U>;
}
```

Going to close this, thanks :)

PR #1674, opened 2022-02-13
Fix incorrect `get_pixel_[mut_]checked` indexing

### LINKED ISSUE #1672: `ImageBuffer::get_pixel_checked` sometimes returns incorrect pixel
In certain cases, the functions `ImageBuffer::get_pixel_checked` and `ImageBuffer::get_pixel_mut_checked` seem to return a shifted pixel reference.

## Reproduction steps

The following code creates a 2x1 RGB image buffer, sets the second pixel to white, then reads its color.
Reading with `get_pixel` succeeds with white but `get_pixel_checked` results in a blue pixel.

(On  [Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f5658a90ff8bf1ba9250b08ee1b9f16c))

```rust
use image::{Rgb, RgbImage};

const WHITE: Rgb<u8> = Rgb([255, 255, 255]);

fn main() {
    let mut img = RgbImage::new(2, 1);

    img.put_pixel(1, 0, WHITE);

    assert_eq!(img.get_pixel(1, 0), &WHITE);
    assert_eq!(img.get_pixel_checked(1, 0).unwrap(), &WHITE);
}
```

The second assert fails:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Rgb([0, 0, 255])`,
 right: `Rgb([255, 255, 255])`', src/main.rs:11:5
```

okaneco:
Add missing pixel channel count multiplication to indexing
Add additional tests to compare methods to `ImageBuffer::get_pixel`

---

This aligns the checked logic with `ImageBuffer::pixel_indices_unchecked` which calculates the minimum index with the following:
```rust
fn pixel_indices_unchecked(&self, x: u32, y: u32) -> Range<usize> {
    let no_channels = <P as Pixel>::CHANNEL_COUNT as usize;
    // If in bounds, this can't overflow as we have tested that at construction!
    let min_index = (y as usize * self.width as usize + x as usize) * no_channels;
    min_index..min_index + no_channels
}
```

Open to suggestions for other/better tests.

Closes #1672

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -735,7 +735,8 @@ where
         let num_channels = <P as Pixel>::CHANNEL_COUNT as usize;
         let i = (y as usize)
             .saturating_mul(self.width as usize)
-            .saturating_add(x as usize);
+            .saturating_add(x as usize)
+            .saturating_mul(num_channels);
 
         self.data
             .get(i..i + num_channels)
@@ -910,7 +911,8 @@ where
         let num_channels = <P as Pixel>::CHANNEL_COUNT as usize;
         let i = (y as usize)
             .saturating_mul(self.width as usize)
-            .saturating_add(x as usize);
+            .saturating_add(x as usize)
+            .saturating_mul(num_channels);
 
         self.data
             .get_mut(i..i + num_channels)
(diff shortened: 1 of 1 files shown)


okaneco:
I had the same test fail locally. The `test_defaults` runner failed.
```bash
check_references tests/reference/jpg/progressive/3.jpg.79e90ab0.png
thread 'check_references' panicked at 'The decoded image's hash does not match (expected = 79e90ab0, actual = c959cb95).', tests/reference_[images.rs:279](http://images.rs:279/):13

failures:
    check_references

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.40s
```

HeroicKatora (approved):
Test failure is caused by "jpeg-decoder = 0.2.2", now using SIMD code with slightly different results.

HeroicKatora:
Thanks!

PR #1675, opened 2022-02-15
0.24.1 bugfix release

fintelia:
We should probably release a version incorporating https://github.com/image-rs/image/pull/1674 sooner rather than later.


fintelia:
@HeroicKatora do you know what's going on with that one test failure?

HeroicKatora (commented):
Test failure is on `check_references tests/reference/jpg/progressive/3.jpg.79e90ab0.png`. I suspsect this is linked to the jpeg-decoder upgrade and it now utilizing SIMD instructions. We don't guarantee bit-for-bit exact results with that code, that might have broken the test which is using crc32—not a perceptual distance test of sorts.

PR #1678, opened 2022-02-23
Add DX10 extended header support for DDS

PowerBall253:
Allows the DDS decoder to load DXT-format files with the additional DX10 header, as described in <https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dx-graphics-dds-pguide#dds-variants>.

The DX10 header struct was taken from [Microsoft's DX10 docs](https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dds-header-dxt10), and the equivalent DXT1/3/5 DXGI types were taken from [here](https://docs.microsoft.com/en-us/windows/win32/direct3d11/texture-block-compression-in-direct3d-11) and [here](https://docs.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format).

### DIFF
--- a/src/codecs/dds.rs
+++ b/src/codecs/dds.rs
@@ -30,6 +30,8 @@ enum DecoderError {
 
     /// Invalid DXGI format in DX10 header
     DxgiFormatInvalid(u32),
+    /// Invalid resource dimension
+    ResourceDimensionInvalid(u32),
     /// Invalid flags in DX10 header
     Dx10FlagsInvalid(u32),
     /// Invalid array size in DX10 header
@@ -53,6 +55,9 @@ impl fmt::Display for DecoderError {
             }
             DecoderError::DxgiFormatInvalid(df) => {
                 f.write_fmt(format_args!("Invalid DDS DXGI format: {}", df))
+            }
+            DecoderError::ResourceDimensionInvalid(d) => {
+                f.write_fmt(format_args!("Invalid DDS resource dimension: {}", d))
             }
             DecoderError::Dx10FlagsInvalid(fs) => {
                 f.write_fmt(format_args!("Invalid DDS DX10 header flags: {:#010X}", fs))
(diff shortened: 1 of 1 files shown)


src/codecs/dds.rs
```diff
@@ -176,6 +204,25 @@ impl<R: Read> DdsDecoder<R> {
                 b"DXT1" => DxtVariant::DXT1,
                 b"DXT3" => DxtVariant::DXT3,
                 b"DXT5" => DxtVariant::DXT5,
+                b"DX10" => {
+                    let dx10_header = DX10Header::from_reader(&mut r)?;
+                    match dx10_header.dxgi_format {
+                        70 | 71 | 72 => DxtVariant::DXT1,
+                        73 | 74 | 75 => DxtVariant::DXT3,
+                        76 | 77 | 78 => DxtVariant::DXT5,
```
HeroicKatora, line 212: Can you insert comments (with links to documentation) regarding these constants?  A link to 
<https://docs.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format>
and the general information that these refer to subvariants with different sample interpretation:
 typeless, unorm, unorm_srgb

src/codecs/dds.rs
```diff
@@ -71,6 +71,16 @@ struct Header {
...
 
+/// Extended DX10 header used by some DDS image files
+#[derive(Debug)]
+struct DX10Header {
+    dxgi_format: u32,
+    resource_dimension: u32,
+    misc_flag: u32,
+    array_size: u32,
+    misc_flags_2: u32,
+}
```
HeroicKatora, line 82: Is there any of these fields that should require some form of validation? If not, why? They are being read but disregarded which is not very forward compatible and could be a slight hazard if that causes parsing result to change unexpectedly in the future.

src/codecs/dds.rs
HeroicKatora: Links from the PR comments would also be good for this purpose! Those comments should belong in the commit message or code, not merely github's issue system.

src/codecs/dds.rs
PowerBall253: Should I add the validation for the fields in the `from_reader` function?

HeroicKatora: A separate `fn validate(&self) -> Result<()>` should suffice. At time we'll likely want to change it to return some kind of parsing result (i.e. feed it to the color type/metadata recorder) and then it's a bit easier if it's separate from reading.

src/codecs/dds.rs
```diff
@@ -171,13 +178,43 @@ impl DX10Header {
...
+    }
+
+    fn validate(&self) -> Result<(), ImageError> {
+        // Note: see https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dds-header-dxt10 for info on valid values
+        if dxgi_format < 0 || dxgi_format > 132 {
+            // Invalid format
+            return Err(DecoderError::DxgiFormatInvalid(dxgi_format).into());
+        }
+
+        if misc_flag != 0x4 && misc_flag != 0x4 {
```
HeroicKatora:
```suggestion
        if misc_flag != 0x0 && misc_flag != 0x4 {
```

PowerBall253:
@HeroicKatora I have implemented the DX10 header validation and added comments for the DXGI constants, will wait for your review.

fintelia (approved):

HeroicKatora (approved):
LGTM. Took a while to get to this review, sorry.

PR #1168, opened 2020-03-25
Support decoding WEBPs in extended format

paolobarbolini:
This adds support for decoding WEBPs encoded in [extended format](https://developers.google.com/speed/webp/docs/riff_container#extended_file_format).

I added tests by copying the two test images in [simple format](https://developers.google.com/speed/webp/docs/riff_container#simple_file_format_lossy), adding a [VP8X chunk](https://developers.google.com/speed/webp/docs/riff_container#extended_file_format) and appended an unknown chunk before and after the 'VP8 ' chunk to make sure the decoder correctly skipped unknown chunks and read the data from the VP8 chunk without going over.

The resulting decoded image is the same as the one from the files from the tests in simple format.

I license past and future contributions under the dual MIT/Apache-2.0 license,
allowing licensees to chose either at their option.

PR #1685, opened 2022-03-09
Extended WebP image decoding

tristanphease:
Allows for decoding of extended WebP headers(VP8X) as [described here](https://developers.google.com/speed/webp/docs/riff_container#extended_file_format), including lossy alpha images and animated WebP files.

Integrates with the current AnimationDecoder trait for animation. One thing that I think would make sense on that trait is a way to get the loop count out of it. Also when testing, I was trying the frame offsets in the Frame::from_parts() which fit what I was doing quite well, but it never seemed to work properly when encoded to a GIF, so it just manually draws the background with the offset.

Ignores ICCP colour profiles, XMP and EXIF metadata. If #1448 happens, maybe we could integrate that in.

Feedback would be appreciated.


fintelia:
Haven't had a chance for a full read-through yet, but could also update the table in src/lib.rs:
https://github.com/image-rs/image/blob/53b9c5cd663cf9527615c8180f20b573a3c3a82a/src/lib.rs#L178-L194

src/codecs/webp/vp8.rs
```diff
@@ -913,6 +913,28 @@ impl Frame {
         }
     }
 
+    /// Fills an rgba buffer by skipping the alpha values
+    pub(crate) fn fill_rgba(&self, buf: &mut [u8]) {
+        for index in 0..self.ybuf.len() {
```
HeroicKatora, line 918: In my opinion, more idiomatic would be

```suggestion
        for (index, rgb_chunk) in (0..self.ybuf.len()).zip(buf.chunks_exact_mut(4)) {
```

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+    Lossy(RgbaImage),
+    Lossless(LosslessFrame),
+}
+
+impl WebPStatic {
+    pub(crate) fn from_alpha_lossy(
+        alpha: AlphaChunk,
+        vp8_frame: VP8Frame,
+    ) -> ImageResult<WebPStatic> {
+        if alpha.data.len() != usize::from(vp8_frame.width * vp8_frame.height) {
```
HeroicKatora, line 323: Here the `usize` conversion is done _after_ multiplication, below it is done on both operands separately. Conversion to the larger type is surely required, the type is `u16`.

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+                let predictor: u8 = WebPStatic::get_predictor(
+                    x.into(),
+                    y.into(),
+                    vp8_frame.width.into(),
+                    alpha.filtering_method,
+                    &image_vec,
+                );
+                let predictor = u16::from(predictor);
+
+                let alpha_index = usize::from(y * vp8_frame.width + x);
```
HeroicKatora, line 343: Similar issue here, this index should be calculated with individual `usize` operands. It will overflow/wrap/panic at the moment.

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+    pub(crate) fn from_alpha_lossy(
+        alpha: AlphaChunk,
+        vp8_frame: VP8Frame,
+    ) -> ImageResult<WebPStatic> {
+        if alpha.data.len() != usize::from(vp8_frame.width * vp8_frame.height) {
+            return Err(DecoderError::SizeMismatch.into());
+        }
+
+        let mut image_vec =
+            vec![0u8; usize::from(vp8_frame.width) * usize::from(vp8_frame.height) * 4];
```
HeroicKatora, line 328: For 32-bit targets this is prone to an overflow with `width = height = u16::MAX` or similar. Prefer checked multiplication and maybe perform that length calculation before the previous statement, as success certifies that the previous is free of overflow as well.

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+use super::vp8::{Frame as VP8Frame, Vp8Decoder};
+use crate::error::DecodingError;
+use crate::image::ImageFormat;
+use crate::{color, Delay, Frame, Frames, ImageError, ImageResult, Rgba, RgbaImage};
+use byteorder::{LittleEndian, ReadBytesExt};
+
+//all errors that can occur while parsing extended chunks in a WebP file
+#[derive(Debug, Clone, Copy)]
+enum DecoderError {
+    ChunkInvalid,
```
HeroicKatora, line 16: Can you add the `WebPRiffChunk` identification as data here, and include it in the error message? This is simple yet possibly effective at identifying a little more information. I wouldn't mind using the chunk's `fmt::Debug` information in the error's `Display` actually.

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+        (u32::from(buffer[2]) << 16) | (u32::from(buffer[1]) << 8) | u32::from(buffer[0]);
+    Ok(value)
+}
+
+fn read_lossy<R: Read>(reader: &mut R) -> ImageResult<VP8Frame> {
+    let (cursor, chunk) =
+        read_chunk(reader)?.ok_or_else(|| Error::from(io::ErrorKind::UnexpectedEof))?;
+
+    if chunk != WebPRiffChunk::VP8 {
+        return Err(DecoderError::ChunkInvalid.into());
```
HeroicKatora, line 556: Maybe this should be a separate error variant?

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+        Some((mut cursor, WebPRiffChunk::ALPH)) => {
+            let alpha_chunk = read_alpha_chunk(&mut cursor, width, height)?;
+
+            let vp8_frame = read_lossy(reader)?;
+
+            let img = WebPStatic::from_alpha_lossy(alpha_chunk, vp8_frame)?;
+
+            Ok(img)
+        }
+        _ => Err(DecoderError::ChunkInvalid.into()),
```
HeroicKatora, line 594: Same here.

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+use crate::error::DecodingError;
+use crate::image::ImageFormat;
+use crate::{color, Delay, Frame, Frames, ImageError, ImageResult, Rgba, RgbaImage};
+use byteorder::{LittleEndian, ReadBytesExt};
+
+//all errors that can occur while parsing extended chunks in a WebP file
+#[derive(Debug, Clone, Copy)]
+enum DecoderError {
+    ChunkInvalid,
+    SizeMismatch,
```
HeroicKatora, line 17: Can you split this into multiple variants as well, depending on how that mismatch occurred.

src/codecs/webp/lossless.rs
```diff
@@ -168,6 +168,30 @@ impl<R: Read> LosslessDecoder<R> {
...
+        let mut buf = Vec::new();
+        self.r.read_to_end(&mut buf)?;
+        self.bit_reader.init(buf);
+
+        self.frame.width = width;
+        self.frame.height = height;
+
+        let mut data = self.decode_image_stream(self.frame.width, self.frame.height, true)?;
+
+        for &trans_index in self.transform_order.iter().rev() {
```
HeroicKatora, line 186: A note that each value in `transform_order` is 2-bit would be nice. It's local information in the other usage but non-obvious for this one, which makes the index below surprising on first sight.

HeroicKatora (changes requested):
So nice to see this PR :pleading_face:

src/codecs/webp/decoder.rs
```diff
@@ -101,78 +157,90 @@ impl<R: Read> WebPDecoder<R> {
...
+pub(crate) fn read_len_cursor<R>(r: &mut R) -> ImageResult<Cursor<Vec<u8>>>
 where
     R: Read,
 {
-    let len = r.read_u32::<LittleEndian>()?;
+    let mut len = u64::from(r.read_u32::<LittleEndian>()?);
+
+    if len % 2 == 1 {
+        len += 1;
+    }
```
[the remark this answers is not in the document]
paolobarbolini: It's explained a bit better in my PR which originally added it https://github.com/image-rs/image/pull/1168/files#diff-5474c83ac44568ce154a80e551838e4538cb1eddd3fded09c0c74d104ae00c14R82-R86

I agree with the fact that the comment should be kept.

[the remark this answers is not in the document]
tristanphease: Sorry, didn't move the comment over when I changed that section. Also yeah, removing the padding byte is probably safer.

src/codecs/webp/extended.rs
tristanphease: Yeah I wasn't really sure about classifying errors, maybe could use the ChunkHeaderInvalid from decoder.rs?

HeroicKatora: I wouldn't be too worried about code reuse here. Rather be as detailed and specific as useful for either certifying _how_  the file standard contradicts the specification (i.e. something you can check on your own) or for locating bugs in the decoder.

HeroicKatora: More code duplication reuse can be helpful. It allows mapping specific error information to a line of code, if this information isn't directly visible from other information. In private projects I almost always include the line of code and even source files. We don't want that kind of information leakage though, but being cautious on reusing error variants already goes most of the way.

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+                            let canvas_f64 = f64::from(canvas[i]);
+                            let buffer_f64 = f64::from(buffer[index + i]);
+
+                            let val = (buffer_f64 * buffer_alpha
+                                + canvas_f64 * canvas_alpha * (1.0 - buffer_alpha / 255.0))
+                                / blend_alpha_f64;
+                            rgb[i] = val as u8;
+                        }
+
+                        rgb
```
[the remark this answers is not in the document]
tristanphease: Done

src/codecs/webp/extended.rs
```diff
@@ -0,0 +1,683 @@
...
+
+impl FilteringMethod {
+    fn from_num(num: u8) -> Self {
+        match num {
+            0 => FilteringMethod::None,
+            1 => FilteringMethod::Horizontal,
+            2 => FilteringMethod::Vertical,
+            3 => FilteringMethod::Gradient,
+            _ => panic!("Invalid filter for alpha chunk in WebP image, should be unreachable"),
+        }
```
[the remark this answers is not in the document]
tristanphease: Yeah, that's better, done.

src/codecs/webp/vp8.rs
tristanphease: Yeah I had intended to rewrite some of this, done now.

src/codecs/webp/extended.rs
tristanphease: Oops, done.

src/codecs/webp/extended.rs
tristanphease: I tried changing all the errors to be more specific to the issue with the image file. With that one, I did use the ChunkHeaderInvalid since I think it characterises the error received.

tristanphease:
Also, with the tests I created an animated image to test but I couldn't figure out how the reference tests check against the frames of animated images if at all. At the moment, it just checks against the first frame of the animation but ideally it would check more.

tristanphease:
@HeroicKatora Sorry, but do you have any further feedback?
I believe all your recommended changes have been resolved.

HeroicKatora (approved):
:+1: LGTM

HeroicKatora:
All files had been marked 'viewed' for me, somehow must have missed pressing `Approve` before :smile:

PR #1688, opened 2022-03-19
impl `GenericImageView` for `SubImage<I>`

JulianKnodt:
Motivation:
While using a 3rd party package which takes a `GenericImageView`, I found that a SubImage does
not actually implement `GenericImageView`, despite being a `GenericImageView` view :).

Thus, I added an implementation, delegating most of the functionality to the inner
`GenericImageView`.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -1164,6 +1164,42 @@ impl<I> SubImage<I> {
     }
 }
 
+impl<I> GenericImageView for SubImage<I>
+where
+    I: Deref,
+    I::Target: GenericImageView,
+{
+    type Pixel = <<I as Deref>::Target as GenericImageView>::Pixel;
+    fn dimensions(&self) -> (u32, u32) {
+        (self.xstride, self.ystride)
+    }
+
+    fn get_pixel(&self, x: u32, y: u32) -> Self::Pixel {
+        if x > self.xstride || y > self.ystride {
+            panic!(
+                "Image index {:?} out of bounds {:?}",
+                (x, y),
+                self.dimensions()
+            );
+        }
+        self.inner.get_pixel(x + self.xoffset, y + self.yoffset)
+    }
+
+    fn bounds(&self) -> (u32, u32, u32, u32) {
+        (
+            self.xoffset,
+            self.yoffset,
+            self.xoffset + self.xstride,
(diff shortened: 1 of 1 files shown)


fintelia:
Thanks for proposing this! The current choice not to the trait is actually intentional for now:
https://github.com/image-rs/image/blob/53b9c5cd663cf9527615c8180f20b573a3c3a82a/src/image.rs#L1089-L1099

JulianKnodt:
Oops sorry for being dumb and proposing something that had been previously rejected, I thought I had looked to see if there was an existing issue for it but did no. This nesting does make sense as a reason to not implement `GenericImageView` here, and I guess changing the trait so that it could be generic over the output type is too costly otherwise.

Would it make sense to implement a flattened structure, which takes a `SubImage<I>`, and returns a `GenericImageView`? This could be fairly easy to implement without modifying the old structure, and this could be done by eagerly flattening any nested `SubImage`s.

i.e. Some functions:
`generic_image_view(&self: SubImage<SubImage<I>>) -> impl GenericImageView`,
and
`generic_image_view(&self: SubImage<I>) -> impl GenericImageView`,
where the first function will call the 2nd one until it resolves to one level of nesting?

HeroicKatora:
I don't see how this would be possible without specialization, unless I'm misunderstanding the proposed function.

JulianKnodt:
I was thinking it would be possible to eagerly flatten the structure, but I guess at compile-time you wouldn't know the number of nested-subviews, so I guess it does make sense you would need specialization if the inner SubImage<Inner> was a SubImage too.

I guess the main difficulty is that you can get these nested subviews, so it would make more sense to change the trait `GenericImageView` to have a `type Sub = Self`, but only override it for SubImage (`type Sub = I::Target`)  so that it always has a flattened structure?

Sorry, about my confusion, for my case I only have 1 level of subview so I didn't think about having a nested type.

fintelia:
What you're describing sounds similar to [how things worked in image 0.23.x](https://docs.rs/image/0.23.14/image/trait.GenericImage.html).

JulianKnodt:
mmmm makes sense, and I guess it was moved away from with good reason. Can't think of any other way to go about this, sorry about the spurious PR!

PR #1689, opened 2022-03-20
Fix Todo and loop colorops.rs

Kallu-A:
This pull request fixes the todo in the code and some other loop that can be upgraded by using .pixels()

As this is my first pull request on this project. I'm not sure if this is done right, so please feel free to start a discussion, if i have done something wrong or missed things.

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -56,7 +56,7 @@ where
     let (width, height) = image.dimensions();
     let mut out = ImageBuffer::new(width, height);
 
-    for pixel in image.pixels() {
+    for (x, y, color) in image.pixels() {
         let grayscale = pixel.2.to_luma_alpha();
         let new_pixel = grayscale.into_color(); // no-op for luma->luma
 
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -1,9 +1,8 @@
 //! Functions for altering and converting the color of pixelbufs
 
 use num_traits::NumCast;
-use std::borrow::{Borrow, BorrowMut};
-use std::cell::RefCell;
 use std::f64::consts::PI;
+use std::ptr::addr_of_mut;
 
 use crate::color::{FromColor, IntoColor, Luma, LumaA, Rgba};
 use crate::image::{GenericImage, GenericImageView};
@@ -70,10 +69,13 @@ where
 /// Invert each pixel within the supplied image.
 /// This function operates in place.
 pub fn invert<I: GenericImage>(image: &mut I) {
-    let mut image = RefCell::new(image);
-    for (x, y, mut pixel) in image.borrow().pixels() {
-        pixel.invert();
-        image.borrow_mut().put_pixel(x, y, pixel);
+    let mut image = image;
+    let image = addr_of_mut!(image);
+    unsafe {
+        for (x, y, mut pixel) in image.as_ref().unwrap().pixels() {
+            pixel.invert();
+            image.as_mut().unwrap().put_pixel(x, y, pixel);
+        }
     }
 }
 
@@ -96,7 +98,7 @@ where
 
     let percent = ((100.0 + contrast) / 100.0).powi(2);
 
-    image.pixels().map(|(x, y, pixel)| {
+    for (x, y, pixel) in image.pixels() {
         let f = pixel.map(|b| {
             let c: f32 = NumCast::from(b).unwrap();
 
@@ -106,7 +108,7 @@ where
             NumCast::from(e).unwrap()
         });
         out.put_pixel(x, y, f);
-    });
+    }
 
     out
 }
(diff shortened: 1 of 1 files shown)


src/imageops/colorops.rs
```diff
@@ -72,16 +69,8 @@ where
...
-
-    for y in 0..height {
-        for x in 0..width {
-            let mut p = image.get_pixel(x, y);
-            p.invert();
-
-            image.put_pixel(x, y, p);
-        }
+    for mut pixel in image.pixels() {
+        pixel.2.invert();
```
HeroicKatora, line 73: This comment was, unfortunately, slightly outdated and the interpretation is incorrect. This will not invert the pixel within the image as `pixels_mut()` would have done, it will invert a single copy of the color.

Unfortunately, `pixels_mut()` does not work as a generic iterator so it had to be removed and deprecated some time ago.

HeroicKatora (commented):
Thank you for addressing this. The loop items can be made a little more idiomatic.

src/imageops/colorops.rs
Kallu-A: I'm not sure where is the issue here, `pixel.invert()` change the value of the pixel and `.pixels()` will iterate on all the pixels. 
Why this will invert a single copy of the color?

fintelia: `.pixels()` is a read-only iterator over the pixels of an image. Modifying the returned value of `pixel` will be visible locally, but won't get written back to the image itself.

Kallu-A: This makes sense, so what is the best way to do it? 
By using `borrow_mut()` the change should persist, or just keep the `put_pixel()`

HeroicKatora: You'll need to use `put_pixel`. The change won't persist otherwise, as the code always operates on a copy of the pixel.

Kallu-A: Okay, I did it, thanks for your help.

src/imageops/colorops.rs
```diff
@@ -72,16 +70,10 @@ where
...
-        for x in 0..width {
-            let mut p = image.get_pixel(x, y);
-            p.invert();
-
-            image.put_pixel(x, y, p);
-        }
+    let mut image = RefCell::new(image);
+    for (x, y, mut pixel) in image.borrow().pixels() {
+        pixel.invert();
+        image.borrow_mut().put_pixel(x, y, pixel);
```
fintelia, line 76: Do any of the tests execute this line? I'd expect it to panic at runtime because the borrow at the top of the loop is still in scope.

src/imageops/colorops.rs
```diff
@@ -104,21 +96,17 @@ where
...
 
-    // TODO use pixels_mut?
-    for y in 0..height {
-        for x in 0..width {
-            let f = image.get_pixel(x, y).map(|b| {
-                let c: f32 = NumCast::from(b).unwrap();
-
-                let d = ((c / max - 0.5) * percent + 0.5) * max;
-                let e = clamp(d, 0.0, max);
+    image.pixels().map(|(x, y, pixel)| {
```
fintelia, line 99: Doesn't this need to be a `for` loop rather than a call to `.map()`?

Kallu-A: in the past code, it's was a map so I kept it but I can use a loop if you want.

HeroicKatora: The past code made use of `Pixel::map`, not `Iterator::map`.

src/imageops/colorops.rs
Kallu-A: No none of the tests were currently running this code. I did some tests function and it didn't work at runtime.  
I'm trying to figure out a solution.

src/imageops/colorops.rs
Kallu-A: Oops. I will correct that

src/imageops/colorops.rs
Kallu-A: It works by using `addr_of_mut!` and some `unsafe` code. I have added a few tests to make sure and it works at both compile-time and runtime.
But I'm not sure if its the best solutions.

src/imageops/colorops.rs
```diff
@@ -125,18 +127,21 @@ where
 
     let percent = ((100.0 + contrast) / 100.0).powi(2);
 
-    let mut image = RefCell::new(image);
-    image.borrow().pixels().map(|(x, y, mut pixel)| {
-        let f = pixel.map(|b| {
-            let c: f32 = NumCast::from(b).unwrap();
+    let mut image = image;
+    let image = addr_of_mut!(image);
+    unsafe {
```
HeroicKatora, line 132: No. This does not come with safety qualifications, and the TODO was looking for a _better_ way. `unsafe` is not a better way. Just leave it as it was.

HeroicKatora (commented):
`unsafe` may _appear_ to work for some examles, but it doesn't, and this is not acceptable as is. An `unsafe` block means the programmer did the compiler's job of proving code sound. However, none of the blocks come with any safety qualifications that actually demonstrate this understanding. They should come with reasoning (quite mathematical proves at best) that convey the logic behind their correctness. No justification is provided and the required reasonsing would certainly not be trivial (`unsafe` in generic code is _never_ trivial) so this is a definitive case of 'not worth the risk'.

src/imageops/colorops.rs
Kallu-A: Okay, I leave it as it was where the borrowing is a problem.

HeroicKatora (approved):
The extra test coverage is an awesome byproduct, that's basically a full second PR :+1:

src/imageops/colorops.rs
```diff
@@ -556,7 +545,38 @@ where
 mod test {
 
     use super::*;
-    use crate::ImageBuffer;
+    use crate::{GrayImage, ImageBuffer};
+
+    macro_rules! assert_pixels_eq {
```
fintelia, line 550: Why does this need to be a macro?

HeroicKatora: I really didn't mind it following `assert_eq!` more closely in form. That said it does evaluate its arguments twice, now that you mention it.

Kallu-A: I have copy-paste `pixels_diffs` and the macro from the `affine.rs` file because `assert_pixel_eq!` is handy, but if it's a problem I can rewrite it as a function or use `assert_eq!`.

HeroicKatora: I gravitate towards 'separate PR' for that change. The specific portion `diff(a, b) -> impl Iterator` could even be useful as a public function.

Kallu-A: I can help to work on that if you want.

PR #1690, opened 2022-03-22
Use consistent spelling of the word 'gray' in the docs

foresterre:
While both are correct (depending on which side of the pond one lives), most documentation uses gray instead of grey, and, the imageops also uses the method name grayscale().

I became aware of the two different spellings used when reading the DynamicImage::grayscale() docs, where both variants were used within the same documentation item :).


HeroicKatora (approved):

fintelia:
Thanks!

PR #1691, opened 2022-03-24
Compensate for overfull buffer in write

HeroicKatora:
Output formats, but mainly PNG, may expect the byte buffer to be
precisely sized according to the image descriptor. The `ImageBuffer`
implementation permits its underlying container to be larger than this
for multiple reasons, and won't check this property in `from_raw`.
Hence, the actual size may be too large. This mismatch can cause
encoding to fail for a curious reason:

	failed to save: IoError(Custom { kind: Other, error: "wrong data size, expected 1288872 got 1289808" })

Too much data should never be harmful here.

We avoid this by pre-selecting only the relevant portion of the
underlying container.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,11 @@ Rust image aims to be a pure-Rust implementation of various popular image format
 
 ### Unreleased
 
+Bug fixes:
+- Calling `DynamicImage`/`ImageBuffer`'s methods `write_to` and `save` will now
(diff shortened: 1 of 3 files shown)


fintelia (approved):

HeroicKatora:
Turns out my regression test was inside a `#[cfg(feature = "benchmark")]` block and not exected when I tried it locally. Whoops. I took the liberty to reorganize that a little bit as well.

Issue #1488, opened 2021-06-01
Progress Callback FnMut & Write Progress Callback?


johannesvollmer:
I wondered whether it is planned to add progress to the writer eventually, and if the progress callback could be made mutable? I'm sure if a user wanted to actually use the progress callback to update a UI Progressbar, an immutable `Fn` would not work, right? 

## Draft

The `Fn` could maybe even changed to `FnMut` without major refactorings. This change could even be backwards-compatible on first glance, as any `Fn` can also be passed to any function expecting `FnMut`.

fintelia:
I don't think it was a deliberate decision not to make the callback `FnMut`, and we should probably correct the `ImageDecoder` trait accordingly. However, I believe the change might break any downstream crates which implement their own decoders, so it probably belongs in the 0.24 branch.

johannesvollmer:
Yes, implementing a decoder will not necessarily compile when Fn is changed to FnMut, I didn't think of that :)

Shnatsel:
The `read_with_progress` family of methods was removed in 0.25.x, so this is no longer relevant.

PR #1694, opened 2022-04-02
make progress callback mutable

johannesvollmer:
see #1488 

which branch should it go?

### DIFF
--- a/src/codecs/openexr.rs
+++ b/src/codecs/openexr.rs
@@ -152,10 +152,10 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for OpenExrDecoder<R> {
     }
 
     // reads with or without alpha, depending on `self.alpha_preference` and `self.alpha_present_in_file`
-    fn read_image_with_progress<F: Fn(Progress)>(
+    fn read_image_with_progress<F: FnMut(Progress)>(
         self,
         unaligned_bytes: &mut [u8],
-        progress_callback: F,
+        mut progress_callback: F,
     ) -> ImageResult<()> {
         let blocks_in_header = self.selected_exr_header().chunk_count as u64;
         let channel_count = self.color_type().channel_count() as usize;
--- a/src/image.rs
+++ b/src/image.rs
@@ -688,10 +688,10 @@ pub trait ImageDecoder<'a>: Sized {
 
     /// Same as `read_image` but periodically calls the provided callback to give updates on loading
     /// progress.
-    fn read_image_with_progress<F: Fn(Progress)>(
+    fn read_image_with_progress<F: FnMut(Progress)>(
         self,
         buf: &mut [u8],
-        progress_callback: F,
+        mut progress_callback: F,
     ) -> ImageResult<()> {
         assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
 


HeroicKatora:
The only case where this is _less_ general is when `F` is recursive, right? We never required `Send` or `Sync` so it's not possible to call the callback from a parallel decoder in the current state anyways.

src/codecs/bmp/decoder.rs
```diff
@@ -1526,7 +1526,7 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for BmpDecoder<R> {
 }
 
 impl<'a, R: 'a + Read + Seek> ImageDecoderRect<'a> for BmpDecoder<R> {
-    fn read_rect_with_progress<F: Fn(Progress)>(
+    fn read_rect_with_progress<F: FnMut(Progress)>(
```
HeroicKatora, line 1529: We can also offer compatibility shims, since the callback is not a guarantee

```rust
let fn_progress = RefCell::new(progress);
read_rect_with_progress(self.0, buf, width, height, color_type, |progress| {
    if let Ok(fn_progress) = fn_progress.try_borrow_mut() {
        fn_progress(progress)
    }
});
```

src/image.rs
```diff
@@ -463,6 +463,11 @@ where
     let mut tmp_scanline = None;
 
     {
+        progress_callback(Progress {
+            current: 0,
+            total: total_bytes,
+        });
+
```
johannesvollmer, line 470: I'm not sure about the recursion you mentioned. But, as the non-mutable callback could be referenced multiple times, but the mutable callback can only be referenced once, it can be used in less situations now. That's why I had to change the order of this call, as it was captured later on, which would attempt to reference the mutable callback twice.

src/codecs/bmp/decoder.rs
johannesvollmer: You mean as an additional helper that could be called by the broken downstream implementations to fix the progress? Yes, this could be a nice help for those implementations.

HeroicKatora: Or as a backwards compatible implementation. Since `FnMut` is a super-trait of `Fn` it is certainly more generic for the caller, which would be encouraged for any new cases of `_with_progress` functions (#1695) in particular. But yet I'm not convinced that this warrants making a breaking change here. There's no logic bug or panicking in the above implementation—indeed the same could be written by any caller. This means that the generalization is merely a convenience, and breaking change for convenience isn't worth it.

So I would propose rather than _changing_ the existing cases we 
* add a new function `with_progress_mut`
* _maybe_ deprecate the existing function, although not immediately
* write it down for `0.25`  where we'll align the interfaces
* for _new_ interfaces we use `with_progress(impl FnMut)`

johannesvollmer: Is there a branch for 0.25?

fintelia: Not yet, but we could make one

johannesvollmer:
There are some errors in code that I didn't touch:
```
warning: unused `Result` that must be used
   --> src/image.rs:869:9
    |
869 |         self.write_image(buf, width, height, color_type);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: this `Result` may be an `Err` variant, which should be handled
```

I cannot find this line of code in my fork. What's happening :o

HeroicKatora:
Well, not sure if that was the right button. My brain assumed it was the `Approve for Run` for reasons of placement… It seems like it wasn't completely wrong though :joy:

src/codecs/openexr.rs
```diff
@@ -152,10 +152,10 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for OpenExrDecoder<R> {
     }
 
     // reads with or without alpha, depending on `self.alpha_preference` and `self.alpha_present_in_file`
-    fn read_image_with_progress<F: Fn(Progress)>(
+    fn read_image_with_progress<F: FnMut(Progress)>(
```
johannesvollmer, line 155: So, @HeroicKatora will there be any problem with recursion of the progress callback? If yes, is it even relevant?

HeroicKatora: No, I don't believe so. Only a really small fraction will be recursive, if any, and our default implementation handles this gracefully.

HeroicKatora (approved):

fintelia:
Is there a particular rush to expose this feature? It feels like a lot of churn to add _read_image_with_progress_mut_ in `0.24.x`, deprecate it in `0.25.x`, and remove in `0.26.x`. So if it is possible to work around this for now (which I think it might be using `Rc<RefCell>` ?), I'd prefer to do that and just change the type signature of _read_image_with_progress_ for `0.25.x`.

johannesvollmer:
No, we could also note this down for the next major release

HeroicKatora:
I've create a new branch `next-version-0.25` for all breaking changes. Feel free to amend everything into the breaking change then.

johannesvollmer:
todo: 
-  just change the type signature of read_image_with_progress for 0.25.x instead of adding a backwards-compatible variant

johannesvollmer:
@fintelia  is the current stance to remove progress completely?

fintelia:
That's the plan at the moment, yes.

It turns out that having a method taking a callback function prevents the `ImageDecoder` trait from being object safe, which is a goal for the 0.25.x release. This however leaves open the possibility of adding an additional non-object safe trait for incremental decoding (like how we also have `ImageDecoderRect`), which we'd be able to do in a backwards compatible way

johannesvollmer:
Maybe removing it now allows someone in the future to add it back in without breaking object safety. Let's remove it for now.

PR #1695, opened 2022-04-02
Image Encoder: write with progress

johannesvollmer:
see #1488 

Note: For backwards compatibility, unlike the reader, the writer adds a default implementation for `write_with_progress`, instead of implementing `write` in terms of `write_with_progress`. This way, we can add progress tracking to the individual image writers one by one.

### DIFF
--- a/src/codecs/farbfeld.rs
+++ b/src/codecs/farbfeld.rs
@@ -259,12 +259,12 @@ impl<W: Write> FarbfeldEncoder<W> {
 
     /// Encodes the image ```data``` (native endian)
     /// that has dimensions ```width``` and ```height```
-    pub fn encode(self, data: &[u8], width: u32, height: u32) -> ImageResult<()> {
-        self.encode_impl(data, width, height)?;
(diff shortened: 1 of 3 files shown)


src/codecs/farbfeld.rs
```diff
@@ -259,12 +259,12 @@ impl<W: Write> FarbfeldEncoder<W> {
 
     /// Encodes the image ```data``` (native endian)
     /// that has dimensions ```width``` and ```height```
-    pub fn encode(self, data: &[u8], width: u32, height: u32) -> ImageResult<()> {
-        self.encode_impl(data, width, height)?;
+    pub fn encode<F: FnMut(Progress)>(self, data: &[u8], width: u32, height: u32, progress: F) -> ImageResult<()> {
```
fintelia, line 262: Isn't this an API breaking change?

johannesvollmer: Sure seems like it. We could rewrite this part such that only new methods are added and none of the existing are modified

Shnatsel:
The `read_with_progress` family of methods was removed in 0.25.x, so this is no longer relevant.

PR #1696, opened 2022-04-11
Fix gif OOM when decoding

5225225:


5225225:
Hmm,, I'm getting a failure

```
check_references tests/reference/jpg/progressive/3.jpg.79e90ab0.png
thread 'check_references' panicked at 'The decoded image's hash does not match (expected = 79e90ab0, actual = c959cb95).', tests/reference_images.rs:279:13


failures:
    check_references

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 16.36s

error: test failed, to rerun pass '--test reference_images'
cargo test  63.29s user 0.27s system 261% cpu 24.314 total
```

Looks to be the `jpeg/platform-independent` feature.

src/codecs/gif.rs
```diff
@@ -154,7 +168,17 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for GifDecoder<R> {
...
             // If the frame does not match the logical screen, read into an extra buffer
             // and 'insert' the frame from left/top to logical screen width/height.
-            let mut frame_buffer = vec![0; self.reader.buffer_size()];
+            let buffer_size = self.reader.buffer_size();
+
+            self.limits.reserve_usize(buffer_size)?;
+
+            let mut frame_buffer = vec![0; buffer_size];
+
+            self.limits.free(
```
5225225, line 177: I think the reserve/free API is good, but I needed to `free` earlier than this buffer is actually freed to avoid the limits not actually being reset.

Maybe an alternate API like a `Limits` being an Allocator wrapper would be useful. Not sure! Probably best to just get more examples of code that use the limits API and then see what fits them the best.

src/codecs/gif.rs
```diff
@@ -154,7 +168,17 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for GifDecoder<R> {
...
+                u64::try_from(buffer_size)
+                    .expect("if buffer_size overflows a usize, we should have returned already"),
```
HeroicKatora, line 179: I don't think this is accurate. [This path](https://github.com/image-rs/image/pull/1696/files#diff-9f2f0567fa73b9d4104f217bfe7c08f7a3eaba4a55e7eba3652e39b041a29d83R136-R138) did not return but `buffer_size` was invalid as `u64`. Should `reserve_usize` return `ImageResult<Option<u64>>`?

src/codecs/gif.rs
HeroicKatora: It's probably correct, an allocator API would be interesting but it would need to interact with all containers. `Vec` and `Box` is potentially a good start but not generic enough. But it's hard to say if it was better to wait on an the allocator interface or to solve it separately.

Given that there is no concurrency in this specific, I would say the order doesn't matter to here. The only effect of free is to allow following `reserve` calls to succeed.

src/codecs/gif.rs
5225225: ah right

hmm..

Could make a `free_usize` which does the u64::try_from().unwrap() but only if a limit is set? And otherwise it's a no-op? Since we should have returned already if a limit is set, but yeah, we might not have if no limit was set.

5225225: Okay, I tried free_usize since it keeps the calling code simpler. `reserve_usize` returning an `Option<u64>` would mean you need to store it and then conditionally `free`. Happy with the current API for now.

HeroicKatora (approved):

PR #1698, opened 2022-04-11
cargo-deny

emilk:
[cargo-deny](https://github.com/EmbarkStudios/cargo-deny) is an amazing tool that protects from:

* duplicated crates (code bloat)
* copy-left licenses in the dependency tree
* RUSTSEC advisories

In this case it discovered the duplication of `jpeg-decoder` and `miniz_oxide`, caused by the `tiff` and `flate2` crates. ~Until new releases of `tiff` and `flate2` are published I opted to ignore these duplicated crates in the `deny.toml` file.~ **EDIT:** new versions of `tiff` and `flate2` have been published, removing the duplication.

Adding cargo-deny to the CI will prevent more dependency duplication, hence this unsolicited PR.

--- 

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -26,7 +26,8 @@ wildcards = "allow" # at least until https://github.com/EmbarkStudios/cargo-deny
 deny = []
 skip = []
 skip-tree = [
-    { name = "criterion" }, # dev-dependnecy
+    { name = "criterion" },  # dev-dependency
+    { name = "quickcheck" }, # dev-dependency
 ]
 
 
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,56 @@
+# https://embarkstudios.github.io/cargo-deny/
+
+targets = [
+    { triple = "aarch64-apple-darwin" },
+    { triple = "aarch64-linux-android" },
+    { triple = "x86_64-apple-darwin" },
+    { triple = "x86_64-pc-windows-msvc" },
+    { triple = "x86_64-unknown-linux-gnu" },
+    { triple = "x86_64-unknown-linux-musl" },
+]
+
+
+[advisories]
+vulnerability = "deny"
+unmaintained = "warn"
+yanked = "deny"
+ignore = []
+
+
+[bans]
+multiple-versions = "deny"
+wildcards = "allow" # at least until https://github.com/EmbarkStudios/cargo-deny/issues/241 is fixed
+deny = []
+skip = [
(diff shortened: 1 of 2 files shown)


fintelia:
How widely used is cargo-deny? The duplicated crate and RUSTSEC detection are both very welcome additions, though I'm slightly less convinced about adopting a policy to ban specific open source licenses.

emilk:
It's hard to gauge how widely used cargo-deny is, but I know a lot of companies use it (and it is backed by one).

[It can be configured extensively](https://embarkstudios.github.io/cargo-deny/index.html) so that that, for instance, the license check is disabled.

fintelia (approved):

deny.toml
```diff
@@ -0,0 +1,37 @@
...
+    { triple = "x86_64-apple-darwin" },
+    { triple = "x86_64-pc-windows-msvc" },
+    { triple = "x86_64-unknown-linux-gnu" },
+    { triple = "x86_64-unknown-linux-musl" },
+]
+
+
+[advisories]
+vulnerability = "deny"
+unmaintained = "warn"
```
HeroicKatora, line 15: Will these show up as a warning on the Action CI step as well?

deny.toml
```diff
@@ -0,0 +1,37 @@
...
+skip-tree = [
+    { name = "criterion" },  # dev-dependency
+    { name = "quickcheck" }, # dev-dependency
+]
+
+
+[licenses]
+unlicensed = "allow"
+allow-osi-fsf-free = "either"
+copyleft = "allow"
```
HeroicKatora, line 37: This is for `MPL-2.0` only, right? I'm not sure if we rather give a specific exception as strong copyleft (GPL) is definitely _not_ intended to be required to build `image` at this moment. The jury is still out on whether building Rust crates compliant with LGPL's linking exception is actually legally tested and possible.

deny.toml
emilk: Yes, unmaintained crates will show up as warnings (but not as errors).
You can also run `cargo deny check` manually

deny.toml
emilk: This allows any and all licenses. If we want to deny GPL we can revert https://github.com/image-rs/image/pull/1698/commits/fec1d0286b3ef73e93253127d9b85e1ec7a979d3 (perhaps in a separate PR?)

emilk:
Thanks for merging this! This will help a lot in keeping my dependency tree nice and tidy, and free of duplicates ❤️

PR #1703, opened 2022-04-16
update openexr 

johannesvollmer:
exr crate license was updated. improves behaviour with cargo-deny.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ rgb = { version = "0.8.25", optional = true }
 mp4parse = { version = "0.12.0", optional = true }
 dav1d = { version = "0.6.0", optional = true }
 dcv-color-primitives = { version = "0.4.0", optional = true }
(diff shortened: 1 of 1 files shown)


fintelia (approved):
Cargo should automatically use newer minor versions when you run `cargo update` or create a new project, so this update isn't required. At the same time it doesn't hurt anything, so :+1:

johannesvollmer:
Even if the exr dependency is specified exactly? I thought this would only work if we use `1.4` instead of `1.4.2` in cargo toml :)

JonasHedEng:
@johannesvollmer It would be the case if it was specified like `=1.4.2` instead of `1.4.2` :)

johannesvollmer:
Ah yes, I see! Thanks :)

johannesvollmer:
Did you check whether cargo-deny accepts the crate now, with the new license?

PR #1710, opened 2022-05-05
Implement TiffReader::set_limits

fintelia:
The tiff crate imposes default limits on decoding, so if the user attempts to override them it is important that we pass that on to the backend. This PR implements `TiffReader::set_limits` to do that.

As an example of things going wrong, if we currently try to load [this TIFF file](https://github.com/image-rs/image/files/8628415/large.zip):
```rust
    #[test]
    fn large_tiff() {
        let mut reader =
            crate::io::Reader::open("/home/jonathan/tmp/large.tiff").unwrap();
        reader.no_limits();
        reader.decode().unwrap();
    }
```
We get this error.
```
---- image::tests::large_tiff stdout ----
thread 'image::tests::large_tiff' panicked at 'called `Result::unwrap()` on an `Err` value: Limits(LimitError { kind: InsufficientMemory })', src/image.rs:1845:25
```
(I'm not adding this as a new testcase because neither having the test include code to extracting from the zip archive nor shipping a ~700KB test image seem like good options)

### DIFF
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -28,7 +28,7 @@ where
 {
     dimensions: (u32, u32),
     color_type: ColorType,
-    inner: tiff::decoder::Decoder<R>,
+    inner: Option<tiff::decoder::Decoder<R>>,
 }
 
 impl<R> TiffDecoder<R>
@@ -74,7 +74,7 @@ where
         Ok(TiffDecoder {
             dimensions,
             color_type,
-            inner,
+            inner: Some(inner),
         })
     }
 }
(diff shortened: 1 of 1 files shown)


src/codecs/tiff.rs
```diff
@@ -173,9 +173,30 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for TiffDecoder<R> {
...
+        let max_alloc = limits.max_alloc.unwrap_or(u64::MAX);
+        let max_intermediate_alloc = max_alloc.saturating_sub(self.total_bytes());
+
+        let mut tiff_limits: tiff::decoder::Limits = Default::default();
+        tiff_limits.decoding_buffer_size =
+            usize::try_from(max_alloc - max_intermediate_alloc).unwrap_or(usize::MAX);
+        tiff_limits.intermediate_buffer_size =
+            usize::try_from(max_intermediate_alloc).unwrap_or(usize::MAX);
+        tiff_limits.ifd_value_size = tiff_limits.intermediate_buffer_size;
+        self.inner = Some(self.inner.take().unwrap().with_limits(tiff_limits));
```
HeroicKatora, line 191: This option-wrapping is avoided if we had a simpler `set_limits`. Do you feel like it's worth it to open an issue on `tiff` upstream? I don't want to delay this PR in any case, we can revert back to a non-option field at a later stage with a bumped dependency.

src/codecs/tiff.rs
```diff
@@ -28,7 +28,7 @@ where
 {
     dimensions: (u32, u32),
     color_type: ColorType,
-    inner: tiff::decoder::Decoder<R>,
+    inner: Option<tiff::decoder::Decoder<R>>,
```
HeroicKatora, line 31: Please add a comment that this is purely so the value can be replaced behind a mutable `&mut self` via `Option::take`.

src/codecs/tiff.rs
fintelia: Not sure it is worth creating an issue now. The tiff crate's interface honestly could use a overhaul overall

HeroicKatora (approved):
Aye on tiff overhaul, also with regards to reading into pre-allocated storage and directly into the DynamicImage as a consequence.

PR #1714, opened 2022-05-22
Optimize Rgba blend

mat-1:
This optimizes the blend method of Rgba<T> by checking for two very common cases. I didn't benchmark this function specifically, but I got my program that makes heavy use of it to be around twice as fast with this change.

Something I'm not sure about is the `other.0[3].to_i32().unwrap() == 255`, since it looks like 255 might not always be the maximum alpha?

---

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -734,7 +734,7 @@ impl<T: Primitive> Blend for Rgba<T> {
         if other.0[3].is_zero() {
             return;
         }
-        if other.0[3].to_i32().unwrap() == 255 {
+        if other.0[3] == T::DEFAULT_MAX_VALUE {
             *self = *other;
             return;
         }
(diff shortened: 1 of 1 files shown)


src/color.rs
```diff
@@ -731,6 +731,14 @@ impl<T: Primitive> Blend for Rgba<T> {
     fn blend(&mut self, other: &Rgba<T>) {
         // http://stackoverflow.com/questions/7438263/alpha-compositing-algorithm-blend-modes#answer-11163848
 
+        if other.0[3].is_zero() {
+            return;
+        }
+        if other.0[3].to_i32().unwrap() == 255 {
```
HeroicKatora, line 737: I believe this comparison should be `other.0[3]` against `T::DEFAULT_MAX_VALUE` as this isn't only being used for `T=u8`.

HeroicKatora: This should work because the impl requires `T: Primitive` which includes `T: PartialOrd+PartialEq`. We might also use a greater-equal comparison instead, but I'm not certain how that actually will work with floating point. Maybe check if the math for floats even works right now?

mat-1: Thanks for the suggestion, I forgot DEFAULT_MAX_VALUE exists 😄

fintelia (approved):

Issue #1715, opened 2022-05-25
Missing big-endian support for exr


botovq:
The image crate 0.24.x fails to compile with default features big-endian platforms. This is unfortunate since various things, like [clamav][1] or [czkawka][2] started depending on the latest image crate, so they need patching to keep working on big-endian architectures.

Should a feature that is enabled by default not work on all platforms?

## Expected

The image crate compiles. This works fine for 0.23.x. It was broken with https://github.com/image-rs/image/pull/1475 that added exr to the default features.

## Actual behaviour

Compilation fails with 

```
error: expected `,`, found `function`
  --> /usr/obj/ports/czkawka-4.1.0/czkawka-4.1.0/modcargo-crates/exr-1.4.2/src/compression/rle.rs:39:29
   |
39 |         unimplemented!(this function probably returns little-endian data);
   |                             ^^^^^^^^ expected `,`

error: could not compile `exr` due to previous error
```
According to https://github.com/johannesvollmer/exrs/issues/155#issuecomment-1119288212 the missing quotes are deliberate to cause  a compile time failure rather than a runtime failure.

## Reproduction steps

Attempt to compile on a big-endian platform such as sparc64 or powerpc.

[1]: https://github.com/Cisco-Talos/clamav/blob/main/libclamav_rust/Cargo.toml#L15
[2]: https://github.com/qarmin/czkawka/blob/master/czkawka_gui/Cargo.toml#L32

johannesvollmer:
Luckily, today I have time to try to add big-endian support. It might take me more than one day though. Until it is there, we have several quick fixes available. For example, disabling the `exr` feature on big endian architectures by default in the image crate, until `exrs` supports it. Until this happens, all users of the image library can also disable the exr feature as a work-around.

johannesvollmer:
> the missing quotes are deliberate to cause a compile time failure rather than a runtime failure

Well, the missing quotes are probably not intentional, I apologize. I was sure that I had inserted intentional compile-time errors in other places in the code, but it seems that there are none. Anyways, I'll try to actually finish the big-endian support today :)

johannesvollmer:
see #1771

PR #1716, opened 2022-05-26
Restrict openexr to little endian platforms

HeroicKatora:
Supposed to fix: #1715 

Not certain if this works correctly, at least it appears to not make a difference for little endian architectures. We would do well to test on a big-endian platform in CI in any case.

cc: @botovq

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,12 @@ rgb = { version = "0.8.25", optional = true }
 mp4parse = { version = "0.12.0", optional = true }
 dav1d = { version = "0.6.0", optional = true }
 dcv-color-primitives = { version = "0.4.0", optional = true }
-exr = { version = "1.4.2", optional = true }
 color_quant = "1.1"
+
+# These appear as an empty dependency on other platforms.
+# See: <https://github.com/rust-lang/cargo/issues/1197#issuecomment-901794879>
+[target.'cfg(target_endian = "little")'.dependencies]
+exr = { version = "1.4.2", optional = true }
 
 [dev-dependencies]
 crc32fast = "1.2.0"
--- a/src/codecs/openexr.rs
+++ b/src/codecs/openexr.rs
@@ -20,8 +20,6 @@
 //!     - meta data is lost
 //!     - dwaa/dwab compressed images not supported yet by the exr library
 //!     - (chroma) subsampling not supported yet by the exr library
-
-extern crate exr;
 use exr::prelude::*;
 
 use crate::error::{DecodingError, EncodingError, ImageFormatHint};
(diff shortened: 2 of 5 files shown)


HeroicKatora:
@johannesvollmer No need to rush anything, but if you have strategies to CI test on big endian then I'd be happy if you could transfer them here.

botovq:
Thanks.

Unfortunately, that doesn't quite work: the default feature list

https://github.com/image-rs/image/blob/0842cce4c743f4c6af4e72dda8b064b0ecd974db/Cargo.toml#L65

enables `feature = openexr`, which attempts to pull in `exr`, which isn't defined on big-endian.

```
error[E0463]: can't find crate for `exr` 314/396: serde, image, tiff, lofty, jpeg-decoder, serd...
  --> /usr/ports/pobj/czkawka-4.1.0/czkawka-4.1.0/modcargo-crates/image-0.24.1/./src/codecs/openexr.rs:24:1
   |
24 | extern crate exr;
   | ^^^^^^^^^^^^^^^^^ can't find crate
```

You can simulate this by flipping the endianness of your `Config.toml` patch:

```
[target.'cfg(target_endian = "big")'.dependencies]
```

I haven't managed to figure out how to enable a feature only for certain target platforms. [This thread](https://github.com/rust-lang/cargo/issues/1197#issuecomment-901794879) seems to suggest that this still isn't possible, but it's long and confusing, so perhaps I'm missing something.


[1]: https://github.com/image-rs/image/blob/0842cce4c743f4c6af4e72dda8b064b0ecd974db/Cargo.toml#L65

HeroicKatora:
Can you try just removing `extern crate exr;`? It shouldn't be required at all.

botovq:
Right. The problem of using an undefined dependency remains:

```
error[E0433]: failed to resolve: use of undeclared crate or module `exr`
  --> /usr/ports/pobj/czkawka-4.1.0/czkawka-4.1.0/modcargo-crates/image-0.24.2/./src/codecs/openexr.rs:24:5
   |
24 | use exr::prelude::*;
   |     ^^^ use of undeclared crate or module `exr`

error[E0433]: failed to resolve: use of undeclared crate or module `exr`
```

HeroicKatora:
Nevermind, I understand what you mean. There are also several other explicit references to the `exr` crate that would be enabled and fail to resolve. I'm looking into a clever way to solve this if we can detect if the actual `exr` feature is available.

HeroicKatora:
@botovq I tried cfg-gating all reference. May have missing some, actually, but could you give it a try?

botovq:
This works. Thank you!

fintelia (approved):
If this has been tested to both work for users of exr on little endian and compiles properly on big endian, then looks good to me!

Issue #1707, opened 2022-04-29
huge_files_return_error fails on 32-bit architectures.


plugwash:
While working on fixing the rust-image package in Debian I discovered that the huge_files_return_error test fails on 32-bit architectures.

https://ci.debian.net/data/autopkgtest/testing/i386/r/rust-image/21005810/log.gz

    ---- codecs::bmp::encoder::tests::huge_files_return_error stdout ----
    thread 'codecs::bmp::encoder::tests::huge_files_return_error' panicked at 'attempt to multiply with overflow', src/codecs/bmp/encoder.rs:299:31
    stack backtrace:
       0: rust_begin_unwind
                 at /usr/src/rustc-1.58.1/library/std/src/panicking.rs:498:5
       1: core::panicking::panic_fmt
                 at /usr/src/rustc-1.58.1/library/core/src/panicking.rs:107:14
       2: core::panicking::panic
                 at /usr/src/rustc-1.58.1/library/core/src/panicking.rs:48:5
       3: image::codecs::bmp::encoder::tests::huge_files_return_error
                 at ./src/codecs/bmp/encoder.rs:299:31
       4: image::codecs::bmp::encoder::tests::huge_files_return_error::{{closure}}
                 at ./src/codecs/bmp/encoder.rs:297:5
       5: core::ops::function::FnOnce::call_once
                 at /usr/src/rustc-1.58.1/library/core/src/ops/function.rs:227:5
       6: core::ops::function::FnOnce::call_once
                 at /usr/src/rustc-1.58.1/library/core/src/ops/function.rs:227:5
    note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

Note: the version I was packaging in Debian was 0.23.14 but the code in question does not seem to have changed since then, so I'm 99% sure this will also affect the current version.

The offending line is.

    let image = vec![0u8; 3 * 40_000 * 40_000];

While it's not well-documented, it looks like this form of the vec! macro takes a usize, hence the overflow.

I believe the correct fix is simply to skip the test on 32-bit architectures. Most if not all 32-bit systems simply do not have enough address space to produce an image that will overflow the limits of the bmp format. A patch doing that can be found at https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/image/debian/patches/avoid-test-overflow-32-bit.patch

fintelia:
That patch looks reasonable to me. Want to create a PR and we can get this fixed?

[...]

PR #1717, opened 2022-05-26
Dont run huge_files_return_error test on 32-bit to avoid overflow

HeroicKatora:
Origin: <https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/image/debian/patches/avoid-test-overflow-32-bit.patch>
Reviewed-By: Andreas Molzer <HeroicKatora@users.noreply.github.com>
Last-Update: 2022-05-26

Closes: #1707, Thanks for the patch.

---

~~I had some trouble with `git am` because apparently it _must_ be ran interactively if it every tries to access gpg, and will fail with an error that indicates it doesn't find committer credentials otherwise. A bit confusing.~~ No, still sometimes works and sometimes doesn't. Im super confused.

### DIFF
--- a/src/codecs/bmp/encoder.rs
+++ b/src/codecs/bmp/encoder.rs
@@ -319,6 +319,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(target_pointer_width = "64")]
     fn huge_files_return_error() {
         let mut encoded_data = Vec::new();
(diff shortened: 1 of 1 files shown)


HeroicKatora:
> fatal: empty ident name (for <>) not allowed


Thanks for not telling me anything, git..

PR #1718, opened 2022-05-26
Buffer based on canvas

197g:
Draft for comments on the approach.

Add a byte-based buffer which is color space aware and tracks the underlying color and texel layout with fields, instead of encoding them into the type. It offers the same raw bytes access as `DynamicImage` but avoids the clunky matching required to access particulars. This drastically improves ergonomics where you'd otherwise need `N×M` _type_ combinations to fully express all color spaces with all pixel layouts. This clunkiness is at least one of the causes of having support for only _very few_ of these combinations in `ColorType` and even `ExtendedColorType`.

The new approach successfullly and succinctly implements n-to-n-conversion of any input combination to any output combination. It can dynamically execute optimized code (e.g skip color conversion) and architecture specific code (e.g. SIMD shuffles) with little effort because those are very local decisions and don't propagate through interfaces by type-level differences.

We don't expose all combinations in `image` (yet) because they aren't equally well implemented, tested and performance is not yet optimized for many of them.


HeroicKatora:
The main *future* use case for the buffer is as a separate decoding and encoding target, such that `CMYK` and `YUV` inputs can be decoded with a delayed conversion. This provides three main advantages:
1. We can represent many new input layouts directly without having to convert them into an rgb-ish representation enforced by `DynamicImage` and its supported variants.
2. The conversion may need to be aware of additional parameters such as ICC-information that is supplied later in decoding (e.g. we may convert directly into a display output color space without ever storing pixels in profile connection space; which is faster and more accurate)
2. Conversion can happen on GPU hardware instead of being forced to execute on a slow CPU. This, again, is more efficient.

fintelia:
I think it would be really cool if we could _replace_ the current enum based implementation of `DynamicImage` with one that wrapped an `image_canvas::Canvas` but was guaranteed to always be in a `ColorType` (either by carrying it as a separate field or deducing like is done in this draft PR).

If we want to support delayed color conversion from a wide range of pixel layouts/color spaces, that sounds like a really great expose functionality that operates on `image_canvas::Canvas`'s directly.

HeroicKatora:
I chose not to replace `DynamicImage` right away because of compatibility, i.e. we obviously can't change the public enum to an opaque struct. Having a compatibility type sounds appealing though—a canvas guaranteed to perform constant-cost conversion to a `DynamicImage`.

> If we want to support delayed color conversion from a wide range of pixel layouts/color spaces, that sounds like a really great expose functionality that operates on image_canvas::Canvas's directly.

Come again, there's a portion missing here?

fintelia:
> I chose not to replace DynamicImage right away because of compatibility, i.e. we obviously can't change the public enum to an opaque struct. Having a compatibility type sounds appealing though—a canvas guaranteed to perform constant-cost conversion to a DynamicImage.

This feels like a big enough feature addition to justify a major version bump, so I'm not too worried about the API breakage aspect. Obviously, we shouldn't go out of our way to break things, but I think the majority of the existing API can be implemented with this alternative approach.  

Plus, there is significant benefit to having a single `DynamicImage` type that is the right choice for almost all uses. Especially if it can have zero cost conversion to/from `image_canvas::Canvas` for its current color type, and optimized color conversions for others. 

> > If we want to support delayed color conversion from a wide range of pixel layouts/color spaces, that sounds like a really great _[edit: **reason to**]_ expose functionality that operates on image_canvas::Canvas's directly.
>
> Come again, there's a portion missing here?

Sorry this was in reference to "The main future use case for the buffer is as a separate decoding and encoding target, such that CMYK and YUV inputs...".

src/buffer/canvas.rs
```diff
@@ -0,0 +1,639 @@
...
+use image_canvas::Canvas as Inner;
+use image_canvas::{color::Color, layout::Block};
+
+/// A byte buffer for various color and layout combinations.
+///
+/// Note that it is possible to convert most [`ImageBuffer`] as well as [`DynamicImage`] instances
+/// into a canvas that represents the same pixels and color. See `From` implementations below.
+pub struct Canvas {
+    inner: Inner,
+}
```
HeroicKatora, line 22: @fintelia Alright with the name? Other ideas that have been floating around in my head:

- `Picture` for a less dumb buffer that also contains the metadata.
- How do we treat the current `ImageBuffer`, I really want to preserve the 'simple' pixel array container because it can be much simpler to work with and requires less unwrapping. For instance, we couldn't offer `rows` on `Canvas`. But the naming is odd now. Maybe `PixelBuffer` is alright?

fintelia:
The main question for me is when should a user prefer `Canvas` over the other existing image representations we support? It is probably worth writing a paragraph explaining that as part of the type's rustdoc documentation, and then including a short bit in the top-level crate documentation as well. One possible answer is that Canvas is better for writing image processing operations that are generic over pixel types.

A follow question would be whether there's any cases currently filled by `DynamicImage` that `Canvas` isn't suited for?

src/buffer.rs
```diff
@@ -1,4 +1,19 @@
...
+//! * The generic [`ImageBuffer`] struct, a simplified pixel container with a standard matrix
+//!   layout of uniform channels. The strength is convenient access for modifications with a direct
+//!   mapping and access to the underlying data.
+//! * The [`DynamicImage`] enumeration builds on `ImageBuffer` to provide a standard selection of
+//!   basic channel representations, and conventionally in sRGB color space. This is usually enough
+//!   to represent any source data without loosing much precision. This makes is suitable for
+//!   basic, but generic, image operations.
+//! * The [`Canvas`] struct as general texel container of general layouts as a buffer. The strength
+//!   is the generality (and extensibility) that allows its use for interchange of image data and
+//!   reduced processing cost if the convenience above is not required.
```
HeroicKatora, line 13: Is this somewhat clear motivation?

fintelia: Sorry for the delay responding. That explains what sorts of images the different types can store, but doesn't really think through what the user can _do_ with them. My rough thinking:

* `ImageBuffer` provides both a collection of image processing methods (resize, crop, etc.) as well as ergonomic pixel getters and setters to implement your own image processing. The serious drawback is that the color type is determined at compile time. One consequence is that saving works fine but loading files as ImageBuffers isn't directly supported.
* `DynamicImage` supports even more built in image processing as well as loading/saving a whole range of file formats. Conversion between color types is supported. The enum is non-exhaustive but users can implement their own algorithms by matching on the current variants with special case logic for each one.
* `Canvas` holds and converts between a broader range of color types. No other image processing algorithms are provided and users who want to write their own must... rely on raw access to aligned byte slices? Cast them to FlatSamples first? That's perhaps a bit unfair of a characterization since the type could still be useful as a common "interchange" representation for images data in Rust. However, I suspect it would have a better chance of catching on if users didn't have to immediately convert the Canvas into something else to operate on it.

HeroicKatora: > No other image processing algorithms are provided and users who want to write their own must... rely on raw access to aligned byte slices?

Precisely. The best reason to use the buffer is uploading it to a GPU texture as directly as possible; for which only the layout and raw bytes are necessary and sufficient, and all further CPU processing is less precise and more costly than necessary. (And I stopped expanding the interface, before anyone could get the idea of making this PR contingent on a corresponding `wgpu` GPU framework).

The lack of direct operations is by design. Precisely because we may not have interpreted the correct color information (color space, or ICC, or otherwise) yet, just the need to store some numeric texel matrix anyways. There's just nothing sensible to do with the matrix entries, at that point.  It should get support for Yuv layouts and CMYK rather soon to fix the tragedy of wrongly applying incorrect ICC tables while loading jpeg, WebP, png and tiff, as we are doing currently... But that's for another PR in my opinion.

197g:
This would need to be heavily rebased on the newest version of `image-canvas` as well as the metadata. I'm closing it to reopen a more recent design. The discussion was highly useful to sharpen the use-case of `DynamicImage` that made the color-aware design possible.

Issue #1712, opened 2022-05-09
Cannot decode a specific WebP image


svercl:
This happens when decoding the WebP image from https://httpbin.org/image/webp

## Expected

It decodes the image successfully

## Actual behaviour

`Format error decoding WebP: Invalid Chunk header: [0x92, 0x9F, 0x00, 0x9D]`

## Reproduction steps

Try loading the image from memory like so:
```
image::load_from_memory_with_format(include_bytes!("PATH_TO_WEBP_FROM_HTTPBIN"), image::ImageFormat::WebP)
    .expect("loading webp");
```

ava57r:
Hello.
I have a problem with `webp`.

```log
thread '<unnamed>' panicked at 'source slice length (738720) does not match destination slice length (915840)', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.24.2/./src/codecs/webp/extended.rs:458:21
```

mockfox:
Attempting to open the image that @bradms linked results in this error (0.24.2):
`thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Decoding(DecodingError { format: Exact(WebP), underlying: Some(ChunkHeaderInvalid([146, 159, 0, 157])) })', src/main.rs:2:30`

ava57r:
I have the image
[test-1.dat.zip](https://github.com/image-rs/image/files/8698438/test-1.dat.zip)
 
and code 

```rust
use std::{vec, io::{Read, Cursor}, fs::File};

fn main() {
    let args  = std::env::args().collect::<Vec<_>>();
    if args.len() == 1 {
        println!("file path unrepresented");
        return;
    }
    let file_path = &args[1];
    let mut rdr = File::open(file_path).unwrap();
    let mut raw= vec![];
    rdr.read_to_end(&mut raw).unwrap();

    let image_opt = image::io::Reader::new(Cursor::new(&raw))
            .with_guessed_format()
            .map_err(image::ImageError::IoError)
            .and_then(|rdr| {
                rdr.format()
                    .map(|format| rdr.decode().map(|img| (format, img)))
                    .transpose()
            }).unwrap();
    
    if let Some((format, _img)) = image_opt {
        println!("image format: {:?}", format);
    } else {
        print!("Unknown image format");
    }
    
}
```

output

```log
[...]

PR #1721, opened 2022-06-06
Fixes some lossy extended WebP issues

tristanphease:
There were two errors with the implementation with the extended WebP images that affected lossy(VP8) WebP images without alpha

1. Tried to read the VP8 chunk header twice
2. Tried to fill a rgbaimage buffer with rgbimage size

This PR fixes those issues, so the image in #1712 is decoded successfully.


HeroicKatora (approved):
:+1: with rustfmt fixed

tristanphease:
Oops, sorry.
rustfmt run now.

PR #1735, opened 2022-06-21
fix #1733

### LINKED ISSUE #1733: bmp Decoder RLE bug for "encoded mode"
When decoding bmp files that are encoded with RLE (in my case RLE8), the "Encoded mode" part of RLE compression is not done correctly.
Description of Encoded mode: https://docs.microsoft.com/en-us/windows/win32/gdi/bitmap-compression

## Expected

When reading `RLEInsn::Delta(x_delta, y_delta)` the imaginary cursor painting the image should jump from its current position to a new position, delta pixels away.

This should work `for x_delta` and `y_delta` both be zero or non-zero or any combination of that.

## Actual behaviour

If (and only if) `y_delta` is non-zero, the behaviour is not a "jump of painter position" but something different:
- The current row gets filled with black for `x_delta` many pixels.
- The next `y_delta - 1` rows get filled with black.

This leads to "skips" in the resulting image (see attachment).

As far as I would go the solution could roughly look like this:
1. it should be `0..y_delta`, not `1..y_delta` in the for loop
2. iff there is any y_delta > 0, the current row has to be filled with black completely
3. after doing row skips, return to the x-position you were at before, then add `x_delta` to this position

## Reproduction steps
Read and write the attached image with example code. See, that there are skips (look at the feet - the lowest row is snipped away).

### Example Code:
```rust
use image::GenericImageView;

fn main() {
    // Use the open function to load an image from a Path.
    // `open` returns a `DynamicImage` on success.
    let img = image::open("image.bmp").unwrap();

    // Write the contents of this image to the Writer in PNG format.
    img.save("test.png").unwrap();
}
```

## problematic code that I am talking about here
from codecs\bmp\decoder.rs:1351-1382
```rust
RLEInsn::Delta(x_delta, y_delta) => {
                                if y_delta > 0 {
                                    for n in 1..y_delta {
                                        if let Some(row) = row_iter.next() {
                                            // The msdn site on bitmap compression doesn't specify
                                            // what happens to the values skipped when encountering
                                            // a delta code, however IE and the windows image
                                            // preview seems to replace them with black pixels,
                                            // so we stick to that.
                                            for b in row {
                                                *b = 0;
                                            }
                                        } else {
                                            delta_pixels_left = x_delta;
                                            // We've reached the end of the buffer.
                                            delta_rows_left = y_delta - n;
                                            break 'row_loop;
                                        }
                                    }
                                }

                                for _ in 0..x_delta {
                                    if let Some(pixel) = pixel_iter.next() {
                                        for b in pixel {
                                            *b = 0;
                                        }
                                    } else {
                                        // We can't go any further in this row.
                                        break;
                                    }
                                }
                            }
``
[image.zip](https://github.com/image-rs/image/files/8941744/image.zip)
`

silentbat:
fixes #1733 

but to be totally honest with you - I do not really understand it. How does skipping x_delta pixels help to get to the correct x position? I thought we need the _current_ x position + x_delta for this??

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -1350,6 +1350,9 @@ impl<R: Read + Seek> BmpDecoder<R> {
                             }
                             RLEInsn::Delta(x_delta, y_delta) => {
                                 if y_delta > 0 {
+                                    // we skip at least a line -> clear the rest of the current line
+                                    blank_bytes(pixel_iter);
+
                                     for n in 1..y_delta {
                                         if let Some(row) = row_iter.next() {
                                             // The msdn site on bitmap compression doesn't specify
@@ -1361,12 +1364,13 @@ impl<R: Read + Seek> BmpDecoder<R> {
                                                 *b = 0;
                                             }
                                         } else {
-                                            delta_pixels_left = x_delta;
                                             // We've reached the end of the buffer.
                                             delta_rows_left = y_delta - n;
                                             break 'row_loop;
                                         }
                                     }
+                                    delta_pixels_left = x_delta;
+                                    break 'rle_loop;
                                 }
 
                                 for _ in 0..x_delta {


silentbat:
That is totally fine, I have something in the working - but currently I am only able to do it working directly on the pixelbuffer, not using the nice iterator patterns.

The "fix" above is just fixing it for some specific cases sadly.
(It fixes the combination of "newline" + "delta jump", but not when there are multiple jumps needed because the real jump would be more far away than the possible 255 pixels)

fintelia:
I'm going to close this for now. It seems too likely to cause regressions given that we don't have a good understanding of what's going on. If someone wants to figure out how the code is supposed to work, please feel free to open a new PR

Issue #1752, opened 2022-07-05
Lossy image loading


demurgos:
I would like to be able to load malformed images, such as truncated files. In particular, once the pixel buffer is allocated, return it even on error so it's possible to recover at least part of the data.

My specific use case for this functionality is to generate thumbnails for a set of files where some of them may be partially invalid (e.g. truncated).

This is more generally applicable to implement more reliable image handling. Instead of failing the whole image load, missing pixels can use a default value. The idea is to handle images similarly to browsers: they are able to display images even if they have some issues (as long as the header is valid).

## Draft

The main change would be in the `image::decoder_to_vec` function:

```diff
 /// Reads all of the bytes of a decoder into a Vec<T>. No particular alignment
 /// of the output buffer is guaranteed.
 ///
 /// Panics if there isn't enough memory to decode the image.
 ///
+/// If an error occurs after the pixel buffer was created, it is returned. Note
+/// however that it may be incomplete. Pixels that were not read use their
+/// default "zero" value (the memory is initialized and safe to access).
-pub(crate) fn decoder_to_vec<'a, T>(decoder: impl ImageDecoder<'a>) -> ImageResult<Vec<T>>
+pub(crate) fn decoder_to_vec<'a, T>(decoder: impl ImageDecoder<'a>) -> Result<Vec<T>, (Option<Vec<T>>, ImageError)> 
 where
     T: crate::traits::Primitive + bytemuck::Pod,
 {
     let total_bytes = usize::try_from(decoder.total_bytes());
     if total_bytes.is_err() || total_bytes.unwrap() > isize::max_value() as usize {
-        return Err(ImageError::Limits(LimitError::from_kind(
-            LimitErrorKind::InsufficientMemory,
-        )));
+        return Err((None, ImageError::Limits(LimitError::from_kind(
+            LimitErrorKind::InsufficientMemory,
+        ))));
     }

     let mut buf = vec![num_traits::Zero::zero(); total_bytes.unwrap() / std::mem::size_of::<T>()];
-    decoder.read_image(bytemuck::cast_slice_mut(buf.as_mut_slice()))?;
-    Ok(buf)
+    match decoder.read_image(bytemuck::cast_slice_mut(buf.as_mut_slice())) {
+        Ok(()) => Ok(buf),
+        Err(e) => Err((Some(buf), e))
+    }
 }
```

On error, return the pixel buffer anyway and let the consumer decide what do.

[...]

PR #1753, opened 2022-07-05
feature: Implement `load_lossy`

demurgos:
Add support for lossy image loading.

The new `load_lossy` function adds error recovery.

If loading reaches the step where it allocates the pixel buffer, then it always succeeds: any error following this step will result in a partially filled buffer where the missing pixels will use their default "zero" value.

This allows for example to read truncated files or other minor issues.

- Closes image-rs/image#1752

---

### DIFF
--- a/src/io/reader.rs
+++ b/src/io/reader.rs
@@ -6,6 +6,7 @@ use crate::dynimage::DynamicImage;
 use crate::error::{ImageFormatHint, UnsupportedError, UnsupportedErrorKind};
 use crate::image::ImageFormat;
 use crate::{ImageError, ImageResult};
+use crate::io::free_functions::LoadErrorHandling;
 
 use super::free_functions;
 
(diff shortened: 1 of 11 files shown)


src/io/reader.rs
```diff
@@ -225,7 +226,7 @@ impl<R: BufRead + Seek> Reader<R> {
     /// If no format was determined, returns an `ImageError::Unsupported`.
     pub fn decode(mut self) -> ImageResult<DynamicImage> {
         let format = self.require_format()?;
-        free_functions::load_inner(self.inner, self.limits, format)
+        free_functions::load_inner(self.inner, self.limits, format, LoadErrorHandling::Strict)
     }
```
HeroicKatora, line 230: Preferrably you'd add `fn decode_lossy` for `Reader` instead of `image::load_lossy`. Use of the module functions is somewhat discouraged due to lack of limit handling and imprecise error reporting.

HeroicKatora (commented):
Succinct and little surface changes, very neat. I'd be in favor. The only concern that comes to mind would be that completely discarding the error information may be suboptimal for few cases. But we can take a look later, I'd generally treat them as warnings and defer this to warning reports in either case.

fintelia:
This PR assumes that `ImageDecoder::read_image` already implements the desired behavior of reading as much pixel data as possible and leaving any corrupt pixels in the output buffer untouched. However, this isn't actually specified anywhere. And AFAIK for many decoders this isn't actually true at all: they try to bail with an error as soon as they discover that any pixel data is corrupt, and some even use the output buffer as scratch memory while they do decoding!

I like the overall API/feature, but I wouldn't be comfortable merging without both (a) changing the spec for `ImageDecoder` to provide the needed guarantees, and (b) someone going through and auditing all our existing decoders to make sure they either already implement lossy decoding or updating them so they do.

PR #1757, opened 2022-07-08
Fix arithmetic overflow in TIFF error reporting

### LINKED ISSUE #1750: Debug-mode arithmetic overflow in TIFF decoding (found by fuzzer)
This happens in image from git on commit 04052e64c9a94606efc8bd3d87d5f3e0f566774e

Decoding the attached file results in an arithmetic overflow, causing a panic in debug mode:

`thread '<unnamed>' panicked at 'attempt to multiply with overflow', image/./src/codecs/tiff.rs:72:51`

## Reproduction steps

`cargo +nightly fuzz run fuzzer_script_tiff path/to/file`

The sample triggering it: [overflow.tiff.gz](https://github.com/image-rs/image/files/9046405/overflow.tiff.gz)

fintelia:
Fixes #1750

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,6 +22,7 @@ jobs:
       env:
         FEATURES: ${{ matrix.features }}
     - name: test
+      if: ${{ matrix.rust == 'stable' || matrix.rust == 'beta' || matrix.rust == 'nightly' }}
       run: >
         cargo test -v --no-default-features --features "$FEATURES" &&
         cargo doc -v --no-default-features --features "$FEATURES"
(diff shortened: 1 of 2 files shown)


fintelia:
Apparently CI is failing because plotters (which we depend on transitively via our dev-dependency criterion) [increased their MSRV in a minor version](https://github.com/plotters-rs/plotters/pull/362).... which they did so they could use the latest version of this crate :upside_down_face:

HeroicKatora:
Aren't circular dependencies a charm? Well, then let's exercise our new policy of indicating MSRV in Cargo.toml's field and bump it in a minor version as well. Should be less controversial with this new feature, and we'll want to do so anyways. It's just a little unfortunate that we can't indicate an intrinsic msrv (based on what's used by *this* specific crate) and a different, more recent extrinsic one (based on dependencies). That will have to wait on cargo-resolver integration of the field, I suppose.

HeroicKatora (approved):

Issue #1738, opened 2022-06-30
Resize check


rcastill:
I would like to be able to resize to a target resolution and don't need to check if the target resolution is already the image's resolution.

My specific use case for this functionality is that I found this to be a huge bottleneck in a real time object detection application. Resize itself took some time, so I pre-processed the image so it came already resized from another thread. Performance improved, but profiling still showed that the guts of resize where being called (vertical/horizontal_sample). Checking if it was already resized, hugely improved my application's performance.

This is more generally applicable to people that could face the same problem. Maybe it's a better default to check first.

## Draft

It's a simple check using GenericImageView::dimensions().

---

I also understand it may be a design decision to leave this check to the user. But I wanted to raise the question.

fintelia:
This seems reasonable to me. The cost of checking should be a tiny fraction of the total time spent resizing an image

Shnatsel:
Done in #1817

PR #1763, opened 2022-07-24
Check if resizing to the same size

drewcassidy:
Addresses #1738 by checking if resizing is to the same size the image already is. 

there's a lot more optimizations that could be made here, such as detecting integer scaling or a Pillow-style ["reduce gap"](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.resize) method (maybe as an additional filter mode, since it only makes sense for bilinear). However from #1718 it looks like you may be planning a refactor of the backing data structures for pixels?

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -888,6 +888,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "png")]
     fn test_resize_same_size() {
         use std::path::Path;
         let img = crate::open(&Path::new("./examples/fractal.png")).unwrap();
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -7,7 +7,7 @@ use std::f32;
 
 use num_traits::{NumCast, ToPrimitive, Zero};
 
-use crate::image::{GenericImageView, GenericImage};
+use crate::image::{GenericImage, GenericImageView};
 use crate::traits::{Enlargeable, Pixel, Primitive};
 use crate::utils::clamp;
 use crate::{ImageBuffer, Rgba32FImage};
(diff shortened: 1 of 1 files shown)


src/imageops/sample.rs
```diff
@@ -750,16 +750,28 @@ where
...
 where
-    I::Pixel: 'static,
-    <I::Pixel as Pixel>::Subpixel: 'static,
+    I: GenericImageView<Pixel = P>,
+    P: Pixel<Subpixel = S> + 'static,
+    S: Primitive + 'static,
 {
+    // check if the new dimensions are the same as the old. if they are, make a copy instead of resampling
+    if (nwidth, nheight) == image.dimensions() {
+        let mut tmp = ImageBuffer::<P, Vec<S>>::new(image.width(), image.height());
```
fintelia, line 766: I think the compiler should infer the type here so you should be able to do `ImageBuffer::new(...)`

src/imageops/sample.rs
```diff
@@ -750,16 +750,28 @@ where
...
+        match tmp.copy_from(image, 0, 0) {
+            Ok(_) => {
+                return tmp;
+            }
+            Err(_) => {} // something has gone wrong doing a direct copy, continue with normal path
+        };
```
fintelia, line 772: GenericImage::copy_from is documented to only return an error if the destination isn't large enough to hold the image. Could you replace this with:
```suggestion
        tmp.copy_from(image, 0, 0).unwrap();
        return tmp;
```

src/imageops/sample.rs
```diff
@@ -875,6 +887,29 @@ mod tests {
         b.bytes = 800 * 800 * 3 + 200 * 200 * 3;
     }
 
+    #[test]
+    #[cfg(feature = "png")]
+    fn test_resize_same_size() {
+        use std::path::Path;
+        let img = crate::open(&Path::new("./examples/fractal.png")).unwrap();
+        let resize = img.resize(img.width(), img.height(), FilterType::Triangle);
+        assert!(img.pixels().eq(resize.pixels()))
```
fintelia, line 896: This would be a bit cleaner as `assert!(img.pixels() == resize.pixels());` or `assert_eq!(img.pixels(), resize.pixels());`

drewcassidy: It actually cant be that, iterators don't implement cmp::Eq, whereas eq is [part of the iterator trait](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.eq)

src/imageops/sample.rs
```diff
@@ -750,16 +750,24 @@ where
 /// Resize the supplied image to the specified dimensions.
 /// ```nwidth``` and ```nheight``` are the new dimensions.
 /// ```filter``` is the sampling filter to use.
-pub fn resize<I: GenericImageView>(
+pub fn resize<I, P, S>(
```
fintelia, line 753: These other generic arguments aren't needed now either, right?

drewcassidy: I guess not

PR #1771, opened 2022-08-19
Add big-endian CI

johannesvollmer:
- [x] taken from the working big-endian CI in [exrs](https://github.com/johannesvollmer/exrs/blob/master/.github/workflows/rust.yml)
- [x] added all currently supported image formats to list of features tested in CI
- [x] use new version and re-enabled `exrs` for big endian architecture (tested on big-endian locally, compiles and passes)
- [x] openexr tests currently relies on hdr feature being turned on, but the ci features matrix only tests isolated features
- [ ] decide whether to also run doc tests & clippy and any of the other workflows for big endian systems

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,11 +6,12 @@ on:
     branches: [ master, next ]
 jobs:
   build:
+    name: Build and Run the features using a combination of features and rust toolchains
     runs-on: ubuntu-latest
     strategy:
       matrix:
         rust: ["1.51.0", stable, beta, nightly]
-        features: [ gif, jpeg, png, pnm, tga, dds, tiff, webp, hdr, farbfeld, '' ]
+        features: [ gif, jpeg, png, tiff, ico, pnm, tga, webp, bmp, hdr, dxt, dds, farbfeld, openexr, jpeg_rayon, '' ]
     steps:
     - uses: actions/checkout@v2
     - uses: actions-rs/toolchain@v1
@@ -28,6 +29,43 @@ jobs:
         cargo doc -v --no-default-features --features "$FEATURES"
       env:
         FEATURES: ${{ matrix.features }}
+
+  test_defaults_big_endian:
+    # github actions does not support big endian systems directly, but it does support QEMU.
+    # so we install qemu, then build and run the tests in an emulated mips system.
+    # note: you can also use this approach to test for big endian locally.
+    runs-on: ubuntu-20.04
+    name: Run tests on an emulated big endian mips system using stable toolchain (exluding doc tests)
+    strategy:
+      matrix:
+        features: [ gif, jpeg, png, tiff, ico, pnm, tga, webp, bmp, hdr, dxt, dds, farbfeld, openexr, jpeg_rayon, '' ]
+
+    # we are using the cross project for cross compilation to mips:
+    # https://github.com/cross-rs/cross
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install or use cached cross-rs/cross
+        uses: baptiste0928/cargo-install@v1
+        with:
+          crate: cross
+
+      - name: Start Docker (required for cross-rs)
+        run: sudo systemctl start docker
+
+      - name: Cross-Compile project to mips-unknown-linux-gnu for ${{ matrix.features }}
+        run: |
+          cross build --target=mips-unknown-linux-gnu --verbose -v --no-default-features --features "$FEATURES"
+        env:
+          FEATURES: ${{ matrix.features }}
+
+      # https://github.com/cross-rs/cross#supported-targets
+      - name: Cross-Run Tests in mips-unknown-linux-gnu using Qemu for ${{ matrix.features }}
+        run: |
+          cross test --target mips-unknown-linux-gnu --verbose -v --no-default-features --features "$FEATURES"
+        env:
+          FEATURES: ${{ matrix.features }}
+
   test_defaults:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


johannesvollmer:
some of these matrix variations get cancelled even though there is no timeout specified in the workflow. why else would github cancel these runs?

johannesvollmer:
the normal openexr test suite fails on CI but runs locally. CI error is:
```
Compiling nanorand v0.7.0
...
error: edition 2021 is unstable and only available with -Z unstable-options.
error: could not compile `nanorand`
```

.github/workflows/rust.yml
```diff
@@ -6,11 +6,12 @@ on:
     branches: [ master, next ]
 jobs:
   build:
+    name: Run tests and doctests
     runs-on: ubuntu-latest
     strategy:
       matrix:
         rust: ["1.51.0", stable, beta, nightly]
```
HeroicKatora, line 13: Feel free to bump MSRV here, adjust `Cargo.toml:package.rust-version` accordingly.

johannesvollmer:
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

The number of concurrent jobs you can run in your account depends on your GitHub plan, as indicated in the following table. If exceeded, any additional jobs are queued. Free Plan has 20 jobs total, macos 5, in parallel.

Job matrix - A job matrix can generate a maximum of 256 jobs per workflow run. This limit applies to both GitHub-hosted and self-hosted runners.

Workflow run queue - No more than 500 workflow runs can be queued in a 10 second interval per repository. If a workflow run reaches this limit, the workflow run is terminated and fails to complete.

.github/workflows/rust.yml
johannesvollmer: My understanding is that, as this should be as low as possible for maximum compatibility, we should aim to never increase it. It is a minimum version. I suggest we only bump that version when someone wants to use a special new feature which is not supported in `1.51.0`. What do you think?

fintelia: We also have to increase our minimum supported version when our dependencies do. From the error posted above, it looks like nanorand now requires Rust 2021 which corresponds to Rust 1.56.0 or later

fintelia:
> some of these matrix variations get cancelled even though there is no timeout specified in the workflow. why else would github cancel these runs?

GitHub cancels all the jobs in a matrix (or maybe all the jobs in an entire CI run?) when one of them fails.

johannesvollmer:
Ooooh I see. Didn't use matrices in workflows before. This would explain it.

johannesvollmer:
The exr library does not support nightly currently. So we either need to exclude it or make it work in the dependencies from `exr` to `flume` to `nanorand`

.github/workflows/rust.yml
HeroicKatora: Your understanding does agree with mine, but I've come to accept that _in practice_ we'll want to bump this version quicker and with our dependencies at least until `cargo` ships with an msrv-aware resolver algorithm. It also seems more brittle to claim compatibility with an untested Rust version, or to implement such resolver by-hand somehow. And, lastly, doing this is a useful testcase for the practability of our policy amyways. In particular if the risk of actual breakage is low, as it is here, because we do it by choice and without inherent need.

fintelia:
Running `cargo +nightly-2022-08-20 test --features exr` works fine for me. Are you sure nightly is the problem? The CI results seem to indicate that Rust 1.51.0 (i.e. our current minimum supported rust version) is the one that's failing

johannesvollmer:
Oh that would be neat. I guess I missed something in the error log

.github/workflows/rust.yml
johannesvollmer: Oh, I didn't see the `Cargo.toml` actually has a different version than the CI right now, of course now this makes sense!

johannesvollmer: `cargo-msrv` says
```
Check for toolchain '1.57.0-x86_64-pc-windows-msvc' succeeded                                                                                                     2
Check for toolchain '1.56.1-x86_64-pc-windows-msvc' succeeded                                                                                                     7
   Finished The MSRV is: 1.56.1

```

johannesvollmer:
Great News! The big-endian test suite seems to work now. It has just uncovered some untested behaviour. The first one shown is in the gif format, a test failed:

```
thread 'main' panicked at 'attempt to multiply with overflow', /cargo/registry/src/github.com-1ecc6299db9ec823/gif-0.11.4/src/reader/mod.rs:404:9


failures:
    bad_gif_oom

```

johannesvollmer:
The question is, would it be fine to merge this PR and fix the uncovered issues in separate pull requests? I won't know if I'm able to resolve all the actual issues in the image formats

HeroicKatora:
That would work for me :+1: We should temporarily mark the runs in big endian as [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) though.

johannesvollmer:
tests run with continue-on-error now, but the build does not, so we should wait to verify that the build succeeds for every feature on the big endian target. If some fail, we should also add continue-on-error for the build step.

johannesvollmer:
surprisingly, gif test passes now on big endian. i couldn't find any PR fixing it. was it fixed or is something wrong with the pipeline?

HeroicKatora (approved):
The [`gif`](https://github.com/image-rs/image/runs/8051828074?check_suite_focus=true) suite is still failing on big endian, but no missing checkmark to indicate that. Still, seems okay.
> 
failures:
    bad_gif_oom

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s

Issue #1775, opened 2022-08-22
Panic when an animated webp image is loaded


PhilipK:
This happens in BoilR
https://github.com/PhilipK/BoilR/issues/203

Whenever an animated webp image is loaded it will lead to a panic:

## Expected

Expected the image to load a static image, or just give an Result::Error instead of panic

## Actual behaviour

A panic is thrown

## Reproduction steps

Download an animated image, eaxmple: https://cdn2.steamgriddb.com/file/sgdb-cdn/logo/42cd4e870371f116e5695912004e1b5a.webp
Load the image with the following code:

```rust
  #[test]
    pub fn image_decode_webp_animated() {
        let path = std::path::Path::new("src/testdata/hollow.webp");
        if let Ok(reader) = image::io::Reader::open(path){
            //This should give an error but instead panics
            if let Ok(image)  = reader.decode(){
                println!("It worked");
            }
        }
    }
```
The code will now panic

Denzo77:
I have a similar but subtley different issue (running v0.24.4, but it looks like this still applies to the patch above). The reproduction steps are the same, but I think it's specific to animated webps with no transparency such as this example:
https://www.tutorialexample.com/wp-content/uploads/2020/09/gif-test-2.webp

I get the following panic:
```
thread 'main' panicked at 'source slice length (750000) does not match destination slice length (1000000)', .cargo/registry/src/github.com-1ecc6299db9ec823/image-0.24.4/./src/codecs/webp/extended.rs:462:21
```

The slice lengths looked suspiciously like it's trying to copy RGB values into an RGBA buffer and panicking because the lengths don't match. Inspecting the header with webpinfo, the VP8X header has Alpha set to true but the actual frames all have it set to false.
```
$ webpinfo resources/gif-test-2.webp
resources/gif-test-2.webp
RIFF HEADER:
  File size: 813092
Chunk VP8X at offset     12, length     18
  ICCP: 0
  Alpha: 1   <=== TRUE HERE
  EXIF: 0
  XMP: 0
  Animation: 1
  Canvas size 500 x 500
Chunk ANIM at offset     30, length     14
  Background color:(ARGB) 00 18 18 18
  Loop count      : 0
Chunk ANMF at offset     44, length  15832
  Offset_X: 0
  Offset_Y: 0
  Width: 500
  Height: 500
  Duration: 0
  Dispose: 0
  Blend: 1
Chunk VP8  at offset     68, length  15808
  Width: 500
  Height: 500
[...]

PR #1776, opened 2022-08-22
Error when decoding unsupported animated webp, instead of panic

PhilipK:
This solves a panic that would happen if a frame was not the same height and width as the image info

This solves the following issue:
https://github.com/image-rs/image/issues/1775

### DIFF
--- a/src/codecs/webp/decoder.rs
+++ b/src/codecs/webp/decoder.rs
@@ -322,6 +322,17 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for WebPDecoder<R> {
                 lossless_frame.fill_rgba(buf);
             }
             WebPImage::Extended(extended) => {
+                if !extended.is_supported_format() {
+                    use crate::error::*;
+                    let format_hint = ImageFormatHint::Exact(ImageFormat::WebP);
+                    return ImageResult::Err(
+                        ImageError::Unsupported(
+                        UnsupportedError::from_format_and_kind(
+                            format_hint.clone(),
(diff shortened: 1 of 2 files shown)


PhilipK:
The panic will only happen if the frame dimension size is different from the info dimension size.

HeroicKatora (commented):
This looks on first sight like a fairly light stop-gap, not a fix of the underlying cause. Somewhere we've made an improper assumption about width or height and this should be fixed. This treats the symptom of one example, at least a regression test and test suite should be added.

It is apparently allowed for subimages to have differing dimensions; are all image sequences with differences in size now handled properly? I'd propose rather clearly not.

Also, similar to `gif` the canvas is by definition the 'size' of the image. We shouldn't return an incorrect result just because our code doesn't handle the proper result.

PhilipK:
I agree. This is a patch to stop a panic, not a rewrite of the underlying problem. 

As a user of a rust library a panic is just about the worst case scenario, because there is really nothing you can do to handle it or prevent it. 

Another suggestion here is to gracefully return an error instead of a panic. But I opted for the option that will result in a "good enough for now" result, and then we can work on the real solution.

HeroicKatora:
Gracefully returning an error would be a better way towards working to a real solution. Wrong results easily become (improperly) relied upon before they are fixed. The library explicitly has an `Unsupported` error variant for these cases. So changing it to return that instead would be appreciated :+1:

PhilipK:
An error is not returned instead of a panic now. I added a "is_supported" function, but this can ofc be completely removed once the real cause of the problem is solved.

PR #1780, opened 2022-09-01
QOI Support

DigitalJokerMan:
[QOI (Quite OK Image) format](https://qoiformat.org/) is a very niche but interesting image format that competes with PNG. The Rust ecosystem already has crates supporting this format, I decided to use [aldanor's implementation](https://crates.io/crates/qoi) as it proved to be the most effective & straight-forward crate to use. This would add encoding + decoding of the format to mainstream image.


src/codecs/qoi.rs
```diff
@@ -0,0 +1,96 @@
...
+        let data = qoi::encode_to_vec(&buf, width, height)
+            .map_err(|e| ImageError::Encoding(EncodingError::new(ImageFormat::Qoi.into(), e)))?;
+
+        // Write data to buffer
+        {
+            self.writer.write_all(&data[..])?;
+            self.writer.flush()?;
+            Ok(())
+        }
+        .map_err(ImageError::IoError)?;
```
HeroicKatora, line 77: This sequence doesn't look right to me, it's effectively `Ok(()).map_err` which is no-op.

HeroicKatora (commented):
Seems to be the right crate

> Zero unsafe code.
> Roundtrip-tested vs the reference C implementation; fuzz-tested.

src/codecs/qoi.rs
fintelia: Also, with the right Into implementation, we should be able to avoid all the map_err calls and just rely on the `?` operator.

fintelia:
How much uptake has QOI had so far? From a quick Google search, I see a couple news articles about the initial release and a bunch of people excited about writing their own implementations in various languages, but no one actually saying they're using it for anything. I'd feel better if we could point to at least a few folks saying they are actually producing/processing images with the format.

The PR itself looks good to me, although I would like to see a direct implementation of `ImageDecoder::read_image` so we don't end up with an extra copy for that operation (which for a format as fast as QOI is a significant cost!)

DigitalJokerMan:
I have experimented with using QOI as a format for storing sprites in pixel-art based games. It's pretty promising. As for the errors in my code, my apologies. I'm kinda new to this whole "contributing-to-open-source" thing. I'm not sure how I would go to resolve the 2 pointed out issues in the code. 😅

src/codecs/qoi.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+impl<W: Write> ImageEncoder for QoiEncoder<W> {
+    fn write_image(
+        mut self,
+        buf: &[u8],
+        width: u32,
+        height: u32,
+        _color_type: ColorType,
```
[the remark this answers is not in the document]
fintelia: To clarify: the simplest fix here would be to return an error if `color_type != ColorType::Rgb8 && color_type != ColorType::Rgba8`.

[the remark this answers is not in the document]
fintelia: Another required check we'll need is to ensure that `buf.len() == width * height * 3` if the color type is RGB or `width * height * 4` if the color type is RGBA.

src/codecs/qoi.rs
fintelia: Here's an example of the trait impl I was talking about:

fintelia, line 77: The `?` operator does automatic conversion, so you can just do:
```suggestion
        self.writer.write_all(&data[..])?;
        self.writer.flush()?;
```

src/codecs/qoi.rs
```diff
@@ -0,0 +1,96 @@
...
+            self.writer.write_all(&data[..])?;
+            self.writer.flush()?;
+            Ok(())
+        }
+        .map_err(ImageError::IoError)?;
+
+        Ok(())
+    }
+}
+
```
fintelia, line 82: Adding something like this would let you remove a couple of the map_err calls elsewhere in the file:
https://github.com/image-rs/image/blob/7f5291d72704777f1e171b7366c9ae5590613c87/src/codecs/webp/decoder.rs#L57-L61

src/lib.rs
```diff
@@ -246,6 +246,8 @@ pub mod codecs {
     pub mod tiff;
     #[cfg(feature = "webp")]
     pub mod webp;
+    #[cfg(feature = "qoi")]
+    pub mod qoi;
```
fintelia, line 250: rustfmt wants these kept in alphabetical order. (That's why our CI is reporting "Some checks were not successful")

HeroicKatora:
> I'm not sure how I would go to resolve the 2 pointed out issues in the code.

Process wise, you make another commit to your branch wherein the issues are addressed. You push the commit and Github will automatically update the pull request status accordingly. Then we give it another look. No worries about any errors, we do code review and CI because everyone makes those :grinning:

mattfbacon:
Has this lost steam?

Issue #1778, opened 2022-08-29
Panic while allocating big vector in fuzzer_script_exr.rs


anfedotoff:
Hi!
We were doing some fuzzing with our tool Sydr and libFuzzer.  We  caught a  panic [here](https://github.com/image-rs/image/blob/master/fuzz/fuzzers/fuzzer_script_exr.rs#L25), trying to allocate a huge vector.

## Expected
I think, if we failed to allocate this we should continue fuzzing. Using [try_reserve](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.try_reserve) doesn't help because, we catch sanitizers error:  `AddressSanitizer: requested allocation size 0xffffffffffffffff`

But maybe it is a bug in parsing image? `decoder.total_bytes()` returns a strange value, it's larger than input file:
[crash-sydr_21c91c10c5bd6a850f36ff6683ce16b4e9089c35_int_overflow_4.txt](https://github.com/image-rs/image/files/9446294/crash-sydr_21c91c10c5bd6a850f36ff6683ce16b4e9089c35_int_overflow_4.txt)

Also, I found some work around [here](https://github.com/image-rs/image/blob/6e0cd31a5287dd589d2e78ae33c1f720c77a6863/fuzz/fuzzers/fuzzer_script_tga.rs#L12).

## Actual behaviour
```
Running: /fuzz/exr-out/crashes/crash-sydr_21c91c10c5bd6a850f36ff6683ce16b4e9089c35_int_overflow_4
thread '<unnamed>' panicked at 'capacity overflow', library/alloc/src/raw_vec.rs:518:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==13== ERROR: libFuzzer: deadly signal
    #0 0x55ee086dba61 in __sanitizer_print_stack_trace /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_stack.cpp:87:3
    #1 0x55ee0a25bd80 in fuzzer::PrintStackTrace() /root/.cargo/registry/src/github.com-1ecc6299db9ec823/libfuzzer-sys-0.3.5/libfuzzer/FuzzerUtil.cpp:210:38
    #2 0x55ee0a26b67a in fuzzer::Fuzzer::CrashCallback() /root/.cargo/registry/src/github.com-1ecc6299db9ec823/libfuzzer-sys-0.3.5/libfuzzer/FuzzerLoop.cpp:233:18
[...]

PR #1781, opened 2022-09-01
Fix unbound allocation in fuzzer_script_exr

anfedotoff:
Fix for #1778

### DIFF
--- a/fuzz/fuzzers/fuzzer_script_exr.rs
+++ b/fuzz/fuzzers/fuzzer_script_exr.rs
@@ -1,28 +1,41 @@
 #![no_main]
-#[macro_use] extern crate libfuzzer_sys;
+#[macro_use]
+extern crate libfuzzer_sys;
 extern crate image;
 
-use std::io::Cursor;
-use image::ImageResult;
 use image::codecs::openexr::*;
-use std::io::Seek;
-use std::convert::TryFrom;
+use image::ColorType;
 use image::ImageDecoder;
 use image::ImageEncoder;
-use image::ColorType;
-use std::io::Write;
+use image::ImageResult;
+use std::convert::TryFrom;
+use std::io::Cursor;
 use std::io::Read;
+use std::io::Seek;
+use std::io::SeekFrom;
+use std::io::Write;
 
 // "just dont panic"
 fn roundtrip(bytes: &[u8]) -> ImageResult<()> {
-
     /// Read the file from the specified path into an `Rgba32FImage`.
     // TODO this method should probably already exist in the main image crate
-    fn read_as_rgba_byte_image(read: impl Read + Seek) -> ImageResult<(u32,u32,Vec<u8>)> {
+    fn read_as_rgba_byte_image(read: &mut (impl Read + Seek)) -> ImageResult<(u32, u32, Vec<u8>)> {
+        let current_pos = read.stream_position()?;
+        let total_size = read.seek(SeekFrom::End(0))?;
+        let _ = read.seek(SeekFrom::Start(current_pos))?;
+
         let decoder = OpenExrDecoder::with_alpha_preference(read, Some(true))?;
         let (width, height) = decoder.dimensions();
 
-        let mut buffer = vec![0; usize::try_from(decoder.total_bytes()).unwrap()];
+        let decoded_size = usize::try_from(decoder.total_bytes()).unwrap();
+        if total_size - current_pos < decoded_size as u64 {
+            return Err(image::ImageError::Limits(
+                image::error::LimitError::from_kind(
+                    image::error::LimitErrorKind::InsufficientMemory,
+                ),
+            ));
+        }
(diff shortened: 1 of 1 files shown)


fuzz/fuzzers/fuzzer_script_exr.rs
```diff
@@ -1,28 +1,41 @@
...
+        let current_pos = read.stream_position()?;
+        let total_size = read.seek(SeekFrom::End(0))?;
+        let _ = read.seek(SeekFrom::Start(current_pos))?;
+
         let decoder = OpenExrDecoder::with_alpha_preference(read, Some(true))?;
         let (width, height) = decoder.dimensions();
 
-        let mut buffer = vec![0; usize::try_from(decoder.total_bytes()).unwrap()];
+        let decoded_size = usize::try_from(decoder.total_bytes()).unwrap();
+        if total_size - current_pos < decoded_size as u64 {
```
fintelia, line 31: I don't think this is the logic we want. The EXR format supports compression, so the decoded size is _usually_ going to be larger than the file size. Instead, for fuzzing purposes we could just impose a maximum size of say 256 MB. 

(As a smaller point, we should probably check decoded_size before calling usize::try_from(...).unwrap() so that we don't fail there on 32-bit systems)

HeroicKatora: If we upper bound the decoded size against anything, then comparing against the limits configured for the reader would also make sense (i.e. `Limits::max_alloc). We *are* doing another immediate allocation afterall.

HeroicKatora: If it's unclear, I'm suggesting that properly supporting `<OpenExrDecoder as ImageDecoder>::set_limits` and then using that function to configure the limit would be a much more thorough fix. That way, we get better feedback if all allocations are accounted for or if the fuzzer manages to find additional ones inside the decoder.

anfedotoff: > I don't think this is the logic we want. The EXR format supports compression, so the decoded size is _usually_ going to be larger than the file size. Instead, for fuzzing purposes we could just impose a maximum size of say 256 MB.
> 
> (As a smaller point, we should probably check decoded_size before calling usize::try_from(...).unwrap() so that we don't fail there on 32-bit systems)

Thanks, I didn't know that there is a compression:). I changed fix on what you offer.

anfedotoff: > If it's unclear, I'm suggesting that properly supporting `<OpenExrDecoder as ImageDecoder>::set_limits` and then using that function to configure the limit would be a much more thorough fix. That way, we get better feedback if all allocations are accounted for or if the fuzzer manages to find additional ones inside the decoder.

Sounds pretty good, I need a little bit more time to look at code, maybe find some examples for this

anfedotoff: I've provided an implementation for `OpenExrDecoder::set_limits` function. In this function I check some common cases and a case with `total_bytes()`. Maybe we need to save `Limits` in `Decoder`, and check them sometimes. But I don't familiar with exr decoding for now:). What do I think?

src/codecs/openexr.rs
```diff
@@ -135,6 +135,23 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for OpenExrDecoder<R> {
         }
     }
 
+    fn set_limits(&mut self, limits: crate::io::Limits) -> ImageResult<()> {
```
fintelia, line 138: This can be left with the [default implementation](https://github.com/image-rs/image/blob/master/src/image.rs). We intentionally don't count the total bytes required for the output buffer towards the max_alloc limit because that is the caller's responsibility.

fuzz/fuzzers/fuzzer_script_exr.rs
```diff
@@ -1,45 +1,56 @@
...
 
-        Ok((width, height, buffer))
+            Ok((width, height, buffer))
+        } else {
+            Err(image::ImageError::Limits(
+                image::error::LimitError::from_kind(
+                    image::error::LimitErrorKind::InsufficientMemory,
+                ),
+            ))
+        }
```
fintelia, line 40: This could be something like:
```rust
match use::try_from(decoder.total_bytes()) {
    Ok(decoded_size) if decoded_size <= 256 * 1024 * 1024 => {
        ...
    }
    _ => return Err(image::ImageError::Limits(
        image::error::LimitError::from_kind(
            image::error::LimitErrorKind::InsufficientMemory,
        ),
    ))
};
```

fuzz/fuzzers/fuzzer_script_exr.rs
fintelia: @HeroicKatora the unbounded allocation that's being fixed here is actually in the fuzz harness itself, which set_limits isn't designed to address

fintelia (approved):

Issue #582, opened 2016-10-25
[Suggestion] WebP Encoding Support


liftoff:
The crate currently supports WebP decoding but it would be nice if it also supported encoding as well.  I'd definitely take advantage of it in one of my forthcoming projects.

liftoff:
For reference purposes (in case anyone wants to know how to encode WebP in the mean time), here's how to encode an `ImageBuffer` to WebP via libwebp/libc:

```rust
extern crate libc;
use libc::{c_int, c_float, size_t};
use std::io::{Write, BufWriter};
use std::fs::File;
use std::path::Path;

#[link(name = "webp")]
extern {
// Here's the C lib's function def:
//     size_t WebPEncodeRGB(const uint8_t* rgb, int width, int height, int stride, float quality_factor, uint8_t** output);
//     size_t WebPEncodeLosslessRGB(const uint8_t* rgb, int width, int height, int stride, uint8_t** output);

    fn WebPEncodeRGB(rgb: *const u8, width: c_int, height: c_int, stride: c_int, quality_factor: c_float, output: *mut *mut u8) -> size_t;
    fn WebPEncodeLosslessRGB(rgb: *const u8, width: c_int, height: c_int, stride: c_int, output: *mut *mut u8) -> size_t;
    fn WebPFree(ptr: *const u8);
}

let filetype = "webp";
let ref path_str = format!("/tmp/test.{}", filetype);
let path = &Path::new(path_str);
// Pretend this screenshot() function returns ImageBuffer<Rgb<u8>, Vec<u8>>:
let img = screenshot();
let stride = width * 3;
let lossless = false; // Set to true for lossless (Warning: CPU intensive/slow)
let quality: c_float = 75.0; // Quality level of the WebP image
let mut output: *mut u8 = std::ptr::null_mut();
let raw = img.clone().into_raw();
unsafe {
    let length: usize;
    if lossless {
        length = WebPEncodeLosslessRGB(raw.as_ptr(), width as c_int, height as c_int, stride as c_int, &mut output);
    } else {
        length = WebPEncodeRGB(raw.as_ptr(), width as c_int, height as c_int, stride as c_int, quality, &mut output);
    }
    let mut fout = BufWriter::new(File::create(&path).unwrap());
    let slice = std::slice::from_raw_parts(output, length);
    let _ = fout.write_all(slice).unwrap();
    let _ = fout.flush();
    let _ = WebPFree(output); // IMPORTANT: Make sure we free that memory
}
```

[...]

PR #1784, opened 2022-09-07
Add WebP encoding support

cycraig:
# Description
This PR adds `WebPEncoder`, which wraps the [Simple Encoding API](https://developers.google.com/speed/webp/docs/api#simple_encoding_api) from the native libwebp library using the [webp](https://crates.io/crates/webp) crate.

The encoding functionality is gated behind a new `webp-encoder` feature, which is not enabled by default and completely separate from the `webp` feature. This follows the convention of the current `avif`/`avif-encoder`/`avif-decoder` features which are separate due to the native depedency.

With `webp-encoder` enabled, the following just works now:

```rust
let im = image::open("Lenna.png").unwrap();
im.save("Lenna.webp").unwrap();
```

The above defaults to a lossy encoding value of 80. The value was chosen since `image` defaults to 75 for JPEG encoding, and [this blog post](https://www.industrialempathy.com/posts/avif-webp-quality-settings/) suggests an equivalent quality level for WebP would be around 77, hence rounded up to 80. On the other hand, the official [cwebp encoder utility ](https://developers.google.com/speed/webp/docs/cwebp) uses a default quality of 75. Both options still output smaller sizes than JPEG.

`WebPEncoder` can also be used directly, to set the desired quality:

```rust
let file = File::create("Lenna.webp").unwrap();
let quality = WebPQuality::lossless(); // or WebPQuality::lossy(70) etc.
let encoder = WebPEncoder::new_with_quality(file, quality);
encoder.write_image(im.as_bytes(), im.width(), im.height(), im.color()).unwrap();
```

## Motivation

WebP encoding support has been a long-requested feature for `image` (#582), and it would be good to provide canonical support for it.

It's been 6 years since the original issue with no sign of a pure Rust implementation for WebP encoding (it's a lot of work), so using bindings to libwebp seems like the only viable approach for the foreseeable future. The [Simple Encoding API](https://developers.google.com/speed/webp/docs/api#simple_encoding_api) is sufficient for basic use-cases (no animation etc.) and is unlikely to change, so this shouldn't require much maintenance.

The [webp](https://crates.io/crates/webp) crate (which uses [libwebp-sys](https://crates.io/crates/libwebp-sys)) was chosen because it's maintained and "just works", in that it builds and runs on every target and architecture I've tested on without any problems or additional dependencies required, unlike Avif.

Even though formats like Avif are intended to supplant WebP, they still lack the complete browser support that WebP has, making it preferable for a while longer.

## Relevant issues

This fixes the following issue: #582.

## Type of change

- Non-breaking addition.

## How the change has been tested

- Added new unit tests: `cargo test --no-default-features --features webp-encoder`
- Built and tested on Ubuntu, MacOS, Windows: https://github.com/cycraig/image/actions/runs/3010465918
- Statically cross-compiled to musl: `cargo build --release --target=x86_64-unknown-linux-musl --no-default-features --features webp-encoder` (requires musl-gcc).
- Used `im.save` to convert images on my website from JPEG to WebP successfully.

## License

Note also the licenses of the new dependencies:
- [webp](https://crates.io/crates/webp) crate is dual licensed under MIT/Apache-2.0.
- [libwebp-sys](https://crates.io/crates/libwebp-sys) crate uses the MIT license.
- The native [libwebp](https://chromium.googlesource.com/chromium/src/third_party/+/master/libwebp/LICENSE) library from Google (compiled via `libwebp-sys`) uses a modified BSD 3-clause license. It's still permissive, just with a non-endorsement clause. Again, this only affects developers that ship code with the non-default `webp-encoder` feature enabled, and it's mostly compatible as far as I know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ All image processing functions provided operate on types that implement the `Gen
 | BMP    | Yes | Rgb8, Rgba8, Gray8, GrayA8 |
 | ICO    | Yes | Yes |
 | TIFF   | Baseline(no fax support) + LZW + PackBits | Rgb8, Rgba8, Gray8 |
-| WebP   | Yes | No |
+| WebP   | Yes | Rgb8, Rgba8 |
 | AVIF   | Only 8-bit | Lossy |
 | PNM    | PBM, PGM, PPM, standard PAM | Yes |
 | DDS    | DXT1, DXT3, DXT5 | No |
--- a/README.md
+++ b/README.md
@@ -26,13 +26,16 @@ All image processing functions provided operate on types that implement the `Gen
 | BMP    | Yes | Rgb8, Rgba8, Gray8, GrayA8 |
 | ICO    | Yes | Yes |
 | TIFF   | Baseline(no fax support) + LZW + PackBits | Rgb8, Rgba8, Gray8 |
-| WebP   | Yes | Rgb8, Rgba8 |
-| AVIF   | Only 8-bit | Lossy |
+| WebP   | Yes | Rgb8, Rgba8 \* |
+| AVIF   | Only 8-bit \*\* | Lossy |
 | PNM    | PBM, PGM, PPM, standard PAM | Yes |
 | DDS    | DXT1, DXT3, DXT5 | No |
 | TGA    | Yes | Rgb8, Rgba8, Bgr8, Bgra8, Gray8, GrayA8 |
 | OpenEXR  | Rgb32F, Rgba32F (no dwa compression) | Rgb32F, Rgba32F (no dwa compression) |
 | farbfeld | Yes | Yes |
+
+- \* Requires the `webp-encoder` feature, uses the libwebp C library.
+- \*\* Requires the `avif-decoder` feature, uses the libdav1d C library.
 
 ### The [`ImageDecoder`](https://docs.rs/image/*/image/trait.ImageDecoder.html) and [`ImageDecoderRect`](https://docs.rs/image/*/image/trait.ImageDecoderRect.html) Traits
 
--- a/src/codecs/webp/encoder.rs
+++ b/src/codecs/webp/encoder.rs
@@ -1,0 +1,238 @@
+//! Encoding of WebP images.
+///
+/// Uses the simple encoding API from the [libwebp] library.
+///
+/// [libwebp]: https://developers.google.com/speed/webp/docs/api#simple_encoding_api
+use std::io::Write;
+
+use libwebp::{Encoder, PixelLayout, WebPMemory};
+
+use crate::error::{
+    EncodingError, ParameterError, ParameterErrorKind, UnsupportedError, UnsupportedErrorKind,
+};
+use crate::{ColorType, ImageEncoder, ImageError, ImageFormat, ImageResult};
+
+/// WebP Encoder.
+pub struct WebPEncoder<W> {
+    inner: W,
+    quality: WebPQuality,
+}
+
+/// WebP encoder quality.
+#[derive(Debug, Copy, Clone)]
+pub struct WebPQuality(Quality);
+
+#[derive(Debug, Copy, Clone)]
+enum Quality {
+    Lossless,
+    Lossy(u8),
+}
+
+impl WebPQuality {
+    /// Minimum lossy quality value (0).
+    pub const MIN: u8 = 0;
+    /// Maximum lossy quality value (100).
+    pub const MAX: u8 = 100;
+    /// Default lossy quality (80), providing a balance of quality and file size.
+    pub const DEFAULT: u8 = 80;
+
+    /// Lossless encoding.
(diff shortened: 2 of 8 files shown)


cycraig:
Rebased the changes after #1787 was merged, the CI should pass now.

src/codecs/webp/encoder.rs
```diff
@@ -0,0 +1,238 @@
...
+        let encoded: WebPMemory = match self.quality.0 {
+            Quality::Lossless => encoder.encode_lossless(),
+            Quality::Lossy(quality) => encoder.encode(quality as f32),
+        };
+
+        // The simple encoding API in libwebp does not return errors.
+        if encoded.is_empty() {
+            return Err(ImageError::Encoding(EncodingError::new(
+                ImageFormat::WebP.into(),
+                "encoding failed, output empty",
```
HeroicKatora, line 114: Oh wow, I thought there was at least some error reporting but [no](https://github.com/webmproject/libwebp/blob/main/src/webp/encode.h#L43-L45). Okay, fine.

HeroicKatora (approved):
:+1: with nits addressed (some just warrant short discussion). Reasoning: for accepting: The native dependency isn't required by default, and we keep it an implementaiton detail in terms of API. So yeah, C dependency is okayish and as you say, most practical.

src/codecs/webp/encoder.rs
```diff
@@ -0,0 +1,238 @@
...
+    Lossy(u8),
+}
+
+impl WebPQuality {
+    /// Minimum lossy quality value (0).
+    pub const MIN: u8 = 0;
+    /// Maximum lossy quality value (100).
+    pub const MAX: u8 = 100;
+    /// Default lossy quality (80), providing a balance of quality and file size.
+    pub const DEFAULT: u8 = 80;
```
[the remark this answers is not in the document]
cycraig: Not really, I just thought the constants might be nice-to-have, e.g. I imagined setting slider values in a GUI.

Only disadvantage of `MIN/MAX/DEFAULT` being `Self` is that there is currently no way to access the inner `u8` value from a lossy `WebPQuality` to get the constants, if they are needed for whatever reason (they're not particularly interesting or likely to change between versions either way).

Happy to change it if you still think it's better?

src/codecs/webp/encoder.rs
```diff
@@ -0,0 +1,238 @@
...
+                        ImageFormat::WebP.into(),
+                        UnsupportedErrorKind::Color(color.into()),
+                    ),
+                ))
+            }
+        };
+
+        // Validate dimensions upfront to avoid panics.
+        let expected_len = stride * (width * height) as u64;
+        if expected_len > data.len() as u64 {
```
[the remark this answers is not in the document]
cycraig: Thanks for the catch! `fits` is far more convenient. I didn't realise while cleaning up that the whole point I cast it to `u64` was lost...

src/codecs/webp/encoder.rs
cycraig: So the "advanced WebP encoding API" does return errors, which the next release of the `webp` crate looks like it will include. When/if it's released, the implementation here can be swapped out internally to return errors (not that any are expected unless more configuration options are exposed from `image`) with no breaking changes.

src/codecs/webp/encoder.rs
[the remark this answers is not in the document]
HeroicKatora: Not necessary to change. Exposing the bounds for the constructor to enable some instrumentation is decent reasoning.

README.md
```diff
@@ -26,7 +26,7 @@ All image processing functions provided operate on types that implement the `Gen
 | BMP    | Yes | Rgb8, Rgba8, Gray8, GrayA8 |
 | ICO    | Yes | Yes |
 | TIFF   | Baseline(no fax support) + LZW + PackBits | Rgb8, Rgba8, Gray8 |
-| WebP   | Yes | No |
+| WebP   | Yes | Rgb8, Rgba8 |
```
fintelia, line 29: Might be worth adding asterisks to denote the cases that require a C dependency that isn't enabled by default

fintelia (approved):

README.md
cycraig: Good point, is this okay?

https://github.com/cycraig/image/blob/master/README.md

| Format | Decoding | Encoding |
| ------ | -------- | -------- |
| WebP   | Yes | Rgb8, Rgba8 \* |
| AVIF   | Only 8-bit \** | Lossy |

- \* requires the `webp-encoder` feature, uses the libwebp C library.
- \** requires the `avif-decoder` feature, uses the libdav1d C library.

djc:
Friendly ping, I think this is ready to go?

djc:
Thanks! Is there any known ETA for getting this out to crates.io?

HeroicKatora:
@djc  The evening of saturday, ping me if not. There's another release outstanding too

PR #1790, opened 2022-09-17
fix: implement size_hint for ImageBuffer iterators

### LINKED ISSUE #1789: Implementation of `size_hint` for `ImageBuffer` iterators should return the same as the `ExactSizeIterator::len` impl.
The [docs for `std::iter::ExactSizeIterator`](https://doc.rust-lang.org/nightly/std/iter/trait.ExactSizeIterator.html) state that

> When implementing an ExactSizeIterator, you must also implement [Iterator](https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html). When doing so, the implementation of [Iterator::size_hint](https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html#method.size_hint) must return the exact size of the iterator.

however the current implementation of `Iterator` for the `ImageBuffer` iterators do not have such behaviour and use the default implementation that returns `(0, None)` even though these iterators do implement `std::iter::ExactSizeIterator`.

## Expected

Every iterator that implements `std::iter::ExactSizeIterator` should also return the exact size in the `size_hint` function.

## Actual behaviour

Implementations currently return the default `(0, None)`.

## Reproduction steps

The following test fails:

```rust
let image = RgbImage::from_raw(1, 1 vec![0; 3]).unwrap();
let iter = image.pixels();
let exact_len = ExactSizeIterator::len(&iter);
assert_eq!(iter.size_hint(), (exact_len, Some(exact_len)));
```

GabrielDertoni:
fixes #1789

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -33,6 +33,12 @@ where
     fn next(&mut self) -> Option<&'a P> {
         self.chunks.next().map(|v| <P as Pixel>::from_slice(v))
     }
+
+    #[inline(always)]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = self.len();
+        (len, Some(len))
+    }
 }
 
 impl<'a, P: Pixel + 'a> ExactSizeIterator for Pixels<'a, P>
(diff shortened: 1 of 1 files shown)


HeroicKatora (approved):
Good catch, thanks!

PR #1802, opened 2022-10-10
Add Rgb64FImage and Rgba64FImage

oberien:

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1401,6 +1401,14 @@ pub type Rgb32FImage = ImageBuffer<Rgb<f32>, Vec<f32>>;
 /// where the backing container is a flattened vector of floats.
 pub type Rgba32FImage = ImageBuffer<Rgba<f32>, Vec<f32>>;
 
+/// An image buffer for 64-bit float RGB pixels,
+/// where the backing container is a flattened vector of floats.
+pub type Rgb64FImage = ImageBuffer<Rgb<f64>, Vec<f64>>;
+
+/// An image buffer for 64-bit float RGBA pixels,
+/// where the backing container is a flattened vector of floats.
+pub type Rgba64FImage = ImageBuffer<Rgba<f64>, Vec<f64>>;
+
 #[cfg(test)]
 mod test {
     use super::{GrayImage, ImageBuffer, ImageOutputFormat, RgbImage};
--- a/src/color.rs
+++ b/src/color.rs
@@ -30,6 +30,10 @@ pub enum ColorType {
     Rgb32F,
     /// Pixel is 32-bit float RGBA
     Rgba32F,
+    /// Pixel is 64-bit float RGB
+    Rgb64F,
+    /// Pixel is 64-bit float RGBA
+    Rgba64F,
 }
 
 impl ColorType {
@@ -44,6 +48,8 @@ impl ColorType {
             ColorType::Rgba16 => 8,
             ColorType::Rgb32F => 3 * 4,
             ColorType::Rgba32F => 4 * 4,
+            ColorType::Rgb64F => 3 * 8,
+            ColorType::Rgba64F => 4 * 8,
         }
     }
 
@@ -51,8 +57,8 @@ impl ColorType {
     pub fn has_alpha(self) -> bool {
         use ColorType::*;
         match self {
-            L8 | L16 | Rgb8 | Rgb16 | Rgb32F => false,
-            La8 | Rgba8 | La16 | Rgba16 | Rgba32F => true,
+            L8 | L16 | Rgb8 | Rgb16 | Rgb32F | Rgb64F => false,
+            La8 | Rgba8 | La16 | Rgba16 | Rgba32F | Rgba64F => true,
         }
     }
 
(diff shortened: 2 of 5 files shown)


fintelia:
Could you say a bit about what use cases you have for 64-bit floating point images?

oberien:
I'm fiddling around with astrophotography stacking. I need to perform operations on tens on thousands of images. In that order float imprecision errors start to creep up. To reduce these losses, I'd like to work on `f64` images.

HeroicKatora (approved):
Seems good to me, it doesn't complicate color representation too much and conversion where it was previously possible with `f32` should work. LGTM.

fintelia:
I looked at this a bit more closely. [It is already possible to work with 64-bit ImageBuffers](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5fbc0be578fccd295207ff023af4cd5f) either by writing out the full type name or by defining `Rgb64FImage` and `Rgba64FImage`  aliases yourself.

The main thing this PR adds is a `PixelWithColorType` implementation and `ColorType`/`DynamicImage` variants for dealing with f64 images. Those additions make me somewhat more nervous because they grow the API surface yet might be used rarely/never. (It isn't clear to me they'd even help with your image stacking task to dynamically switch color types at runtime since the types involved should be known statically at compile time)

HeroicKatora:
The traits `FromPrimitive` and `EncodableLayout` aren't implemented on these types, yet. It could just be those for a start and not the `ImageCanvas` extension. [This issue](https://github.com/python-pillow/Pillow/issues/1659) gives other precedent that the pure picture notion is too narrow and types beyond the defaults should be considered.

The concern about `DynamicImage` sounds valid, though. There's no guarantee that any of the operations we already define for it will compute in `f64` themselves which could be confusing. That said, I'd treat these as a normal bug or even lower priority as GPU optimizations are surely worse for accuracy on textures than what we're doing. Being able to represent the values in the buffer is a necessary precursor to correctness here in any case and that at least allows them to be implemented externally. We've had `DynamicImage` as a `non_exhaustive` enumeration nd this sounds as good a time to make use of it as any.

oberien:
Before creating this PR, I did in fact write my own `type Rgb64FImage = ImageBuffer<Rgb<f64>, Vec<f64>>`. However, some utility functions, helper functions and conversions aren't usable from a third party context due to the sealed traits `SealedPixelWithColorType` and `EncodableLayout`. That is why I though it was a good idea to add this directly to the `image` library.

In my particular use-case I was missing image conversion from `RgbImage` to `Rgb64FImage` (loading pngs) and vice versa (saving the result back as a png) and needed to manually implement these conversions.

Regarding adding `DynamicImage::ImageRgb64F`, in addition to the link regarding tiff files linked by @HeroicKatora, I plan on supporting [FITS images](https://heasarc.gsfc.nasa.gov/docs/heasarc/fits_overview.html) in the future, which allow "unsigned 8-bit bytes, 16 and 32-bit signed integers, and 32 and 64-bit single or double precision floating point reals. FITS can also store 16 and 32-bit unsigned integers."

As FITS support is still quite a way in the future, I'd be fine with leaving out `DynamicIage::ImageRgba?64F` for now if that makes the PR easier to merge.

fintelia:
Adding an `FromPrimitive` should unlock `ImageBuffer::convert` which seems worthwhile. The `EncodableLayout` is only useful if we also add a corresponding `PixelWithColorType` implementation (because both are required to unlock `ImageBuffer::{save, save_with_format, write_to}`).

I view `ColorType`/`DynamicImage` as trying to be a small set of color types that are well supported throughout the crate, including both pairwise conversions between all of them and reasonable handling by the all the encoders and decoders. 

More unusual color types can be used by interacting with the format specific crates directly, like how the TIFF crate already supports 64-bit float images. If applicable, we can also have decoders/encoders transparently convert images in unsupported color types into ones we do support. (You could argue this is happening with JPEG when YCbCr colors get mapped to RGB.) 

In the case of Pillow, I believe that [this is the analogous listing of color types](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes).

ripytide:
I think if we decide to add `Rgba64F` image types we might as well just add every fixed intrinsic type for completeness, `{u8,i8,u16,i16,u32,i32,u64,i64,u128,i128,f32,f64}`, that way this type of issue: "my type is missing from `DynamicImage`", cannot come up anymore as every type is already added.

For example:
```rust


pub enum DynamicImage {
     Luma(LumaDynamicImage),
     Rgb(RgbDynamicImage),
     Rgba(RgbaDynamicImage),
     ...,
}

pub enum LumaDynmaicImage {
    U8(Image<Luma<u8>>),
    U16(Image<Luma<u16>>),
    ...,
    F64(Image<Luma<f64>>),
}

...
```

This shouldn't make conversions too much harder as we can just add more `FromPrimitive` trait impls.

I think this would be the "Final Form", per se, of the idea behind the current `DynamicImage`.

mqudsi:
`image` already has a dependency on `num-traits`, perhaps this could be more elegantly solved with generics?

ripytide:
If you don't mind generics then you can use `ImageBuffer` (which is what `DynamicImage` already used in each enum variant), but generics are a leaky abstraction which can cause issues in some use-cases such as de-serialization and some others.

ardijanr:
@ripytide I agree, currently working on some satellite data and its all f32 tiff images which would be luma, but its not supported.

Would be nice to have L32, L64, L32F and L64F

PR #1806, opened 2022-10-12
Fix webp transform panic

### LINKED ISSUE #1740: WebP decoding panics on libwebp test data: index out of bounds in codecs/webp/lossless_transform.rs:45:36
This happens in image from git 04052e64c9a94606efc8bd3d87d5f3e0f566774e (latest commit as of this writing)

## Expected

No panics

## Actual behaviour

A panic:

```
thread '<unnamed>' panicked at 'index out of bounds: the len is 390 but the index is 400', image/./src/codecs/webp/lossless_transform.rs:45:36
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f2d93935ffba3ab9d7ccb5300771a2d29b4c8bf3/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/f2d93935ffba3ab9d7ccb5300771a2d29b4c8bf3/library/core/src/panicking.rs:142:14
   2: core::panicking::panic_bounds_check
             at /rustc/f2d93935ffba3ab9d7ccb5300771a2d29b4c8bf3/library/core/src/panicking.rs:84:5
   3: image::codecs::webp::lossless_transform::TransformType::apply_transform
   4: image::codecs::webp::lossless::LosslessDecoder<R>::decode_frame
   5: image::codecs::webp::decoder::WebPDecoder<R>::read_frame
   6: image::codecs::webp::decoder::WebPDecoder<R>::new
   7: image::io::free_functions::load_decoder
   8: rust_fuzzer_test_input
   9: __rust_try
  10: LLVMFuzzerTestOneInput
  11: _ZN6fuzzer6Fuzzer15ExecuteCallbackEPKhm
  12: _ZN6fuzzer10RunOneTestEPNS_6FuzzerEPKcm
  13: _ZN6fuzzer12FuzzerDriverEPiPPPcPFiPKhmE
  14: main
  15: __libc_start_main
  16: _start
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
## Reproduction steps

I have copied the webp images from https://github.com/dvyukov/go-fuzz-corpus/tree/master/webp/corpus and https://chromium.googlesource.com/webm/libwebp-test-data/  to `fuzz/corpus/fuzzer_script_webp` and ran `cargo +nightly fuzz run fuzzer_script_webp`.

The fuzzer crashed immediately, in the stage of running the provided samples through the decoder.

The file triggering this panic is attached: [panic.webp.gz](https://github.com/image-rs/image/files/9035648/panic.webp.gz)

But I am sure there are many more to come if you seed the fuzzer with this corpus and run it.

fisherdarling:
Fixes #1740 

webp transforms would assume that the given width and height match the image data's dimensions and then panic when applying the transform to the image data. We can't error if `data.len() != width * height` when decoding an image, since apparently that's valid for webp (a test failed if that was checked). Instead, we early return out of the transform if it would cause a panic.


src/codecs/webp/lossless_transform.rs
```diff
@@ -32,6 +32,10 @@ impl TransformType {
                 let width = usize::from(width);
                 let height = usize::from(height);
 
+                if image_data.len() < width * height {
+                    return;
+                }
```
HeroicKatora, line 37: I don't feel comfortable fully ignoring the transform. It's correct that we can't assume equality in general but should we error _here_  specifically instead?

fisherdarling: Makes sense. I didn't want to make the fn return a result since it didn't before, but it does seem like an error to have a transform that would be out-of-bounds.

fisherdarling:
In my first PR you said something about prioritization and I could definitely use some help :) 

Right now I'm just looking through the issue tracker for bugs that seem pretty reproducible / easy to fix. Let me know if there's something else I should be looking at :) I'm also on the gitter (matrix).

HeroicKatora (approved):

PR #1807, opened 2022-10-12
use checked_adds in the bmp writer when calculating `file_size`

### LINKED ISSUE #1708: integer overflow in bmp writer.
While writing up https://github.com/image-rs/image/issues/1707 I noticed that the calculation of image_size in encode_with_palette is protected against integer overflow but the calculation of file_size is not. 

Therefore, if i'm reading the code right, while an image with 2³² or more bytes of image data will produce an error an image with slightly less than 2³² bytes of image data will result in a panic in debug mode or the creation of a bmp file with a bogus file_size value in the header in release mode.

fisherdarling:
Fixes  #1708

From @plugwash in #1708, file_size calculations when writing BMP did not check for overflow. This would cause images with slightly less than 2^32 bytes of data to either panic in debug mode or have a bogus file_size in release mode.

This PR makes the addition checked and errors on overflow.

### DIFF
--- a/src/codecs/bmp/encoder.rs
+++ b/src/codecs/bmp/encoder.rs
@@ -68,7 +68,15 @@ impl<'a, W: Write + 'a> BmpEncoder<'a, W> {
                 ))
             })?;
         let palette_size = palette_color_count * 4; // all palette colors are BGRA
-        let file_size = bmp_header_size + dib_header_size + palette_size + image_size;
+        let file_size = bmp_header_size
+            .checked_add(dib_header_size)
+            .and_then(|v| v.checked_add(palette_size))
+            .and_then(|v| v.checked_add(image_size))
+            .ok_or_else(|| {
+                ImageError::Parameter(ParameterError::from_kind(
+                    ParameterErrorKind::DimensionMismatch,
+                ))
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/encoder.rs
```diff
@@ -68,7 +68,15 @@ impl<'a, W: Write + 'a> BmpEncoder<'a, W> {
...
             })?;
         let palette_size = palette_color_count * 4; // all palette colors are BGRA
-        let file_size = bmp_header_size + dib_header_size + palette_size + image_size;
+        let file_size = bmp_header_size
+            .checked_add(dib_header_size)
+            .and_then(|v| v.checked_add(palette_size))
+            .and_then(|v| v.checked_add(image_size))
+            .ok_or_else(|| {
+                ImageError::Parameter(ParameterError::from_kind(
+                    ParameterErrorKind::DimensionMismatch,
```
fintelia, line 77: Is this the right error to return here?

fisherdarling: Hey! I switched it to be an encoding error with a descriptive message :)

fintelia (approved):

PR #1812, opened 2022-10-25
Add PngDecoder::with_limits

### LINKED ISSUE #1788: InsufficientMemory error while trying to decode
This happens in random pngs, like [this](https://github.com/negezor/image-rs-decode-png-repro/blob/main/broken.png)

## Expected

Decoded png

## Actual behaviour

```
Err(Limits(LimitError { kind: InsufficientMemory }))
```

## Reproduction steps

[Repository](https://github.com/negezor/image-rs-decode-png-repro)

```rs
use std::io::Read;

fn main() {
    let path_to_file = "./broken.png";

    let mut f = std::fs::File::open(path_to_file).expect("no file found");
    let metadata = std::fs::metadata(&path_to_file).expect("unable to read metadata");
    let mut buffer = vec![0; metadata.len() as usize];
    f.read(&mut buffer).expect("buffer overflow");

    let format = image::guess_format(&buffer).unwrap();

    let mut reader = image::io::Reader::with_format(std::io::Cursor::new(buffer), format);

    reader.no_limits();

    let im = reader.decode().unwrap();
}

```

fintelia:
This isn't the most elegant solution given that other decoders use `ImageDecoder::set_limits`, but idiosyncrasies of the PNG format make that interface non-ideal. Specifically, finding out whether the image is animated (contains an acTL chunk) or has a "simple alpha channel" (contains a tRNS chunk) potentially requires scanning a nearly unbounded number of chunks. Worse, we currently don't have a Seek bound on the reader, so we have to _store_ all the chunks we read in case we need them later.

Fixes #1788

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ num-rational = { version = "0.4", default-features = false }
 num-traits = "0.2.0"
 gif = { version = "0.11.1", optional = true }
 jpeg = { package = "jpeg-decoder", version = "0.2.1", default-features = false, optional = true }
-png = { version = "0.17.0", optional = true }
+png = { version = "0.17.6", optional = true }
 scoped_threadpool = { version = "0.1", optional = true }
 tiff = { version = "0.7.1", optional = true }
 ravif = { version = "0.8.0", optional = true }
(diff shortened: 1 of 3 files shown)


fintelia:
@HeroicKatora does this look reasonable to you?

HeroicKatora (approved):
It kind of does, we'll just see about `with_limits` and the need to change it at runtime. This is certainly the more cautious approach.

PR #1817, opened 2022-10-30
Check dimensions of dynamic image before resizing it

Flashover89:
Should fix the issue: https://github.com/image-rs/image/issues/1738

Not super happy with it since I'm cloning self but I didn't want to change the return type to &DynamicImage e.g.
Let me know if you have a better idea on how to make this more efficient :)

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -668,7 +668,7 @@ impl DynamicImage {
             return self.clone();
         }
         let (width2, height2) =
-        resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
+            resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
 
         self.resize_exact(width2, height2, filter)
     }
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -658,22 +658,17 @@ impl DynamicImage {
     }
 
     /// Resize this image using the specified filter algorithm.
+    /// Returns a new image. The image's aspect ratio is preserved.
     /// The image is scaled to the maximum possible size that fits
     /// within the bounds specified by `nwidth` and `nheight`.
-    /// The image's aspect ratio is preserved.
-    /// Returns a new image if the requested image size is different than the already existing one.
-    /// Otherwise the function returns a copy of self.
     pub fn resize(&self, nwidth: u32, nheight: u32, filter: imageops::FilterType) -> DynamicImage {
         if (nwidth, nheight) == self.dimensions() {
-            // Image is already in the desired dimensions --> no need to resize.
             return self.clone();
         }
-        else {
-            let (width2, height2) =
-            resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
+        let (width2, height2) =
+        resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
 
-            return self.resize_exact(width2, height2, filter);
-        }
+        self.resize_exact(width2, height2, filter)
     }
 
     /// Resize this image using the specified filter algorithm.
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -658,14 +658,22 @@ impl DynamicImage {
     }
 
     /// Resize this image using the specified filter algorithm.
-    /// Returns a new image. The image's aspect ratio is preserved.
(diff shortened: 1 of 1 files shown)


src/dynimage.rs
```diff
@@ -658,14 +658,22 @@ impl DynamicImage {
     }
 
     /// Resize this image using the specified filter algorithm.
-    /// Returns a new image. The image's aspect ratio is preserved.
     /// The image is scaled to the maximum possible size that fits
     /// within the bounds specified by `nwidth` and `nheight`.
+    /// The image's aspect ratio is preserved.
+    /// Returns a new image if the requested image size is different than the already existing one.
+    /// Otherwise the function returns a copy of self.
```
fintelia, line 665: Other than performance, the caller shouldn't notice a difference, right? If so, I think we should leave off the discussion of returning a new image vs. a copy. That way users don't have to think about it

src/dynimage.rs
```diff
@@ -658,14 +658,22 @@ impl DynamicImage {
...
     pub fn resize(&self, nwidth: u32, nheight: u32, filter: imageops::FilterType) -> DynamicImage {
-        let (width2, height2) =
+        if (nwidth, nheight) == self.dimensions() {
+            // Image is already in the desired dimensions --> no need to resize.
+            return self.clone();
+        }
+        else {
```
fintelia, line 671: Could you remove this else block and just leave the rest of the method intact? Given that the if statement above contains a return, the behavior should be unchanged

src/dynimage.rs
```diff
@@ -658,14 +658,22 @@ impl DynamicImage {
...
-    /// Returns a new image. The image's aspect ratio is preserved.
     /// The image is scaled to the maximum possible size that fits
     /// within the bounds specified by `nwidth` and `nheight`.
+    /// The image's aspect ratio is preserved.
+    /// Returns a new image if the requested image size is different than the already existing one.
+    /// Otherwise the function returns a copy of self.
     pub fn resize(&self, nwidth: u32, nheight: u32, filter: imageops::FilterType) -> DynamicImage {
-        let (width2, height2) =
+        if (nwidth, nheight) == self.dimensions() {
+            // Image is already in the desired dimensions --> no need to resize.
```
fintelia, line 668: More of a nit, but I'd usually suggest not including comments like this unless if the code is particularly confusing.

fintelia (commented):
I left a couple stylistic suggestions, but the code itself looks good

fintelia:
As a side note, trying to change the return type to `&DynamicImage` would cause borrow checker errors in the more common case where the dimensions do not match. There's even fancier ways to work around that, but they would make things harder for end users. Ultimately, I think keeping the types as they are is the right decision.

Flashover89:
> As a side note, trying to change the return type to `&DynamicImage` would cause borrow checker errors in the more common case where the dimensions do not match. There's even fancier ways to work around that, but they would make things harder for end users. Ultimately, I think keeping the types as they are is the right decision.

Yeah, for the return by reference to work, I would have needed to change the common case when the dimensions do not match. I didn't want to go down that way.
Thanks for your explanations :)

fintelia:
There's now just a small rustfmt issue to resolve:
```diff
/home/runner/.cargo/bin/cargo fmt -- --check
Diff in /home/runner/work/image/image/src/dynimage.rs at line 668:
             return self.clone();
         }
         let (width2, height2) =
-        resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
+            resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
 
         self.resize_exact(width2, height2, filter)
     }
```

Flashover89:
> There's now just a small rustfmt issue to resolve:
> 
> ```diff
> /home/runner/.cargo/bin/cargo fmt -- --check
> Diff in /home/runner/work/image/image/src/dynimage.rs at line 668:
>              return self.clone();
>          }
>          let (width2, height2) =
> -        resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
> +            resize_dimensions(self.width(), self.height(), nwidth, nheight, false);
>  
>          self.resize_exact(width2, height2, filter)
>      }
> ```

Should be fixed now

fintelia:
Thanks!

PR #1824, opened 2022-11-15
QOI Support, attempt 2

mattfbacon:
License consent as usual. I fixed the issues described in the initial PR.

Backreference: #1780


src/codecs/qoi.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+impl<W: Write> ImageEncoder for QoiEncoder<W> {
+    fn write_image(
+        mut self,
+        buf: &[u8],
+        width: u32,
+        height: u32,
+        _color_type: ColorType,
```
HeroicKatora, line 71: This one is still open.

> Ignoring color isn't correct here. Rather, QOI supports only 8-bit rgb-ish color (with optional colorspace). We'll want to either convert here, or raise on error on unsupported formats.


https://github.com/image-rs/image/pull/1780#discussion_r960338581

HeroicKatora (changes requested):
That's my only remaining concern here.

mattfbacon:
Fixed

HeroicKatora (approved):

PR #1837, opened 2022-12-17
Support ICO files with dimensions larger than 256

### LINKED ISSUE #1836: Some modern ico-formats will be rejected
Hello,

when i try to load:
`let img = image::open(&Path::new(&iconPath)).unwrap();`
and the ico is an 512x512 or with an png in the stack
it comes:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Decoding(DecodingError { format: Exact(Ico), underlying: Some(ImageEntryDimensionMismatch { format: Png, entry: (256, 256), image: (512, 512) }) })', src\EXE.rs:40:54
```

the PNG entry and the image entry are correct modern ico-spezifications

thx

fintelia:
[Wikipedia claims](https://en.wikipedia.org/wiki/ICO_(file_format)#Structure_of_image_directory) that a value of 0 for the width or height in an image directory means 256 _or more_. This crate previously assumed it meant exactly 256. If anyone knows of a more definitive spec reference, that would be nice to include.

Fixes #1836

### DIFF
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -231,7 +231,8 @@ impl DirEntry {
     }
 
     fn matches_dimensions(&self, width: u32, height: u32) -> bool {
-        u32::from(self.real_width()) == width && u32::from(self.real_height()) == height
+        u32::from(self.real_width()) == width.min(256)
+            && u32::from(self.real_height()) == height.min(256)
     }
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
I see this has been open for a while with no blockers, and no reviewers assigned. 

Is there anything blocking it, or should it be just merged before conflicts appear?

Issue #1760, opened 2022-07-18
Extremely slow converting to avif


negezor:
This happens in any image, example https://imgur.com/a/5YNiz89

## Expected

Quickly converted image to avif

## Actual behaviour

More than a minute and a half conversion time for an example image. In this case, only one core is used. At the same time, converting to other types of images occurs instantly.

```sh
cargo run --bin avif_repro --release
    Finished release [optimized] target(s) in 0.16s
     Running `target/release/avif_repro`
project master took 1m56s
```

## Reproduction steps

Run with `--release`
```rs
fn main() {
    let im = image::open("fox.png").unwrap();

    im.save("fox.avif").unwrap();
}
```

## Additional info
- OS: `WSL Arch Linux`
- Kernel: `5.10.60.1-microsoft-standard-WSL2`
- CPU: `AMD Ryzen 9 3900X`
- Rust: `1.61.0`
- Crate version: `0.24.2`
- Enabled crate features: `["png", "avif"]`

If you need information that I can provide, just let me know.

fintelia:
I fear this may just be a consequence of avif being slower to encode than other formats. As you note, it should be possible to parallelize the encoding process (PRs welcome!), but beyond that I'm not sure whether there's really much more low hanging fruit to improve things

mothsART:
![compare_image-rs--vs--imagemagick](https://user-images.githubusercontent.com/10601196/185371147-1130850f-8de6-4acd-bfda-d7227c3d456c.png)

I think, encoding algorithm must be compare to imagemagick. More than 400x slower and same checksum at the end.

anarchodin:
> I fear this may just be a consequence of avif being much slower to encode than other formats.

I just ran a test on an image I have. I killed the binary using this crate after more than seven minutes with no results. That same photo recodes in [Squoosh](https://squoosh.app) in 2.7 seconds on the default settings. Clearly not something inherent to AVIF.

Which got me thinking - there's absolutely no access to configuration options with the simple API, so just what _are_ the options being passed to the encoder?

[...]

PR #1838, opened 2022-12-27
AVIF encoder changes.

anarchodin:
As noted in #1760, AVIF encoding is very slow. In looking at _why_, I noticed that the default encoder settings are very different from what the `cavif` CLI provided by upstream uses - in particular, upstream documentation suggests that a quality setting of 100 is not a useful configuration.

The majority of the changes here are to upgrade the `ravif` dependency. The images I was using for testing appear to compress _significantly_ better using the new version.

The project requests this statement to be made:

> I license past and future contributions under the dual MIT/Apache-2.0 license,
> allowing licensees to choose either at their option.

I would like to add that I do not believe the changes in question to be significant enough to warrant copyright.

### DIFF
--- a/src/codecs/avif/encoder.rs
+++ b/src/codecs/avif/encoder.rs
@@ -17,7 +17,7 @@ use crate::{ImageError, ImageResult};
 
 use bytemuck::{try_cast_slice, try_cast_slice_mut, Pod, PodCastError};
 use num_traits::Zero;
-use ravif::{encode_rgb, encode_rgba, Config, Img, RGB8, RGBA8};
+use ravif::{Encoder, Img, RGB8, RGBA8};
 use rgb::AsPixels;
 
 /// AVIF Encoder.
@@ -26,7 +26,7 @@ use rgb::AsPixels;
 pub struct AvifEncoder<W> {
     inner: W,
     fallback: Vec<u8>,
-    config: Config,
+    encoder: Encoder,
 }
 
 /// An enumeration over supported AVIF color spaces
@@ -56,7 +56,7 @@ enum RgbColor<'buf> {
 impl<W: Write> AvifEncoder<W> {
     /// Create a new encoder that writes its output to `w`.
     pub fn new(w: W) -> Self {
-        AvifEncoder::new_with_speed_quality(w, 1, 100)
+        AvifEncoder::new_with_speed_quality(w, 4, 80) // `cavif` uses these defaults
     }
 
     /// Create a new encoder with specified speed and quality, that writes its output to `w`.
@@ -67,24 +67,22 @@ impl<W: Write> AvifEncoder<W> {
         let quality = min(quality, 100);
         let speed = min(speed, 10);
 
+        let encoder = Encoder::new()
+            .with_quality(f32::from(quality))
+            .with_alpha_quality(f32::from(quality))
+            .with_speed(speed);
+
         AvifEncoder {
             inner: w,
             fallback: vec![],
-            config: Config {
-                quality: f32::from(quality),
-                alpha_quality: f32::from(quality),
-                speed,
-                premultiplied_alpha: false,
-                color_space: ravif::ColorSpace::RGB,
-                // match core count
-                threads: 0,
-            },
+            encoder,
         }
     }
 
     /// Encode with the specified `color_space`.
     pub fn with_colorspace(mut self, color_space: ColorSpace) -> Self {
-        self.config.color_space = color_space.to_ravif();
+        let encoder = self.encoder;
+        self.encoder = encoder.with_internal_color_space(color_space.to_ravif());
         self
     }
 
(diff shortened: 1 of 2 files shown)


anarchodin:
If I understand things correctly, the failure(s) here are due to an external issue, namely cross-rs/cross#1177. I hadn't bothered with figuring that out until now, because we're still in a part of the year where fairly little happens. :)

src/codecs/avif/encoder.rs
```diff
@@ -101,20 +99,18 @@ impl<W: Write> AvifEncoder<W> {
         color: ColorType,
     ) -> ImageResult<()> {
         self.set_color(color);
-        let config = self.config;
+        let encoder = self.encoder.clone();
```
fintelia, line 102: I think this clone could be avoided

anarchodin: It can. I put the clone there because I thought there was a good reason for the mutable borrow the compiler complained about, but looking closer I don't think there is.

src/codecs/avif/encoder.rs
```diff
@@ -124,7 +119,7 @@ impl<W: Write> AvifEncoder<W> {
     }
 
     fn encode_as_img<'buf>(
-        &'buf mut self,
+        fallback: &'buf mut Vec<u8>,
```
fintelia, line 122: At this point, `fallback` could just be allocated locally in this function.

src/codecs/avif/encoder.rs
```diff
@@ -101,20 +97,19 @@ impl<W: Write> AvifEncoder<W> {
...
         // owned version in our own buffer or zero-copy if possible by using the input buffer.
         // This requires going through `rgb`.
-        let result = match self.encode_as_img(data, width, height, color)? {
-            RgbColor::Rgb8(buffer) => encode_rgb(buffer, &config).map(|(data, _color_size)| data),
-            RgbColor::Rgba8(buffer) => {
-                encode_rgba(buffer, &config).map(|(data, _color_size, _alpha_size)| data)
-            }
+        let mut fallback = vec![]; // This vector is used if we need to do a color conversion.
+        let result = match Self::encode_as_img(&mut fallback, data, width, height, color)? {
+            RgbColor::Rgb8(buffer) => encoder.encode_rgb(buffer),
```
fintelia, line 106: The function call could just be `self.encoder.encode_rgb(buffer)` instead of borrowing `encoder` above.

src/codecs/avif/encoder.rs
```diff
@@ -67,24 +66,21 @@ impl<W: Write> AvifEncoder<W> {
...
-            },
+            encoder,
         }
     }
 
     /// Encode with the specified `color_space`.
     pub fn with_colorspace(mut self, color_space: ColorSpace) -> Self {
-        self.config.color_space = color_space.to_ravif();
+        let encoder = self.encoder;
+        self.encoder = encoder.with_internal_color_space(color_space.to_ravif());
```
fintelia, line 83: Does the borrow checker allow you to write `self.encoder = self.encoder.with_internal_color_space(...)`? If so, that would be cleaner.

fintelia (approved):
Looks good to me. I added a few comments suggesting style changes

src/codecs/avif/encoder.rs
anarchodin: I thought so initially, but it turns out that doesn't work. The return value is a reference. When no colour conversion is required that reference is into `data`, which is borrowed anyway, so that's fine. When the input isn't RGB(A), though, the reference is into `fallback`, which means it can't be deallocated on function return.

There's _probably_ a way around this, but `ravif` is very specific about the type of its input data and getting that to align with maybe-owned-maybe-borrowed values is beyond my Rust experience to solve rapidly. :)

src/codecs/avif/encoder.rs
anarchodin: It could. That's a purely style change, since the borrow would happen with the function call anyway, according to the times the borrow checker yelled at me here.

src/codecs/avif/encoder.rs
anarchodin: It does.

anarchodin:
I made an additional change, which is that the encoder now declares that it implements the `ImageEncoder` trait ... for some reason, that wasn't there before. I figure it's better to add it here than to make a different pull request, seeing as it's _literally_ just the declaration.

src/codecs/avif/encoder.rs
```diff
@@ -71,48 +71,49 @@ impl<W: Write> AvifEncoder<W> {
...
-        let encoder = self.encoder;
-        self.encoder = encoder.with_internal_color_space(color_space.to_ravif());
+        self.encoder = self
+            .encoder
+            .with_internal_color_space(color_space.to_ravif());
         self
     }
+}
 
+impl<W: Write> ImageEncoder for AvifEncoder<W> {
```
fintelia, line 86: Oh, good catch!

PR #1840, opened 2023-01-08
Consistent definition of Rgba

0xack13:
Proposing a change to make `Rgba` type definition similar to how `LumaA` is defined. It makes it easier to distinguish the difference with `Rgb` type.


HeroicKatora (approved):
Sure, sounds good.

PR #1844, opened 2023-01-18
Improve performance of grayscale function and better clone standards

MASACR99:
Hey! I was working with your library and found out that the grayscale function was too slow, specially, compared to the examples you use, for example:
`DynamicImage::ImageLuma8(image.into_luma8());`

After some investigation I found the issue inside the grayscale function, when using the code above there's a call to the dynamic map function which works much faster than the imageops::grayscale generic function.

I also changed a few unneeded conversions to just clone since the type is already the same (in this case for example ImageLumaA8 was being converted with imageops even though it's not needed)

After those changes the DynamicImage grayscale function is now faster than the example code above.

Pictures proving the new grayscale is faster: 

Old performance:
OldPerformance

New performance:
NewPerformance

Edit: Forgot to remove comment from license agreement

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -629,27 +629,24 @@ impl DynamicImage {
     pub fn grayscale(&self) -> DynamicImage {
         match *self {
             DynamicImage::ImageLuma8(ref p) => DynamicImage::ImageLuma8(p.clone()),
-            DynamicImage::ImageLumaA8(ref p) => {
-                DynamicImage::ImageLumaA8(imageops::grayscale_alpha(p))
-            }
-            DynamicImage::ImageRgb8(ref p) => DynamicImage::ImageLuma8(imageops::grayscale(p)),
-            DynamicImage::ImageRgba8(ref p) => {
-                DynamicImage::ImageLumaA8(imageops::grayscale_alpha(p))
-            }
+
+            DynamicImage::ImageLumaA8(ref p) => DynamicImage::ImageLumaA8(p.clone()),
+
+            DynamicImage::ImageRgb8(_) => dynamic_map!(*self, |ref p| DynamicImage::ImageLuma8(p.convert())), 
+
+            DynamicImage::ImageRgba8(_) => dynamic_map!(*self, |ref p| DynamicImage::ImageLumaA8(p.convert())),
+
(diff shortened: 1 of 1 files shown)


src/dynimage.rs
```diff
@@ -629,27 +629,24 @@ impl DynamicImage {
...
-            DynamicImage::ImageRgb8(ref p) => DynamicImage::ImageLuma8(imageops::grayscale(p)),
-            DynamicImage::ImageRgba8(ref p) => {
-                DynamicImage::ImageLumaA8(imageops::grayscale_alpha(p))
-            }
+
+            DynamicImage::ImageLumaA8(ref p) => DynamicImage::ImageLumaA8(p.clone()),
+
+            DynamicImage::ImageRgb8(_) => dynamic_map!(*self, |ref p| DynamicImage::ImageLuma8(p.convert())), 
+
+            DynamicImage::ImageRgba8(_) => dynamic_map!(*self, |ref p| DynamicImage::ImageLumaA8(p.convert())),
```
fintelia, line 637: I'm not sure why the dynamic_map is useful here. The dynamic_map! macro just expands to a match statement exactly like the one we're currently in:
https://github.com/image-rs/image/blob/37d3de296cf256c17e14de71d212849cbe026542/src/dynimage.rs#L107-L118

MASACR99: Also, just saw that I didn't remove the markdown comment to accept the license agreement, I edited the main comment to include it

fintelia: I would expect that this would be at least as fast as your dynamic_map! version:
```
DynamicImage::ImageRgba8(p) => DynamicImage::ImageLumaA8(p.convert()),
```

fintelia:
I'd feel better about this if we had some theory for why `convert` is faster than `imageops::grayscale`. Are we sure they're producing identical results?

HeroicKatora:
The conceptual difference: `imageops::grayscale` does an external loop over pixels, `convert` is the interior iteration version of this loop. Due  `imageops` being generic it can't statically know the best iteration strategy for the layout, nor can it communicate the same bounds checks to the compiler. It's not very surprising to me.

MASACR99:
@fintelia Just ran a few tests to check that the resulting images of grayscale are correct. They are! I'll add a few images of the result of the "new" function
grayscale_function_1
grayscale_function
grayscale_function

PR #1848, opened 2023-01-27
Exhaustively match in ColorType -> DynamicImage conversion

fintelia:
This ensures that `ColorType` and `DynamicImage` variants don't accidentally get out of sync


HeroicKatora (approved):
Slightly quirky (in the sense of, never seen that done before) random thought, is there merit in:

```rust
impl From<core::mem::Discriminant<DynamicImage>> for ColorType { … }
```

This PR is definitely good on its own though.

fintelia:
I never seen a From implementation like that, but I wouldn't be opposed

PR #1851, opened 2023-02-01
Feature: methods for getting a pixel using signed coordinates

rostyq:
This PR added two methods, each with default implementation to `GenericImageView` trait: `checked_get_pixel` and `saturating_get_pixel`. Both allow to get a pixel from an image view using signed coordinates. That can be convenient for different computer vision algorithms like object detection or tracking.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -1,13 +1,11 @@
 #![allow(clippy::too_many_arguments)]
-use std::convert::{TryFrom, TryInto};
+use std::convert::TryFrom;
 use std::ffi::OsStr;
 use std::io;
 use std::io::Read;
 use std::ops::{Deref, DerefMut};
 use std::path::Path;
 use std::usize;
-
-use exr::image;
 
 use crate::color::{ColorType, ExtendedColorType};
 use crate::error::{
@@ -1388,7 +1386,7 @@ mod tests {
     };
     use crate::color::Rgba;
     use crate::math::Rect;
-    use crate::{GrayImage, ImageBuffer, Luma};
+    use crate::{GrayImage, ImageBuffer};
 
     #[test]
     #[allow(deprecated)]
@@ -1527,7 +1525,7 @@ mod tests {
 
     #[test]
     fn test_saturating_get_pixel() {
-        let source: ImageBuffer<Luma<u8>, Vec<u8>> = ImageBuffer::from_vec(3, 3, vec![
+        let source = GrayImage::from_vec(3, 3, vec![
             0u8, 1u8, 2u8,
             3u8, 4u8, 5u8,
             6u8, 7u8, 8u8,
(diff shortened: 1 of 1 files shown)


fintelia:
This could be simplified using the max() and min() methods. For instance, `x.max(0).min(width - 1) as u32`

Also worth double checking that zero-sized images are either prohibited by GenericImage or properly handled by this patch

src/image.rs
```diff
@@ -921,6 +921,65 @@ pub trait GenericImageView {
...
+                }).unwrap_or(0u32)
+        }
+
+        unsafe {
+            self.unsafe_get_pixel(
+                convert(x, self.width()),
+                convert(y, self.height())
+            )
+        }
+    }
```
HeroicKatora, line 981: > Also worth double checking that zero-sized images are either prohibited by GenericImage or properly handled by this patch

They are not. Panic in those cases would be permissible behavior but may it's not desirable. The methodical evaluation of how we wrote the trait is, however:

> [Safety](https://docs.rs/image/latest/image/trait.GenericImageView.html#safety)
> * The coordinates must be [in_bounds](https://docs.rs/image/latest/image/trait.GenericImageView.html#method.in_bounds) of the image.

Note that `in_bounds` is both safe, and the trait itself is safe. Thus, we can't make any assumption about its implementation other than the method signature. Since we haven't called the method here either, we can not argue to have fulfilled the precondition properly. I don't think this method possible in a generic, sound manner either way.

However, we _can_ add specialized implementations of these two new methods for `ImageBuffer`. Since the `in_bounds` method is local, we can use its implementation details and rely on them being correct.

HeroicKatora (changes requested):

src/image.rs
HeroicKatora: This is similar to `get_pixel_unchecked`. Note that the _generic_ implementation provided by the trait does not in fact call any unsafe method, it defaults to a purely safe fallback. This is correct and intended; since there's nothing `unsafe` about the trait, the implementor has no obligations to 'do the right thing'. There's no inherent property of `in_bounds` that the generic implementation could rely on for any effect; so calling the safe fallback is indeed the best useful implementation we can provide.

Note to self: `ImageBuffer` is still dubious if the user chooses a non-default container. Damn.

strasdat:
When designing image types (e.g. https://github.com/strasdat/Sophus/blob/23.04-beta/cpp/sophus/image/image_view.h#L125), I came to the same conclusion that signed integer pixel coordinates are beneficial. For instance, it is quite useful/common to ask whether a 3d point almost projects into a camera's field of view - hence talking about a pixel coordinate of e.g. (-5, -10) can indeed by meaningful.

rostyq:
If anyone is interested, I've come up with an idea to use the trait approach for implementing this logic. See `ExtendedImageView` trait in [pixelutil-image-rs](https://github.com/rostyq/pixelutil-image-rs/).

Issue #772, opened 2018-05-11
Can the HDR backend use rayon instead of scoped_threadpool?


icefoxen:
That would mean that between it and the `jpeg_rayon` feature this crate only bothers using one multithreading implementation.

Also `hdr_decode.rs` includes things like `let mut pool = Pool::new(8);`, which is kind of sub-ideal and using rayon would make it easier to be nicer.

crazymykl:
It's at least not a drop-in replacement, since rayon's scope method runs on the pool, not the main thread.

james7132:
I know this issue is old, but found this after checking why this crate has this dependency. It'd be ideal to provide a way to let users to handle the decoding themselves in their own multithreaded executors (i.e. rayon, tokio's blocking threads, etc.).

We're using this crate for Bevy's renderer, and one of the key things we're trying to avoid is thread contention and overhead from spawning too many. We're using our own thread pool, and would like for it to be the only one running over the lifetime of the app: spawning and joining 8 threads to decode one HDR image is generally undesirable on our end.

fintelia:
Honestly in the short term it might make sense to just change the decoder to operate single threaded. Pulling in scoped_threadpool for a single rarely used format is a bit unfortunate. If it is too slow without multithreading (which I kind of doubt given that HDR isn't that complex a format), there's a bunch of other optimizations we could try.

Do you want to make a PR to get rid of the multithreading?

PR #1857, opened 2023-02-09
Remove multithreading in HDR decoding

bash:
As discussed in https://github.com/image-rs/image/issues/772#issuecomment-1382406241. This also makes it work on WASM.

---

### DIFF
--- a/src/codecs/hdr/decoder.rs
+++ b/src/codecs/hdr/decoder.rs
@@ -462,12 +462,12 @@ impl<R: BufRead> HdrDecoder<R> {
 
         let mut buf = vec![Default::default(); self.width as usize];
         for chunk in chunks_iter {
+            // read_scanline overwrites the entire buffer or returns an Err,
+            // so not resetting the buffer here is ok.
             read_scanline(&mut self.r, &mut buf[..])?;
             for (dst, &pix) in chunk.iter_mut().zip(buf.iter()) {
                 *dst = f(pix);
             }
-            buf.clear();
-            buf.resize(self.width as usize, Default::default());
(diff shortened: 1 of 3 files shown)


src/codecs/hdr/decoder.rs
```diff
@@ -460,21 +459,16 @@ impl<R: BufRead> HdrDecoder<R> {
...
+        let mut buf = vec![Default::default(); self.width as usize];
+        for chunk in chunks_iter {
+            read_scanline(&mut self.r, &mut buf[..])?;
+            for (dst, &pix) in chunk.iter_mut().zip(buf.iter()) {
+                *dst = f(pix);
             }
-            Ok(())
-        }) as Result<(), ImageError>)?;
+            buf.clear();
+            buf.resize(self.width as usize, Default::default());
```
fintelia, line 470: Would `buf.fill(Default::default())` work instead?

fintelia: Actually, why is this needed at all? Shouldn't read_scanline overwrite the entire buffer?

fintelia:
The same CI failure seems to be happening on main: https://github.com/image-rs/image/actions/runs/4067817240/jobs/7156440418

src/codecs/hdr/decoder.rs
bash: Removed (and commented) in [7df5cb1](https://github.com/image-rs/image/pull/1857/commits/7df5cb142318525ff1c544361656157c6bb70be6)

fintelia (approved):

PR #1858, opened 2023-02-10
Add ICC profile to ImageDecoder

### LINKED ISSUE #1832: Access ICC profile data
I would like to be able to convert images to sRGB (or other profiles in the future.) The specific `png` and `jpeg-decoder` crates support accessing `icc_profile()` that I can use with the `lcms2` crate but, they do not support `DynamicImage`. The `image` crate does not support accessing the color profile information but supports `DynamicImage`.

My specific use case for this functionality is writing a new image viewer for GNOME.

This might be more generally applicable to other features like EXIF block data or background colors (bKGD).

## Draft

1. Add those features to `ImageDecoder` returning `None` if not applicable.
2. Add `DynamicImage` support to the specific crates. Webp already has that.
3. Add those features to the format-specific decoders like `image::codecs::png::PngDecoder`. However, not sure if that's possible for EXIF where the end of the file is relevant?

sophie-h:

---

Giving this a go. Not sure why `&mut self` was suggested in the issue

Closes #1832

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -237,6 +237,10 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for PngDecoder<R> {
         self.color_type
     }
 
+    fn icc_profile(&mut self) -> Option<Vec<u8>> {
+        self.reader.info().icc_profile.as_ref().map(|x| x.to_vec())
+    }
+
     fn into_reader(self) -> ImageResult<Self::Reader> {
         PngReader::new(self.reader)
     }
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -27,6 +27,7 @@ pub struct AvifDecoder<R> {
     inner: PhantomData<R>,
     picture: dav1d::Picture,
     alpha_picture: Option<dav1d::Picture>,
+    icc_profile: Option<Vec<u8>>,
 }
 
 impl<R: Read> AvifDecoder<R> {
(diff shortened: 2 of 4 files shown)


src/codecs/png.rs
```diff
@@ -131,6 +132,8 @@ impl<R: Read> PngDecoder<R> {
         let info = decoder.read_header_info().map_err(ImageError::from_png)?;
         limits.check_dimensions(info.width, info.height)?;
 
+        let icc_profile = info.icc_profile.as_ref().map(|x| x.to_vec());
```
fintelia, line 135: This will always be None. `read_header_info` never returns ICC information, you have to call `info()` on the reader returned from `read_info()` instead.

fintelia:
Thanks for this PR! The reason to require `&mut self` is because some future decoders might require it. In particular, we don't want to force every decoder to parse the ICC profile in its new() method.

src/codecs/png.rs
sophie-h: Ah, thanks. I don't even have to store it. Fixed.

sophie-h:
Changed all to `&mut`

HeroicKatora:
:+1: Though I now loathe how `ImageDecoder::read_image` takes `self` instead of `&mut self`. For instance, in the PNG decoder the ICC profile may not be part of the header but be placed at some arbitrary location in the data stream. The caller can not generically _know_ if the information has already been read until the image parsing was 'done'. The whole design deserves a good rework. 

Going to ignore the deny warnings for this PR, it looks like rav1e and criterion had new releases we should bump to.

sophie-h:
> in the PNG decoder the ICC profile may not be part of the header

I guess I should mention in the docs that it could only be available after completely reading the image?

> The whole design deserves a good rework.

Not sure if that is about this MR? I mean, that sounds like a fundamental issue. But for ICC profiles, it seems like it's fine to only have them after the complete image is decoded.

For EXIF, it would be much more relevant to know it as early as possible because knowing the size and orientation of the image is very helpful for the presentation, even before complete decoding.

HeroicKatora (approved):
>  Not sure if that is about this MR? I mean, that sounds like a fundamental issue. But for ICC profiles, it seems like it's fine to only have them after the complete image is decoded.

It's not, PR itself looks good to me.

PR #1860, opened 2023-02-10
Force install AFL in github workflow

fintelia:


fintelia:
Force install doesn't seem to work... the issue is being tracked in https://github.com/rust-fuzz/afl.rs/issues/299

PR #1863, opened 2023-02-16
Adds Into<Container> for ImageBuffer

coreylowman:

# Summary

Hello, I'd like to be able to transform `ImageBuffer` into `Vec<u8>` generically. As of now there is the `to_vec` and `into_vec` methods, but they don't necessarily work in a generic context.

More concretely I'd like to be able to use ImageBuffer with anything that requires a `T: Into<Vec<u8>>`. For example:

```rust
pub fn do_something<T: Into<Vec<u8>>>(t: T) -> Vec<u8> { t.into() }
```

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -659,6 +659,14 @@ pub struct ImageBuffer<P: Pixel, Container> {
     data: Container,
 }
 
+// transformation into the container - useful for things that want to
+// transform a buffer back into a Vec
+impl<P: Pixel, Container> Into<Container> for ImageBuffer<P, Container> {
+    fn into(self) -> Container {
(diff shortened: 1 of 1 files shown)


fintelia:
There's already `ImageBuffer::into_raw` which produces the underlying container. I wouldn't be opposed to having the same functionality exposed via an `Into<>` implementation. However, based on the CI run there may be trait conflict issues that prevent that from working?

coreylowman:
Yeah similar to `to_vec`/`into_vec`, `into_raw` can't be used in a generic context. I see the conflict, usually it tells you where the other impl is though. Unless I'm not seeing it?

Issue #1184, opened 2020-04-03
no_std support


vadixidav:
I would like `image` to support no_std.

I am using it in the [`akaze` crate](https://github.com/rust-cv/akaze). To be able to support embedded robotics/drone targets, it needs to support no_std. However, `image` is used for the input types to integrate into the ecosystem more cleanly.

This is important for utilizing image across the whole ecosystem, including in embedded targets, where they may have images coming in from attached cameras or even just when reading/writing images being sent over a bus to a screen.

The most basic way to implement this functionality is just to create a `std` feature that is included by default to allow backwards compatibility. This feature would gate all functions that are specific to an OS. This includes `image::open`, which can only execute with a file system.

I would also recommend adding a GitHub action to test that no_std support actually works in CI. A good example of this is in some crates I have recently worked on such as [`eight-point`](https://github.com/rust-cv/eight-point/blob/master/.github/workflows/no-std.yml).

HeroicKatora:
Depending `alloc` in a `#[no_std]` crate was stabilized only in `1.36`. It's not completely unlikely that we'll run into some MSRV problem. That could defer completion until the next major release, there's also an appropriate tag to set if any issues are encountered. (I'm assuming you're not asking for

fintelia:
This issue is even harder than it might initially seem because the `Read` and `Write` standard library traits our encoders/decoders depend on [don't work with no_std](https://github.com/rust-lang/rust/issues/48331).

vadixidav:
Is it possible to feature gate any code that requires `Read` and `Write` behind the `std` feature? I only need image for its image types, for instance. I am using them (with liballoc) to act as a standard way of passing the image data around and doing conversions. Once `Read` and `Write` are supported in no_std, we can just add those impls/fns later.

HeroicKatora:
It's not just the methods but even the ubiquitous [`Error` type uses `std::io::Error`](https://docs.rs/image/0.23.3/image/error/enum.ImageError.html#variant.IoError) and thus is on its own currently incompatible with `no_std`. This then 'bleeds' into many other interfaces that use that central type directly, in particular even into `GenericImage`. So this would leave a lot less remaining than it would remove.

This is however useful to plan for `0.24`, the earliest where we could do the required interface changes. First, we'd need to identify methods that should use more concrete error types. These could then change in the next release and bring us closer to a useful `#[no_std]` portion.

luojia65:
[...]

PR #1868, opened 2023-03-01
basic no_std support

n-prat:
Hello,

Following up on https://github.com/image-rs/image/issues/1184#issuecomment-1443804914 I am opening a PR to discuss
about potential `no_std` support.

Copy-pasting and rewriting my comment:

TODO?:
- it still needs more work: I had to butcher error.rs, but I can probably do better.
  - I could optionally move the whole "error.rs refactor" into a separate PR [e.g. replace custom errors by Snafu which supports `no_std`] 
- I had to be liberal with e.g. `#[cfg(feature = "std")]` but I can probably replace most of them by a new e.g. `io` feature.
  - instead of deactivating all io related code, there are other options: https://docs.rs/binrw/latest/binrw/ and https://github.com/dataphract/acid_io But I do not know these libs so not sure if going that way is a good idea?
- Right now I just aimed to make the core lib compile b/c that is my use case [I need this crate b/c it is a dependency of `imageproc` which is the one I use directly] → making each and every feature work under `no_std` is a different beast I am guessing

I'm open to feedback!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -121,6 +121,19 @@ jobs:
       env:
         SYSTEM_DEPS_DAV1D_BUILD_INTERNAL: always
 
+  no_std_check:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: nightly
+        override: true
+    - name: build
+      run: |
+        cargo install cargo-no-std-check
+        cargo no-std-check --no-default-features --features=alloc
+
   clippy:
     runs-on: ubuntu-20.04
     steps:


src/error.rs
```diff
@@ -23,56 +27,83 @@ use crate::image::ImageFormat;
...
     /// requirements set by the caller.
-    Decoding(DecodingError),
+    ///
+    /// An error was encountered while decoding an image.
+    ///
+    /// This is used as an opaque representation for the [`ImageError::Decoding`] variant. See its
+    /// documentation for more information.
+    ///
+    /// [`ImageError::Decoding`]: enum.ImageError.html#variant.Decoding
+    Decoding { format: ImageFormatHint },
```
HeroicKatora, line 44: Is all of this diff necessary for `no_std`? Seems a little weird. In particular the comments are no longer accurate. Enum variants are _not_ opaque. It's definitely preferrable to have opaque types to allow changing the field representation / error messages / internal variants.

HeroicKatora (commented):
Seems actually smaller than expected. Without regards to the _fundamental_ issue with the `ImageDecoder` trait, it seems almost quite feasible to do in single major version. 

Except: Getting rid of/bounding `DynamicImage` to `std` is quite a deal breaker for adoption. It would seem that an alloc-but-no_std should be capable of preserving the type, or not?

HeroicKatora:
One comment I want to add outside the workflow of the PR review: If we _need to_  get rid of a core buffer data structure, rewrite the `DecoderTrait`, etc. completely. Then I'm pretty open to that in a major version as long as there is a clear plan on how to push such rework forward and keep the test suite in tact while doing it (in a separate branch). That is, somehow keeping track of features that need to be ported. But a new major version is still better than needing an ecosystem split to a different crate, no?

src/image.rs
```diff
@@ -647,7 +649,10 @@ impl Progress {
 /// The trait that all decoders implement
 pub trait ImageDecoder<'a>: Sized {
     /// The type of reader produced by `into_reader`.
-    type Reader: Read + 'a;
+    #[cfg(feature = "std")]
+    type Reader: std::io::Read + 'a;
+    #[cfg(not(feature = "std"))]
+    type Reader: 'a;
```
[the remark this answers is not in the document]
fintelia: I'd be OK with just limiting ImageDecoder to std. All the implementations have a constructor that takes a `io::Read` so there's going to be little use for it in no_std settings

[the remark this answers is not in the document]
nathanprat: I don't think I have gated `ImageDecoder` behind `std`. Technically, only its impl using `Reader` are?

src/error.rs
nathanprat: Yes I think it is...
Previously error types were custom one using `std::error::Error` but that can not work in `no_std`.
The only way to make it work is to use a no_std-compatible lib to handle `Error` and I only found `Snafu`.

But I had a feeling it was messy to bundle the `error.rs` refactor with this PR.
Maybe we can do it in two steps:
- first PR to refactor `error.rs` to use `Snafu`
- "main" PR for `no_std`

It is a big diff but not a complicated one b/c it is mainly code removal:
- no need to `impl Error` when using Snafu
- no need to `impl fmt::Display` when using Snafu

Question: not sure if what I did is a breaking change?

nathanprat:
> Except: Getting rid of/bounding `DynamicImage` to `std` is quite a deal breaker for adoption. It would seem that an alloc-but-no_std should be capable of preserving the type, or not?

I will try to make it work under `no_std` but my initial test shows it is not easy.
Indeed it requires eg `decoder_to_vec` which in turn requires `ImageDecoder::read_image_with_progress` which in turn requires `type Reader: std::io::Read + 'a`
I do agree that it should ideally be refactored b/c it makes no sense for "fn with slice input" to require `std::io`.

But why would it be a dealbreaker?
- people using `default` feature would not see a difference
- people requiring `no_std` don't really expect 100% of the feature to available
   Eg for my use case: I need `image` as a transitive dep of `imageproc`; and the parts of `imageproc` I use works apparently fine without `DynamicImage` cf https://github.com/Interstellar-Network/imageproc/tree/sgx-nostd-compat for the fork, and https://github.com/Interstellar-Network/lib-garble-rs/blob/main/lib-garble-rs/src/watermark.rs for how I use it

src/error.rs
fintelia: These changes to ImageError are a breaking change because the variants of public enums are public. What's not clear to me is whether the changes here are actually necessary. Can Snafu really not derive its trait for enums with tuple variants?

src/image.rs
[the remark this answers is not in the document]
fintelia: What I was saying is that `ImageDecoder` should be gated behind `std` because some of its associated types / methods require standard library traits.

src/error.rs
fintelia: What happens if we just gate the `impl std::error::Error` parts behind `std`? Would that cause any problems? Because it would be nice to avoid the code churn if possible

src/image.rs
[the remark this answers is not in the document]
nathanprat: Ah, ok! Done.

src/error.rs
nathanprat: Wouldn't that make all the errors not proper errors in `no_std` context?

But we can go with your idea and instead[starting by reversing all changes in `error.rs`]:
- with `std`: `use std::error::Error`, same as before
- `no_std`: `use core::error::Error`
   That would make `no_std` builds require nightly for now; but that is fine for my use case.

Would that be ok for you?

For completeness: there is also the "duplicated code" alternative:
- move original `error.rs` -> `error/error_compat.rs`
- rename this forks `error.rs` -> `error/error_new.rs`
- in `error/mod.rs` switch impl based on std vs no_std

197g:
See https://github.com/image-rs/image/pull/2462#issuecomment-3643079460 but of course this is far older. There's not much to gain from leaving it open, unfortunately. Hopefully other libraries provide the compatibility in `no_std` you were looking for.

PR #1869, opened 2023-03-03
Improve ImageDecoder trait

fintelia:
This sketches out possible changes we may want to make to the ImageDecoder trait in the next major release:
* Makes the trait object safe
* ~~Adds a constructor (all non-deprecated implementations had one, but it wasn't part of the trait).~~
* ~~Combines `ImageDecoder` and `ImageDecoderRect`~~
* Removes the `into_reader`, `*_with_progress`, and `scanline_bytes` methods. This is the aspect I'm least sure about, but few backends had meaningful implementations of these and it cuts down on the number of methods on the trait. Some decoders will still be able to support incremental decoding by doing multiple `read_rect` calls 

The object safety change in particular makes our internal usage of the trait much nicer. Instead of having to work with a DecoderVisitor trait, we can rely on dynamic dispatch.

~~For now I've held on updating the various format decoders. That'll be a bunch of work so it makes sense to settle on what the API should be first.~~ 

_**Edit:** scaled back some of these changes and updated all the format decoders._

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        features: ['', default, gif, jpeg, png, tiff, ico, pnm, tga, webp, bmp, hdr, dxt, dds, farbfeld, openexr, jpeg_rayon, webp-encoder]
+        features: ['', default, gif, jpeg, png, tiff, ico, pnm, tga, webp, bmp, hdr, dds, farbfeld, openexr, jpeg_rayon, webp-encoder]
     steps:
     - uses: actions/checkout@v4
     - uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,8 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ["1.61.0", stable, beta, nightly]
-        features: [gif, jpeg, png, tiff, ico, pnm, tga, webp, bmp, hdr, dds, farbfeld, openexr, jpeg_rayon, webp-encoder, '']
+        features: ['', default, gif, jpeg, png, tiff, ico, pnm, tga, webp, bmp, hdr, dxt, dds, farbfeld, openexr, jpeg_rayon, webp-encoder]
     steps:
     - uses: actions/checkout@v4
     - uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,7 +14,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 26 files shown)


sophie-h:
While working on image-rs I got the idea that this trait could go into something like an `image-decoder` crate.

This would allow the format-specific crates to already implement `ImageDecoder`, which could then be used by image-rs. That could bring more of the format-specific code into one place and would give more overall flexibility with coupling things into image-rs and having a common abstraction trait already usable when a format crate is not yet used in image-rs.

fintelia:
Yeah, it would be nice to restructure that way. The challenge is that we have to be _very_ confident in the trait definitions we pick, because making any backwards incompatible changes to them would cause tons of ecosystem churn

HeroicKatora:
The impact from a breaking change in another trait could be mitigated by some choices in usage. It's technically possible to depend on multiple versions of a single crate. What's hard is using the trait only in such ways that bumping to a new version is not a breaking change downstream.

For instance, consider an opaque wrapper that _happens to implement_ `From<Box<dyn  image_core::ImageDecoder>>`. Then adding a new version of the trait would be akin to having to add `From<Box<dyn image_core_v2::ImageDecoder>>` where `image_core_v2` is just an alias for the new major version of the crate. And then we can be cautious with slowly deprecating the old versions. The main costs are compiling with multiple old versions that must be managed (and this pattern interacts badly with `cargo vendor` which may be used in some distributions? not quite sure).

johannesvollmer:
I think a progress callback is an essential feature for such a mature library. It's fine to not have it for small prototypes, but any real ui app or game would want to display some kind of progress. Instead, we should work on adding it to the decoders and encoders. For hdr, for example, it would be trivial to implement. 

I think it's more valuable to have it, even if support is rare, than to remove it and add it later again. 

Here's something we could try: Code up a wrapper around `std::io::Write` and `std::io::Read` that tracks the number of bytes being written or read. Many codecs could use this to implement the progress callback quickly, albeit imprecise. I could try to sketch out that idea

johannesvollmer:
There is a way to tack-on progress on existing decoders: Observe the bytes that they read. But its only meaningful if we can make two assumptions:

- the decoders implementation reads the file bytes incrementally
- the compressed file is roughly as large as the in-memory representation

if the assumptions don't hold, the progress can still be useful, but will be imprecise

johannesvollmer:
i sketched the idea at #2014 

about your second point: i think a `progress` struct is nice. a percentage float would probably be enough though, but only to reduce the complexity in the repo. because otherwise we can just tack on a `ratio_0_to_1() ->f32` on the `Progress` struct.

fintelia:
Another thing that makes it annoying to convert decoders is that you want to reasonably throttle how often you call the progress callback. Too often and you'll destroy performance. Not often enough and the user won't get progress indications. Even just checking "should I invoke the callback" too often might interfere with SIMD and slow down our fastest encoders (which can hit multiple hundred million pixels per second!)

I also realized that the current function signature _won't_ work when trying to make the trait object safe. No methods with generic arguments can be called on a trait object. At which point, perhaps we might as well make a separate `ImageDecoderProgress` trait

PR #1870, opened 2023-03-05
Update dav1d and dcv-color-primitives

averyanalex:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,8 @@ tiff = { version = "0.8.0", optional = true }
 ravif = { version = "0.11.0", optional = true }
 rgb = { version = "0.8.25", optional = true }
 mp4parse = { version = "0.12.0", optional = true }
-dav1d = { version = "0.6.0", optional = true }
-dcv-color-primitives = { version = "0.4.0", optional = true }
+dav1d = { version = "0.9.3", optional = true }
+dcv-color-primitives = { version = "0.5.2", optional = true }
 color_quant = "1.1"
 exr = { version = "1.5.0", optional = true }
 qoi = { version = "0.4", optional = true }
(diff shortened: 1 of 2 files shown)


HeroicKatora (approved):

averyanalex:
`primary_decoder.get_picture()` always returns `dav1d::Error::Again`. Adding `primary_decoder.flush()` doesn't help.

Quackdoc:
I was looking into this the other day myself

I was able to get decoding working by using something along the lines of this. though it was a little different, however I started running into issues with dcp so I stopped working on this. it doesnt seem like dcp supports greater then 8bit yet either so I dont really have a lot of motiviation here. I did want to look at possibly using the yuvxyb crate for format conversion



```diff
-        let mut primary_decoder = dav1d::Decoder::new();
-        let coded = ctx.primary_item_coded_data();
-        primary_decoder
-            .send_data(coded, None, None, None)
-            .map_err(error_map)?;
+        let mut dsettings = dav1d::Settings::new();
+        dsettings.set_n_threads(1);
+        let mut primary_decoder = dav1d::Decoder::with_settings(&dsettings).unwrap();
+        let coded = ctx.primary_item_coded_data().to_owned();
+        primary_decoder.send_data(coded, None, None, None).map_err(error_map)?;
```

HeroicKatora:
Alright, this should be at least covered by the test suite. Does that mean you have a reproduction you could contribute @Quackdoc ?

Quackdoc:
sorry, this got burried in my mail, I think it would probably be best to swap out `dcv-color-primitives` for `yuvxyb` crate since that would get us 422 support and 444 support.

here is a patch I used to decode the avif
[dav1d.patch](https://github.com/image-rs/image/files/11349349/dav1d.patch)
and here is a sample avif that doesn't work 
https://files.catbox.moe/vmkzie.avif

currently it's failing at 

```rust
            dcp::convert_image(
                width,
                height,
                &src_format,
                Some(strides),
                &src_buffers,
                &dst_format,
                None,
                dst_buffers,
            )
            .map_err(error_map)?;
            ```

jmonteiro64:
> as far as I can tell the avif should be compatible with what is there since it's 8bit yuv420p.

It looks like dcv-color-primitives only supports converting from `I420` to `Bgra`, not to `Rgba` (which is what's requested), see [here](https://docs.rs/dcv-color-primitives/0.5.4/dcv_color_primitives/fn.convert_image.html#errors).

fintelia:
This ended up getting implemented in a different PR, but thanks for raising this!

PR #1898, opened 2023-03-21
Release 0.24.6

fintelia:


HeroicKatora:
@fintelia Is it intention to do this before merging #1877 as a fallback? If so, looks good :+1:

fintelia:
Yep. Exactly

Issue #1381, opened 2020-11-28
Make DynamicImage serializable via serde


fintelia:
Requested by @GalAster in [this comment](https://github.com/image-rs/image/issues/741#issuecomment-735258185):

> I really need serde features, I'm using bincode to transfer textures.
> 
> My current solution is to convert as bytes with img.write_to(&mut buf, ImageOutputFormat::Png) before transmission.
> 
> Then import bytes with image::load_from_memory_with_format(buf, ImageFormat::Png) after the transfer is complete.
> 
> In most cases, this is lossless, but I hope there is a better solution, and we can experiment with a feature flag that is turned off by default.

slackers123:
Are there any plans to continue this?

oovm:
I added a basic support in #1903.

Png is not always available, so I added a type field to safely convert the byte stream to the corresponding type

oovm:
I'm not familiar with bytemuck, is there an interface to cast `Vec<u8>` as `Option<Vec16>>`?

slackers123:
Do you mean when the Vec<u8> is empty it is None and otherwise Some(Vec<i16>)?

fintelia:
Nope. Even if the `Vec<u8>` is properly aligned, the allocator may treat `[u8]` and `[u16]` differently so dropping the `Vec<u16>` may misbehave (because the slice was allocated with type `u8` but freed via the `u16` codepath).

PR #1903, opened 2023-04-10
Basic serde support

oovm:
Releated: https://github.com/image-rs/image/issues/1381

## License

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ color_quant = "1.1"
 exr = { version = "1.5.0", optional = true }
 qoi = { version = "0.4", optional = true }
 libwebp = { package = "webp", version = "0.2.2", default-features = false, optional = true }
+serde = { version = "1.0.159", optional = true }
 
 [dev-dependencies]
 crc32fast = "1.2.0"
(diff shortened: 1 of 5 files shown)


fintelia:
Is there a reason that this cannot use Serde derives? Do the non-exhaustive annotations cause issues or something?

Cargo.toml
```diff
@@ -88,6 +89,8 @@ avif = ["avif-encoder"]
 avif-encoder = ["ravif", "rgb"]
 # Non-default, even in `avif`. Requires stable Rust and native dependency libdav1d.
 avif-decoder = ["mp4parse", "dcv-color-primitives", "dav1d"]
+# Enable serde for dependent crates.
+serde = ["dep:serde", "rgb/serde", "num-complex/serde", "num-rational/serde"]
```
cycraig, line 93: Use `rgb?/serde` to avoid pulling in the optional `rgb` dependency unintentionally.

Coincidentally, `rgb?/serde` is the official example for [optional dependency features](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features).

(I am not a maintainer)

PR #1907, opened 2023-04-13
Add option to set the background color of animated webp images

### LINKED ISSUE #1906: Add option to ignore/replace background color for extended webp
I would like to be able to disable the background color when decoding extended webp images/animations. 

Some webp animations use a white background by default, and transparent pixels. When the frame does not cover the whole canvas, the white background will appear. 

[Example webp](https://cdn.7tv.app/emote/62a8dfd6aa031674e785e8a1/1x.webp)

The webp specification says that the background color value should be treated as a hint, and does not need to be used, so it would be nice to have the option to make the background transparent, or choose it's color completely.

## Draft
```rust
// WebPDecoder
pub fn set_background_color(&mut self, color: Rgba<u8>) -> ImageResult<()> {
    match &mut self.image {
        WebPImage::Extended(image) => image.set_background_color()?
        _ => Err(...) 
    }
}

// ExtendedImage
pub(crate) fn set_background_color(&mut self, color: Rgba<u8>) -> ImageResult<()> {
    match &mut self.image {
        ExtendedImageData::Animation { anim_info, .. } => {
            anim_info.background_color = color;
            Ok(())
        }
        _ => Err(...),
    }
}


```

Nogesma:
Fixes #1906.

I'm not sure I chose the right error type, tell me if there's one that suits this type of error better, maybe `UnsuportedError` ?

---

### DIFF
--- a/src/codecs/webp/decoder.rs
+++ b/src/codecs/webp/decoder.rs
@@ -4,9 +4,9 @@ use std::io::{self, Cursor, Error, Read};
 use std::marker::PhantomData;
 use std::{error, fmt, mem};
 
-use crate::error::{DecodingError, ImageError, ImageResult};
+use crate::error::{DecodingError, ImageError, ImageResult, ParameterError, ParameterErrorKind};
 use crate::image::{ImageDecoder, ImageFormat};
-use crate::{color, AnimationDecoder, Frames};
+use crate::{color, AnimationDecoder, Frames, Rgba};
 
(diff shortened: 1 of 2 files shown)


sophie-h:
Thanks! Loupe [also wants this](https://gitlab.gnome.org/Incubator/loupe/-/issues/159).

fintelia (approved):

PR #1910, opened 2023-04-20
Fix `pixel_indices` bound check

Coca162:
Currently the bounds check for `pixel_indices` is:
```rs
if x >= self.width || y >= self.height {
    return None;
}
```

Having this as greater then or equal causes issues if you attempt to get the edge of a image, I noticed my program was crashing when I sent my cursor coordinates to get a pixel color from the edge of my screen. Changing this to be only `>` fixed the issue.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -810,7 +810,7 @@ where
 
     #[inline(always)]
     fn pixel_indices(&self, x: u32, y: u32) -> Option<Range<usize>> {
-        if x >= self.width || y >= self.height {
+        if x > self.width || y > self.height {
             return None;
         }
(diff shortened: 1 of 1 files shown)


fintelia:
I don't think this is right. Shouldn't the edges of the image be at `0` and `width-1` / `0` and `height-1`?

HeroicKatora (changes requested):
This is a very soundness critical assertion, too. And @fintelia is correct. This method must stay consistent with `in_bounds` which will return false for those cases mentioned. You'll want your program to be prepared for handling `None` returns properly here, to fix the crash instead.

PR #1912, opened 2023-04-20
improve JPEG encode_block perf by about 25% with faster integer encoding

mcroomp:

Also add benches for tracking performance of JPEG block encoding. 

pre:
`test codecs::jpeg::encoder::tests::bench_encode_block      ... bench:         548 ns/iter (+/- 131) = 135 MB/s`
post:
`test codecs::jpeg::encoder::tests::bench_encode_block      ... bench:         410 ns/iter (+/- 140) = 180 MB/s`

### DIFF
--- a/src/codecs/jpeg/encoder.rs
+++ b/src/codecs/jpeg/encoder.rs
@@ -1,7 +1,6 @@
 #![allow(clippy::too_many_arguments)]
 
 use std::borrow::Cow;
-use std::convert::TryFrom;
 use std::io::{self, Write};
 use std::num::NonZeroI32;
 


mcroomp:
Looks like NonZeroU32::unsigned_abs was stabilized in 1.64... let's see if there's a workaround

HeroicKatora:
We can also bump MSRV for a perf win, it's not a set-in-stone restriction. Bump here and as indicated in the comment to do it with the PR:

https://github.com/image-rs/image/blob/master/Cargo.toml#L7-L8

Shnatsel:
I see MSRV has been bumped to 1.67 since, which is higher than this PR's requirement of 1.64.

I understand there is nothing blocking this, and it can be merged? 

The change is simple, covered by tests, and the performance gains are tantalizing.

fintelia:
Will look into reviving this PR once `0.25.0` is out

mcroomp:
Rebased since this change had sat there for a while

mcroomp:
There's still a CI test_toolchain with 1.63 that needs to be upgraded before this can be integrated

src/codecs/jpeg/encoder.rs
```diff
@@ -765,7 +855,8 @@ fn build_quantization_segment(m: &mut Vec<u8>, precision: u8, identifier: u8, qt
     }
 }
 
-fn encode_coefficient(coefficient: i32) -> (u8, u16) {
+#[cfg(feature = "benchmarks")]
+fn encode_coefficient_old(coefficient: i32) -> (u8, u16) {
```
fintelia, line 859: I don't think we should keep around an old implementation simply to benchmark against

fintelia (commented):
Thanks for this PR

I think it is kind of doing a lot and would be better split into a bunch of PRs. Perhaps:

1. Adding benchmarks to give a baseline performance number and justify subsequent PRs
2. Optimizations to `BitWriter`
3. New version of `write_block`
4. New version of `encode_coefficient`

Shnatsel:
This is superseded by #2636 which replaced the in-tree JPEG encoder with the `jpeg-encoder` crate.

PR #1920, opened 2023-04-30
Overhaul Pixel methods on `ImageBuffer` and `GenericImage`/`GenericImageView` traits

okaneco:
_#1853 for background_

Best viewed commit by commit. Most of the changes are in the first commit.

This PR adds `Option`-returning methods to the main image traits. Duplicate methods that `ImageBuffer` implemented were removed.

I migrated functions that used the old methods like `get_pixel` internally to the newer `Option` methods where applicable so they could eventually be phased out. Where possible, I kept the old behavior by using `Index` access on `ImageBuffer`s. Since this is a breaking change, I also removed the deprecated functions from the traits.

- Add `pixel[_unchecked]` to GenericImageView
- Add `pixel_mut[_unchecked]` to GenericImage
- Remove `get_pixel_mut`, `blend_pixel` from GenericImage
- Remove `get_pixel[_checked]`, `get_pixel_mut[_checked]` from ImageBuffer
- Move the inner implementation of `get_pixel[_mut]` to Index, IndexMut
- Add `pixel_i64` to GenericImageView, `pixel_mut_i64` to GenericImage

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ let (width, height) = img.dimensions();
 let pixel = img[(100, 100)];
 
 // Or use the `get_pixel` method from the `GenericImage` trait.
-let pixel = *img.get_pixel(100, 100);
+let pixel = img.get_pixel(100, 100);
 
 // Put a pixel at coordinate (100, 100).
 img.put_pixel(100, 100, pixel);
@@ -218,7 +218,7 @@ fn main() {
                 i += 1;
             }
 
-            let pixel = imgbuf.get_pixel_mut(x, y);
+            let pixel = &mut imgbuf[(x, y)];
             let image::Rgb(data) = *pixel;
             *pixel = image::Rgb([data[0], i as u8, data[2]]);
         }
--- a/examples/fractal.rs
+++ b/examples/fractal.rs
@@ -2,6 +2,8 @@
 extern crate image;
 extern crate num_complex;
 
+use image::{GenericImage, ImageBuffer, Rgb};
+
 fn main() {
     let imgx = 800;
     let imgy = 800;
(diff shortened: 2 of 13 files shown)


fintelia:
There's a lot going on in this PR and it will potentially break a lot of existing user code. Thanks for putting this all together, but in its current form, it is probably too big to review and merge.

Some more detailed thoughts:
* `get_pixel_mut` and `blend_pixel` are already deprecated. Makes sense to remove them.
* Not sure we should be adding `pixel_mut[_unchecked]` without fixing the problems with `get_pixel_mut`.
* The `pixel[_mut]_i64` changes seem separate from the rest. Not yet convinced they're worth adding, but either way they would be better as a standalone PR.
* I'm not really convinced by the Index / IndexMut changes. I haven't seen that pattern used much in Rust, and for such sweeping changes I want to be very sure we're making things more idiomatic rather than less

okaneco:
I understand, also don't want to cause too much unnecessary breakage. I'm not sure if sealed extension traits would make sense for forward compatibility and being able to implement newer methods like these.

- I'd like to see the problems fixed with `get_pixel_mut` but not sure how that can be solved for different containers or layouts.

- Agree that the i64 can be a separate change.

- The indexing method was listed earlier in the readme for the buffer example, so I never saw it as unidiomatic but that might not be a commonly held opinion.
https://github.com/image-rs/image/blob/139420ddfcf5d4b937987a09769e5f201b7a2a29/README.md?plain=1#L101-L105

During this PR, I remembered when I was first learning Rust and how I struggled with understanding traits and this crate. Sometimes I had to dereference `get_pixel` and sometimes not, eventually understanding the problem. 
```rust
// ImageBuffer
fn get_pixel(&self, x: u32, y: u32) -> &P
// GenericImageView
fn get_pixel(&self, x: u32, y: u32) -> P
```
The readme needs to be corrected because it's not using the trait method but the `ImageBuffer` version so it has to dereference on L105. The current state can be confusing to new users but anything to address or deprecate the method will likely cause a lot of churn with CI warnings.
https://docs.rs/image/latest/image/struct.ImageBuffer.html#method.get_pixel
https://docs.rs/image/latest/image/struct.ImageBuffer.html#method.get_pixel-1

PR #1924, opened 2023-05-05
webp: Decode animations frame by frame

sophie-h:
Previously, animated WebPs could take some time
until they started playing and would consume more memory.

### DIFF
--- a/src/codecs/webp/extended.rs
+++ b/src/codecs/webp/extended.rs
@@ -1,5 +1,5 @@
 use std::convert::TryInto;
-use std::io::{self, Cursor, Error, Read};
+use std::io::{self, Cursor, Error, Read, Seek};
 use std::{error, fmt};
 
 use super::decoder::{
@@ -68,7 +68,8 @@ pub(crate) struct WebPExtendedInfo {
 #[derive(Debug)]
 enum ExtendedImageData {
     Animation {
-        frames: Vec<AnimatedFrame>,
+        frames: Vec<Vec<u8>>,
+        first_frame: AnimatedFrame,
         anim_info: WebPAnimatedInfo,
     },
     Static(WebPStatic),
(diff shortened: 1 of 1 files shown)


sophie-h:
We can also drop the "skip invalid frames" part. But since the iterator is silently ending with errors I thought we might as well skip invalid frames.

src/codecs/webp/extended.rs
```diff
@@ -633,7 +674,7 @@ pub(crate) fn read_extended_header<R: Read>(reader: &mut R) -> ImageResult<WebPE
 }
 
 fn read_anim_frame<R: Read>(
-    mut reader: R,
+    mut reader: &mut R,
```
fintelia, line 677: This change shouldn't be needed since there is a blanket `impl<R: Read> Read for &mut R` 😄

src/codecs/webp/extended.rs
```diff
@@ -112,18 +113,45 @@ impl ExtendedImage {
...
-                            )
+                if let ExtendedImageData::Animation {
+                    frames,
+                    anim_info,
+                    first_frame,
+                } = &self.image.image
+                {
+                    // Loop until a valid frame is found to skip invalid frames
+                    loop {
+                        let frame = frames.get(self.index);
```
fintelia, line 124: If you use the `?` operator (i.e. `frames.get(self.index)?`) then there is no need for the match below

fintelia (commented):
This code will change a bunch when we switch to `image-webp`. However, that'll have to wait for a major release (the crate relies on being able to `Seek` the underlying reader). So in the meantime I think this change does make sense.

Relatedly, if you have thoughts about API changes we should make for animation decoding, please do share them!

fintelia:
I don't have strong opinions on how invalid frames should be handled, but one thing to be aware of is that frames can be composited onto the image canvas rather than replacing it. This means that portions of the animation may be left unchanged between frames, and thus skipping a single frame can potentially corrupt the displayed contents for all subsequent frames.

sophie-h:
> I don't have strong opinions on how invalid frames should be handled

I dropped it from this MR. I don't know what the reference implementation does.

fintelia (approved):

PR #1925, opened 2023-05-07
WebP Extended animation format needs to be rendered into frames…

### LINKED ISSUE #1856: Lossy WEBP with alpha: source slice length (761600) does not match destination slice length (997920)
Whenever I try to decode [this webp image](https://make.hmc.edu/img/equipment/cricut.webp) I get the following error:
```
panicked at 'source slice length (761600) does not match destination slice length (997920)'
C:\Users\Ethan Vazquez\.cargo\registry\src\github.com-1ecc6299db9ec823\image-0.24.5\./src\codecs\webp\extended.rs:459:21
```
image

bu5hm4nn:
before it can be copied.
Previous code was copying raw webp frames which can be smaller than the canvas.
Fixes #1856.


src/codecs/webp/extended.rs
```diff
@@ -324,10 +372,13 @@ impl ExtendedImage {
     pub(crate) fn fill_buf(&self, buf: &mut [u8]) {
         match &self.image {
             // will always have at least one frame
-            ExtendedImageData::Animation { frames, .. } => &frames[0].image,
-            ExtendedImageData::Static(image) => image,
+            ExtendedImageData::Animation { .. } => {
+                if let Some(Ok(frame)) = self.as_frames().into_iter().nth(0) {
```
fintelia, line 376: This should probably assert that frames isn't empty, as the old code does

bu5hm4nn: I don't understand, you want it to crash if frames is empty?

fintelia: The current code does `frames[0]`, so it crashes if frames is empty. Which based on the code comment should never happen. It is better to fail loudly if something unexpected happens rather than just silently doing nothing

bu5hm4nn: That is entirely debatable. I default to code that should not crash. (Kind of the idea of Rust me thinks)

I actually don't care at all how it's fixed but this code currently crashes a lot of programs due to the buffer size mismatch and it needs fixed.

fintelia:
The CI seems to detect one failure with `cargo test -v --release`. Should just be a matter of changing the "expected" hashes for any images that are now decoded differently

src/codecs/webp/extended.rs
fintelia: The idea is to _handle_ errors. Anyway, I'd like to avoid taking in a PR that fixes one bug but accidentally conceals another one. So for now, could you add an `assert!` or `expect` or similar? That and the cargo test failure, and I think this PR is ready to merge

src/codecs/webp/decoder.rs
```diff
@@ -321,16 +321,35 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for WebPDecoder<R> {
     }
 
     fn read_image(self, buf: &mut [u8]) -> ImageResult<()> {
-        assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
```
fintelia, line 324: Let's not get rid of this assert. The spec for read_image [guarantees a panic](https://docs.rs/image/latest/image/trait.ImageDecoder.html#panics) if this condition doesn't hold.

Theoretically the values of `{vp8_frame, lossless_frame, extended}.get_buf_size()` should match, but there could be bugs

sophie-h:
I still think that the whole concept of the WebP animation decoder, decoding all frames before being able to show them, is fundamentally broken for image viewers. See #1924 and https://gitlab.gnome.org/GNOME/Incubator/loupe/-/issues/160

Not sure if it's in the scope of this MR to fix it, but a bunch of things happening in this MR would have to be changed again when adding frame-by-frame decoding.

src/codecs/webp/extended.rs
```diff
@@ -147,6 +147,54 @@ impl ExtendedImage {
         Frames::new(Box::new(frame_iter))
     }
 
+    pub(crate) fn as_frames<'a>(&'a self) -> Frames<'a> {
```
sophie-h, line 150: Thanks for adding the `as_frames` API. That's very useful. However, does the code have to be duplicated with `into_frames`, or could `into_frames` cust call `as_frames`?

sophie-h: @fintelia I think having `as_frames` for #1869 would be good. Should it maybe take a `&mut self` just to be safe?

fintelia:
Superseded by #1959

Issue #1905, opened 2023-04-10
Using values in `[0,1] : f32` to sample image?


JulianKnodt:
Forgive me, I'm pretty sure this operation was already asked for, but I was wondering if there's an API that supports sampling from an image (similar to [texture in OpenGL](https://registry.khronos.org/OpenGL-Refpages/gl4/html/texture.xhtml). I suspect this was not added because it's not clear what the behavior between pixels should be, but similar to `FilterType`, I think it would make sense to add an enum for how to perform interpolation.

If I missed existing functionality, is it possible to reference it in the docs of `get_pixel`? If not, is there a reason it does not exist?

Overall, I'm trying to achieve resolution independent sampling. I _could_ implement it where I require it, but I imagine this is probably a common operation that needs to be performed.

JulianKnodt:
hmmmm I see, I can propose some design in a PR, at least for the implementation of bit depth and pixel types.
I think the general design is that we want to have some sampling filter (nearest, bilinear, others),

JulianKnodt:
🎉 I'm not going to close this yet, will also add nearest sampling

PR #1929, opened 2023-05-16
Add sampling using [0,1] for `GenericImageView`

JulianKnodt:
***WIP***
Not sure whether this should directly go on `GenericImageView`, and where the enums should go.
Also need to add tests but put the general idea up for review.

---

Adds a way to sample images in [0,1], which is a common use case in graphics.

Currently adds non-exhaustive enums based on OpenGL's that allow for simple sampling of nearby coordinates.

Nearest sampling can produce artifacts depending on the sampling resolution of the UV whereas bilinear will smoothly interpolate.

Addresses #1905

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -304,6 +304,101 @@ where
     out
 }
 
+/// Linearly bisample from an image using coordinates in [0,1].
+pub fn sample_bilinear<P: Pixel>(
+    img: &impl GenericImageView<Pixel = P>,
+    u: f32,
+    v: f32,
+) -> Option<P> {
+    if ![u, v].iter().all(|c| (0.0..=1.0).contains(c)) {
+        return None;
+    }
+
+    let (w, h) = img.dimensions();
+    if w == 0 || h == 0 {
+        return None;
+    }
+
+    let ui = w as f32 * u - 0.5;
+    let vi = h as f32 * v - 0.5;
+    interpolate_bilinear(
+        img,
+        ui.max(0.).min((w - 1) as f32),
+        vi.max(0.).min((h - 1) as f32),
+    )
+}
+
+/// Linearly bisample from an image using coordinates in [0,w-1] and [0,h-1].
+pub fn interpolate_bilinear<P: Pixel>(
+    img: &impl GenericImageView<Pixel = P>,
+    x: f32,
+    y: f32,
+) -> Option<P> {
+    let (w, h) = img.dimensions();
+    if w == 0 || h == 0 {
+        return None;
+    }
+    if !(0.0..=((w - 1) as f32)).contains(&x) {
+        return None;
+    }
+    if !(0.0..=((h - 1) as f32)).contains(&y) {
+        return None;
+    }
+
+    let uf = x.floor();
+    let vf = y.floor();
+    let uc = (x + 1.).min((w - 1) as f32);
+    let vc = (y + 1.).min((h - 1) as f32);
+
+    // clamp coords to the range of the image
+    let coords = [[uf, vf], [uf, vc], [uc, vf], [uc, vc]];
+
+    assert!(coords
+        .iter()
+        .all(|&[u, v]| { img.in_bounds(u as u32, v as u32) }));
+    let samples = coords.map(|[u, v]| img.get_pixel(u as u32, v as u32));
+    assert!(P::CHANNEL_COUNT <= 4);
+
+    // convert samples to f32
+    // currently rgba is the largest one,
+    // so just store as many items as necessary,
+    // because there's not a simple way to be generic over all of them.
+    let [sff, sfc, scf, scc] = samples.map(|s| {
+        let mut out = [0.; 4];
+        for (i, c) in s.channels().iter().enumerate() {
+            out[i] = c.to_f32().unwrap();
+        }
+        out
+    });
+    // weights
+    let [ufw, vfw] = [x - uf, y - vf];
+    let [ucw, vcw] = [1. - ufw, 1. - vfw];
+
+    // https://en.wikipedia.org/wiki/Bilinear_interpolation#Weighted_mean
+    // the distance between pixels is 1 so there is no denominator
+    let wff = ucw * vcw;
+    let wfc = ucw * vfw;
+    let wcf = ufw * vcw;
+    let wcc = ufw * vfw;
+    assert!(f32::abs((wff + wfc + wcf + wcc) - 1.) < 1e-3);
+
+    // hack to get around not being able to construct a generic Pixel
+    let mut out = samples[0];
+    for (i, c) in out.channels_mut().iter_mut().enumerate() {
+        let v = wff * sff[i] + wfc * sfc[i] + wcf * scf[i] + wcc * scc[i];
+        // this rounding may introduce quantization errors,
+        // but cannot do anything about it.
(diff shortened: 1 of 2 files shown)


fintelia (commented):
Looking at these changes, I think adding bilinear sampling to this crate is valuable but I worry that a generic sample method is going to add too much complexity. There's just so many different ways to sample an image with a `(f32, f32)`. For instance, there's the question of whether (0,0) is the center of the top left pixel (like in this PR) or the corner of that pixel (like the usual convention in OpenGL/Vulkan/etc.) and that's on top of questions about wrapping mode and filtering method. And no single option is "right"

What would you think of adding just a `bilinear_sample` method that always clamped to edge and used the same coordinates as `get_pixel` (ie from `0` to `size-1` with integer values being pixel centers). It shouldn't be too hard for users to implement wrapping or normalized coordinates based on that if they want

JulianKnodt:
I think that's definitely a step forward.
I do think that being able to switch between different sampling modes is very convenient, so I think keeping the SampleKind is good, and would let additional sample kinds be added more easily in the future.

I think that clamping against the edge quietly is probably going to lead to unexpected behavior for users. Instead, we should probably return an `Option<Pixel>`, and that will be immediately clear to the user that they've sampled out of bounds. As long as we guarantee that all values in `[0,1]` do not return `None`.

JulianKnodt:
Just wanted to bump this, while I am pushing back on some ideas you suggested, if you feel strongly about them I will change it. I would rather have sampling in some form than no sampling.

fintelia:
Let's go with one method per supported sampling type. If users want to dynamically pick between sampling types, they can implement it themselves, but this way there's no extra branching if the sampling type is known in advance.

Thinking about it, the methods probably belong in the [`imageops`](https://docs.rs/image/latest/image/imageops/index.html) module with all the other methods implemented for `GenericImage[View]`

JulianKnodt:
Sounds good, it may take me a bit to implement but I'll add them in.

src/imageops/sample.rs
```diff
@@ -304,6 +304,71 @@ where
     out
 }
 
+/// Linearly bisample from an image using coordinates in [0,1].
+pub fn bilinear_sample<P: Pixel>(
+    img: &impl GenericImageView<Pixel = P>,
+    uv: [f32; 2],
+) -> Option<P> {
+    if !uv.iter().all(|c| (0.0..=1.0).contains(c)) {
```
fintelia, line 312: This method currently treats the center of the top left pixel as `(0, 0)`, while 3D graphics APIs generally specify that the center of that pixel is at `(0.5 / width, 0.5 / height)`.

Let's have the input range go from `(0.0, 0.0)` to `(width-1, height-1)` instead. That way we sidestep the question and avoid confusion. Conveniently it is already the range needed for `ui` and `vi`. 

In any case, there should be more documentation on the method explaining exactly how the coordinates map to pixels.

src/imageops/sample.rs
```diff
@@ -304,6 +304,71 @@ where
...
+
+    let max: f32 = P::Subpixel::DEFAULT_MAX_VALUE.to_f32().unwrap();
+    // convert samples to f32
+    // currently rgba is the largest one,
+    // so just store as many items as necessary,
+    // because there's not a simple way to be generic over all of them.
+    let [s0, s1, s2, s3] = samples.map(|s| {
+        let mut out = [0.; 4];
+        for (i, c) in s.channels().iter().enumerate() {
+            out[i] = c.to_f32().unwrap() / max;
```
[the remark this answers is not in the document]
JulianKnodt: that's true, I'm used to having colors in the range [0,1], so I just did it instinctively

src/imageops/sample.rs
JulianKnodt: do you mean the input range for uv? the point of having it be in the range 0-1 was so that it could be independent of the resolution of the image. Or do you mean the validation should be done on `ui` and `vi`?

This is intended to closely mirror the API of glsl's `texture` function. If we instead make the function coordinates in the  range `(0, w/h-1)`, then any user who operates in `0,1` space will have to multiply by the coordinate indices, which is a bit of a pain and can lead to subtle off by 1 errors (multiplying by w instead of w-1). My use case for these functions is I'm using it to sample textures from a mesh, and for that coordinates are entirely in 0-1 space

src/imageops/sample.rs
```diff
@@ -304,6 +304,71 @@ where
...
+        }
+        out
+    });
+    // weights
+    let [ufw, vfw] = [ui - uf, vi - vf];
+    let [ucw, vcw] = [1. - ufw, 1. - vfw];
+
+    let u02 = add2(s0.map(|c| c * ucw), s2.map(|c| c * ufw));
+    let u13 = add2(s1.map(|c| c * ucw), s3.map(|c| c * ufw));
+    let interp = add2(u02.map(|c| c * vcw), u13.map(|c| c * vfw));
```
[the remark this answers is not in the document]
JulianKnodt: Mm not quite sure what you mean? Do you mean just interpolate the weights then multiply and sum those?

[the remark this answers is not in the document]
fintelia: Yeah, I think that might produce more understandable code

src/imageops/sample.rs
fintelia: Hm, matching GLSL's `texture` function does sound useful, especially given all the subtlety around pixel corners vs. centers. We could break the method into two pieces:
* `interpolate_bilinear` which used coordinates from (0, 0) to (width-1, height-1), and contained the actual interpolation logic, and
* `sample_bilinear` which uses (0, 0) to (1, 0) coordinates matching GLSL style, and is implemented by calling into the first method.

fintelia: Also, the proper conversion from [0,1] -> [0, width-1] is `(u * width - 0.5).max(0.0).min(width-1)`

src/imageops/sample.rs
```diff
@@ -891,6 +958,26 @@ mod tests {
         assert!(img.pixels().eq(resize.pixels()))
     }
 
+    #[test]
+    #[cfg(feature = "png")]
+    fn test_bilinear_sample() {
+        use std::path::Path;
+        let img = crate::open(&Path::new("./examples/fractal.png")).unwrap();
+        assert!(sample_bilinear(&img, 0., 0.).is_some());
```
fintelia, line 966: Would probably be good to also have a few test that make sure the interpolation produces the expect values

JulianKnodt:
Cheers, was in Alaska but have fixed most of the comments

src/imageops/sample.rs
```diff
@@ -891,6 +986,60 @@ mod tests {
...
+        assert!(sample_bilinear(&img, 1.2, 0.5).is_none());
+        assert!(sample_bilinear(&img, 0.5, 1.2).is_none());
+        assert!(sample_bilinear(&img, 1.2, 1.2).is_none());
+
+        assert!(sample_bilinear(&img, -0.1, 0.2).is_none());
+        assert!(sample_bilinear(&img, 0.2, -0.1).is_none());
+        assert!(sample_bilinear(&img, -0.1, -0.1).is_none());
+    }
+    #[test]
+    #[cfg(feature = "png")]
```
fintelia, line 1009: nit: PNG feature isn't needed for this test

fintelia (approved):

JulianKnodt:
@fintelia You're gonna have to merge it, I don't have the ability to do so

PR #1935, opened 2023-05-27
Add to_mime_type for ImageFormat

### LINKED ISSUE #1722: Would ImageFormat::to_mime_type be valuable for this crate?
This crate already provides a `ImageFormat::from_mime_type` function, but there is no counterpart (like `to_mime_type`).

There are circumstances this would be very handy, like image manipulation web services, which need to provide MIME types as request/response headers.

If maintainers agree but don't have time, I can propose a PR.

## Draft

```rust
impl ImageFormat {
  // <snip>

  pub fn to_mime_type(&self) -> Option<&'static str>
    {
      match self {
        ImageFormat::Png => Some("image/png"),
        // <snip>
    }
}
```

101100:

This adds `to_mime_type` as suggested in #1722.

I attempted to imitate the style of other comments, including having an example.

Since I found a MIME type for all of the formats, I made the method not return an `Option`, but I can change that if you prefer.

Fixes #1722.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -187,6 +187,44 @@ impl ImageFormat {
         }
     }
 
+    /// Return the MIME type for this image format.
+    ///
+    /// Sometimes multiple MIME types exist for a format, so this method attempts to return the one
+    /// used most often.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use image::ImageFormat;
+    ///
+    /// let mime_type = ImageFormat::Png.to_mime_type();
+    /// assert_eq!(mime_type, "image/png");
+    /// ```
+    pub fn to_mime_type(&self) -> &'static str {
+        match self {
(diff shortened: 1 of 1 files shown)


fintelia (changes requested):
This looks good to me. Could you just move this method so that it is directly above/below `from_mime_type`? That'll make it easier to keep them in sync

101100:
@fintelia OK. I've moved the method. Not sure if you were looking while I was making the other change away from `Option`, though.

fintelia:
Moving away from option seems good, though could you add a note in the doc comment that "application/octet-stream" is used as a fallback?

101100:
I've added a mention of the fallback MIME type as well as a description for all of the ambiguous ones and which is returned.

fintelia (approved):

101100:
@fintelia Thanks for the quick review and merge!

Issue #1936, opened 2023-05-29
`.hdr` images decoded as LDR.


pema99:
In my renderer, I was using the following code to load radiance HDR image (`.hdr` extension), but noticed that all values were getting clamped to [0; 1] range:

```rs
pub fn load_dynamic_image(path: &str) -> Option<DynamicImage> {
    image::io::Reader::open(path).ok()?.decode().ok()
}
```

I was able to fix it by adding a special case:
```rs
pub fn load_dynamic_image(path: &str) -> Option<DynamicImage> {
    // Image crate does not by default decode .hdr images as HDR for some reason
    if path.ends_with(".hdr") {
        let hdr_decoder = image::codecs::hdr::HdrDecoder::new(std::io::BufReader::new(
            std::fs::File::open(&path).unwrap(),
        )).ok()?;
        let width = hdr_decoder.metadata().width;
        let height = hdr_decoder.metadata().height;
        let buffer = hdr_decoder.read_image_hdr().ok()?;
        return Some(DynamicImage::ImageRgb32F(image::ImageBuffer::from_vec(
            width,
            height,
            buffer.into_iter().flat_map(|c| vec![c[0], c[1], c[2]]).collect(),
        )?));
    }

    image::io::Reader::open(path).ok()?.decode().ok()
}
```

I noticed it seems to be caused by this line `HdrAdapter::read_image_data`:
```rs
fn read_image_data(&mut self, buf: &mut [u8]) -> ImageResult<()> {
        assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
        match self.inner.take() {
            Some(decoder) => {
                let img: Vec<Rgb<u8>> = decoder.read_image_ldr()?; // <-- clamp to LDR
                ...
```
I guess the simple fix would just be to swap the call to `read_image_hdr` instead. I'm mainly opening this bug report to verify if this is intentional or indeed a bug. It seems like too obvious of a mistake to me.

## Expected
`.hdr` image is decoded as HDR values, nothing is clamped.

## Actual behaviour
`.hdr` image is decoded such that all pixels above 1.0 are clamped to 1.0.

## Reproduction steps
Run the code from the first snippet, passing a path to an `.hdr` image. Read the pixel values and observe all will be clamped. Attaching the image file I used to test.
https://dl.polyhaven.org/file/ph-assets/HDRIs/hdr/2k/clarens_midday_2k.hdr

fintelia:
[...]

PR #1937, opened 2023-05-29
Fix .hdr decoding

pema99:
Fixes https://github.com/image-rs/image/issues/1936. First time contributor, so unsure if this is the right way to fix, but in my testing it seemed to work and causes .exr and .hdr image files to have the same behavior when decoding.

### DIFF
--- a/src/codecs/hdr/decoder.rs
+++ b/src/codecs/hdr/decoder.rs
@@ -157,9 +157,9 @@ impl<R: BufRead> HdrAdapter<R> {
         assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
         match self.inner.take() {
             Some(decoder) => {
-                let img: Vec<Rgb<u8>> = decoder.read_image_ldr()?;
+                let img: Vec<Rgb<f32>> = decoder.read_image_hdr()?;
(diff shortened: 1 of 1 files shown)


johannesvollmer:
I think the test fails, because the reference image are u8 values that are gamma corrected, but f32 values from the hdr decoder are linear. This means the image returned by the hdr decoder is too dark, and therefore comparison with the reference png fails.

- hdr decoder returns linear f32 values
- test reference png image has non-linear u8 values
- test suite currently can't compare f32 with u8 images (easily fixable)

In #1599, the solution was to make the hdr decoder return non-linear f32 values. However, in hindsight, it seems weird, as in my experience, one would always f32 values to be linear, as this generally seems to be the convention in image processing.

In addition, the tests fail because they can't compare the existing u8 png images with the f32 images from the hdr decoder. This means we must adjust the tests. Maybe the best would be to make the tests perform gamma conversion before comparing the images? Maybe DynamicImage<u8>::to_f32() should perform gamma conversion?! 

Maybe we need to define within the image crate for f32 images whether they are always gamma or linear?

fintelia:
Merged via #2150

PR #1938, opened 2023-05-30
Add `track_caller` attribute on `get_pixel`, `get_pixel_mut`, `put_pixel`

MrGunflame:
The `track_caller` attribute allows tracking the panic location of the caller, instead of location where the panic was invoked, giving more useful panic information.


HeroicKatora (approved):
LGTM. Faintly remember not adding it due to MSRV at some point but that concern no longer applies. Thanks :+1:

PR #1942, opened 2023-06-09
Fix to prevent blocking file::open that may happen on some file types. Replacing File::Open() with OpenOptions()

visnkmr:
While iterating over a list of files and trying to list dimensions if images. I found that the execution was indefinitely stuck in the part of code that used image crate, I used img::open(), instead of img::image_dimensions(). Upon further lookup found that File::open was the culprit as it can block execution when opening a pipe file since it blocks until there is some data to read. If the pipe is empty or not ready, the program will wait indefinitely.  To prevent this we can pass custom flags to OpenOptions, which is what File::Open is an abstraction of. 
The code changes replace File::open() with OpenOptions(). The changes allow for optionally setting custom flags when opening a file using : 
image::open_with_custom_flags()
image::image_dimensions_with_custom_flags()
reader::open_with_custom_flags()
functions that take a path and customflag as arguments. 
This has been done in a way so as to not introduce any braking changes.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -1097,7 +1097,21 @@ where
     P: AsRef<Path>,
 {
     // thin wrapper function to strip generics before calling open_impl
-    free_functions::open_impl(path.as_ref())
+    free_functions::open_impl_with_custom_flags(path.as_ref(),0)
+}
+/// Open the image located at the path specified with the flags specified.
+/// The image's format is determined from the path's file extension.
+///
+/// Try [`io::Reader`] for more advanced uses, including guessing the format based on the file's
+/// content before its path.
+///
+/// [`io::Reader`]: io/struct.Reader.html
+pub fn open_with_custom_flags<P>(path: P,custom_flags:i32) -> ImageResult<DynamicImage>
+where
+    P: AsRef<Path>,
+{
+    // thin wrapper function to strip generics before calling open_impl
+    free_functions::open_impl_with_custom_flags(path.as_ref(),custom_flags)
 }
 
 /// Read a tuple containing the (width, height) of the image located at the specified path.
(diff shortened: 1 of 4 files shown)


sophie-h:
Just a tangent: When you say pipe file, are you talking of something like [UnixStream](https://doc.rust-lang.org/std/os/unix/net/struct.UnixStream.html)? Because strictly you can't use those with `io::Reader` since some decoders require `Seek` for the source.

fintelia:
Is there a reason you cannot open the file yourself and then pass the reader to `io::Reader::new`?

visnkmr:
> Just a tangent: When you say pipe file, are you talking of something like [UnixStream](https://doc.rust-lang.org/std/os/unix/net/struct.UnixStream.html)? Because strictly you can't use those with `io::Reader` since some decoders require `Seek` for the source.

Yes something like Unixstream, I encountered the issue when i tried to open polybar's message queue file while iterating over files in ramdisk, this happened because i tried image::open() on each file in a parallel iterator which never returned anything seemed like an infinite loop @ first later realised it was a file blocking the execution.

> Is there a reason you cannot open the file yourself and then pass the reader to `io::Reader::new`?
> 
> Also, since I'm not familiar with the details of OpenOptions, what specific option do you intend to pas to solve your usecase?

Yes opening the file and passing the reader to io::Reader::new() does not cause the issue, the issue is that the normal entry point of image::open() does not error out when trying to read files that blocks execution. For my use case image::dimensions() would suffice for now since it checks file type before opening.

File::open() is an alias of OpenOptions(), in my case setting the custom flag to libc::O_NONBLOCK or 2048 flag fixed the issue when opening a file, it makes the file operation return immediately if the file is not ready or available, instead of blocking until it is. In the my case erroring out immediately is the desired behaviour which this provided.

fintelia:
Thanks for the suggestion! Since it sounds like the same functionality can already be achieved with our existing `io::Reader::new` and `io::Reader::into_dimensions` API, I'm going to go ahead and close this. Please feel free to ask questions if you have trouble getting that to work.

PR #1945, opened 2023-06-13
JXL: initial support

Quackdoc:
The work done here was entirely @EugeneVert, I just cleaned it up a bit and intend to implement missing features at some point, this is a draft because it needs cleaned up and is missing a few features 

- [ ] Animation
- [ ] CMYK (this may be handled internally by jxl-oxide in the future)

jxl-oxide also has some issues decoding large image's often running into OOM issues if the image is large enough. 

currently it's usable enough for a good chunk of images out there. decided to post it both to get comments on the current state as well since I'm using it in it's current state, it may as well be in the open

### DIFF
--- a/src/codecs/jxl/decoder.rs
+++ b/src/codecs/jxl/decoder.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use std::io::{Cursor, Read};
 
-use jxl::{JxlImage, PixelFormat, Render};
+use jxl_oxide::{JxlImage, PixelFormat, Render};
 
 /// JPEG XL decoder
 pub struct JxlDecoder<R> {
@@ -21,11 +21,12 @@ pub struct JxlDecoder<R> {
 
 impl<R: Read> JxlDecoder<R> {
     /// Creates a new decoder that decodes from the stream ```r```
+     
     pub fn new(r: R) -> ImageResult<JxlDecoder<R>> {
         let mut image = JxlImage::from_reader(r).map_err(|_| {
             ImageError::Decoding(DecodingError::new(
                 ImageFormat::Jxl.into(),
-                "Failsed to parse image",
+                "Failed to parse image",
             ))
         })?;
         let mut keyframes = Vec::new();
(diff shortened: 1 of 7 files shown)


EugeneVert:
I license past and future contributions to image-rs under the dual MIT/Apache-2.0 license,

src/codecs/jxl/decoder.rs
```diff
@@ -0,0 +1,163 @@
...
+        self.colortype
+    }
+
+    fn into_reader(self) -> ImageResult<Self::Reader> {
+            let data = decoder_to_vec(self)?;
+            Ok(Cursor::new(data))
+    }
+
+    fn icc_profile(&mut self) -> Option<Vec<u8>> {
+        self.image.embedded_icc().map(Vec::from)
```
tirr-c, line 111: Decoded images are in `decoded_icc()` profile, but it seems like `JxlImage` lacks the method. I'll fix that in the next release of jxl-oxide.

fintelia (commented):
Thanks for opening this PR, it is off to a good start!

I left some minor comments, but the one big thing I'd really like to see before we enable jxl by default is to implement [`ImageDecoder::set_limits`](https://docs.rs/image/latest/image/trait.ImageDecoder.html#method.set_limits). Supporting resource limits for decoding can be rather tricky to implement if you don't have it right from the start (as we're unfortunately finding with some of our other decoders) but we can't really do fuzzing without them

src/codecs/jxl/decoder.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+impl<R: Read> JxlDecoder<R> {
+    /// With limits
+     
+    pub fn new(r: R) -> ImageResult<JxlDecoder<R>> {
+        Self::with_limits(r, Limits::default())
+    }
+    /// Creates a new decoder that decodes from the stream ```r```
+    pub fn with_limits(r: R, limits: Limits) -> ImageResult<JxlDecoder<R>> {
```
fintelia, line 31: If possible, implementing `set_limits` should be preferred over `with_limits`

Quackdoc:
sorry haven't had a lot of time to work on this, ill try to make an update soon if I get the time

markg85:
This issue is closed but it doesn't look like it's merged. Was there an issue left that requires resolving before this can be merged?

fintelia:
The main blocker is described in #1979. Another recent development is the work on `jxl-rs`

markg85:
@fintelia Ahh, I see! I hope the JPEG XL format is going to be open (or open enough) to be included in this library at some point in time.

Performance wise (just using the `jpegxl_rs` crate) i can say that it's very fast! For example, using this `image` crate and just resizing a jpg image to a smaller version (in my test 4k to 400x400) is orders of magnitude faster in jpegxl_rs. However, when using  `turbojpeg` then the scale tips again, it's 2x faster then jxl.

By which i just mean to say that `jpegxl_rs` is quite fast!

PR #1951, opened 2023-06-24
Update CI workflow

fintelia:
Update CI workflow to use https://github.com/dtolnay/rust-toolchain rather than the [unmaintained](https://github.com/actions-rs/toolchain/issues/216) actions-rs.

Also switches a few places `ubuntu-20.04` -> `ubuntu-22.04`.


fintelia:
Seems a transitive dependency just updated their MSRV to `1.70.0` (released earlier this month!) and I don't have the energy to deal with it tonight. I'll get back to this at some point

PR #1955, opened 2023-07-01
Escape types more correctly in hdr docstrings

linkmauve:
Some types weren’t in backquotes, making `rustdoc` interpret them as unclosed HTML tags.

Also unify the backquote format to a single one for single words.


fintelia:
Thanks!

PR #1962, opened 2023-07-18
Add sample/interpolate nearest

JulianKnodt:
This looks very similar to the prior PR, but I'm not sure if the two sample modes should be merged into one?
That was originally why I had included an enum to indicate which to perform, but it's fine to keep them separate as well.

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -304,7 +304,7 @@ where
     out
 }
 
-/// Linearly bisample from an image using coordinates in [0,1].
+/// Linearly sample from an image using coordinates in [0,1].
 pub fn sample_bilinear<P: Pixel>(
     img: &impl GenericImageView<Pixel = P>,
     u: f32,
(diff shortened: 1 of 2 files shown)


src/imageops/sample.rs
```diff
@@ -328,6 +328,50 @@ pub fn sample_bilinear<P: Pixel>(
...
+    v: f32,
+) -> Option<P> {
+    if ![u, v].iter().all(|c| (0.0..=1.0).contains(c)) {
+        return None;
+    }
+
+    let (w, h) = img.dimensions();
+    if w == 0 || h == 0 {
+        return None;
+    }
```
fintelia, line 344: This check is redundant with the one below and can be removed
```suggestion
```

src/imageops/sample.rs
```diff
@@ -328,6 +328,50 @@ pub fn sample_bilinear<P: Pixel>(
...
+    if !(0.0..=((w - 1) as f32)).contains(&x) {
+        return None;
+    }
+    if !(0.0..=((h - 1) as f32)).contains(&y) {
+        return None;
+    }
+
+    Some(img.get_pixel(x.round() as u32, y.round() as u32))
```
fintelia, line 372: I think we should be rounding before checking whether the index is in bounds. For instance, (-0.3, -0.3) should round to (0,0) and thus be accepted

src/imageops/sample.rs
JulianKnodt: It's not completely redundant, in debug mode it will cause a panic due to overflow in the following lines with `w-1`, but if that's changed to a `saturating_sub(1)` then it's fine to remove

src/imageops/sample.rs
JulianKnodt: If that's the case then, I think that we should also do the same for bilinear interpolation, otherwise it would lead to inconsistent boundary conditions.

Not that we have to go with the convention of glsl, but for reference
https://www.khronos.org/opengl/wiki/Sampler_Object#:~:text=If%20GL_NEAREST%20is%20used%2C%20then,between%20the%20nearest%20adjacent%20samples.

If a user were to implement wrapping outside of this, then `-0.3` could be considered as both `0` and `w-1`

fintelia: Actually, I'm realizing that `sample_linear` is already inconsistent. Bilinear sampling at uv coordinates (0,0) should produce the top left pixel's color if doing "clamp-to-edge" and should be the average of the four corner pixels for "wrapping".

JulianKnodt: you mean inconsistent with interpolate_bilinear or sample_nearest?

It only clamps to edge to handle numerical inconsistencies, but both methods should return none if it's out of bounds.

fintelia: I mean that `sample_linear` already can't be used to implement wrapping sample, since not matter what arguments the user calls it with, they won't be able to interpolate between pixels on opposite sides of the image.

JulianKnodt: That's true... One alternative is to provide a separate function but that seems not as clean (akin to wrapping add, saturating, checked).
There was also providing an enum as an option, which forces the user to think about it, but may be unfriendly to people learning the library.

I looked at how Vulkan handles sampling, and they have an object with a number of settings to sample from an image. Maybe it makes sense to attach these to an object which has sensible defaults (clamp to border), but can be changed if the user needs precise control. It will also be easier to modify in the future from a backcompat perspective if new features need to be added.

PR #1965, opened 2023-07-19
Remove some allocations from HDR Decoder

### LINKED ISSUE #1882: Unnecessary `memset()`+`memcpy()` in HDR decoding
As of image 0.24.5, the HDR decoding needlessly copies the entire output in memory, acquiring a `Vec` and then calling `copy_from_slice` to create another `Vec`:

https://github.com/image-rs/image/blob/55732d14434cf639f2e79890f43a1d48bfb8617b/src/codecs/hdr/decoder.rs#L156-L171

The following function can directly return the `Vec` it gets from the underlying decoder:

```rust
    /// Read the actual data of the image, and store it in Self::data.
    fn get_image_data(&mut self, buf: &mut [u8]) -> ImageResult<Vec<u8>> {
        assert_eq!(u64::try_from(buf.len()), Ok(self.total_bytes()));
        match self.inner.take() {
            Some(decoder) => {
                let img: Vec<Rgb<u8>> = decoder.read_image_ldr()?;
                Ok(bytemuck::allocation::cast_vec(img))
            }
            None => Err(ImageError::Parameter(ParameterError::from_kind(
                ParameterErrorKind::NoMoreData,
            ))),
        }
    }
```
..or it could if `image`'s `Rgb` struct implemented `bytemuck::Pod`. Right now it doesn't, which causes the above to fail to compile. I understand this would require an `unsafe impl` since a `#[derive]` won't work on a generic struct.

Once the `Pod` implementation is sorted, the HDR decoding can be specialized to return the underlying Vec where possible similar to how it's done for JPEG in #1881

JulianKnodt:
A larger change was proposed in #1599,
but this is intended to fix #1882 precisely.

It adds extra trait implementations to `Pixels`, and removes the allocation while keeping the API the same.


fintelia:
I've only skimmed it, but this PR looks kinda hard to review because moves around a bunch of finicky code. 

One thing that particularly stands out to me though is that it uses a macro to generate generic, unsafe trait impls, on a public type. That's both a public API addition and inserts new unsafe code to the library. Potentially hard to justify for a minor perf improvement

JulianKnodt:
The alternative is to create a clone of the `image_transform` which instead returns the colors in place.

Other than exposing an `unsafe impl` directly to the users, it's also possible to directly transmute the underlying types at the point where the bytemuck cast was added. I wasn't sure if that would be more or less preferred.

fintelia:
Looking though this now, the PR basically just changes a memcpy into an unsafe transmute? That's not an improvement IMO. We'd ideally like this crate to be `#[forbid(unsafe_code)]`, and while we're not quite there yet, I'd really rather not add any more unsafe in the meantime

JulianKnodt:
For me, as long as the unsafe code is entirely within the confines of the API and with controlled types, then it's fine to introduce. But that's a personal opinion, so I've ripped it out duplicated code instead.

197g:
I'm sorry we never got around to rebasing this or integrating it otherwise. I think it got superseded by #2707.

JulianKnodt:
All good! Glad to see progress however it happens.

PR #1976, opened 2023-08-05
Simplify cargo features exposed by the crate

fintelia:


src/image.rs
```diff
@@ -403,7 +403,7 @@ pub enum ImageOutputFormat {
     /// An Image in BMP Format
     Bmp,
 
-    #[cfg(feature = "farbfeld")]
+    #[cfg(feature = "ff")]
```
kornelski, line 406: This is a cryptic feature name

fintelia: All the image format feature names were changed to be their file extension. However, since approximately nobody uses `.ff` files, that one ends up being cryptic. I'll add a note in the Cargo.toml so at least it'll be clear to people that look there.

(That does raise the question of whether we should even support this format given its obscurity, but someone took the effort to add the feature, so presumably there's at least a bit of interest in keeping it)

PR #1978, opened 2023-08-09
Webp encode

fintelia:
The goal of this PR is to have "good enough" pure-Rust webp encoding to enable it by default and remove our dependency on libwebp.

My plan is to completely ignore lossy webp encoding (we already have pure-Rust lossy jpeg and avif encoders) and only focus on lossless mode. ~~Right now, the code is able to produce valid lossless webp's, but they're completely uncompressed. Actually compressing the input will require building huffman trees based on symbol frequency.~~

I have however implemented hard-coded "subtract green" and "predictor" transforms which should make the data at least somewhat compressible. That's in contrast to libwebp which tries a whole bunch of different transform configurations to see what produces the best compression. Nonetheless, it might be enough to get compression ratios on par with PNG

*Update: Huffman compression is implemented.*

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ qoi = ["dep:qoi"]
 # Enables WebP decoder support.
 webp = []
 # Non-default, not included in `webp`. Requires native dependency libwebp.
-webp-encoder = ["libwebp"]
+webp-encoder = ["libwebp", "webp"]
 
 # Enables multi-threading.
 # Requires latest stable Rust.
--- a/src/codecs/webp/encoder.rs
+++ b/src/codecs/webp/encoder.rs
@@ -482,8 +482,7 @@ impl<W: Write> ImageEncoder for WebPEncoder<W> {
 
 #[cfg(test)]
 mod tests {
-    use crate::codecs::webp::{WebPEncoder, WebPQuality};
-    use crate::{ColorType, ImageEncoder};
+    use crate::ImageEncoder;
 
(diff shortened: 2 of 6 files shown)


fintelia:
Compression ratio on the QOI benchmark suite is currently 29%, compared to about 17% when using libwebp. On the plus side, this encoder is something like 60x faster.

fintelia:
At this point, the compression speed is still about 230M pixels/s, which is right on par with QOI and our fast PNG encoding. And the run length compression got the compression ratio to just under 23%, which is way better than the 27-28% ratios those other encoders get on this corpus. This encoder is coming together better than I was expecting!

sophie-h:
Would it maybe make sense to move all the webp code to a web-rs crate at some point?

fintelia:
I started splitting off the webp decoder into its own crate [here](https://github.com/image-rs/image-webp), but it is proving to be rather involved. I do want to continue with that process, but in the meantime I think it makes sense to go ahead and merge this PR

fintelia:
In the 0.24.9 release, both the pure-Rust webp encoder and the C based encoder are used. The pure-Rust one is used by the high level methods that don't take options, or if you call `WebPEncoder::new_lossless`. The C-based one is used if you call `WebPEncoder::{new, new_with_quality}`.

In 0.25.0-preview.0, only pure-Rust encoding is used. Though it actually relies on a copy of this code that has been moved into the `image-webp` crate.

PR #1986, opened 2023-08-26
Remove `num-rational`

CryZe:
This removes `num-rational` as a dependency. As it turns out, `num-rational`, based on which features are active, can very easily end up on the critical path during compilation, meaning that it is the crate that determines how long it takes to compile the `image` crate. This is because `cargo` can parallelize compilation of many crates, but crates with a long dependency chains (rather than wide), are the ones that in the end determine the compilation speed. This is one of the learnings of the `serde` drama, where it has been noticed that allowing `serde` and `serde_derive` to compile in parallel reduces its long dependency chain, resulting in faster compile times. This introduces the same kind of optimization here, where for all the following features, `num-rational` turns out to be the longest dependency chain:

`png,jpeg,gif,bmp,ico,pnm,tga,tiff,webp,hdr,dxt,dds,farbfeld,qoi`

The critical path is the following:

`autocfg` → `num-traits` compile build.rs → `num-traits` run build.rs → `num-traits` → `num-integer` → `num-rational` → `image`

As it turns out, the only thing really used in `num-rational` is `Ratio`, which can be easily replicated.

This cuts out both `num-integer` and `num-rational` cutting compile times by around 0.5 seconds on a debug build.

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ path = "./src/lib.rs"
 [dependencies]
 bytemuck = { version = "1.7.0", features = ["extern_crate_alloc"] } # includes cast_vec
 byteorder = "1.3.2"
-num-rational = { version = "0.4", default-features = false }
 num-traits = "0.2.0"
 gif = { version = "0.12", optional = true }
 jpeg = { package = "jpeg-decoder", version = "0.3.0", default-features = false, optional = true }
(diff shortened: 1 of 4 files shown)


CryZe:
I didn't port the ratio simplification, but turns out the tests check for it. Should I port it or is it not relevant and the tests can be adjusted?

PR #1987, opened 2023-08-26
Replace MIPS targets with PowerPC

CryZe:
The MIPS targets are broken and partially don't come prebuilt anymore for those reasons. PowerPC is a more stable architecture for big endian.


CryZe:
nvm, looks like mips is broken in general anyway. probably makes sense to switch to powerpc.

fintelia:
Thanks!

PR #2007, opened 2023-09-19
Add PngDecoder::gamma_value method

### LINKED ISSUE #2001: Can `image::codecs::png::PngDecoder` somehow preserve `png::Info::source_gamma`?
My goal is to replace `libpng` with a PNG decoder implemented in memory-safe Rust (e.g. no FFI, limited use of `unsafe`, etc.). One problem I've encountered is replicating gamma processing of the old code:

* The old code [asks `libpng`](https://source.chromium.org/chromium/chromium/src/+/main:ui/gfx/codec/png_codec.cc;l=197-207;drc=ccf1fbee1a326b812933b1524052a0a94af88026) to transform the output pixels using the gamma information from a PNG header.
* In theory Rust exposes gamma information in [`png::Info::gama_chunk`](https://docs.rs/png/latest/png/struct.Info.html#structfield.gama_chunk), but:

    - The gamma value is *not* exposed via `image::codecs::png::PngDecoder`
    - Even if the gamma value was exposed, I wasn't able to identify Rust crate that implements an equivalent of [`png_do_gamma`](https://github.com/glennrp/libpng/blob/f8e5fa92b0e37ab597616f554bee254157998227/pngrtran.c#L3944).  Maybe this part is a *separate* feature request for the `image` crate but for `imageproc`?

AFAICT the problem can be reproed with `pngsuite/basn2c08.png`.

fintelia:
The `Result` return type was chosen so that the method can return I/O errors once the PNG decoder no longer reads all metadata in the `new()` method. And an `f64` was used so that all gamma values can be exactly represented (PNG encodes the value as a fixed precision decimal stored as a u32 which doesn't directly map to a f32).

cc: @anforowicz 

Fixes #2001

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -193,6 +193,27 @@ impl<R: Read> PngDecoder<R> {
         Ok(PngDecoder { color_type, reader })
     }
 
+    /// Returns the gamma value of the image or None if no gamma value is indicated.
+    ///
+    /// If an sRGB chunk is present this method returns a gamma value of 0.45455 and ignores the
+    /// value in the gAMA chunk. This is the recommended behavior according to the PNG standard:
+    ///
+    /// > When the sRGB chunk is present, [...] decoders that recognize the sRGB chunk but are not
+    /// > capable of colour management are recommended to ignore the gAMA and cHRM chunks, and use
+    /// > the values given above as if they had appeared in gAMA and cHRM chunks.
+    pub fn gamma_value(&self) -> ImageResult<Option<f64>> {
+        let gamma = if self.reader.info().srgb.is_some() {
+            Some(0.45455)
+        } else {
+            self.reader
+                .info()
+                .gama_chunk
+                .map(|x| x.into_scaled() as f64 / 100000.0)
+        };
+
+        Ok(gamma)
+    }
+
     /// Turn this into an iterator over the animation frames.
     ///
     /// Reading the complete animation requires more memory than reading the data from the IDAT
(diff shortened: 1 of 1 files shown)


src/codecs/png.rs
```diff
@@ -193,6 +193,27 @@ impl<R: Read> PngDecoder<R> {
...
+    ///
+    /// If an sRGB chunk is present this method returns a gamma value of 0.45455 and ignores the
+    /// value in the gAMA chunk. This is the recommended behavior according to the PNG standard:
+    ///
+    /// > When the sRGB chunk is present, [...] decoders that recognize the sRGB chunk but are not
+    /// > capable of colour management are recommended to ignore the gAMA and cHRM chunks, and use
+    /// > the values given above as if they had appeared in gAMA and cHRM chunks.
+    pub fn gamma_value(&self) -> ImageResult<Option<f64>> {
+        let gamma = if self.reader.info().srgb.is_some() {
+            Some(0.45455)
```
anforowicz, line 206: I wonder why you have chosen to manually consider `srgb` and `gama_chunk`, instead of using [Info::source_gamma](https://docs.rs/png/0.17.10/png/struct.Info.html#structfield.source_gamma), which AFAICT takes both `srgb` ([here](https://github.com/image-rs/image-png/blob/cbea7ffd093573a76a406a145d776acf7a1d5baa/src/decoder/stream.rs#L1078)) and `gama_chunk` ([here](https://github.com/image-rs/image-png/blob/cbea7ffd093573a76a406a145d776acf7a1d5baa/src/decoder/stream.rs#L1052)) into account.  FWIW, [`png::srgb::substitute_gamma`](https://github.com/image-rs/image-png/blob/cbea7ffd093573a76a406a145d776acf7a1d5baa/src/srgb.rs#L6) is also `45455`

fintelia: That's a good point point... I didn't notice the `png` crate already implemented that logic

anforowicz:
Thanks - LGTM!

PR #2013, opened 2023-09-21
Implement ImageEncoder for hdr where color type is Rgba32F

johannesvollmer:
- impl encoder for hdr where rgba32f for now
- add new output format 

this unlocks:
- ImageBuffer.write_with_encoder, which unlocks DynamicImage.write_with_encoder
- free_functions::write_buffer_impl, which unlocks all the sweet stuff (DynamicImage.write_to, ImageBuffer.write_to)

### DIFF
--- a/src/codecs/hdr/encoder.rs
+++ b/src/codecs/hdr/encoder.rs
@@ -47,7 +47,7 @@ impl<W: Write> HdrEncoder<W> {
     /// that has dimensions ```width``` and ```height```.
     /// The callback must return the color for the given flattened index of the pixel (row major).
     pub fn encode_pixels(mut self, mut flattened_rgbe_pixels: impl ExactSizeIterator<Item=Rgbe8Pixel>, width: usize, height: usize) -> ImageResult<()> {
-        assert!(flattened_rgbe_pixels.len() >= width * height, "not enough pixels provided"); // this might elide some bounds checks
+        assert!(flattened_rgbe_pixels.len() >= width * height, "not enough pixels provided"); // bonus: this might elide some bounds checks
 
         let w = &mut self.w;
         w.write_all(SIGNATURE)?;
@@ -57,8 +57,8 @@ impl<W: Write> HdrEncoder<W> {
         w.write_all(format!("-Y {} +X {}\n", height, width).as_bytes())?;
 
         if !(8..=32_768).contains(&width) {
-            for _ in 0 .. width * height {
-                write_rgbe8(w, flattened_rgbe_pixels.next().expect("not enough pixels provided"))?;
+            for pixel in flattened_rgbe_pixels {
+                write_rgbe8(w, pixel)?;
             }
         } else {
             // new RLE marker contains scanline width
@@ -70,6 +70,8 @@ impl<W: Write> HdrEncoder<W> {
             let mut bufe = vec![0; width];
             let mut rle_buf = vec![0; width];
             for _scanline_index in 0 .. height {
+                assert!(flattened_rgbe_pixels.len() >= width); // may reduce the bound checks
+
                 for (((r, g), b), e) in bufr
                     .iter_mut()
                     .zip(bufg.iter_mut())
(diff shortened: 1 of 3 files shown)


johannesvollmer:
please give me some hints as to what kind of tests I can add to make sure this works as expected :)

src/codecs/hdr/encoder.rs
```diff
@@ -17,8 +44,24 @@ impl<W: Write> HdrEncoder<W> {
...
+    pub fn encode(self, rgb: &[Rgb<f32>], width: usize, height: usize) -> ImageResult<()> {
+        self.encode_pixels(rgb.iter().map(|&rgb| to_rgbe8(rgb)), width, height)
+    }
+
+    /// Encodes the image ```data```
+    /// that has dimensions ```width``` and ```height```.
+    /// The callback must return the color for the given flattened index of the pixel (row major).
+    pub fn encode_pixels(
+        mut self,
+        mut flattened_rgbe_pixels: impl ExactSizeIterator<Item = Rgbe8Pixel>,
```
johannesvollmer, line 56: i needed to add a function that uses an iterator, because the byte buffer might not be aligned, so we just read individual values from the buffer with bytemuck in an iterator

a future performance improvement might be to check whether the buffer is aligned, and then cast the whole thing (which will be the case nearly all the time)

src/codecs/hdr/encoder.rs
```diff
@@ -17,8 +44,24 @@ impl<W: Write> HdrEncoder<W> {
...
+        width: usize,
+        height: usize,
+    ) -> ImageResult<()> {
+        assert!(
```
johannesvollmer, line 60: this should probably return an error isntead of panicking.

johannesvollmer:
can we add a reference test that compares images written with HDR to images written with EXR? How does that work in the current test suite?

fintelia:
The relevant file `tests/reference_images.rs`. I'm not super familiar with how it works, but shouldn't be too hard to figure out from there

fintelia (approved):

fintelia:
If you're interested in reviving this PR, I think it is in good shape. Just needs a rebase and ideally also updates to `ImageFormat::{can_write, writing_enabled}`

src/codecs/hdr/encoder.rs
fintelia: We actually added similar asserts on the other encoders, let's keep this. If anything it should actually `assert_eq!` to catch dimensions being too small

zhouhang95:
> If you're interested in reviving this PR, I think it is in good shape. Just needs a rebase and ideally also updates to `ImageFormat::{can_write, writing_enabled}`

Sorry, I'm not good at github workflow. I don't konw how to change this pr.

src/codecs/hdr/encoder.rs
```diff
@@ -39,20 +82,23 @@ impl<W: Write> HdrEncoder<W> {
...
+
+                for (((r, g), b), e) in bufr
                     .iter_mut()
                     .zip(bufg.iter_mut())
                     .zip(bufb.iter_mut())
                     .zip(bufe.iter_mut())
-                    .zip(scanline.iter())
                 {
-                    let cp = to_rgbe8(pix);
+                    let cp = flattened_rgbe_pixels.next().unwrap(); // we know it's here because the length is checked earlier
```
fintelia, line 94: Can this be included in the chain of `zip` calls above? Roughly the same as the old version, but passing `flattened_rgbe_pixels` instead of `scanline`

fintelia:
@zhouhang95 GitHub only lets the original PR author or a repository maintainer update a PR

src/codecs/hdr/encoder.rs
johannesvollmer: this should be possible, maybe this was done to minimize the diff for an easier review. unfortunately i don't have time at the moment :/

johannesvollmer:
but it should be possible to checkout the branch of this pr and then open a new pr with your own branch :)

ripytide:
This can be closed now as it was merged in #2246

fintelia:
Thanks everyone!

PR #2014, opened 2023-09-22
generic progress callback implementation on a byte-level

johannesvollmer:
this approach would be imprecise, but better than nothing

- the decoders and encoders could internally use this to add support for progress callbacks (thus `pub(crate)`)
- they can replace it with real progress tracking implementation any time

### DIFF
--- a/.idea/image.iml
+++ b/.idea/image.iml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="EMPTY_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/benches" isTestSource="true" />
+      <sourceFolder url="file://$MODULE_DIR$/examples" isTestSource="false" />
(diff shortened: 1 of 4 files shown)


johannesvollmer:
attempting tiff encoder. currently impossible without unsafe code because:

- we would need to wrap the real byte source upon creation of the tiff reader
- the progress callback is only available later in load_with_progress, but we cannot mutate the reader at that point, it's immutable

fintelia:
I wonder if this is something that would make more sense living outside this crate. Given a `R: Read + Seek`, a user would be able to construct a `BufReader<TrackProgressReader<R>>` and then pass it when constructing the image decoder. And the stream length could be obtained exactly using `Seek::stream_len` (or a manual implementation until it stabilizes)

johannesvollmer:
Using Seek to find the length fits well to the core idea. 

However, this byte-level tracking only works if assumptions about the internal implementation of the decoder holds. So, you can't really rely on it to work in general. I'd rather author this on a per-decoder basis, because it won't work for some decoders.

My goal was to provide a helper that works with the current image api. I consider this experiment a success, the answer is: it's not possible right now.

Issue #1956, opened 2023-07-08
webp: Lossless images should use table-based huffman decoder


fintelia:
The current lossless webp decoding logic does a branch for almost every single _bit_ in the file:

https://github.com/image-rs/image/blob/038bc30b9b9b1daf809452523b557db58f9e781b/src/codecs/webp/huffman.rs#L189-L192

Switching to a more efficient huffman implementation would likely be orders of magnitude faster

PR #2020, opened 2023-09-29
Use SIMD to speed up clamping

### LINKED ISSUE #2019: `image::utils::clamp` could be better optimized
`image::utils::clamp` is callled by `image::codecs::webp::vp8::Frame::fill_single` and accounts for the vast majority of its runtime.

Due to how slow other parts of WebP decoding are (see #1956 ) this is not currently a bottleneck for WebP decoding, only accounting for 6% of the execution time; but it would become a significant contributor to the runtime in the future.

I've tried some things to get the compiler to autovectorize clamping, to no avail: https://godbolt.org/z/813sEnGs4

The portable SIMD codegen for clamping is also quite absurd: https://godbolt.org/z/x6zszG1jf

But it does appear to be possible even without intrinsics in another LLVM-based language, Zig. With AVX-512 it's very short: https://godbolt.org/z/cMz6foPjv
But even the baseline x86_64 produces vector instructions: https://godbolt.org/z/j8j34GoG3

Shnatsel:
Leverage autovectorization to speed up clamping.

Improves end-to-end WebP decoding performance by 2% on x86_64 baseline; should be even better with more recent instructions, and even benefits from AVX-512. 

Not a whole lot of an improvement because Huffman decoding takes up so much time that everything else is trivial by comparison, see #1956; but this will become a significant improvement in the future once Huffman is sped up.

Fixes #2019

I'm opening this to get feedback on the basic direction. If you're OK with the approach I can add it to other parts of the codebase for a few % speedups in other places too.

### DIFF
--- a/src/codecs/webp/vp8.rs
+++ b/src/codecs/webp/vp8.rs
@@ -25,7 +25,7 @@ use crate::error::{
 };
 use crate::image::ImageFormat;
 
-use crate::utils::clamp;
+use crate::utils::{clamp, clamp_simd};
 
 const MAX_SEGMENTS: usize = 4;
 const NUM_DCT_TOKENS: usize = 12;
@@ -929,19 +929,15 @@ impl Frame {
         let d: i32 = i32::from(u) - 128;
         let e: i32 = i32::from(v) - 128;
 
-        let r: u8 = clamp((298 * c + 409 * e + 128) >> 8, 0, 255)
-            .try_into()
-            .unwrap();
-        let g: u8 = clamp((298 * c - 100 * d - 208 * e + 128) >> 8, 0, 255)
-            .try_into()
-            .unwrap();
-        let b: u8 = clamp((298 * c + 516 * d + 128) >> 8, 0, 255)
-            .try_into()
-            .unwrap();
+        let r = (298 * c + 409 * e + 128) >> 8;
+        let g = (298 * c - 100 * d - 208 * e + 128) >> 8;
+        let b = (298 * c + 516 * d + 128) >> 8;
 
-        rgb[0] = r;
-        rgb[1] = g;
-        rgb[2] = b;
+        let clamped = clamp_simd([r, g, b, 0], 0, 255);
+
+        rgb[0] = clamped[0].try_into().unwrap();
+        rgb[1] = clamped[1].try_into().unwrap();
+        rgb[2] = clamped[2].try_into().unwrap();
     }
 
     /// Gets the buffer size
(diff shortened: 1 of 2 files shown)


Shnatsel:
Godbolt version instantiated with concrete types so you can see the assembly at various ISA levels: https://godbolt.org/z/dvaMTbPqq

fintelia:
I've been in the process of forking the WebP codec out into its own crate here: https://github.com/image-rs/image-webp, so that would be a better place for non-urgent changes to the codec. There's also an in progress draft PR that (among many other things) incidentally modifies `fill_single` to have bit exact output matching libwebp

Overall, I'd say I'm hesitant but not opposed to making the code messier in order to make it vectorize better

Shnatsel:
I see that https://github.com/image-rs/image-webp/pull/2 refactors this function. Should I make this change to the version currently in master, or build on top that PR?

fintelia:
I finished that other PR, so you can now work directly again the `main` branch there.

Though one thing to point out is that lossless and lossy WebP are basically separate formats and have different performance bottlenecks (though the alpha channel of lossy WebP can be encoded with the lossless format).

Lossy involves:
1. **VP8 decoding** - Haven't worked on this part of the code, so I'm not actually sure how optimal this is.
2. **YUV -> RGB** - This is where clamping happens. I believe the reference encoder does this entirely with SIMD.

Lossless has a few bottlenecks. They'd all need to be addressed to make a decoding truly fast:
1. **Read bits** - The current strategy is pointlessly inefficient, [this blog](https://fgiesen.wordpress.com/2018/09/27/reading-bits-in-far-too-many-ways-part-3/) describes how to do better.
2. **Huffman decoding** - Decoding tables will help, but the format allows images to have _thousands_ of huffman trees so if we aren't careful the precomputations could make things worse rather than better.
3. **Predictor** - This is analogous to PNG's unfiltering. Unfortunately WebP lets images switch predictors for each 4 pixel block, so using SIMD may be even harder (the reference decoder doesn't even try).

Shnatsel:
In my benchmarks the lossy decoding is 2x slower than `dwebp` with assembly optimizations disabled, or 4x slower than `dwebp` with assembly optimizations enabled.

Fortunately WebP images are usually rather small (unlike JPEG which is commonly used for large photos), so it is not very noticeable in practice in interactive use cases.

Shnatsel:
The rewritten YUV to RGB conversion is not so amenable to optimization, probably due to the additional branches introduced there. When inlined, clamping does not emit SIMD instructions; and once I did enough transforms to the code to make it emit them again, llvm-mca claimed I actually made the code slower.

PR #2021, opened 2023-09-29
Use SIMD via autovec in `macroblock_filter`

Shnatsel:
Ekes out a slight but measurable performance improvement (at around 1%) on top of #2020 because this function is not very hot.

Depends on #2020 which should be reviewed first.

This might not be worth it because it would make the scalar fallback slower, due to the one extra element being processed, but since I wrote it I thought I might as well show it.


Shnatsel:
Tests fail so I'm going to close this

Issue #1954, opened 2023-06-30
Decode AVIF reports image -11 Error Code EDEADLOCK


charlie-HiveAi:
This happens in AVIF image decoder

When I run 
`image::image_dimensions(&path);`
I get `Decoding(DecodingError { format: Exact(Avif), underlying: Some(Error(-11)) })`

After some digging, the error is returned from this line [here](https://github.com/image-rs/image/blob/master/src/codecs/avif/decoder.rs#L43):
`let picture = primary_decoder.get_picture().map_err(error_map)?;`

I have to think the reported deadlock has to do with the mutex implementation of `dav1d`. I wonder if anyone else has run into the same issues.

## Reproduction steps
Download the 8-bit AVIF image from: "https://aomediacodec.github.io/av1-avif/testFiles/Link-U/hato.profile0.8bpc.yuv420.no-cdef.avif". 
Then simply run the following code
```
let im_path = "avif-test.avif";
let buffered_read = std::io::BufReader::new(std::fs::File::open(im_path).unwrap());
let image = load(buffered_read, ImageFormat::Avif).unwrap();
```
And we get
`Decoding(DecodingError { format: Exact(Avif), underlying: Some(Error(-11)) })`

197g:
This was eventually addressed in: #2043

PR #2023, opened 2023-10-03
avif-decoder: Send pending data to dav1d when Err(Again)

metamuffin:
image-rs currently errors with `Try Again` (returned from `get_picture`) when decoding AVIF images.
This is because dav1d needs to be explicitly flushed via `send_data` to ensure the stream end is processed. 
 For reference see [dav1d-rs "example" usage](https://github.com/rust-av/dav1d-rs/blob/25075aabaefd41ae16d90fcc50a277a7eefa819a/tools/src/main.rs#L116)
This PR adds a call to `send_pending_data` when required to do exactly that, fixing the error.

This PR resolves https://github.com/image-rs/image/issues/1954.

### DIFF
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -40,7 +40,14 @@ impl<R: Read> AvifDecoder<R> {
         primary_decoder
             .send_data(coded.to_vec(), None, None, None)
             .map_err(error_map)?;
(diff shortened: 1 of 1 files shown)


fintelia:
Should the `Error::Again` case be handled in a loop? The example seems to repeatedly send data when it gets rather, rather than only doing so once

PR #2032, opened 2023-10-17
Fix `<flat::View(Mut) as GenericImageView>::bounds` implementations

zetanumbers:
`flat::View(Mut)` have incorrect implementations of `GenericImageView::bounds`, given that it should return `(x, y, w, h)`, which i inferred from `GenericImageView::view`'s parameters.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1169,10 +1169,6 @@ where
 {
     type Pixel = P;
 
-    fn dimensions(&self) -> (u32, u32) {
-        self.dimensions()
-    }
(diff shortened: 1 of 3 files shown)


fintelia:
I think it might be better just to deprecate `GenericImageView::bounds` given that it has inconsistent implementations. That would give a clear warning to anyone currently relying on it, without potentially breaking anyone depending on the current behavior

fintelia:
Superseded by https://github.com/image-rs/image/pull/2036

PR #2034, opened 2023-10-20
chg: update webp dependency (MEMORYLEAK)

cosinusalpha:
This update of the webp dependency fixes a memory leak when saving images as ".webp".

The memory leak can be easily simulated with the following code:

```rust

const IMAGE: &str = "input.png";
const OUTPATH: &str = "out.webp";

fn main() {
    println!("Hello, world!");


    // do this 1 M times
    for i in 0..1_000_000 {
        println!("{}", i);
        let img = ImageReader::open(IMAGE).unwrap().decode().unwrap();

        // save image as webp
        img.save(OUTPATH).unwrap();
    }
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ dcv-color-primitives = { version = "0.5.4", optional = true }
 color_quant = "1.1"
 exr = { version = "1.5.0", optional = true }
 qoi = { version = "0.4", optional = true }
-libwebp = { package = "webp", version = "0.2.2", default-features = false, optional = true }
+libwebp = { package = "webp", version = "0.2.6", default-features = false, optional = true }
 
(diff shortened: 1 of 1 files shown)


fintelia:
Thanks for raising this. In Cargo.toml, `version = "0.2.2"` actually means `version = "^0.2.2"`, i.e. any version 0.2.x where x>=2. You should be able to pull in the newer version of libwebp by running `cargo update`.

What this means is that bumping to semver compatible versions in our Cargo.toml doesn't do anything. Users would have to do an update to get the newer version of this crate, at which point they'd end pull in the newer libwebp at the same time (regardless of whether our Cargo.toml changed).

cosinusalpha:
Thanks for the quick reply!

Issue #1994, opened 2023-08-28
Function to ask for supported image formats/extensions


emilk:
I would like to be able to ask `image` what file extensions it supports _give its current feature flags_.

For instance: `println("{:?}", image::extensions());` could print `["jpg", "jpeg", "png"]` if only the `image` and `jpeg` features of `image` is turned on.

My specific use case for this functionality is adding file extension filters to a file dialog.

fintelia:
I think it makes sense to expose this information. Maybe something like:

```rust
impl ImageFormat {
    pub fn reading_enabled(&self) -> bool;
    pub fn writing_enabled(&self) -> bool; 
}
```
Though we should probably provide a way to iterate over all the formats. I wonder if a method like `fn all() -> impl Iterator<Item=ImageFormat>` would be good?

emilk:
Oh yeah, great catch on the read vs write difference. That API looks solid to me!

fintelia:
Added in #2038

PR #2038, opened 2023-10-24
#1994 - Support for iterating enabled image formats

xweskingx:
Add functions to enumerate all ImageFormat, and determine availability based on configured feature flags.

Example Usage:

```
let available_extensions = ImageFormat::all()
            .filter(ImageFormat::reading_enabled)
            .flat_map(ImageFormat::extensions_str)
            .collect();


fintelia (approved):

fintelia:
Looks like there's a few clippy issues. Specifically, the `all()` method needs to be marked `pub` and there's a few instances of `assert_eq!(false, ...)` that can be simplified to `assert!(!...)`.

fintelia:
Thanks!

PR #2040, opened 2023-10-25
:pencil2: Fix typos

CAESIUS-TIM:


fintelia:
I see some references online to `<!---` (with three dashes) being a special comment format for markdown, so I think this is correct as is

CAESIUS-TIM:
I'm grateful for being corrected and learning new [markdown syntax](https://www.jamestharpe.com/markdown-comments/).

Issue #2016, opened 2023-09-24
Two arithmetic overflow found in RgbImage


thdusdl1219:
I found two arithmetic overflow cases in the latest version. They look same but exist in different location.

### Case 1

This happens in 

https://github.com/image-rs/image/blob/f1e34f365133a33a27dac5bc59c93d1cfb34a37e/src/buffer.rs#L792

### Case 2

This happens in 

https://github.com/image-rs/image/blob/f1e34f365133a33a27dac5bc59c93d1cfb34a37e/src/buffer.rs#L970

## Expected

Assert or use the checked operation instead.

### Case 1
`thread 'main' panicked at 'attempt to add with overflow', .cargo/registry/src/index.crates.io-6f17d22bba15001f/image-0.24.7/./src/buffer.rs:792:21`

### Case 2
`thread 'main' panicked at 'attempt to add with overflow', .cargo/registry/src/index.crates.io-6f17d22bba15001f/image-0.24.7/./src/buffer.rs:970:25`

### Case 1

```
    let ImageBuffer_self_obj = image::RgbImage::new(u32::MAX - 1, 1);
    let y = u32::MAX;
    let x = u32::MAX - 2;
    ImageBuffer_self_obj.get_pixel_checked(x, y);
```

### Case2

```
    let mut ImageBuffer_self_obj = image::RgbImage::new(u32::MAX - 1, 1);
    let y = u32::MAX;
    let x = u32::MAX - 2;
    ImageBuffer_self_obj.get_pixel_mut_checked(x, y);
```

fintelia:
Could you make a pull request?

JulianKnodt:
@fintelia would it be alright if I fixed these?

thdusdl1219:
Sure! I didn't have enough time to look into them lately, so feel free to do so :)

JulianKnodt:
This can be closed

Issue #2015, opened 2023-09-24
Two arithmetic overflow found in DynamicImage::thumbnail_exact


thdusdl1219:
I found two arithmetic overflow cases in the latest version.

### Case 1

This happens in 

https://github.com/image-rs/image/blob/f1e34f365133a33a27dac5bc59c93d1cfb34a37e/src/imageops/sample.rs#L579

### Case 2

This happens in 

https://github.com/image-rs/image/blob/f1e34f365133a33a27dac5bc59c93d1cfb34a37e/src/imageops/sample.rs#L586

## Expected

Assert or size check before subtraction or use the checked operation instead.

### Case 1
`thread 'main' panicked at 'attempt to subtract with overflow', .cargo/registry/src/index.crates.io-6f17d22bba15001f/image-0.24.7/./src/imageops/sample.rs:579:54`

### Case 2
`thread 'main' panicked at 'attempt to subtract with overflow', .cargo/registry/src/index.crates.io-6f17d22bba15001f/image-0.24.7/./src/imageops/sample.rs:586:54`

### Case 1

```
    let dyimg = DynamicImage::default();
    dyimg.thumbnail_exact(0, 1);
```

### Case2

```
    let dyimg = DynamicImage::new_rgb8(0, 1);
    dyimg.thumbnail_exact(1, 1);
```

fintelia:
Pull requests would be appreciated!

fintelia:
Fixed by #2041

PR #2041, opened 2023-10-26
Fix overflows

### LINKED ISSUE #2017: Two arithmetic overflow found in SubImage::view
I found two arithmetic overflow cases in the latest version. They followed by two assertions but look not working to prevent them.

They happen in L1284 and L1285

https://github.com/image-rs/image/blob/f1e34f365133a33a27dac5bc59c93d1cfb34a37e/src/image.rs#L1280-L1285

## Expected

Assert or use the checked operation instead.

### Case 1
`thread 'main' panicked at 'attempt to add with overflow', .cargo/registry/src/index.crates.io-6f17d22bba15001f/image-0.24.7/./src/image.rs:1271:17`

### Case2
`thread 'main' panicked at 'attempt to add with overflow', .cargo/registry/src/index.crates.io-6f17d22bba15001f/image-0.24.7/./src/image.rs:1272:17`

### Case 1

```
    let mut ImageBuffer_self_obj = image::RgbImage::new(1, 1);
    let SubImage_self_obj = image::SubImage::new(&ImageBuffer_self_obj, u32::MAX, u32::MAX, u32::MAX, u32::MAX);
    SubImage_self_obj.view(1, 1, 1, 1);
```

### Case2

```
    let mut ImageBuffer_self_obj = image::RgbImage::new(1, 1);
    let SubImage_self_obj = image::SubImage::new(&ImageBuffer_self_obj, u32::MAX, u32::MAX, u32::MAX, u32::MAX);
    SubImage_self_obj.view(0, 1, 0, 1);
```

JulianKnodt:
cc @thdusdl1219

Closes #2017, #2016, #2015

If you found these overflows using a compiler tool, maybe it's worth linking to it so they can be found easily in the future by crate authors.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -789,7 +789,7 @@ where
             .saturating_mul(num_channels);
 
         self.data
-            .get(i..i + num_channels)
+            .get(i..i.checked_add(num_channels)?)
             .map(|pixel_indices| <P as Pixel>::from_slice(pixel_indices))
     }
 
@@ -967,7 +967,7 @@ where
             .saturating_mul(num_channels);
 
         self.data
-            .get_mut(i..i + num_channels)
+            .get_mut(i..i.checked_add(num_channels)?)
             .map(|pixel_indices| <P as Pixel>::from_slice_mut(pixel_indices))
     }
 
--- a/src/image.rs
+++ b/src/image.rs
@@ -1350,8 +1350,8 @@ where
         use crate::GenericImageView as _;
         assert!(x as u64 + width as u64 <= self.inner.width() as u64);
         assert!(y as u64 + height as u64 <= self.inner.height() as u64);
-        let x = self.inner.xoffset + x;
-        let y = self.inner.yoffset + y;
+        let x = self.inner.xoffset.saturating_add(x);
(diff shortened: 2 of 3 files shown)


fintelia:
I don't think `saturating_add` is the right option for most of these. They should probably either panic or return `None` depending on the function.

For `SubImage::view`, I think the bug is actually in the `SubImage::new` method. You shouldn't be able to create a SubImage that extends outside the bounds of the underlying image, but the `new` method lets you. Once you have a `SubImage` with invalid bounds, the assertions in the `view` method are no longer fully meaningful and that's why an overflow can happen.

JulianKnodt:
If I were to check that the `SubImage::new` calls all properly enforced that the view was within bounds of the original, it would be a breaking change to the API as there was no previous trait bound on `I` in `SubImage<I>::new`. Is that alright?

I would note that these conditions are in my head impossible to hit in practice unless being maliciously exploited. I mean that if a buffer where these values actually mattered could not be stored in memory, so the only use case to consider is that this library is embedded in an application, and someone is trying to crash it by passing invalid values. As long as we guard against that case, then I think that's fine, or at least that's how I see these in my head.

PR #2043, opened 2023-11-01
Handle non-fatal error dav1d::Error::Again

FlareFlo:
Issue location:
`src/codecs/avif/decoder.rs` line 43 and 50

Currently, when `send_data` is called, its non-fatal error variant `Again` is treated as a fatal one, where this may be treated by `send_pending_data`, as mentioned by the [dav1d-rs docs](https://imgur.com/a/itNVERV) (dav1d-rs docs [fail to build](https://github.com/rust-av/dav1d-rs/issues/45) so i used this screenshot from my local build).

https://github.com/image-rs/image/pull/2023 addresses this issue as well, but does not handle the case of Alpha channels and only calls `send_pending_data` once.

### DIFF
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -68,6 +68,7 @@ impl<R: Read> AvifDecoder<R> {
 
 /// Wrapper struct around a `Cursor<Vec<u8>>`
 pub struct AvifReader<R>(Cursor<Vec<u8>>, PhantomData<R>);
+
 impl<R> Read for AvifReader<R> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         self.0.read(buf)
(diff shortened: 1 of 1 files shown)


src/codecs/avif/decoder.rs
```diff
@@ -172,3 +172,30 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for AvifDecoder<R> {
         Ok(())
     }
 }
+
+/// `get_picture` and `send_pending_data` yield `Again` as a non-fatal error requesting more data is sent to the decoder
+/// This ensures that in the case of `Again` all pending data is submitted
+/// This should be called after `send_data` (which does not yield `Again` when called the first time)
+fn read_until_ready(decoder: &mut dav1d::Decoder) -> ImageResult<dav1d::Picture> {
+    loop {
+        match decoder.get_picture() {
```
fintelia, line 181: This match could be simplified to:
```rust
match decoder.get_picture() {
    Err(dav1d::Error::Again) => { ... }
    r => return r.map_err(error_map),
}
```

FlareFlo: Applied in bffd087d6a31f2f991e2c0d868000d5c9ebf8dc6

fintelia:
Thanks!

PR #2046, opened 2023-11-03
Bump `webp` version

MidasLamb:
There is a security advisory for the `libwebp-sys` that `webp` uses. It's suggested to upgrade to `libwebp-sys` 0.9.3, which happens in `webp` 0.2.6


MidasLamb:
Seems that this would also raise the MSRV to 1.64, I'm not sure if that's fine?

fintelia:
In a Cargo.toml `version = "0.2.2"` actually means "any version 0.2.x where x>=2". Thus, even without this change anyone who runs `cargo update` will pull in the latest version of `webp`. And if we published a new version of this crate, users would still need to update to get that newly published version

PR #2048, opened 2023-11-03
Bump dcv-color-primitives to 0.6.1

FlareFlo:
Note: `dcp::initialize` no longer needs to be called, as its global state is [initialized automatically when required](https://github.com/aws/dcv-color-primitives/pull/78).


fintelia:
Thanks!

Issue #2049, opened 2023-11-04
Promote assertion to Error


FlareFlo:
https://github.com/image-rs/image/blob/eb05a39313093bf37b7fb435dffa48f0ad122112/src/codecs/avif/decoder.rs#L159
Promoting this assertion to an error such as `UnsupportedErrorKind::Format` would let callers catch the error.

fintelia:
Please open a PR!

FlareFlo:
Fixed in #2050

PR #2050, opened 2023-11-05
Replace assert with proper Unsupported error

FlareFlo:
PR for #2049.

I am not quite sure if this is the correct format for the error message of this kind.

### DIFF
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -9,7 +9,7 @@ use std::io::{self, Cursor, Read};
 use std::marker::PhantomData;
 use std::mem;
 
-use crate::error::{DecodingError, ImageFormatHint, UnsupportedError, UnsupportedErrorKind};
+use crate::error::{DecodingError, UnsupportedError, UnsupportedErrorKind};
 use crate::{ColorType, ImageDecoder, ImageError, ImageFormat, ImageResult};
 
 use dav1d::{PixelLayout, PlanarImageComponent};
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -9,7 +9,7 @@ use std::io::{self, Cursor, Read};
 use std::marker::PhantomData;
 use std::mem;
 
-use crate::error::DecodingError;
+use crate::error::{DecodingError, ImageFormatHint, UnsupportedError, UnsupportedErrorKind};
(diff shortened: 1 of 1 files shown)


FlareFlo:
Maybe it should include that the alpha must be in I400 format?

fintelia:
It should be possible to somewhat simplify the error construction. Here's an example from another part of the codebase:
https://github.com/image-rs/image/blob/eb05a39313093bf37b7fb435dffa48f0ad122112/src/codecs/dds.rs#L274-L282

It would be also good to add a few words to the error string explaining what isn't supported. If nothing else, it gives a string for folks to search for to find out where the error is coming from.

fintelia:
Oh, and to clarify `UnsupportedErrorKind::Format` means that the image file format isn't support (like say "AVIF isn't supported at all") whereas `UnsupportedErrorKind::GenericFeature` means that the overall format is supported but some specific feature isn't supported, which is the case here.

FlareFlo:
Cargo build did not warn about that unused import :/

fintelia:
To double check, is I400 required by the AVIF spec or just something our decoder needs?

FlareFlo:
> To double check, is I400 required by the AVIF spec or just something our decoder needs?

I400 is monochrome, which i would assume is the only sensible format for the alpha plane

fintelia:
If other layouts aren't allowed, this should be a DocodingError rather than an UnsupportedError.

FlareFlo:
[The note on the AVIF spec](https://aomediacodec.github.io/av1-avif/v1.1.0.html#auxiliary-images) mentions a 3-component image of sorts?

FlareFlo:
I stumbled upon this assertion after encountering an image with a I420 alpha mask, so there might be more to this.

FlareFlo:
I believe the monochrome mention of 3-components relates to the Y C and B mapping for the alpha layer, meaning that 3 seperate alpha masks could be stored via I444 (or any format for that matter). Its possible the assertion checks against I400 because image only supports one alpha layer?

fintelia:
Thanks!

FlareFlo:
Are there any other formats that have more than one alpha layer?

fintelia:
TIFF and probably EXR support arbitrary numbers of layers. Other than that, I don't think there are other common formats that support multiple alpha layers

PR #2052, opened 2023-11-09
Do not round floats in interpolate

### LINKED ISSUE #2045: Sample/interpolate gives rounded values, which is unexpected when dealing with f32 images
I was trying to manipulate HDR pictures by using `sample_bilinear` on RgbF32Images but I think this function, as well as `interpolate_bilinear`, were not made with this kind of image in mind.

## Expected

When I use `sample_bilinear` on an RgbF32Image, I expected a correct, not rounded output

## Actual behaviour

The values for each channel are rounded to the nearest integer. The rounding happens here: https://github.com/image-rs/image/blob/01c6c9975c320e95f36e708f0ab773b207f68ed5/src/imageops/sample.rs#L435

## Reproduction steps

Little code that demonstrate the issue:
```rust
use image::{ImageBuffer, Rgb32FImage, Rgb, imageops};

fn main() {
    // Make a 3-pixel graident that goes black->gray->white
    let mut image: Rgb32FImage = ImageBuffer::new(3, 1);
    image[(0, 0)] = Rgb([0.0, 0.0, 0.0]);
    image[(1, 0)] = Rgb([0.5, 0.5, 0.5]);
    image[(2, 0)] = Rgb([1.0, 1.0, 1.0]);

    // Sample at non-integer coordinates using sample_bilinear
    let num_samples = 10;
    for i in 0..=num_samples {
        let u = i as f32 / num_samples as f32;
        let sampled = imageops::sample_bilinear(&image, u, 0.0).unwrap();
        println!("u={u}, value={sampled:?}");
    }
}
```
It prints the following:
```
u=0, value=Rgb([0.0, 0.0, 0.0])
u=0.1, value=Rgb([0.0, 0.0, 0.0])
u=0.2, value=Rgb([0.0, 0.0, 0.0])
u=0.3, value=Rgb([0.0, 0.0, 0.0])
u=0.4, value=Rgb([0.0, 0.0, 0.0])
u=0.5, value=Rgb([1.0, 1.0, 1.0])
u=0.6, value=Rgb([1.0, 1.0, 1.0])
u=0.7, value=Rgb([1.0, 1.0, 1.0])
u=0.8, value=Rgb([1.0, 1.0, 1.0])
u=0.9, value=Rgb([1.0, 1.0, 1.0])
u=1, value=Rgb([1.0, 1.0, 1.0])
```
However what I believe is a correct output is a smooth gradient from 0 to 1

JulianKnodt:
I filed this issue in num_traits to control rounding behavior when converting:
https://github.com/rust-num/num-traits/issues/296

I don't expect it to be fixed for a while though. This is a super hacky (but generic) way to ensure that floats aren't rounded in interpolate.

How do other methods handle this? I see a FIXME in gaussian blur so I suspect other methods currently also miss this

Closes #2045

cc @alucas2


src/imageops/sample.rs
```diff
@@ -430,15 +430,26 @@ pub fn interpolate_bilinear<P: Pixel>(
...
+                    P::Subpixel::DEFAULT_MIN_VALUE
+                } else {
+                    P::Subpixel::DEFAULT_MAX_VALUE
+                }
+            })
+        };
+        let new_val = convert(v);
+        // super hacky check to see if primitive is an integer or a float
+        // note that it should always be convertible to f32, so this unwrap is alright.
+        *c = if new_val.to_f32().unwrap() != v {
```
fintelia, line 445: `DEFAULT_MAX_VALUE` is defined to be 1.0 if the type is floating point and all integer types have larger values. Perhaps this check could be:
```rust
if P::Subpixel::DEFAULT_MAX_VALUE.to_f32() == 1.0 { ... }
```
That could be resolved at compile-time and thus avoid any branch overhead

(Equality comparisons are usually dangerous for floats, but all small integers can be exactly represented by an f32 so there's no risk here)

PR #2056, opened 2023-11-13
TGA: RLE encoding

marvin-j97:

---

This is my first pass of RLE encoding support for TGA encoding. It currently supports ImageType::RunTrueColor and ImageType::RunGrayScale.

### Here's a run down:

- Modified `Header::from_pixel_info` to support an `use_rle` flag and set image_type accordingly
- Add `use_rle` to enable RLE on a `TgaEncoder`
- `TgaEncoder::new` defaults to rle = false
- Add `TgaEncoder::write_rle_encoded_packet` to write a run-length encoded packet (high bit = 1) to a writer
- Add `TgaEncoder::run_length_encode` which does the actual run-length encoding
- Modified `TgaEncoder::encode` to check for the RLE flag and then use RLE
- Moved old encoding tests to a test submodule, and added the equivalent RLE encoding tests
- Added a test to check for saturating the run-length encoding counter (`round_trip_bw`), I added a test file inside the `tests/` folder, I hope that's okay

### Considerations

`RunColorMap` falls back to uncompressed encoding... should this throw an Unsupported error, because it does not fulfill user expectations?


src/codecs/tga/encoder.rs
```diff
@@ -34,12 +37,60 @@ impl error::Error for EncoderError {}
 /// TGA encoder.
 pub struct TgaEncoder<W: Write> {
     writer: W,
+
+    /// Run-length encoding
+    use_rel: bool,
```
fintelia, line 42: Why "rel" rather than "rle"?

marvin-j97: I don't know how I typed that 20 times without noticing.

marvin-j97:
> Looking at this code, I wonder whether it would make sense to do a prepass over the image to determine whether RLE compression would produce a smaller output than storing the pixels uncompressed. With that logic implemented, it might even make sense to not even expose the `use_rel` switch and just automatically compress if it would help

Even in run-length encoded TGA images there can be packets without RLE (high bit = 0).
So you can have a mix of run-length encoded packets and raw packets.
That would require some lookahead or buffer, basically as long as pixels are different, try to put them in raw packets, and when pixels are repeating, write a run-length packet instead.
Like the RLE packet, the raw packet can have 128 items max, so after 128 different pixels, it would need to be spilled to the writer and then continue with a new packet.

Could look like this:

[RLE packet header with counter = 4] PIXEL
[Raw packet header with counter = 50] PIXEL 1, PIXEL 2, ..., PIXEL 50

fintelia:
I think we should probably add raw packets then. Having your RGB image 0.3% larger because you enabled compression is probably OK, while having it become 33% larger is more frustrating

marvin-j97:
Got it working, I think. My (very noisy) test image now only goes to 8.8 MB, instead of 10.8 MB.

fintelia:
Could you make RLE the default but instead of removing uncompressed encoding, swap the builder method to be `disable_rle`? That way if anyone needs uncompressed they still have a way to get it, but calling `DynamicImage::save_with_format` will produce compressed TGAs.

src/codecs/tga/encoder.rs
```diff
@@ -71,22 +176,48 @@ impl<W: Write> TgaEncoder<W> {
...
-        match color_type {
-            ColorType::Rgb8 | ColorType::Rgba8 => {
-                let mut image = Vec::from(buf);
+        let image_type = ImageType::new(header.image_type);
 
-                for chunk in image.chunks_mut(usize::from(color_type.bytes_per_pixel())) {
-                    chunk.swap(0, 2);
-                }
+        match image_type {
+            //TODO: support RunColorMap, and change match to image_type.is_encoded()
```
fintelia, line 185: nit: is_encoded got renamed

fintelia (approved):

src/codecs/tga/encoder.rs
marvin-j97: Has it? Where exactly, because I didn't change it

fintelia: I think it was in an earlier version of the PR. Don't worry about it

PR #2058, opened 2023-11-17
Add `rayon` parallel iterators

ScottyThePilot:
This adds variants of the `Pixels`/`PixelsMut`/`EnumeratePixels`/`EnumeratePixelsMut` iterators that implement `ParallelIterator`/`IndexedParallelIterator` from rayon, as well as a `from_par_fn` function on `ImageBuffer`. This will allow large image buffers to be manipulated with multi-threading easily. Everything is also under a new optional `rayon` feature flag. This PR doesn't currently do it, but it might be worth considering re-exporting rayon somewhere.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ color_quant = "1.1"
 exr = { version = "1.5.0", optional = true }
 qoi = { version = "0.4", optional = true }
 libwebp = { package = "webp", version = "0.2.2", default-features = false, optional = true }
+rayon = { version = "1.8.0", optional = true }
 
 [dev-dependencies]
 crc32fast = "1.2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ color_quant = "1.1"
 exr = { version = "1.5.0", optional = true }
 qoi = { version = "0.4", optional = true }
 libwebp = { package = "webp", version = "0.2.2", default-features = false, optional = true }
-rayon = { version = "1.8.0", optional = true }
+rayon = { version = "1.7.0", optional = true }
 
 [dev-dependencies]
 crc32fast = "1.2.0"
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -888,7 +888,7 @@ where
     Container: Deref<Target = [P::Subpixel]> + DerefMut,
 {
     // TODO: choose name under which to expose.
-    fn inner_pixels_mut(&mut self) -> &mut [P::Subpixel] {
+    pub(crate) fn inner_pixels_mut(&mut self) -> &mut [P::Subpixel] {
         let len = Self::image_buffer_len(self.width, self.height).unwrap();
         &mut self.data[..len]
     }
--- a/src/buffer_par.rs
+++ b/src/buffer_par.rs
@@ -8,6 +8,7 @@ use crate::traits::Pixel;
 use crate::ImageBuffer;
 
 /// Parallel iterator over pixel refs.
+#[derive(Clone)]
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -43,6 +43,7 @@ color_quant = "1.1"
 exr = { version = "1.5.0", optional = true }
 qoi = { version = "0.4", optional = true }
 libwebp = { package = "webp", version = "0.2.2", default-features = false, optional = true }
+rayon = { version = "1.8.0", optional = true }
```
fintelia, line 46: Are you able to depend on 1.7.0 instead? We currently target a lower MSRV than the 1.8.0 series.

To get the CI to pass, you probably will also have to run something like:
```bash
mv Cargo.lock.msrv Cargo.lock
cargo update -p rayon --precise 1.7.0
mv Cargo.lock Cargo.lock.msrv
```

src/buffer.rs
```diff
@@ -888,7 +888,7 @@ where
     Container: Deref<Target = [P::Subpixel]> + DerefMut,
 {
     // TODO: choose name under which to expose.
-    fn inner_pixels_mut(&mut self) -> &mut [P::Subpixel] {
+    pub(crate) fn inner_pixels_mut(&mut self) -> &mut [P::Subpixel] {
```
fintelia, line 891: Mostly a note to myself: this method should theoretically be the same as the `DerefMut` implementation below, but actually behaves slightly differently. I want to look into why, and maybe even eliminate the duplication

src/buffer_par.rs
```diff
@@ -0,0 +1,523 @@
...
+    }
+
+    fn with_producer<CB: ProducerCallback<Self::Item>>(self, callback: CB) -> CB::Output {
+        self.chunks
+            .map(|v| <P as Pixel>::from_slice(v))
+            .with_producer(callback)
+    }
+}
+
+impl<P> Clone for PixelsPar<'_, P>
```
fintelia, line 62: Can this implementation be derived automatically?

src/buffer_par.rs
```diff
@@ -0,0 +1,523 @@
...
+    P::Subpixel: Sync,
+{
+    fn clone(&self) -> Self {
+        PixelsPar {
+            chunks: self.chunks.clone(),
+        }
+    }
+}
+
+impl<P> fmt::Debug for PixelsPar<'_, P>
```
fintelia, line 74: This also might be automatically derivable?

fintelia (commented):
Thanks for this PR! I think this makes a nice addition to the existing API.

I haven't gotten a chance to fully review the change, but I commented on a few things that I noticed. I think you'll also need to run `cargo fmt` to fix some of the formatting so that the CI passes

src/buffer_par.rs
ScottyThePilot: It looks like `derive(Clone)` on this automatically puts a `Clone` bound on `P`. Since `Pixels`/`EnumeratePixels` doesn't have this bound, I assume it's unnecessary.

src/buffer_par.rs
ScottyThePilot: Just like clone, `derive(Debug)` also seems to put unnecessary bounds on `P` (when it should be just `P::Subpixel`, I think), and also breaks on `EnumeratePixelsPar` and `EnumeratePixelsMutPar` for some reason.

src/buffer_par.rs
fintelia: The `P: Pixel` bounds implies `P: Clone` (see https://docs.rs/image/latest/image/trait.Pixel.html)

PR #2059, opened 2023-11-17
pnm: Scale samples according to specified maximum

sophie-h:
Some formats allow to specify a maximum value for samples. Before, samples were not scaled respective to this white point, decoding some images incorrectly dimmed.

### PAM (P7)

I think the tests were wrong before. But I'm not 100% confident since I haven't found something to test this format in practice.

### Example

Comparison of `image-rs` and correct rendering.

Screenshot from 2023-11-17 23-32-57 Screenshot from 2023-11-17 23-33-06

```
P3 3 3 20
20 0 0
0 20 0
0 0 20
20 20 0
0 20 20
20 0 20
20 20 20
0 0 0
0 0 0
```

### DIFF
--- a/src/codecs/pnm/decoder.rs
+++ b/src/codecs/pnm/decoder.rs
@@ -60,6 +60,8 @@ enum DecoderError {
     UnexpectedByteInRaster(u8),
     /// Specified sample was out of bounds (e.g. >1 in B&W)
     SampleOutOfBounds(u8),
+    /// The image's maxval is zero
+    MaxvalZero,
     /// The image's maxval exceeds 0xFFFF
     MaxvalTooBig(u32),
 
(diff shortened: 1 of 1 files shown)


src/codecs/pnm/decoder.rs
```diff
@@ -659,10 +667,35 @@ impl<R: Read> PnmDecoder<R> {
...
+            } else if S::sample_size() == 2 {
+                for chunk in buf.chunks_exact_mut(2) {
+                    let v = NativeEndian::read_u16(chunk);
+                    NativeEndian::write_u16(chunk, (v as f32 * factor).round() as u16);
+                }
             }
-            SampleEncoding::Ascii => self.read_ascii::<S>(buf),
         }
+
+        dbg!(buf);
```
fintelia, line 696: nit: remove

fintelia:
Does the decoder return an error if the image sets maxval=0? If not, we need to add that or we'll end up with a divide by zero

sophie-h:
CI failures seem to be hiccups?

fintelia:
Yep, everything passed when I tried rerunning. Thanks!

PR #2065, opened 2023-12-09
refactor: replace the (dublicated) impl blocks with the equivalent macros

0x61nas:
Apply DRY (Don't Repeat Yourself) rule and replace the (simirer) impl blocks with equivalent macros


0x61nas:
Fair point

PR #2066, opened 2023-12-10
Show features in documentation

DerZade:

---

This PR will improve the crate's docs.rs page by showing everything that is exported (even those things which are not included in the default features). Everything that is behind a feature flag will be marked as such.

---

For anyone, who wants to try this out locally:
```
cargo +nightly rustdoc --all-features --open -- --cfg docsrs
```


Cargo.toml
```diff
@@ -55,6 +55,10 @@ criterion = "0.5.0"
 # feature when testing, so `cargo test` works correctly.
 jpeg = { package = "jpeg-decoder", version = "0.3.0", default-features = false, features = ["platform_independent"] }
 
+[package.metadata.docs.rs]
+all-features = true
```
fintelia, line 59: I think we have to avoid enabling the `avif` and `benchmarks` feature or we'll have build issues when docs.rs tries to compile the crate. The one feature we do want to enable however is rayon:

```suggestion
features = ["rayon"]
```

fintelia:
Thanks for this PR! I think the approach is nicer than in #2063

DerZade:
> Thanks for this PR! I think the approach is nicer than in #2063

Oh I didn't realize that there was already a PR. Tbh I didn't even check 🙈. I stumbled upon this because I wanted to use the WebPEncoded, which is not (entirely) documented atm, since it is not included in the default features. Instead of complaining I just thought I'd copy paste the solution from one of my projects open a PR 😇

Cargo.toml
DerZade: _Just having looked at the code for a tiny bit_

I don't think `benchmarks` should be an issue since everything that depends on `benchmark` is behind `#[cfg(test)]` anyway. A quick local rustdocs build confirmed that or am I missing something here?

Why is `aviv` a problem? Because it depends on the `dav1d` system lib? Tbh I don't really have experience with docs.rs and system libraries, but I would hope that sys libraries are not an issue for docs.rs, since there are many crates with system lib dependencies on docs.rs, right? That is just me assuming tho 😅, do you have any experience with that?

DerZade: So looks like `dav1d` does not work with docs.rs ([see here](https://docs.rs/crate/dav1d/latest)), which is kind of a shame. I wish they would just handle the docs.rs environment in their build script, like suggested [here](https://docs.rs/about/builds#detecting-docsrs). Maybe I'll add a PR for the `dav1d-sys` crate in the next few days (don't know if I'll get to it tho).

Apart from that, there apparently is [a way to add system libraries to the docs.rs build environment](https://forge.rust-lang.org/docs-rs/add-dependencies.html) as well as [a way to test the build in the same environment as docs.rs](https://forge.rust-lang.org/docs-rs/add-dependencies.html#testing-the-image). I'll see if I can find some time to test it in the next few days, but again, I can't make any promises.

DerZade:
For the record I just opened a PR for `dav1d-rs` (https://github.com/rust-av/dav1d-rs/pull/90#pullrequestreview-1780269161) to fix their docs.rs build.

fintelia:
Ok, lets wait for `dav1d-rs` to release.

I figured out how to test how docs.rs will build a crate:

```bash
# Pull the docker container
docker pull ghcr.io/rust-lang/crates-build-env/linux:latest

# Run docker
docker run --name docsrs -it -p 127.0.0.1:8000:8000 ghcr.io/rust-lang/crates-build-env/linux:latest

# Install rust (be sure to select 'nightly')
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

# Download the crate
git clone ...

# Compile (the 'RUSTDOCFLAGS' and '--all-features' are based on the contents in Cargo.toml)
RUSTDOCFLAGS="--cfg docsrs" cargo rustdoc --all-features

# Launch a webserver to display the docs
cd target/doc/image && python3 -m http.server
```

DerZade:
I updated `dav1d` to the latest version (which works on `docs.rs`) and  added the `package.metadata.docs.rs` section to the `Cargo.toml`.

I also tested the docs build with the container as suggested by @fintelia

PR #2068, opened 2023-12-11
Fix the supported image formats table

### LINKED ISSUE #2067: Radiance HDR is missing in the supported image formats table
`ImageFormat` and `ImageOutputFormat` are defined as follows:

<https://github.com/image-rs/image/blob/81b3fe66fba04b8b60ba79b3641826df22fca67e/src/image.rs#L28-L73>
<https://github.com/image-rs/image/blob/81b3fe66fba04b8b60ba79b3641826df22fca67e/src/image.rs#L382-L439>

The supported image formats table is as follows:

<https://github.com/image-rs/image/blob/81b3fe66fba04b8b60ba79b3641826df22fca67e/README.md?plain=1#L23-L38>
<https://github.com/image-rs/image/blob/81b3fe66fba04b8b60ba79b3641826df22fca67e/src/lib.rs#L211-L226>

I think Radiance HDR is not listed in the supported image formats table. Also, other image formats defined in `ImageFormat` and `ImageOutputFormat` seems to be listed.

sorairolake:
Add Radiance HDR.

Closes #2067

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -208,22 +208,23 @@ pub mod flat;
 ///
 /// <!--- NOTE: Make sure to keep this table in sync with the README -->
 ///
-/// | Format | Decoding | Encoding |
-/// | ------ | -------- | -------- |
-/// | AVIF   | Only 8-bit | Lossy |
-/// | BMP    | Yes | Rgb8, Rgba8, Gray8, GrayA8 |
-/// | DDS    | DXT1, DXT3, DXT5 | No |
-/// | Farbfeld | Yes | Yes |
-/// | GIF    | Yes | Yes |
-/// | ICO    | Yes | Yes |
-/// | JPEG   | Baseline and progressive | Baseline JPEG |
-/// | OpenEXR  | Rgb32F, Rgba32F (no dwa compression) | Rgb32F, Rgba32F (no dwa compression) |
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -208,22 +208,23 @@ pub mod flat;
...
+/// | DDS          | DXT1, DXT3, DXT5                          | No                                      |
+/// | Farbfeld     | Yes                                       | Yes                                     |
+/// | GIF          | Yes                                       | Yes                                     |
+/// | ICO          | Yes                                       | Yes                                     |
+/// | JPEG         | Baseline and progressive                  | Baseline JPEG                           |
+/// | OpenEXR      | Rgb32F, Rgba32F (no dwa compression)      | Rgb32F, Rgba32F (no dwa compression)    |
+/// | PNG          | All supported color types                 | Same as decoding                        |
+/// | PNM          | PBM, PGM, PPM, standard PAM               | Yes                                     |
+/// | QOI          | Yes                                       | Yes                                     |
+/// | Radiance HDR | Yes                                       | No                                      |
```
fintelia, line 224: Could you just call this "HDR" rather than "Radiance HDR"? Also we do support encoding: https://docs.rs/image/latest/image/codecs/hdr/struct.HdrEncoder.html

fintelia:
Thanks!

PR #2075, opened 2023-12-18
tiff: Support CMYK images

### LINKED ISSUE #2074: Support CMYK conversion (for TIFFs)
The following image 

- https://exoplanets.nasa.gov/rails/active_storage/blobs/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaXdGIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--bb2d9b418904bbf57e31960b5174b487a5da70de/Dark_Matter_27x39_ENGLISH_copy_UPDATED.tif

gives this error with image-rs

```
The decoder for Tiff does not support the color type `Unknown(32)`
```

I'm not sure if that's the exact same missing feature as image-rs/image-tiff#202 since instead of "palette", this image reports "separated."

```
=== TIFF directory 0 ===
TIFF Directory at offset 0x8 (8)
  Subfile Type: (0 = 0x0)
  Image Width: 8100 Image Length: 12021
  Resolution: 300, 300 pixels/inch
  Bits/Sample: 8
  Compression Scheme: LZW
  Photometric Interpretation: separated
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 4
  Rows/Strip: 8
  Planar Configuration: single image plane
  Software: Adobe Photoshop 22.0 (Windows)
```

sophie-h:
- [x] Cleanup tiff code

Closes #2074


src/traits.rs
```diff
@@ -44,12 +44,22 @@ pub trait Primitive: Copy + NumCast + Num + PartialOrd<Self> + Clone + Bounded {
     const DEFAULT_MIN_VALUE: Self;
 }
 
+pub trait IntoF32 {
```
sophie-h, line 47: We could also use `NumCast` instead but that would add some unwraps.

src/color.rs
```diff
@@ -682,6 +688,57 @@ where
...
+    T: FromPrimitive<S>,
+{
+    fn from_color(&mut self, other: &Cmyk<S>) {
+        let [r, g, b] = &mut self.0;
+
+        (*r, *g, *b) = cmyk_to_rgb(other.0);
+    }
+}
+
+fn cmyk_to_rgb<S: Primitive, T: Primitive>([c, m, y, k]: [S; 4]) -> (T, T, T) {
```
sophie-h, line 702: That's the best I could come up with. The intermediate conversion to f32 seems weird, but I also don't know how to avoid it.

src/codecs/tiff.rs
```diff
@@ -74,12 +76,29 @@ where
...
             color_type,
+            original_color_type,
             inner: Some(inner),
         })
     }
+
+    fn total_buffer_bytes(&self) -> u64 {
+        let dimensions = self.dimensions();
+        let total_pixels = u64::from(dimensions.0) * u64::from(dimensions.1);
+        let bytes_per_pixel = if self.original_color_type == ExtendedColorType::Cmyk8 {
```
sophie-h, line 95: That needs a `bytes_per_pixel` for `ExtendedColorType`

sophie-h (commented):
Maybe someone give initial feedback if my plans are viable before I finish this,

src/color.rs
```diff
@@ -367,6 +371,8 @@ define_colors! {
     pub struct Rgba<T: Primitive Enlargeable>([T; 4, 1]) = "RGBA";
     /// Grayscale colors + alpha channel
     pub struct LumaA<T: Primitive>([T; 2, 1]) = "YA";
+    /// CMYK
+    pub struct Cmyk<T: Primitive>([T; 4, 1]) = "CMYK";
```
sophie-h, line 375: This would all be much more straightforward if we don't implement Cmyk as a type. Is it helpful for other things to have it or is it more of an annoyance?

src/codecs/tiff.rs
fintelia: We could have a `ExtendedColorType::bits_per_pixel` method, but thanks to variants like `ExtendedColorType::L1` having an integers bytes per pixel isn't going to work :)

fintelia:
I'd feel better restricting this PR to just the changes in tiff.rs and maybe the addition of `ExtendedColorType::Cmyk8`. The `read_image` method could directly call a crate private `cmyk_to_rgb` method.

src/color.rs
fintelia: I'd rather not add this struct

src/codecs/tiff.rs
```diff
@@ -234,6 +258,16 @@ impl<'a, R: 'a + Read + Seek> ImageDecoder<'a> for TiffDecoder<R> {
             .read_image()
             .map_err(ImageError::from_tiff_decode)?
         {
+            tiff::decoder::DecodingResult::U8(v)
+                if self.original_color_type == ExtendedColorType::Cmyk8 =>
+            {
+                let mut in_cur = Cursor::new(v);
+                let mut out_cur = Cursor::new(buf);
+                let mut cmyk = [0_u8; 4];
+                while in_cur.read_exact(&mut cmyk).is_ok() {
```
fintelia, line 267: Could you instead use `chunks_exact` for this iteration?

```rust
for (rgb, cmyk) in buf.chunks_exact(3).zip(v.chunks_exact(4)) {
    rgb.copy_from_slice(&cmyk_to_rgb(cmyk.try_into().unwrap());
}
```

(This would be nicer with array_chunks but that's not stable yet)

sophie-h: Thanks, I always forget which ones are stabilized at all.

PR #2076, opened 2023-12-19
implement From<DynamicImage> for all image types

drewcassidy:
Allows for generically converting from a dynamicImage. This is something that would be handy to have in another project, and I realized it would be easy to add myself.


fintelia (approved):

fintelia:
I feel like I'm missing something about we haven't added this before, but if no one raises any concerns then I'm OK with merging

Issue #1052, opened 2019-10-02
Large CPU and memory consumption on decoding a crafted GIF file


Shnatsel:
Tested on latest git commit as of this writing, ac93e755b51def00c93e7fadd33a5e5593afa1e2

## Actual behavior

Decoding a crafted file mere 469 bytes in size takes over 7 gigabytes of memory. And not just virtual memory - it does get actually used and backed by physical memory.

This sample does not pose a problem as is, but it's possible that by applying the same technique it's possible to create very small files that OOM the system.

## Expected

imagemagick rejects these files immediately. For the exact reason for each file see https://ncry.pt/p/6bPn#OPffqnjbN7RwGKJMwA0BQl3wuNlynKoUxw2dXZx9UKY

## Reproduction steps

Same reproduction code as in #876

I'm attaching all samples classified as hangs by AFL so that you have the most complete info. The file `id:000002,src:000000+000052,op:splice,rep:128` is the most egregious in terms of time and memory consumption.

[image_gif_timeouts.zip](https://github.com/image-rs/image/files/3683273/image_gif_timeouts.zip)

HeroicKatora:
Woah! `gif` only has 16-bit sizes so this is incredibly excessive.

HeroicKatora:
Or just enough to be possible with 2^32 * 4 bytes (`width×height×bpp`). Either way, it's a good reminder to have better limits.

Shnatsel:
For more on memory limits see #938

fintelia:
In fact, the code from #876 *does* implement a memory limit. `width * height < 2_000_000_000` at 4 bytes per pixel enforces a limit of about 7.45 GiB. I don't think this particular case is especially worrying, but we should have a more general solution. 

I alluded to this in that other thread, but perhaps our new Reader API should have:
```rust
/// Set the limit on total memory use. Attempting to read a file which would take more 
/// space than this will fail with ImageError::InsufficientMemory. Defaults to 128MB, set to 
/// `usize::MAX` to disable.
Reader::with_memory_limit(bytes: usize);
/// Set the limit on image size. Defaults to 16,384 x 16,384.
Reader::with_size_limit(width: u32, height: u32);
```

HeroicKatora:
Good catch, I had for some reason thought the limit was `2_000_000` pixels and there was a verification bug in the existing code. I'm glad it's just a little less bad.

Shnatsel:
[...]

PR #2083, opened 2024-01-03
Fix `GifDecoder::with_limits` to raise an error when limits are exceeded

Shnatsel:
Partly addresses #1052

### DIFF
--- a/src/codecs/gif.rs
+++ b/src/codecs/gif.rs
@@ -69,10 +69,16 @@ impl<R: Read> GifDecoder<R> {
         let mut decoder = gif::DecodeOptions::new();
         decoder.set_color_output(ColorOutput::RGBA);
 
-        Ok(GifDecoder {
+        let mut img_decoder = GifDecoder {
             reader: decoder.read_info(r).map_err(ImageError::from_decoding)?,
-            limits,
-        })
+            limits: Limits::default(),
+        };
+
+        // call `.set_limits()` instead of just setting the field directly
+        // so that we raise an error in case they are exceeded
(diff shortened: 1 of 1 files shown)


fintelia (commented):
Could you instead reuse the logic from `new`?

```rust
#[deprecated("Use `new` followed by `set_limits` instead")]
pub fn with_limits(r: R, limits: Limits) -> ImageResult<GifDecoder<R>> {
    let mut decoder = Self::new(r)?;
    decoder.set_limits(limits)?;
    Ok(decoder)
}

Shnatsel:
Done. Do you want me to deprecate `with_limits` on `PngDecoder` as well?

Shnatsel:
At a glance, the `PngDecoder` does not seem to implement `set_limits`, so the `with_limits` constructor is the only way to set them reliably. That's probably also something worth fixing, if that's actually true.

Shnatsel:
You know what, let's just merge this as-is since it's a clear fix, and I'll do my own research on the limits across all decoders. I don't want to distract you from the reviews on the `png` crate and your work on WebP.

fintelia:
`PngDecoder` should also be changed. But sadly we can't support `set_limits` there without a breaking change.

The issue is a subtle problem with the type signature of `PngDecoder::is_apng` that doesn't let us return limit or parsing errors. Instead we have to do metadata parsing ahead of time in `new`/`with_limits` so that we can report those errors. But once we've parsed the metadata the crate doesn't let us set the allocation limit (and that's the correct approach because we want to use the limit to avoid allocating an unbounded amount of space for storing metadata!)

PR #2085, opened 2024-01-04
Add tests for resource limits

Shnatsel:
This PR adds tests verifying that all decoding APIs enforce the resource limits.


Shnatsel:
The fix actually broke some other stuff, so I'm going to make this PR purely about tests and then chip away at the test failures in subsequent PRs.

Shnatsel:
This is now ready for review.

Since this only adds tests, the changes should be uncontroversial and won't break anything.

I'll propose fixes in subsequent PRs once this is merged.

tests/limits.rs
```diff
@@ -0,0 +1,228 @@
...
+    // PngDecoder
+    let mut decoder = PngDecoder::new(Cursor::new(&image)).unwrap();
+    assert!(decoder.set_limits(width_height_limits()).is_err());
+    // let mut decoder = PngDecoder::new(Cursor::new(&image)).unwrap();
+    // assert!(decoder.set_limits(allocation_limits()).is_err()); // BROKEN!
+    // The proper fix is known to require an API break: https://github.com/image-rs/image/issues/2084
+
+    // Custom constructor on PngDecoder
+    assert!(PngDecoder::with_limits(Cursor::new(&image), width_height_limits()).is_err());
+    //assert!(PngDecoder::with_limits(Cursor::new(&image), allocation_limits()).is_err()); // BROKEN!
```
fintelia, line 121: This isn't quite how the allocation limit is expected to work. The limit only counts bytes allocated by the decoder. And since the output buffer is provided by the caller for `ImageDecoder::read_image`, it doesn't count towards the limit. Only intermediate buffers count.

Further, decoders are generally expected to return allocation errors when `read_image` is called, rather than requiring them to anticipate how much memory will be required based only on the file header. And since many decoders attempt to read in somewhat of a streaming fashion, they may use very little memory even when decoding large images. (And conversely, crafted images with small total dimensions could take large amounts of memory to decode)

Shnatsel: The generic `image::io::Reader` actually deducts the output buffer size from the allocation limit.[*] So `.set_limits()` clearly sets the total allocation budget, _including_ the output buffer size.

There is nothing on the `with_limits()` constructor to indicate that it would behave differently to `set_limits`, they both accept the same `Limits` struct. Perhaps changing the behavior would be a breaking change, but if we consider this intended behavior, at least it must be clearly documented.

[*] this causes a double-counting bug in TIFF decoder because it deducts the output buffer size from the limits _again._

fintelia: Sorry, I didn't mean to say there was something special about `with_limits` vs `set_limits`. In both cases they shouldn't deduct the output buffer size.

The reason `io::Reader` deducts the output buffer size is because its `decode` method actually allocates a buffer of that size. By contrast, `read_image` only requires a slice of that size to be passed in as an argument. Someone has no right to complain if they get back a huge number of `total_bytes` and then go off and allocate that many bytes own their own.

At the point that `set_limits` is called, we don't even know if the user will try to decode the whole image at once. Depending on the decoder, they may instead opt to call `read_rect` and decode only a subset of the image. Or maybe they want the allocation limit only for querying other image metadata and don't plan to decode image contents at all.

> [*] this causes a double-counting bug in TIFF decoder because it deducts the output buffer size from the limits again.

Sadly this isn't a bug. The TIFF decoder internally allocates a working buffer the same size as the output due to API mismatches between the `image` and `tiff` crates.

Shnatsel: Ahhh, I see. That makes sense. 

So to fix #1052 I'll have to modify APIs such as `into_frames`.

I'll drop the erroneous tests from this PR and see if I can add tests for `into_frames` and such. 

Thank you for the review!

Shnatsel:
The tests now cover the existing API surface and verify that the parts that are already in place work correctly.

I'll leave tests for and implementation of resource limits in animation to follow-up PRs.

Issue #1890, opened 2023-03-15
Document that not calling `decoder.set_limits()` means no limits, not default limits


Shnatsel:
Continuing from [here](https://github.com/image-rs/image/issues/1889#issuecomment-1469127654)

> better document that not calling `set_limits` is implicitly using `Limits::no_limits` rather than `Limits::default` as you might expect

PR #2090, opened 2024-01-09
Fix setting GIF limits, implement limits for GIF animations

### LINKED ISSUE #1052: Large CPU and memory consumption on decoding a crafted GIF file

Shnatsel:
1. Fix `.set_limits()` on the GIF decoder to actually set limits
2. Implement limit tracking in the GIF animation decoder, so APIs such as `.into_frames()` now also obey both image size and allocation limits, including for intermediate buffers
3. Add tests to verify limits in animation work correctly, which found (1)

At long last, fixes #1052


src/codecs/gif.rs
```diff
@@ -58,6 +58,12 @@ impl<R: Read> GifDecoder<R> {
...
 
+        // We previously used `Limits::default()`, but they weren't actually enforced.
+        // Suddenly enabling enforcement while keeping default limits would break existing users.
+        // Therefore, set no limit at least until the next semver break, 0.25.x
+        let mut limits = Limits::default();
+        limits.max_alloc = None;
+
         Ok(GifDecoder {
             reader: decoder.read_info(r).map_err(ImageError::from_decoding)?,
             limits: Limits::default(),
```
kornelski, line 69: should the struct be given the limits variable?

Shnatsel: :facepalm: 

Good catch! Done.

kornelski (approved):

src/codecs/gif.rs
```diff
@@ -58,9 +58,15 @@ impl<R: Read> GifDecoder<R> {
         let mut decoder = gif::DecodeOptions::new();
         decoder.set_color_output(ColorOutput::RGBA);
 
+        // We previously used `Limits::default()`, but they weren't actually enforced.
+        // Suddenly enabling enforcement while keeping default limits would break existing users.
+        // Therefore, set no limit at least until the next semver break, 0.25.x
+        let mut limits = Limits::default();
+        limits.max_alloc = None;
```
fintelia, line 65: I guess we forgot to follow up on https://github.com/image-rs/image/issues/1890, but we should continue the convention defined there of having the `new()` method initialize to `no_limits()` rather than `default()`

Shnatsel: I've updated the comments to reflect that and simplified the code a bit. This PR should be good to go.

Issue #2091, opened 2024-01-09
Add caching to the CI fuzzing target


Shnatsel:
CI fuzz job takes a long time not because it fuzzes for a while, but because it takes a long time to compile. 

It should be made to use caching, just like the rest of CI.

fintelia:
Fixed in #2157. The build itself isn't cached, but the installation of cargo-fuzz is. Also switched to "check" instead of a full build which further sped things up

PR #2107, opened 2024-01-14
Cache Cargo dependencies for fuzzing targets in CI

astral4:
Addresses #2091.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -129,6 +129,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: dtolnay/rust-toolchain@nightly
+    - name: cache Cargo dependencies
+      uses: Swatinem/rust-cache@v2
+      with:
+        cache-on-failure: true
     - name: install-deps
       run: sudo apt-get -y install clang llvm
     - name: build
@@ -148,6 +152,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: dtolnay/rust-toolchain@nightly
+    - name: cache Cargo dependencies
+      uses: Swatinem/rust-cache@v2
+      with:
+        cache-on-failure: true
     - name: build
       run: |
         cargo install cargo-fuzz


fintelia:
I don't think this works. 

I tried triggering a re-run of the afl target to see if it would use the cache, and it seems to have both downloaded/re-compiled all the dependencies and then failed saying:
```
    Checking image-fuzz-afl v0.0.1 (/home/runner/work/image/image/fuzz-afl)
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
AFL LLVM runtime was not built for Rust rustc-1.77.0-nightly-d78329b; run `cargo afl config --build` to build it.
Error: Process completed with exit code 1.
```

astral4:
Seems like `cargo-fuzz` re-compiles dependencies as well, even though there is no error.

astral4:
I'm not sure how to get caching to work for `cargo-afl`. It seems that they include the source of `afl++` at build time (see [here](https://github.com/rust-fuzz/afl.rs/blob/e25e65a63fcd978e4b2fc444268059bfda7eea66/cargo-afl/build.rs)), which doesn't work nicely with caches in GitHub Actions like `Swatinem/rust-cache@v2` or `actions/cache@v3`.

The current build times for `cargo-afl` might be acceptable, since that job doesn't seem to be the bottleneck (i.e. the job that finishes last) for CI times.

astral4:
Other people seem to experience long `cargo-fuzz` compile times as well - see https://github.com/rust-lang/rust/issues/95240.

Changing [this line](https://github.com/image-rs/image/blob/778f18565b40bdc99b8aafef7d631d8fe7a99413/.github/workflows/rust.yml#L158) seems to help a little:
```diff
- cargo fuzz run "$format" -- -runs=0;
+ RUSTFLAGS="-C codegen-units=16" cargo fuzz run "$format" -- -runs=0;
```
(I picked `codegen-units=16` because it's the default value for non-incremental builds, according to the rustc [docs](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units).)

In the end, if the goal is to reduce CI time, I am no longer sure that caching dependencies is the way to go.

Shnatsel:
> Since the cargo-afl and cargo-fuzz jobs use the latest nightly toolchain, dependencies will usually have to be re-compiled between runs anyway.

We could switch to a specific, pinned nightly version to mitigate that.

fintelia:
I imagine we'd then have to periodically bump the pinned version when it got too old and CI started failing. Not sure that's worth it

PR #2110, opened 2024-01-16
Use `chunks_exact_mut` in PNG decoder

Shnatsel:
`chunks_mut` can return chunks with sizes 1 or 2, while `chunks_exact_mut` always returns chunks of size 2. This should result in better codegen.

This certainly could be vectorized further, but this codepath is only taken for 16-bit images which are uncommon, so I'm not going to invest a lot of my time into optimizing it.

In the current code a chunk of size 1 being returned would result in a panic, so this should not break anything.

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -676,7 +676,7 @@ impl<W: Write> ImageEncoder for PngEncoder<W> {
                 // yet take Write/Read traits, create a temporary buffer for
                 // big endian reordering.
                 let mut reordered = vec![0; buf.len()];
-                buf.chunks(2)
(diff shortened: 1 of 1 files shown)


src/codecs/png.rs
```diff
@@ -677,7 +677,7 @@ impl<W: Write> ImageEncoder for PngEncoder<W> {
                 // big endian reordering.
                 let mut reordered = vec![0; buf.len()];
                 buf.chunks(2)
```
fintelia, line 679: Shouldn't this also be `chunks_exact`?

Shnatsel: Good catch! Done!

Issue #2111, opened 2024-01-16
PNG decoding relies on `output_buffer_size()` function that may overflow


Shnatsel:
This happens in image 0.24.8

## Expected

Images with buffer sizes too large for `usize` rejected early

## Actual behaviour

I see PNG decoding code using `output_buffer_size()` from the PNG crate in multiple places. Looking at its implementation, the computation may silently overflow.

This was recently fixed in GIF in #2103, but PNG seems to be affected too.

I haven't actually tested this - there might be some early check that catches this, but even so the code is fragile.

fintelia:
If anyone wants to look at resolving this, the `total_bytes()` method on `PngDecoder` (via the `ImageDecoder` trait) returns the same value, but as a `u64` and computed with saturating multiplies so it doesn't overflow.

Issue #2084, opened 2024-01-04
`PngDecoder` supports setting limits via `with_limits` but not via `set_limits`


Shnatsel:
This happens in `image` v0.24.7.

## Expected

`PngDecoder` should support setting resolution and memory limits via the generic `set_limits` method, not only via its own `with_limits` method.

## Actual behaviour

Only the `with_limits` constructor sets limits correctly; setting limits via `set_limits` is not implemented.

Quoting @fintelia :

> `PngDecoder` should also be changed. But sadly we can't support `set_limits` there without a breaking change.
> 
> The issue is a subtle problem with the type signature of `PngDecoder::is_apng` that doesn't let us return limit or parsing errors. Instead we have to do metadata parsing ahead of time in `new`/`with_limits` so that we can report those errors. But once we've parsed the metadata the crate doesn't let us set the allocation limit (and that's the correct approach because we want to use the limit to avoid allocating an unbounded amount of space for storing metadata!)

mstoeckl:
I may well be wrong here, but would it be better to instead make the functions that create a PngDecoder specify limits immediately? That is, add new `image::open_with_limits`, ~`ImageReader::new_with_limits`, `ImageReader::open_with_limits`~, etc. functions. These could be helpful to library users anyway since the few uses of `ImageDecoder::set_limits` that I've seen do not depend on the image dimensions and color type.

(For context: I recently tried to write an ImageDecoder implementation for the OpenRaster file format. This is a ZIP file which contains layer data plus a `mergedimage.png` composite image ([spec](https://www.openraster.org/)). See https://github.com/mstoeckl/image-extras/blob/ora/src/ora.rs for the implementation. Due to all the layers of abstraction, implementing Seek on a compressed stream is rather awkward and inefficient, and the `zip` crate loads a variable-sized index into memory before the dimensions and color type can be determined.) 

Edit: it turns out that `ImageReader` already supports specifying limits.

Shnatsel:
`ImageReader` already supports limits, as you've already noticed. `image::open` is a high-level convenience function, if you need more control you're expected to create an `ImageReader` and set the options on it, as `image::open` [already documents](https://docs.rs/image/latest/image/fn.open.html).

RunDevelopment:
Fixed by #2672. `PngDecoder::set_limits` now works as expected.

Issue #1748, opened 2022-07-04
OOM in PNG decoding despite a memory limit and dimension limits being set


Shnatsel:
`image` OOMs when attempting to decode a crafted PNG file, discovered by fuzzer.

The fuzzer uses `image::load_from_memory_with_format()` which, according to the documentation, has a default memory limit of 512Mb. However, the decoding process actually allocates way more than that.

I have tried setting the limit explicitly, just in case, and got the same OOM:

```rust
    let mut reader = image::io::Reader::new(Cursor::new(data));
    reader.set_format(image::ImageFormat::Png);
    reader.limits(image::io::Limits::default());
    let _ = reader.decode();
```

This happens on commit 04052e64c9a94606efc8bd3d87d5f3e0f566774e

## Reproduction steps

`cargo +nightly fuzz run --release fuzzer_script_png path/to/file.png`

The file to trigger it: [oom-cf5aeccc25f220c487c913b20c285a860e302ffc.png](https://user-images.githubusercontent.com/291257/177205340-a8f254aa-cb45-444a-bbb2-e46a5c35903a.png)

Shnatsel:
Even setting width and height limits to cap the output at ~130Mb doesn't help. The following code still OOMs:

```rust
    let mut reader = image::io::Reader::new(std::io::Cursor::new(data));
    reader.set_format(image::ImageFormat::Png);
    let mut limits = image::io::Limits::default();
    // add limits to stop the fuzzer from OOMing
    limits.max_image_width = Some(4096);
    limits.max_image_height = Some(4096);
    reader.limits(limits);
    let _ = reader.decode();
```
on this input: [oom-cf5aeccc25f220c487c913b20c285a860e302ffc](https://user-images.githubusercontent.com/291257/177206228-55f4ea34-c3e0-432b-9b98-df2715b85a55.png)

Shnatsel:
FWIW I could not trigger similar failures in JPEG or TIFF.

fintelia:
There's a simple explanation for this issue... PngDecoder ignores any user limits and always creates an underlying decoder that uses an unbounded amount of memory: https://github.com/image-rs/image/blob/c24aecb954b0c9ae96ebdacd023140e323de2d75/src/codecs/png.rs#L120-L123

Unfortunately this is a bit annoying to fix because right now the png crate does potentially large allocations in the constructor, while the ImageDecoder trait allows you to set limits after the object has been created.

fintelia:
[...]

PR #2112, opened 2024-01-16
Support limits in PNG animation decoding

Shnatsel:
Implements memory limit for animation decoding and dimensions limits for individual animation frames, similar to how 
it's done for GIF.

Notable changes:

- Added a helper function `reserve_buffer` on `Limits` to be reused in PNG and GIF animation decoders
- Implemented allocation tracking in `ApngDecoder`
- Changed the PNG decoder's `new()` to default to no limits (consistently with other decoders) now that the limits are actually enforced

Out of scope of this PR:

 - Passing through the limits to the `png` crate for it to limit its own allocations in animation decoding because the required APIs for that are missing from `png`
 - `png` crate actually enforcing the limits it's passed: https://github.com/image-rs/image-png/issues/286 (rendering the previous bullet point moot)
 - #1748 still isn't fixed (I think?). A proper fix would require an API break so that `is_apng` is allowed to return an error (unless we're OK with it just returning `false` on an error or something?), see #2084

Open questions:

 - Does defaulting to no limits make #1748 worse / easier to trigger? It would if `png` crate allocates large amounts of memory when reading just the headers, but I don't know if it does.
 - Should #2111 be fixed in this PR or separately?

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -402,7 +402,7 @@ impl<R: Read> ApngDecoder<R> {
 
         self.animatable_color_type()?;
 
-        // We've initialized them earlier
+        // We've initialized them earlier in this function
         let previous = self.previous.as_mut().unwrap();
         let current = self.current.as_mut().unwrap();
 
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -338,9 +338,8 @@ impl<R: Read> ApngDecoder<R> {
         let has_thumbnail = info.frame_control.is_none();
         ApngDecoder {
             inner,
-            // TODO: should we delay this allocation? At least if we support limits we should.
-            current: None,  //  RgbaImage::new(width, height),
-            previous: None, // RgbaImage::new(width, height),
+            current: None,
+            previous: None,
             dispose: DisposeOp::Background,
             remaining,
             has_thumbnail,
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -118,7 +118,7 @@ pub struct PngDecoder<R: Read> {
 impl<R: Read> PngDecoder<R> {
     /// Creates a new decoder that decodes from the stream ```r```
     pub fn new(r: R) -> ImageResult<PngDecoder<R>> {
-        Self::with_limits(r, Limits::default())
+        Self::with_limits(r, Limits::no_limits())
     }
 
     /// Creates a new decoder that decodes from the stream ```r``` with the given limits.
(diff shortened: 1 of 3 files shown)


src/codecs/png.rs
```diff
@@ -287,6 +292,16 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for PngDecoder<R> {
         let width = self.reader.info().width;
         self.reader.output_line_size(width) as u64
     }
+
+    fn set_limits(&mut self, limits: Limits) -> ImageResult<()> {
+        limits.check_support(&crate::io::LimitSupport::default())?;
+        let info = self.reader.info();
+        limits.check_dimensions(info.width, info.height)?;
+        self.limits = limits;
+        // TODO: add `png::Reader::change_limits()` and call it here
```
fintelia, line 301: We can't really implement a `change_limits` method on the `png` side.

The problem is that many of the allocations in that crate happen when you create the `png::Reader` by calling `png::Decoder::read_info` because that's where the image metadata is read and saved into the inner Info struct. Ironically, the crate allocates relatively little memory for decoding pixel data, because it is quite good at decoding that data in a streaming fashion.

Rather, the current plan is to switch to calling `Decoder::set_limits` in this method, and only later call `read_info` when the user queries metadata beyond what's stored in the IHDR.

An alternative possibility would be for the `png` to actually not store metadata chunks as it reads them, but rather to store the offsets of those chunks, and then `Seek` past them. This would bound the memory used by the decoder while still enabling the metadata to be loaded later. In that case, the limits could safely be set later. Adding the `Seek` bound would, however, be a breaking change both here and in the png crate.  (An added complication is that `BufReader`'s `seek()` implementation has some performance gotchas. It should be possible to work around them once [`Seek::seek_relative`](https://doc.rust-lang.org/beta/std/io/trait.Seek.html#method.seek_relative) is stabilized, but I don't know if/when that'll happen.)

fintelia (approved):

fintelia:
I think that at some point we should probably move the frame compositing logic into the `png` and `gif` crates respectively, which would also solve the issues around sharing the allocation limit across the crate boundary. But that'll be a larger refactor, so I'm OK with merging this for now

src/codecs/png.rs
Shnatsel: The idea is that `change_limits()` would look at the amount of memory already allocated by the `png` crate and report an error if the crate is already over budget. This would let us easily bail out of the total memory usage limit is exceeded (including all the metadata).

I don't think the `Seek` bound is a good idea (too limiting, largely defeats streaming decoding use cases) but a refactor to move the compositing logic to `png` and track the allocations there could be viable.

Shnatsel:
Then let's go ahead and merge this. I plan to do some more work on the `png` decoder and I don't want to run into editing conflicts with myself :laughing:

PR #2116, opened 2024-01-29
Improve panic messages when providing invalid buffer sizes to `ImageEncoder`s

MrGunflame:
Replaces the cryptic `assertion failed` into a more useful panic message when writing a PNG image with invalid buffer/dimensions.

### DIFF
--- a/src/codecs/avif/encoder.rs
+++ b/src/codecs/avif/encoder.rs
@@ -105,9 +105,16 @@ impl<W: Write> ImageEncoder for AvifEncoder<W> {
         height: u32,
         color: ColorType,
     ) -> ImageResult<()> {
+        let expected_buffer_len =
+            (width as u64 * height as u64).saturating_mul(color.bytes_per_pixel() as u64);
         assert_eq!(
-            (width as u64 * height as u64).saturating_mul(color.bytes_per_pixel() as u64),
-            data.len() as u64
+            expected_buffer_len,
+            data.len() as u64,
+            "Invalid buffer length: expected {} got {} for image dimensions ({}, {})",
+            expected_buffer_len,
+            data.len(),
+            width,
+            height,
         );
 
         self.set_color(color);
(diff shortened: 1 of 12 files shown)


fintelia:
Thanks for this suggestion, I think a clearer panic message would be an improvement. One note is that you don't actually have to switch to `if ... { panic!(...) }` since there's a version of the `assert_eq!` macro that takes a format string as its third argument.

As far as the message itself, I think we might be able to be a bit more concise. Perhaps `"Invalid buffer length: expected {} got {}"? The PNG encoding part should be evident from the backtrace into the `PngEncoder::write_image` method.

Finally, this same assertion is actually present in all the other `ImageEncoder` implementations as well, so I think it would make sense to update them all at the same time.

MrGunflame:
> As far as the message itself, I think we might be able to be a bit more concise. Perhaps `"Invalid buffer length: expected {} got {}"? The PNG encoding part should be evident from the backtrace into the `PngEncoder::write_image` method.

I do think it would be helpful to include how the buffer size was derived, i.e. include `width` and `height` and maybe even `ColorType` as that info makes tracking down the panic easier IMO.

kornelski:
Maybe add `#[track_caller]` on these functions, since the assert is not an internal bug, but the fault of the caller.

src/codecs/avif/encoder.rs
```diff
@@ -98,16 +98,24 @@ impl<W: Write> ImageEncoder for AvifEncoder<W> {
...
         color: ColorType,
     ) -> ImageResult<()> {
+        let expected_buffer_len =
+            (width as u64 * height as u64).saturating_mul(color.bytes_per_pixel() as u64);
         assert_eq!(
-            (width as u64 * height as u64).saturating_mul(color.bytes_per_pixel() as u64),
-            data.len() as u64
+            expected_buffer_len,
+            data.len() as u64,
+            "Invalid buffer length: expected {} got {} for image dimensions ({}, {})",
```
fintelia, line 114: Could you change the message to `"Invalid buffer length: expected {} got {} for {}x{} image"`? Beyond that, this looks good to me.

You can also write `"{width}"` and similar instead of passing a separate argument, though that only works for variables, not expressions like `data.len()`

fintelia:
Thanks!

PR #2117, opened 2024-01-30
Update to gif 0.13

torokati44:
To the tune of #1826.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ include = [
 bytemuck = { version = "1.7.0", features = ["extern_crate_alloc"] } # includes cast_vec
 byteorder = "1.3.2"
 num-traits = "0.2.0"
-gif = { version = "0.12", optional = true }
+gif = { version = "0.13", optional = true }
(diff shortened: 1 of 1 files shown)


torokati44:
Looks like the OOM regression test now fails due to some additional checks added by one of the recent PRs of @kornelski.
Could you maybe offer some advice here please?
Should the test data be altered to pass this new check, or should a different error be expected now?
My hunch is on the former, but am not that familiar with the GIF format to know exactly what to change...

kornelski:
The GIF has lzw code size == 0, but the minimum valid value is 2. You can change the data to this:

```
    let data = [
        71, 73, 70, 56, 55, 97, 0, 0, 0, 0, 0, 0, 0, 44, 255, 255, 219, 255, 172, 199, 199, 255,
        216, 255, 255, 0, 0, 48, 230, 2, 195, 195, 195, 195, 255, 239, 0,
    ];
```

and you will get the Limit error due to default memory limit of 50MB.

torokati44:
Awesome, thank you so much!

PR #2121, opened 2024-01-31
Perform MSRV check via `-Zminimal-versions`

### LINKED ISSUE #2120: Use `-Z minimal-versions` to test MSRV
In https://github.com/johannesvollmer/exrs/pull/231 (CC @johannesvollmer) we're stripping out an upper-version bound for a dependency to allow it to be used across other dependencies, and reimplementing the MSRV test by `-Zminimal-versions`.

There was a request to make sure that the `exrs` change in question doesn't affect this `image` crate - especially its MSRV.

It however seems unnecessarily complicated to contribute to `Cargo.lock.msrv` or to reproduce the versions used by it in downstream crates. The file that's checked in seems to be a random snapshot from some moment in time - possibly with manual downgrades via text editing or `cargo update --precise ..` calls to downgrade specific (transitive) dependencies - that is not something a consume or `image` can easily cobble up except for trial-and-error downgrading dependencies, or swinging the big `-Zminimal-versions` hammer.

Even if downstream consumer crates decide to use `-Zminimal-versions`, they'll be greeted by a compiler failure because the `image` crate uses `bytemuck::pow_read_unaligned()` which only became available in `bytemuck 1.8.0`: 61ec877d6d4eda4905354f8a7851600f158bd3bd. Hence the big advantage of this is that, besides satisfying the MSRV check, downstream crates can safely use `-Zminimal-versions` and be sure that at least `image` and its dependencies continue to compile.

This approach only seems to have been discussed very briefly in https://github.com/image-rs/image/issues/1433#issuecomment-956449108 and was deemed _"shouldn't be used in production either"_, perhaps because it is a nightly-only feature?

MarijnS95:
Closes #2120


.github/workflows/rust.yml
```diff
@@ -48,9 +48,13 @@ jobs:
       uses: Swatinem/rust-cache@v2
       with:
         cache-on-failure: true
-    - name: Install MSRV Cargo.lock
```
MarijnS95, line 51: Note that YAML indentation is also "wrong" here :sweat_smile:

fintelia (approved):
Pleasantly surprised that things actually work with `-Zminimal-versions` 😄

MarijnS95:
Note that I didn't go as far as checking if certain version bounds could actually _be lower_ while still compiling successfully... Don't think that's particularly interesting :)

johannesvollmer:
Great work!

PR #2124, opened 2024-02-01
Added offsets to GenericImageView to compensate for the now-deprecated bounds call.

### LINKED ISSUE #2118: No method to get xoffset/yoffset from SubImage after deprecating bounds() in #2036
As of 0.25.0, specifically in MR #2036, the `::bounds(&self)` method has been deprecated. But there has been no way to get the x and y offsets out of that view since that deprecation - is there any way that something can be reintroduced alongside the newly minted `::dimentions(&self)` call?

In my case, I use it in a `::change_bounds()` call in order to shift the bounds around by a relative margin.

brykgroup-andrews:
Fixes #2118


brykgroup-andrews:
Not sure why this closed, this is not merged. Was fixing a bad commit message. Paging @fintelia from #2118

PR #2126, opened 2024-02-02
[PATCH] Added offsets to GenericImageView to compensate for the now-deprecated bounds call.

brykgroup-andrews:
Fixes #2036

@fintelia It worked this time lol


fintelia:
I think the `offsets` method makes sense on the `SubImage` struct, but I'm not really sure we should expose it on `GenericImageView`. The difference is that the `GenericImageView` (for immutable access) and `GenericImage` (for mutable access) traits should be implementable for any sort of image. For most of them, the concept of an offset from a different origin doesn't make a ton of sense.

brykgroup-andrews:
I'm down with that, if that's more reasonable. However the reason I did it on GenericImageView is that the old interface also provided it - it allows the method to be used in the same contexts as the previous one. If you think it would be better to move it still in spite of this, I'm happy to make the change.

fintelia:
Please add the method on `SubImage` but not on `GenericImageView`.

brykgroup-andrews:
@fintelia should be good to go!

fintelia (approved):

PR #2130, opened 2024-02-10
Fix avif odd width/height handling

toppa102:

[Related issue](https://github.com/image-rs/image/issues/1647)

This PR fixes the InvalidValue error in dcv_color_primitives when trying to decode odd width/height AVIF files and changes PixelLayout::I422 to todo!() since it's not supported by dcv_color_primitives [(source)](https://docs.rs/dcv-color-primitives/latest/dcv_color_primitives/fn.convert_image.html).

It's a bit hacky but dav1d's output buffer's width/stride seems to always be a multiple of 128 (not too sure about this though) so no need to account for that just increase the width provided to the convert_image  by 1 if it's odd.

And for the height add a padding row of zeros to the source luma buffer.

And at the end copy only the actual width and height parts of the dst buffer while discarding the padding.

### DIFF
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -113,7 +113,7 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for AvifDecoder<R> {
             let pixel_format = match self.picture.pixel_layout() {
                 PixelLayout::I400 => todo!(),
                 PixelLayout::I420 => dcp::PixelFormat::I420,
-                PixelLayout::I422 => dcp::PixelFormat::I422,
+                PixelLayout::I422 => todo!(),
                 PixelLayout::I444 => dcp::PixelFormat::I444,
             };
             let src_format = dcp::ImageFormat {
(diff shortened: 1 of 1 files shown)


fintelia:
Given that the `dcv-color-primitives` doesn't really seem to match our requirements well (based on the number of todo!()'s and copies involved) I wonder if it would be better just to implement the logic ourselves.

Based on that crates documentation, the relevant formula is:
```rust
r = 1.164 * (y - 16) + 1.596 * (cr - 128)
g = 1.164 * (y - 16) - 0.813 * (cr - 128) - 0.392 * (cb - 128)
b = 1.164 * (y - 16) + 2.017 * (cb - 128)
```

toppa102:
> Given that the `dcv-color-primitives` doesn't really seem to match our requirements well (based on the number of todo!()'s and copies involved) I wonder if it would be better just to implement the logic ourselves.

It's an option, I was also thinking about adding the yuv422 -> rgba and yuv400 -> rgba algorithms to `dcv-color-primitives`. I took a look at the crate and they seem to have a lot of the constants and other needed stuff already.

src/codecs/avif/decoder.rs
```diff
@@ -127,29 +127,52 @@ impl<'a, R: 'a + Read> ImageDecoder<'a> for AvifDecoder<R> {
...
             dcp::convert_image(
-                width,
-                height,
+                rounded_width,
+                rounded_height,
                 &src_format,
                 Some(strides),
                 &src_buffers,
                 &dst_format,
                 None,
```
fintelia, line 164: It may be worth seeing whether a non-None value here would enable support for odd sized images

jinx-eob1:
Any updates on this issue? Lots of avif images out there with odds dimensions.

awxkee:
> Given that the `dcv-color-primitives` doesn't really seem to match our requirements well (based on the number of todo!()'s and copies involved) I wonder if it would be better just to implement the logic ourselves.
> 
> Based on that crates documentation, the relevant formula is:
> 
> ```rust
> r = 1.164 * (y - 16) + 1.596 * (cr - 128)
> g = 1.164 * (y - 16) - 0.813 * (cr - 128) - 0.392 * (cb - 128)
> b = 1.164 * (y - 16) + 2.017 * (cb - 128)
> ```

This is correct only for Bt.601 matrix with limited range which is not really common nowadays, since it was replaced by Bt.709 matrix in most cases.
Here you can take conversion for [yuv400](https://github.com/awxkee/yuvutils-rs/blob/master/yuvs/src/yuv400.rs), [yuv420](https://github.com/awxkee/yuvutils-rs/blob/master/yuvs/src/yuv420.rs), [yuv422](https://github.com/awxkee/yuvutils-rs/blob/master/yuvs/src/yuv422.rs) and [yuv444](https://github.com/awxkee/yuvutils-rs/blob/master/yuvs/src/yuv444.rs). Those methods should be in your requirements and supports any image size combinations, and can be easily adopted for 8+ bit-depth or your specific requirements. Here is [defined standard](https://www.itu.int/rec/T-REC-H.273) if you want to check it.

PR #2138, opened 2024-02-12
Remove ImageOutputFormat

### LINKED ISSUE #1481: ImageOutputFormat should be used everywhere or nowhere
Right now, the `ImageOutputFormat` enum is only used by `DynamicImage::write_to`, whereas everywhere else that we encode image files takes an instance of the `ImageFormat` enum. This distinction has led to confusion (see #1461) and is an API inconsistency. I see two possible ways to resolve this:

* Use ImageOutputFormat everywhere. This was attempted by @johannesvollmer who provided a possible patch in https://github.com/image-rs/image/pull/1477#issuecomment-844069443. May also want to implement #1380 at the same time to avoid more API breakage in the future.
    - Pros: Retains and expands the opportunity for per-format encoding options
    - Cons: More API breakage, more complex API 

* Change `DynamicImage::write_to` to take a ImageFormat, and remove ImageOutputFormat entirely.
   - Pros: decreases API complexity
   - Cons: can't specify JPEG quality without working with a JpegEncoder directly.

fintelia:
Fixes #1481

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1482,8 +1482,8 @@ mod test {
     use super::{GrayImage, ImageBuffer, RgbImage};
     use crate::math::Rect;
     use crate::GenericImage as _;
-    use crate::{color, Rgb};
     use crate::ImageFormat;
+    use crate::{color, Rgb};
 
     #[test]
     /// Tests if image buffers from slices work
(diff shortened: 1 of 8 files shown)


Marcono1234:
@fintelia, it seems there are still a few references to `ImageOutputFormat` in the Rustdoc. For example `ImageBuffer::write_to` still mentions it.

PR #2140, opened 2024-02-12
Clarify 'ff' feature name

fintelia:
cc: @kornelski


kornelski:
Is it important to keep the feature flag short?

The comment helps, but cargo and crates.io don't display it. Non-abbreviated googlable name would be better IMHO. 

Farbfeld is one person's niche format, and googling for "ff file format" finds Call Of Duty Fast File and pages mentioning Firefox.

fintelia:
Yeah, it would be nice if cargo, rustdoc, or crates.io could display feature documentation. But for now, this is a default feature so most people shouldn't need to worry about it, and hopefully anyone who does need to know will look at the Cargo.toml

PR #2141, opened 2024-02-12
Switch to using zune-jpeg for JPEG decoding

### LINKED ISSUE #1845: Allow using `zune-jpeg` for JPEG decoding

fintelia:
This rebases @Shnatsel's #1877 onto the current next branch.

Fixes #1845


Shnatsel:
I understand this is going to ship in a release with an API break? 

If so, I can make a PR to clean up the API once this is merged - e.g. drop the extraneous `PhantomData` that's purely there for backwards-compatibility, and also add tests for decoding limits.

Issue #1952, opened 2023-06-26
GenericImage::copy_from can't be used with RGB, only with RGBa


libbkmz:
I'm trying to create DynamicImage from FlatSamples, which I got from another library.

```
use image::GenericImage;

fn main() {
    let w = 640;
    let h = 480;
    let bits_per_pixel = 3;

    let buffer = image::FlatSamples {
        samples: vec![],
        layout: image::flat::SampleLayout::row_major_packed(bits_per_pixel, w, h),
        color_hint: None,
    };

    let view = buffer.as_view::<image::Rgb<u8>>().unwrap();

    let mut dyn_img = image::DynamicImage::new_rgb8(w, h);
    let _ = dyn_img.copy_from(&view, 0, 0);
}
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1d7da94da0974f1eda3f867cf2e468bf)

This is the compilation error output:
```
error[[E0271]](https://doc.rust-lang.org/stable/error_codes/E0271.html): type mismatch resolving `<View<&[u8], Rgb<u8>> as GenericImageView>::Pixel == Rgba<u8>`
    --> src/main.rs:17:31
     |
17   |     let _ = dyn_img.copy_from(&view, 0, 0);
     |                     --------- ^^^^^ expected `Rgba<u8>`, found `Rgb<u8>`
     |                     |
     |                     required by a bound introduced by this call
     |
     = note: expected struct `Rgba<u8>`
                found struct `Rgb<u8>`
```

How can I create DynamicImage without an Alpha channel?

fintelia:
Construct an RgbImage and then make the DynamicImage from that:

```rust
let mut img = image::RgbImage::new(w, h);
img.copy_from(&view, 0, 0).unwrap();
let dyn_img = DynamicImage::ImageRgb8(img);
```

The fact that DynamicImage has a copy_from method at all is a bug IMO, because as you observed it only works for Rgba8 and not any other color type.

PR #2169, opened 2024-03-17
fix rgba8 image jpeg encode bug

yuanyan3060:


fintelia:
Could you also change `JpegEncoder::encode` to reject `La8` and `Rgba8` color types? Since JPEG cannot represent an alpha channel, we shouldn't be silently allowing it.

Sadly, I think we have to allow RGBA in `JpegEncoder::encode_image` because there are probably existing users calling it with `DynamicImage` and that type [implements `GenericImageView<Pixel=Rgba<u8>>`](https://github.com/image-rs/image/issues/1952).

fintelia (approved):

PR #2172, opened 2024-03-18
Improve error messages

Landay7:
Hello, thank you for your project.
I use it to work with ico files and sometimes error messages don't provide me with full information, which can help.
So, I tried to improve it a bit and add additional information.

### DIFF
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -18,12 +18,12 @@ enum DecoderError {
     /// The ICO directory is empty
     NoEntries,
     /// The number of color planes (0 or 1), or the horizontal coordinate of the hotspot for CUR files too big.
-    IcoEntryTooManyPlanesOrHotspot,
+    IcoEntryTooManyPlanesOrHotspot(u16),
     /// The bit depth (may be 0 meaning unspecified), or the vertical coordinate of the hotspot for CUR files too big.
-    IcoEntryTooManyBitsPerPixelOrHotspot,
+    IcoEntryTooManyBitsPerPixelOrHotspot(u16),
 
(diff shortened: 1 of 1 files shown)


src/codecs/ico/decoder.rs
```diff
@@ -45,14 +45,14 @@ impl fmt::Display for DecoderError {
...
-            DecoderError::IcoEntryTooManyPlanesOrHotspot => {
-                f.write_str("ICO image entry has too many color planes or too large hotspot value")
-            }
-            DecoderError::IcoEntryTooManyBitsPerPixelOrHotspot => f.write_str(
-                "ICO image entry has too many bits per pixel or too large hotspot value",
-            ),
-            DecoderError::PngShorterThanHeader => {
-                f.write_str("Entry specified a length that is shorter than PNG header!")
+            DecoderError::IcoEntryTooManyPlanesOrHotspot(num_color_planes) => {
+                f.write_fmt(format_args!("ICO image entry specifies {}, which exceeds the allowed number of color planes or has an invalid hotspot value", num_color_planes))
```
fintelia, line 49: This message is unclear whether the number indicated is how many planes are in the image, or how many hotspots are

Landay7: Thank you.
I tried to improve it. 
Now users at least can know, depending on which file he or she works on.

fintelia:
I'm not very familiar with ICO files. Is there a reason we can't say whether the field is a color planes count or a hotspots count?

Landay7:
We can by using [_type ](https://github.com/image-rs/image/blob/main/src/codecs/ico/decoder.rs#L160)
(It will 1 for ICO or 2 for CUR, other values will be invalid) 
I can try to improve PR checking the type and improve check and error message correspondingly :)

PR #2181, opened 2024-03-25
Only accept .ff as path extension for Farbfeld

foresterre:
Now only accepts .ff as path extension for Farbfeld

This is consistent with ImageFormat::extensions_str, and usage by other image libraries and tools.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -101,7 +101,7 @@ impl ImageFormat {
                 "hdr" => ImageFormat::Hdr,
                 "exr" => ImageFormat::OpenExr,
                 "pbm" | "pam" | "ppm" | "pgm" => ImageFormat::Pnm,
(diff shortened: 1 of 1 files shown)


fintelia:
Is ".farbfeld" actually an accepted file extension for the format? I tried finding references online, but both [this wiki page](http://justsolve.archiveteam.org/wiki/Farbfeld) and the [Ubuntu man pages](https://manpages.ubuntu.com/manpages/focal/man5/farbfeld.5.html) both just mention ".ff"

foresterre:
That's a valid question. I've not seen .farbfeld completely written out as an extension outside this library. Usually, the header or .ff extension is used.

fintelia:
In that case, could you instead change from_extension to only support ".ff"?

fintelia:
Thanks!

PR #2183, opened 2024-03-26
Disable strict mode for JPEG decoder

### LINKED ISSUE #2180: Disable strict mode on JPEG decoder
I would like to be able to decode JPEG images using HDR functionality to embed "multiple images" without errors.

My specific use case for this functionality is decoding images captured from Instagram for resizing.

This is more generally applicable to JPEG decoding. See https://github.com/etemesi254/zune-image/issues/167.

## Draft

The default JPEG decoder settings should not be strict (in this regard).

djc:

Fixes #2180.

Would be great if we can get a release that includes this.

### DIFF
--- a/src/codecs/jpeg/decoder.rs
+++ b/src/codecs/jpeg/decoder.rs
@@ -29,6 +29,7 @@ impl<R: BufRead + Seek> JpegDecoder<R> {
         let mut r = r;
         r.read_to_end(&mut input)?;
         let options = zune_core::options::DecoderOptions::default()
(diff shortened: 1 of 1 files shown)


djc:
@fintelia sorry for harping on this -- do you have any insight on when you plan to cut a release, so we can decide if we should merge a workaround internally for now?

fintelia:
#2187 is open now. Should be released in the next day or two

djc:
Awesome, thanks!

PR #2193, opened 2024-04-05
Improve encode_gray performance in BmpEncoder

mattatz:
Hi `image-rs` team,

This Pull Request focuses on enhancing the performance of the encode_gray function within BmpEncoder.
It optimizes the buffer writing process, thereby improving the overall execution speed of the function.

This modification has significantly improved the encoding speed for large-resolution BMP files.

### DIFF
--- a/src/codecs/bmp/encoder.rs
+++ b/src/codecs/bmp/encoder.rs
@@ -244,10 +244,21 @@ impl<'a, W: Write + 'a> BmpEncoder<'a, W> {
         for row in (0..height).rev() {
             // from the bottom up
             let row_start = row * y_stride;
-            let row_end = row_start + y_stride;
+
             // color value is equal to the palette index
-            self.writer.write_all(&image[row_start as usize..row_end as usize])?;
+            if bytes_per_pixel == 1 {
+                // improve performance by writing the whole row at once
+                let row_end = row_start + y_stride;
+                self.writer
+                    .write_all(&image[row_start as usize..row_end as usize])?;
+            } else {
+                for col in 0..width {
+                    let pixel_start = (row_start + (col * x_stride)) as usize;
+                    self.writer.write_u8(image[pixel_start])?;
+                }
+            }
             // alpha is never written as it's not widely supported
+
             self.write_row_pad(row_pad_size)?;
         }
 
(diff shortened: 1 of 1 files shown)


fintelia:
Could you share any benchmarks results you've measured from these changes?

mattatz:
The former is a benchmark for encoding BMP files before the change, and the latter is the benchmark after the change.
Significant improvements can be seen in the cases affected by the changes, namely zero-L8-file and zero-L8-rawvec.

before
after

src/codecs/bmp/encoder.rs
```diff
@@ -244,12 +244,20 @@ impl<'a, W: Write + 'a> BmpEncoder<'a, W> {
...
             // from the bottom up
             let row_start = row * y_stride;
-            for col in 0..width {
-                let pixel_start = (row_start + (col * x_stride)) as usize;
-                // color value is equal to the palette index
-                self.writer.write_u8(image[pixel_start])?;
-                // alpha is never written as it's not widely supported
+
+            // color value is equal to the palette index
+            if bytes_per_pixel == 1 {
```
fintelia, line 249: Could you make this `x_stride == 1`?

src/codecs/bmp/encoder.rs
```diff
@@ -244,12 +244,20 @@ impl<'a, W: Write + 'a> BmpEncoder<'a, W> {
...
+                let row_end = row_start + y_stride;
+                self.writer
+                    .write_all(&image[row_start as usize..row_end as usize])?;
+            } else {
+                for col in 0..width {
+                    let pixel_start = (row_start + (col * x_stride)) as usize;
+                    self.writer.write_u8(image[pixel_start])?;
+                }
             }
+            // alpha is never written as it's not widely supported
```
fintelia, line 260: Could you move this comment into the `else` branch?

src/codecs/bmp/encoder.rs
```diff
@@ -244,12 +244,20 @@ impl<'a, W: Write + 'a> BmpEncoder<'a, W> {
...
-                // color value is equal to the palette index
-                self.writer.write_u8(image[pixel_start])?;
-                // alpha is never written as it's not widely supported
+
+            // color value is equal to the palette index
+            if bytes_per_pixel == 1 {
+                // improve performance by writing the whole row at once
+                let row_end = row_start + y_stride;
+                self.writer
+                    .write_all(&image[row_start as usize..row_end as usize])?;
```
fintelia, line 253: Mostly comes down to style, but I think this would be slightly nicer as `.write_all(&image[row_start as usize..][..y_stride as usize]`

mattatz:
@fintelia 
Thank you for the review 👍 
I have incorporated all the feedback.

fintelia:
Thanks!

PR #2198, opened 2024-04-08
Upgrade zune-jpeg

fintelia:
Currently targets `0.5.0-rc0`, will switch to the full `0.5.0` release once it comes out

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,8 +53,7 @@ ravif = { version = "0.11.2", default-features = false, optional = true }
 rayon = { version = "1.7.0", optional = true }
 rgb = { version = "0.8.25", optional = true }
 tiff = { version = "0.9.0", optional = true }
-zune-core = { version = "0.4.11", default-features = false, optional = true }
-zune-jpeg = { version = "0.4.11", optional = true }
+zune-jpeg = { version = "0.5.0-rc0", optional = true }
 
 [dev-dependencies]
 crc32fast = "1.2.0"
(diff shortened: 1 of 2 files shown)


Murad-Awad:
You need to use ZCursors rather than Cursor from what I can tell, did some testing and when decoding images with your implementation on an all black image and saw gray pixels randomly come up (they may have been new pixels), to look at an example of how I made it work (probably has some unnecessary code): https://github.com/image-rs/image/compare/main...Murad-Awad:image:upgrade-zune

Murad-Awad:
> You need to use ZCursors rather than Cursor from what I can tell, did some testing and when decoding images with your implementation on an all black image and saw gray pixels randomly come up (they may have been new pixels), to look at an example of how I made it work (probably has some unnecessary code): [main...Murad-Awad:image:upgrade-zune](https://github.com/image-rs/image/compare/main...Murad-Awad:image:upgrade-zune)

I also tested out a version that is functionally the same as yours outside of setting the color + limits params properly (although limits is probably not needed at all given how it is initialized) + using ZCursor: https://github.com/image-rs/image/compare/main...Murad-Awad:image:mawad/keep-testing-zune

fintelia:
@Murad-Awad could you file an issue against the upstream zune-jpeg crate? My understanding is that this version without ZCursor is intended to decode JPEGs properly, so if you are seeing some that get corrupted, they'll want to know

Murad-Awad:
will do!

Murad-Awad:
issue opened here: [v0.5.0-rc0 jpeg decoder does not work when input is a Cursor (rather than a ZCursor)](https://github.com/etemesi254/zune-image/issues/191#top)

Issue #2171, opened 2024-03-18
WebP encoder sometimes creates invalid WebP files 


KenyC:
The following code produces an 4x4 RGBA image filled with blank pixels in both webp and png format. 

```rust
use image::{ImageBuffer, Rgba};

fn main() {

	let width = 4;
	let height = 4;
	let data = vec![0; width * height * 4];

	let image : ImageBuffer<Rgba<u8>, Vec<u8>> = ImageBuffer::from_vec(
		width as u32, height as u32, 
		data
	).unwrap();

	let mut image_file = std::fs::File::create("/tmp/test.webp").unwrap();
	image.write_to(&mut image_file, image::ImageFormat::WebP).unwrap();

	let mut image_file = std::fs::File::create("/tmp/test.png").unwrap();
	image.write_to(&mut image_file, image::ImageFormat::Png).unwrap();
}
```

## Expected

The code runs without panic and produces the desired two files. Both files can be opened by external image viewers.

## Actual behaviour

The code runs without panic and does produce the desired two files. But the webp image cannot be opened by image viewers (tried Gwenview and Firefox). The png on the other hand opens without issues.

## Reproduction steps

See above.

Observations:

 - changing the width to 100 fixes the problem
 - problem persists if all bytes are set to 0xff

Shnatsel:
I've written a fuzzing harness to test for issues like these. It does appear to be fixed, or at least my fuzz harness doesn't turn up any issues.

I'll open a PR with it once I add RGB support (not just RGBA) and more checks.

PR #2201, opened 2024-04-11
Add WebP roundtrip fuzzing target

Shnatsel:
This will detect issues such as #2171, as well as instances of incorrect encoding or decoding.

It is currently limited to RGBA, since I didn't find an easy way to use RGB format conditionally. The decoding always returns RGBA, which makes comparisons of the data a little tricky.


fintelia:
Looks good to me.

Though I'll note that this actually wouldn't have caught the linked issue. That was caused by a bug in the encoder that caused it to write slightly corrupt images that were missing a trailing padding byte at the end of the file. The reason it slipped through was that out decoder isn't that disciplined about ensuring that extra padding bytes are present if there's enough input data to decode all the pixels

PR #2208, opened 2024-04-16
Upgrade zune

Murad-Awad:


fintelia:
Thanks for this PR. We actually [already have one in progress for upgrading](https://github.com/image-rs/image/pull/2198) zune-jpeg. It is just waiting for the full release of zune 0.5

PR #2209, opened 2024-04-17
Upgrade zune

Murad-Awad:
This is similar to [Upgrade zune-jpeg](https://github.com/image-rs/image/pull/2198#top) but makes use of ZCursor, which prevents some corruption, flupping with zune-jpeg so that they can hopefully fix issues with the use of cursor.


etemesi254:
Hi, what corruption?

Murad-Awad:
@etemesi254 I'll reference the specific corruption in a zune-jpeg issue as discussed!

PR #2212, opened 2024-04-23
Added serialization-deserialization and FITS output.

sunipkm:

Here is a summary of changes:
- Images may now contain optional metadata (minor).
- Implemented serialization and deserialization for the `DynamicImage` object for transport over the network, with Zlib compression of the image container (minor).
- Images can now be saved as compressed/uncompressed FITS files if the `fitsio` feature flag is set (breaking).
  - If the `fitsio` feature is enabled, imposes the additional `WriteImage` trait bound on the `Primitive` trait. 
  This trait bound (and the `fitsio` feature) is optional to compile `image` crate on `wasm` targets, since `fitsio` crate requires the `cfitsio` library. 
  - Removes the `Primitive` trait implementations for `usize`, `isize` and `u64`. `fitsio::images::WriteImage` is not implemented for these types, and these pixel types could be deemed unusable since `image` does not serialize these types natively, and any of the image encoders/decoders do not support these types.
- Exposed the 16-bit buffer types (e.g. `Gray16Image`, minor).


fintelia:
Thanks for your interest in `image-rs`! Changes of this scale need to be discussed before we can get to the stage of reviewing a specific PR. 

We also strongly prefer breaking up changes as much as possible because it makes them faster to review and reduces the amount of back-and-forth interaction to get any individual change merged. Each of your bullet point summary items likely warrant separate discussion issue and (if we decided the feature was something that belonged in this crate) its own PR.

sunipkm:
That’s fair, glad to know you are interested. I will break this up into three and submit tomorrow:

1. The “metadata”. 
2. Serialization-deserialization. 
3. FITS IO.

fintelia:
Before working on PRs could you create issues to discuss each of those? They all seem useful but I'm not yet sold that they are all things we want to add to this specific crate

Issue #2215, opened 2024-04-26
Serialization-Deserialization support for `DynamicImage`


sunipkm:
On a client-server type application where the server can attach to a multitude of sources of image data, a `DynamicImage` is the ideal structure to encapsulate the image data. However, since `DynamicImage` does not support serialization-deserialization, sending the data from one process to the other is not trivial and requires conversion to custom, intermediate formats.

I have attached an implementation:
```rust
use crate::{
    DynamicImage, ImageMetadata, Gray16Image, GrayAlpha16Image,
    GrayAlphaImage, GrayImage, Rgb16Image, Rgb32FImage, RgbImage,
    Rgba16Image, Rgba32FImage, RgbaImage,
};
use serde::{
    de::{self, Visitor},
    ser, Deserialize, Deserializer, Serialize,
};
use std::{
    fmt,
    io::{Read, Write},
};
use base64::{engine::general_purpose::STANDARD_NO_PAD, Engine};

#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
enum DynaColor {
    Luma8,
    LumaA8,
    Rgb8,
    Rgba8,
    Luma16,
    LumaA16,
    Rgb16,
    Rgba16,
    Rgb32F,
    Rgba32F,
}

struct SerialBuffer<'a> {
    data: &'a [u8],
    width: u32,
    height: u32,
    color: DynaColor,
    le: bool,
    meta: Option<&'a ImageMetadata>,
}

struct AllocSerialBuffer {
    data: Vec<u8>,
    width: u32,
    height: u32,
    color: DynaColor,
    le: bool,
    meta: Option<ImageMetadata>,
}

impl<'a> Serialize for SerialBuffer<'a> {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        use flate2::write::ZlibEncoder;
        use serde::ser::SerializeStruct;
        let mut state = serializer.serialize_struct("SerialBuffer", 6)?;
        let mut encoder = ZlibEncoder::new(Vec::new(), flate2::Compression::default());
        let res = encoder.write_all(self.data);
        if res.is_err() {
            return Err(ser::Error::custom("Failed to compress data."));
        }
        let compressed_data = encoder.finish();
        if let Ok(compressed_data) = compressed_data {
            let compressed_data = compressed_data.as_slice();
            let compressed_data = STANDARD_NO_PAD.encode(compressed_data);
[...]

PR #2224, opened 2024-05-02
Serialization-Deserialization for DynamicImage using PNG and TIFF

sunipkm:
Addresses #2215.

Changes:
1. RGB32F and RGBA32F support implemented to facilitate serialization of all variants of DynamicImage to standard formats (PNG for integer pixels, TIFF for float32 pixels).
2. DynamicImage objects are encoded to PNG, or, encoded to TIFF and Zlib::Deflate compressed, and base64-encoded for serialization. Inversion is performed for deserialization.


src/dynimage_serde.rs
```diff
@@ -0,0 +1,182 @@
+use crate::{DynamicImage, ImageFormat};
+use base64::{engine::general_purpose::STANDARD_NO_PAD, Engine};
+use serde::{
+    de::{self, Visitor},
+    ser, Deserialize, Deserializer, Serialize,
+};
+use std::{fmt, io::{Cursor, Write}};
+
+impl<'de> Deserialize<'de> for DynamicImage {
```
fintelia, line 9: This feel like over complicating things. We could instead use serde-derive with a proxy struct:
```rust
mod serialize {
    /// Proxy struct for serializing a DynamicImage
    #[derive(Serialize, Deserialize)]
    struct DynamicImage(Vec<u8>);

    impl<'de> Deserialize<'de> for crate::DynamicImage {
        fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<DynamicImage, D::Error> {
            let img = DynamicImage::deserialize(deserializer)?;
            crate::load_from_memory(&img.0).map_err(de::Error::custom)
        }
    }

    ...
}

File: src/dynimage_serde.rs [hunk 2]
```diff
@@ -0,0 +1,182 @@
...
+
+        let mut imgdata = STANDARD_NO_PAD.decode(res.data.as_bytes()).map_err(|e| {
+            de::Error::custom(format!("Failed to decode base64 string: {}", e))
+        })?;
+        let imgfmt = match res.dtype.as_str() {
+            "png" => Ok(ImageFormat::Png),
+            "tiff" => {
+                use flate2::read::ZlibDecoder;
+                use std::io::Read;
+                let mut decoder = ZlibDecoder::new(imgdata.as_slice());
```
fintelia, line 118: It would be nice to use compression in the TIFF case, but I think that should be added as a feature to the tiff crate rather than doing it specially here.

src/dynimage_serde.rs
```diff
@@ -0,0 +1,182 @@
...
+
+                Ok(DynamicSerialImage { dtype, data })
+            }
+        }
+
+        const FIELDS: &[&str] = &["dtype", "data"];
+        let res =
+            deserializer.deserialize_struct("DynamicSerialImage", FIELDS, SerialBufferVisitor)?;
+
+        let mut imgdata = STANDARD_NO_PAD.decode(res.data.as_bytes()).map_err(|e| {
```
fintelia, line 110: I'd rather skip the base64 coding step. The efficient serialization formats like bincode actually get worse sending base64 strings rather than the raw bytes

fintelia (commented):
I left a few comments, but I'd really encourage you to try to break up PRs more. A couple line PR can often be merged with a single glance once the tests pass. A couple dozen line PR might take a round or two of back-and-forth. When you start getting to multiple hundreds of lines long, they can take many rounds and have a high risk of stalling out

src/codecs/tiff.rs
```diff
@@ -310,6 +339,14 @@ fn u8_slice_as_u16(buf: &[u8]) -> ImageResult<&[u16]> {
     })
 }
 
+fn u8_slice_as_f32(buf: &[u8]) -> ImageResult<&[f32]> {
+    bytemuck::try_cast_slice(buf).map_err(|err| {
+        ImageError::Parameter(ParameterError::from_kind(ParameterErrorKind::Generic(
```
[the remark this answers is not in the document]
sunipkm: 4-byte alignment is required only for RGB32 images (which are being added in), does that break the encoder interface?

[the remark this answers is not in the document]
fintelia: Yes. Other parts of the crate allow 1-byte aligned buffers for storing 16-bit integers or 32-bit floats. For instance, the HDR decoder which decodes f32's into a byte slice.

The reason we allow it is that you can't easily create an over-aligned `Vec<u8>` and we don't want to _force_ users to store image data in typed slices before passing them to encoders/decoders. And it would be too easy of a gotcha to overlook when using the API

197g:
Since there's no recent development and we didn't quite reach a design consensus, what format to use, why not a simpler adapter, I'm closing this. Feel free to reopen it with a serde design that addresses the concerns.

PR #2228, opened 2024-05-04
webp frames iterator is now ending iteration if no more frames are decodable

### LINKED ISSUE #2226: WebPDecoder.into_frames() always returns an error at the end of the file instead of ending iteration normally
Broken in 0.25.x and git head, unsure about when it started since I was hoping to replace the webp-animation crate with this, but it's probably been unused and broken for a while.

## Expected

Frames iterator returns None after the last frame.

## Actual behaviour

Returns either a `DecodingError { format: Exact(WebP), underlying: Some(NoMoreFrames) }` or an IO Error from a BufReader as it tries to read past the end of the file.

## Reproduction steps

Call `into_frames` on an WebPDecoder for an animated image. Here's a test that runs on your own test data.

```
#[test]
fn webp_decode() {
    let decoder = WebPDecoder::new(std::io::Cursor::new(
        std::fs::read("tests/images/webp/extended_images/anim.webp").unwrap(),
    ))
    .unwrap();
    let frames = decoder.into_frames();
    for result in frames {
        result.unwrap();
    }
}
```

```
thread 'webp_decode' panicked at tests/limits_anim.rs:20:16:
called `Result::unwrap()` on an `Err` value: Decoding(DecodingError { format: Exact(WebP), underlying: Some(NoMoreFrames) })
```

EmiOnGit:
Fixes #2226 

The iterator returns `None` as soon as a `DecodingError::NoMoreFrames` is found.

## Debatable
The current implementation returns every `Frame` only once regardless of the specified `LoopCount`. I believe this is what most people would expect, especially since `LoopCount::Forever` is the most used variant.

It would be a bit more performant to return the iterator at the start since the allocation of one `RgbaImage`/`RgbImage` can be saved. 
(Thank you for this awesome project btw :))

### DIFF
--- a/src/codecs/webp/decoder.rs
+++ b/src/codecs/webp/decoder.rs
@@ -79,12 +79,14 @@ impl<'a, R: 'a + Read + Seek> AnimationDecoder<'a> for WebPDecoder<R> {
                     let mut img = RgbaImage::new(width, height);
                     match self.decoder.inner.read_frame(&mut img) {
                         Ok(delay) => (img, delay),
+                        Err(image_webp::DecodingError::NoMoreFrames) => return None,
                         Err(e) => return Some(Err(ImageError::from_webp_decode(e))),
                     }
                 } else {
                     let mut img = RgbImage::new(width, height);
                     match self.decoder.inner.read_frame(&mut img) {
                         Ok(delay) => (img.convert(), delay),
+                        Err(image_webp::DecodingError::NoMoreFrames) => return None,
                         Err(e) => return Some(Err(ImageError::from_webp_decode(e))),
                     }
                 };
--- a/src/codecs/webp/decoder.rs
+++ b/src/codecs/webp/decoder.rs
@@ -68,25 +68,28 @@ impl<'a, R: 'a + Read + Seek> AnimationDecoder<'a> for WebPDecoder<R> {
     fn into_frames(self) -> Frames<'a> {
         struct FramesInner<R: Read + Seek> {
             decoder: WebPDecoder<R>,
+            current: u32,
         }
         impl<R: Read + Seek> Iterator for FramesInner<R> {
             type Item = ImageResult<Frame>;
 
             fn next(&mut self) -> Option<Self::Item> {
+                if self.current == self.decoder.inner.num_frames() {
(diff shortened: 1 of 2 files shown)


HeroicKatora (commented):
There are probably existing image applicable as a test here, e.g. [this one](https://github.com/image-rs/image/blob/main/tests/images/webp/extended_images/anim.webp).

Would you mind expanding the regression test suite for this behavior?

EmiOnGit:
## Changes to the iterator
After benchmarking the iterator on `DecodingError::NoMoreFrames` again returning early I found that the early return is constantly around 8-12 percent faster on a webp animation with 6 frames (tested on two machines and 10_000 iterations). Therefore I changed the implementation to check at the start using a internal counter.

## Testing
I also implemented a simple test which is checking against the internal framecount of the `image_webp::WebPDecoder::num_frames()`. The test is run against the 3 files in `webp/extended_images` which I believe to be  a good representation of the expected behaviour.

[advertises_rgba_but_frames_are_rgb.webp](https://github.com/image-rs/image/blob/main/tests/images/webp/extended_images/advertises_rgba_but_frames_are_rgb.webp): returns the first two frames and the rest as a EOF decoding error. The iterator still exists after the specified frame-count of 12 .
[anim.webp](https://github.com/image-rs/image/blob/main/tests/images/webp/extended_images/anim.webp) returns all frames.
[lossy_alpha.webp](https://github.com/image-rs/image/blob/main/tests/images/webp/extended_images/lossy_alpha.webp): (not animated webp) returns one frame.

Let me know if something else should be implemented let me now. We could also check the first returned frame against the image returned by `image::open`, which would also be expected to be for all inputs. I wasn't sure if more tests are welcomed.

(Also I'm not sure about the clippy check failing. It also fails on my computer on the main branch so I assume it's not my fault)

HeroicKatora (approved):
LGTM

HeroicKatora:
> (Also I'm not sure about the clippy check failing. It also fails on my computer on the main branch so I assume it's not my fault)

clippy is linting that there are some cfg-options that do not correspond to actual features we have in Cargo.toml. Huh, weird, but definitely not your fault.

PR #2236, opened 2024-05-16
Override `clone_from` implementations for a few types

a1phyr:

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -44,7 +44,7 @@ use crate::{Rgb32FImage, Rgba32FImage};
 /// would hardly be feasible as a simple enum, due to the sheer number of combinations of channel
 /// kinds, channel order, and bit depth. Rather, this type provides an opinionated selection with
 /// normalized channel order which can store common pixel values without loss.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Debug, PartialEq)]
 #[non_exhaustive]
 pub enum DynamicImage {
     /// Each pixel in this image is 8-bit Luma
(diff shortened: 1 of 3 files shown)


kornelski (approved):

fintelia:
How common is manually implementing `clone_from`? I don't think I've seen it done in other crates

HeroicKatora (approved):
It's odd that `derive()` does not do it, but it obviously has the potential to save an allocation. Not quite as intricate as [std::vec::Vec](https://doc.rust-lang.org/stable/src/alloc/vec/mod.rs.html#2778) on the dynamic image but effective.

fintelia:
This would avoid the allocation but potentially "leak" memory if the image started out substantially larger than the new size. (Technically not a leak because the memory would still remain referenced but inaccessible.) Though perhaps that's well known behavior and folks using the method would expect that?

kornelski:
I think it's fine that `clone_from` keeps potentially large allocations. It's opt-in, and its purpose is to reuse allocations.

src/dynimage.rs
```diff
@@ -78,6 +78,39 @@ pub enum DynamicImage {
...
+            Self::ImageLumaA8(p) => Self::ImageLumaA8(p.clone()),
+            Self::ImageRgb8(p) => Self::ImageRgb8(p.clone()),
+            Self::ImageRgba8(p) => Self::ImageRgba8(p.clone()),
+            Self::ImageLuma16(p) => Self::ImageLuma16(p.clone()),
+            Self::ImageLumaA16(p) => Self::ImageLumaA16(p.clone()),
+            Self::ImageRgb16(p) => Self::ImageRgb16(p.clone()),
+            Self::ImageRgba16(p) => Self::ImageRgba16(p.clone()),
+            Self::ImageRgb32F(p) => Self::ImageRgb32F(p.clone()),
+            Self::ImageRgba32F(p) => Self::ImageRgba32F(p.clone()),
+        }
```
fintelia, line 94: Would be good to use the `dynamic_map!` macro here.

src/dynimage.rs
```diff
@@ -78,6 +78,39 @@ pub enum DynamicImage {
...
+            (Self::ImageLumaA8(p1), Self::ImageLumaA8(p2)) => p1.clone_from(p2),
+            (Self::ImageRgb8(p1), Self::ImageRgb8(p2)) => p1.clone_from(p2),
+            (Self::ImageRgba8(p1), Self::ImageRgba8(p2)) => p1.clone_from(p2),
+            (Self::ImageLuma16(p1), Self::ImageLuma16(p2)) => p1.clone_from(p2),
+            (Self::ImageLumaA16(p1), Self::ImageLumaA16(p2)) => p1.clone_from(p2),
+            (Self::ImageRgb16(p1), Self::ImageRgb16(p2)) => p1.clone_from(p2),
+            (Self::ImageRgba16(p1), Self::ImageRgba16(p2)) => p1.clone_from(p2),
+            (Self::ImageRgb32F(p1), Self::ImageRgb32F(p2)) => p1.clone_from(p2),
+            (Self::ImageRgba32F(p1), Self::ImageRgba32F(p2)) => p1.clone_from(p2),
+            (this, source) => *this = source.clone(),
```
fintelia, line 109: This match arm should be avoided. The enum is marked non-exhaustive, but within this crate we can still match exhaustively on it

fintelia (approved):

src/dynimage.rs
a1phyr: I am not sure of what you want here. Something like this ?

```rust
match (self, source) {
    (Self::ImageRgb8(p1), Self::ImageRgb8(p2)) => p1.clone_from(p2),
    (Self::ImageRgba8(p1), Self::ImageRgba8(p2)) => p1.clone_from(p2),
    // other formats...

    (this, Self::ImageRgb8(p)) => *this = Self::ImageRgb8(p.clone()),
    (this, Self::ImageRgba8(p)) => this = Self::ImageRgba8(p.clone()),
    // other formats...
}
```

fintelia: Sorry, I didn't read closely. It's fine as is

PR #2239, opened 2024-05-20
Add missing map2/apply2 functions to `Pixel` trait

jnickg:
_New contributor license agreement_: I license past and future contributions under the dual MIT/Apache-2.0 license,

This PR merges quick changes to add a few missing functions to `Pixel` so that basic image math operations are easier  when dealing with alpha-channel images.

## Commit Messages
* Account for alpha channel to simplify some image math operations where naively operating on all pixel components won't work (such as summing images, where usually picking one alpha value from `self` or `other` makes more sense)


HeroicKatora (commented):
The `{apply,map}2_with_alpha` are well-defined extensions and were sketched out already. Overall whether or not they can be accomplished with the existing functions or not wouldn't matter too much here, it's a pretty small change whose minor breakage can be easily part of the next major version. (In theory there could be an approximate default impl based on `as_channels`  and color model but I don't think the risk of modelling error is worth it to mitigate that inconvenience).

I think the opposite applies to `*_without_alpha`. Those can be written by using a specific parameter choice for `g` from the `with_alpha` version: `|lhs, _| lhs`. We don't have the equivalent method for the single pixel function either. Here it might be best to provide them as defaults on the trait instead.

jnickg:
Hi @HeroicKatora , thanks for the review. I might be misunderstanding something though, can you please clarify? I see [these](https://github.com/image-rs/image/pull/2239/files#diff-172747fc43e38d33de743b2d5f564ad6440570670af7aa912235399448a55c7cR329-R345) functions which look like the equivalent single-pixel functions, which is why I added them. I'm happy to elide `*2_without_alpha` functions though (in fact I only added them for parity, and also thought they could just use the closure you mentioned).

Just to clarify, it sounds like you're looking for a code change for the `*2_without_alpha` functions so that they are defaults on the trait. Is that accurate? Would you like the `*2_with_alpha` functions implemented that way as well?

Thanks!

HeroicKatora (approved):
LGTM, it fits the design. Should be considered for the next breaking version change (0.26).

jnickg:
Just to close the loop, I rebased onto the latest `main` and applied the `cargo fmt` changes to make CI happy. Hopefully that is enough to get it merge-able

jnickg:
@HeroicKatora looks like everything is green, but I don't have a "merge" button and can't push a merge commit. Please let me know if there's any further action I should take.

HeroicKatora:
This is still a break change so in light of the issue with packaging, a release of `0.25.2` before this is merged to the main branch for `0.26` then. But should be good to go, nothing more to do from your side I think.

jnickg:
Thank you!!!

ripytide:
I think we should wait to merge this until the the conflict with #2259 has been discussed.

PR #2243, opened 2024-05-26
Rename `Reader`  to `ImageReader`

ripytide:

Following the discussion from #1327

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,11 @@ All image processing functions provided operate on types that implement the `Gen
 
 ## High level API
 
-Load images using [`io::Reader`]:
+Load images using [`ImageReader`]:
 
 ```rust,ignore
 use std::io::Cursor;
-use image::io::Reader as ImageReader;
+use image::ImageReader;
 
 let img = ImageReader::open("myimage.png")?.decode()?;
 let img2 = ImageReader::new(Cursor::new(bytes)).with_guessed_format()?.decode()?;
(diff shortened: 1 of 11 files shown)


fintelia:
Created https://github.com/image-rs/image/issues/2245 to track issues like this that'll need a breaking release

kornelski:
I think it'd be helpful to create a pub use under old name with a deprecated attribute. It makes migration less painful.

fintelia:
I'm not sure you can place a deprecated attribute on a re-export.

But in either case, in the future @kornelski let's avoid merging breaking changes directly to the main branch. I'd prefer to keep the main branch in a state where it is always possible to make a patch release if needed. We can discuss major release strategy more in the linked issue, but I wasn't planning to start it for another 6+ months.

kornelski:
Deprecated attribute works fine on re-exports and correctly fires only when the reexported name is used.

fintelia:
@ripytide could you open a PR with the necessary re-export?

I'd also like to give @HeroicKatora a chance to weigh in. They had some concerns with the renaming before, and while they've been less active recently they _are_ still a maintainer.

HeroicKatora:
The previous design is to large parts modelled around the standard library, that is small modules and avoid prefix duplication (the library is `image`, it shouldn't preface everything with another `Image*`). That decision was before experience of `wgpu` has demonstrated, to me but also in general, that the inverse is also possible and a large top level is still a workable library usage experience. I'd *prefer* to keep the `io` module itself for uncommon types related specifically to it but I don't have any deeper concerns about the renaming.

PR #2251, opened 2024-06-02
Added compression and f32 support for TIFF encoding/decoding.

sunipkm:

f32 encoding/decoding support is added without additional byte alignment requirements. Removed 2-byte alignment requirement for u16 encoding/decoding. Added TIFF compression support.


tests/images/tiff/testsuite/rgb32f_bw.tiff
fintelia, line 1: This is a pretty big image to check into the repository. Could you crop it / make a smaller one?

fintelia:
CI seems to be failing because this exposes the tiff crate's `Compressor` type

tests/images/tiff/testsuite/rgb32f_bw.tiff
sunipkm: Do you want me to break it into F32 stuff and compression?

sunipkm:
I somehow missed this, just pushed changes that hide tiff crate's `Compressor` type.

sunipkm:
@fintelia can you please take a look at this?

fintelia:
I've been taking a hiatus from most PR review and issue triage. Another reviewer might have a chance to review this at some point

sunipkm:
@fintelia I hope you feel better soon!

src/codecs/tiff.rs
```diff
@@ -282,6 +312,61 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
...
+
+impl From<TiffDeflateLevel> for DeflateLevel {
+    fn from(val: TiffDeflateLevel) -> Self {
+        match val {
+            TiffDeflateLevel::Fast => DeflateLevel::Fast,
+            TiffDeflateLevel::Balanced => DeflateLevel::Balanced,
+            TiffDeflateLevel::Best => DeflateLevel::Best,
+        }
+    }
+}
```
HeroicKatora, line 359: This introduces a public dependency on the `tiff` major version. A standard method would be sufficient and does not have that effect.

src/codecs/tiff.rs
```diff
@@ -296,24 +381,91 @@ fn cmyk_to_rgb(cmyk: &[u8]) -> [u8; 3] {
...
-        // `Error` and `Display` trait.
-        // See <https://github.com/Lokathor/bytemuck/issues/22>.
-        ImageError::Parameter(ParameterError::from_kind(ParameterErrorKind::Generic(
-            format!("{:?}", err),
-        )))
-    })
+enum DtypeContainer<'a, T> {
+    Slice(&'a [T]),
+    Vec(Vec<T>),
+}
```
HeroicKatora, line 387: This looks much like `std::borrow::Cow`. What's the difference for not using that standard type?

src/codecs/tiff.rs
```diff
@@ -296,24 +381,91 @@ fn cmyk_to_rgb(cmyk: &[u8]) -> [u8; 3] {
...
+                _ => {
+                    // If the buffer is not the correct length for a f32 slice, err.
+                    Err(ImageError::Parameter(ParameterError::from_kind(
+                        ParameterErrorKind::Generic(format!("{:?}", err)),
+                    )))
+                }
+            }
+        }
+    }
+}
```
HeroicKatora, line 422: I'm surprised this is not a generic method bounded by `Pod`. In particular some of the verbosity around `from_ne_bytes` is simulating an unaligned read, a [method bytemuck provides](https://docs.rs/bytemuck/latest/bytemuck/fn.pod_read_unaligned.html) for the relevant sample types.

HeroicKatora (changes requested):
The bindings themselves look quite alright, I think the lower-level code could be cleaned up quite a lot with help of `bytemuck` and the standard library. Quite boilerplatey.

tests/images/tiff/testsuite/rgb32f_bw.tiff
fintelia: No, just make it like a 10x10 image rather than a 400x400 one

197g: I reduced their size to 32x32 respectively, removed the color profile, exif, and XMP metadata which aren't testing much regarding our decoder.

src/codecs/tiff.rs
```diff
@@ -338,6 +364,61 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
 /// Encoder for tiff images
 pub struct TiffEncoder<W> {
     w: W,
+    comp: Compression,
+}
+
+/// Compression types supported by the TIFF format
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum CompressionType {
```
fintelia, line 373: This should be marked non-exhaustive

src/codecs/tiff.rs
```diff
@@ -338,6 +364,61 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
...
+    Lzw,
+    /// Deflate compression
+    Deflate(TiffDeflateLevel),
+    /// Bit packing compression
+    Packbits,
+}
+
+impl Default for CompressionType {
+    fn default() -> Self {
+        CompressionType::Deflate(Default::default())
```
fintelia: As far as I know, LZW is more common for TIFF files so that should probably be the default.

tests/reference_images.rs
```diff
@@ -39,6 +39,72 @@ where
...
+
+        let mut out_path = base.clone();
+        out_path.push(OUTPUT_DIR);
+        out_path.push("compressed_tiff");
+        out_path.push("deflate");
+        std::fs::create_dir_all(&out_path).unwrap();
+        out_path.push(path.file_name().unwrap());
+        let encoder = TiffEncoder::new_with_compression(
+            fs::File::create(&out_path).unwrap(),
+            CompressionType::Deflate(TiffDeflateLevel::Balanced),
```
fintelia, line 63: How long does this test take? I could imagine that compressing all the input files with balanced compression could take a while, though I don't know how many / how large they are.

src/codecs/tiff.rs
```diff
@@ -352,24 +433,53 @@ fn cmyk_to_rgb(cmyk: &[u8]) -> [u8; 3] {
...
+        TiffEncoder { w, comp }
+    }
+
+    /// Create a new encoder that writes its output to `w` with `CompressionType` `compression`.
+    ///
+    /// It is best to view the options as a _hint_ to the implementation on the smallest or fastest
+    /// option for encoding a particular image. That is, using options that map directly to a TIFF
+    /// image parameter will use this parameter where possible. But variants that have no direct
+    /// mapping may be interpreted differently in minor versions. The exact output is expressly
+    /// __not__ part of the SemVer stability guarantee.
```
fintelia: This seems like a more natural note to put on `CompressionType::Deflate` to explain that the compression levels don't have well defined meaning. Getting the requested compression algorithm seems like something that should be guaranteed.

src/codecs/tiff.rs
```diff
@@ -127,20 +130,47 @@ where
...
+                        "Unhandled TIFF color type {color_type:?}",
+                    )),
+                ),
+            ))
+        }
+    };
+
+    match SampleFormat::from_u16(sample_format) {
+        Some(SampleFormat::Uint) if num_bits <= 16 => Ok(()),
+        Some(SampleFormat::IEEEFP) if num_bits == 32 => Ok(()),
```
fintelia: I don't love the implicitness of "num_bits <= 16 means integer" and "num_bits == 32 means float". But really this should be addressed by having the `tiff` crate expose a way to directly get the sample format, so I'm OK with doing it this way for now.

src/codecs/tiff.rs
```diff
@@ -387,40 +497,48 @@ impl<W: Write + Seek> TiffEncoder<W> {
...
-
         let mut encoder =
             tiff::encoder::TiffEncoder::new(self.w).map_err(ImageError::from_tiff_encode)?;
+
+        match self.comp {
+            Compression::Uncompressed => {}
+            Compression::Lzw | Compression::Deflate(_) | Compression::Packbits => {
+                encoder = encoder.with_compression(self.comp);
+            }
+        }
```
fintelia:
```suggestion
        if self.comp != Compression::Uncompressed {
            encoder = encoder.with_compression(self.comp);
        }
```

fintelia (commented):
This ideally should have been two PRs for f32 support and compression respectively

src/codecs/tiff.rs
197g: It's not implicit, the SampleFormat is itself parsed from a tiff tag. They just happen to not overlap, too. We do not have fp16 support in `tiff` (due to predictor math) and we do not support loading u32/i32 channels into `image`. But the match here would be unique even without the conditions. Nonetheless, the `tiff` API should return all that information in one struct instead of strewn over `ColorType` and manual tag fetching so this will change.

197g:
> This ideally should have been two PRs for f32 support and compression respectively

Alright, I'll split it up. It also seems to me like f32 support was simpler than the compression part.

src/codecs/tiff.rs
fintelia: Implicit is probably the wrong word. The concern is that the `tiff::ColorType::RGB(32) => ColorType::Rgb32F` line above only works because of the checks that are happening here. The result is you have to read the implementation of `check_sample_format` to understand what is being checked, and only afterwards can you understand why `new` can match on color type without checking the sample format.

If `new` had access to both the sample format and sample type, the logic could be more straightforward:

```rust
match sample_format {
    SampleFormat::Uint => match color_type {
        RGB(8) => ColorType::Rgb8,
        ...
    },
    SampleFormat::IEEEFP => match color_type {
        RGB(32) => ColorType::Rgb32F
        ...
    }
}
```

tests/reference_images.rs
197g: `0.35s` and `0.53s` respectively. Without the IO for compressed output it is `0.33` and `0.45` so I'll remove the file write. Seems fine although may we want to restrict it to a particular set of files with enough coverage at some point.

src/codecs/tiff.rs
197g: Defaulting LZW would seem to be more a historical accident. If we're going with libtiff as upstream we might want to _also_ look into zstd support: http://libtiff.maptools.org/v4.0.10.html And I think its default option is no compression?

> Please note that COMPRESSION_ZSTD is self-assigned the id 50000
    by the libtiff project and is not officially registered with Adobe
    since Adobe's registration function is defunct.

Yikes.

We should be making our own comparison in terms of size / speed. Based on the compression test this might favor deflate if `Balanced` achieves comparable size to LZW. Which would not be all that surprising, LZW is the oldest of the bunch.

maxall41:
I think Iv'e found an issue with either this PR or #2534. I started by forking the image-rs main branch and then I used the following commands to merge this PR into the branch `x` on my fork:
```
git remote add sunipkm https://github.com/sunipkm/image-rs.git
git fetch origin
git fetch sunipkm
git checkout origin/main
git checkout -b merge-tiff-compress
git merge sunipkm/tiff-compress
git push origin merge-tiff-compress
```

I then setup my application to use the merge-tiff-compress branch on my fork with the following:
```
image = { git = "https://github.com/maxall41/image.git", branch = "merge-tiff-compress" }
```

With this setup **I am able to successfully load all of the tiff images in the test suite added in PR #2534**. But with the attached image I get the following error while attempting to decode:
```
The encoder or decoder for Tiff does not support the color type `Unknown(32)`
```

Tiff info:
```
❯ tiffinfo reproduce.tif
TIFFReadDirectory: Warning, Unknown field with tag 50838 (0xc696) encountered.
TIFFReadDirectory: Warning, Unknown field with tag 50839 (0xc697) encountered.
=== TIFF directory 0 ===
TIFF Directory at offset 0x8 (8)
  Subfile Type: (0 = 0x0)
  Image Width: 424 Image Length: 546
  Bits/Sample: 32
  Sample Format: IEEE floating point
  Compression Scheme: None
  Photometric Interpretation: min-is-black
  Samples/Pixel: 1
  Rows/Strip: 546
  Planar Configuration: single image plane
  ImageDescription: ImageJ=1.54p
min=0.0
max=1.0

  Tag 50838: 12,22,14,20,8
  Tag 50839: 73,74,73,74,112,114,111,112,0,0,0,4,0,83,0,108,0,105,0,99,0,101,0,95,0,76,0,97,0,98,0,101,0,108,0,99,0,108,0,97,0,115,0,115,0,32,0,49,0,85,0,110,0,105,0,113,0,117,0,101,0,78,0,97,0,109,0,101,0,116,0,114,0,117,0,101
```

Image to reproduce issue: https://drive.google.com/file/d/1qcnFq-LCde0SFxxuKNgCSnr4sq2OnpuS/view?usp=sharing

**I am able to load this image in ImageJ2/Fiji and Apple Preview.**

fintelia:
@maxall41 This PR adds RGB and RGBA 32-bit floating point support, but notably doesn't add support for single-channel floating point images like the one you attached.

Though it would probably be good to adjust the error message...

maxall41:
@fintelia Ah that explains it. I might write a PR to add support...

maxall41:
Just made a PR: #2536

PR #2252, opened 2024-06-02
Add deprecated re-export of ImageReader as Reader

ripytide:
As requested in #2243, some module renaming was required to ensure the current `io` can become private while still re-exporting the deprecated `Reader` type synonym from a top-level module called `io`.


fintelia:
Looking more closely, `io::Limits` and `io::LimitSupport` will also need to be re-exported to avoid it being a breaking change

ripytide:
I've added re-exports for them.

fintelia:
Thanks!

Issue #1499, opened 2021-06-25
A compatible option for decoding to various pixel types


HeroicKatora:
> The only point of these helper methods is to allocate a buffer, such that I don't have to repeat the boilerplate allocation code twice in every test. 
> The Rgb<f32>::SubPixel type will not resolve to an f32 color type. The Pixel trait is only really implemented for Rgb8 and Rgb16, not f32

This helper method as sketched would then be better suited for `impl ImageBuffer`. It is similar to [`DynamicImage::from_decoder`](https://docs.rs/image/0.23.14/image/enum.DynamicImage.html#method.from_decoder). I'm afraid the `Pixel` issue is something we can't easily resolve in this major version of `image`. We could, however, already add an improved trait that is _just_ used for these method. Something like:

```
#[repr(u8)]
enum SampleType {
    U8 = 0,
    U16 = 1,
…
    F32 = 2,
}

trait Sample: Primitive {
    const TYPE: SampleType;
}

mod hidden {
    pub trait PixelV2 { …  }
    impl<T: Sample> PixelV2 for Rgb<T> { … }
}

impl<Px> ImageBuffer<Px, Vec<Px::Subpixel>>
where
    Px: hidden::PixelV2,
{
    fn from_decoder(mut decoder: impl ImageDecoder) -> ImageResult<Self> {
        if Px::COLOR_TYPE != decoder.color_type() { … }
    }
}
```

In any case, I would much rather discuss these helpers in a separate PR _after_ the core EXR functionality has been added together with the new `ColorType` variant :)

_Originally posted by @HeroicKatora in https://github.com/image-rs/image/pull/1475#discussion_r658560179_

fintelia:
The problem with the current Pixel trait is how COLOR_TYPE is implemented. After stripping out the macros, the code looks like:
```rust
impl<T: Primitive + 'static> Pixel for Rgba<T> {
    const COLOR_TYPE: ColorType =
        [ColorType::Rgba8, ColorType::Rgba16][(std::mem::size_of::<T>() > 1) as usize];
}
``` 

At one point I looked into using [`TypeId::of<T>()`](https://doc.rust-lang.org/std/any/struct.TypeId.html#method.of) instead of just matching on the size, but sadly using that function from const contexts is still unstable ([and there doesn't seem to be a ton of motion on changing that](https://github.com/rust-lang/rust/issues/77125)).

johannesvollmer:
Wouldn't it be possible to have one macro call for each color type? Like:

[...]

Issue #1464, opened 2021-04-14
Pixel::map methods should map to arbitrary subpixel type


alexkirsz:
Right now, `Pixel::map` methods map from `I: Pixel` to `I`. These functions would be more powerful if they could map to a `J: Pixel`, where there is no constraint on the type of the projection's subpixel. In essence, I'd like to be able to map from a `Rgb<u8>` to a `Rgb<f32>` generically.

My use case is converting from 8-bit sRGB to f32 linear RGB, resizing in the linear RGB space, then converting back to 8-bit sRGB. See [this article](http://www.ericbrasseur.org/gamma.html?i=1) for an explanation of the *why*.

fintelia:
Seems reasonable to me. As a workaround for now, you should just be able to iterate over the pixels with a for loop.

anforowicz:
Target-subpixel-generic map seems to be relatively straightforward to implement for specific pixel types like `Rgb<T>`, but feels quite tricky for the generic `Pixel` trait.

For same-subpixel-type map, the `Pixel` trait knows that the return type of `map` is `Self` (this is the existing code in traits.rs):

    pub trait Pixel ... {
        ...
        fn map<F>(&self, f: F) -> Self
        where
            F: FnMut(Self::Subpixel) -> Self::Subpixel;

For target-subpixel-generic map, the `Rgb<T>` pixel knows that the return type of `map` is `Rgb<U>` (this is a hypothetical tweak of the code in color.rs):

    impl<T: Primitive + 'static> Pixel for $ident<T> {
        ...
        fn map<F, U>(& self, f: F) -> $ident<U>  // Note the `U` here ...
            where F: FnMut(T) -> U,  // ... and the `U` here.
                       U : Primitive ... {  
            ...
        }

But there is no way in *stable* Rust to spell out the type of target-subpixel-generic map in the `Pixel` trait (this is a hypothetical tweak of the code in traits.rs):

    pub trait Pixel ... {
        ...
        // For Rgb<u8>::Mapped<u16> returns the associated type of Rgb<u16>
        // For Bgr<u8>::Mapped<u16> returns the associated type of Bgr<u16>
        //
        // Note that the line below won't compile in stable Rust - it requires
        // Generic associated types (GATs) feature.
        type Mapped<U>: Pixel;  
        ...
        // If `map` returned a generic `Pixel`, then
        // `pixel.map(identityFunction)` wouldn't round-trip to the same type...
[...]

Issue #1212, opened 2020-05-05
Why is the FromColor trait not public?


Drevoed:
How does one create a type constraint on a generic function that allows an image to be converted into the pixel type that the caller wants to get?

Overall, I'm trying to achieve creating a fully white RGBA image and then return it with the the type that is needed elsewhere. So far I tried this:
```rust
use imageproc::drawing::draw_filled_rect_mut;
use image::buffer::ConvertBuffer;
use image::{GenericImage, ImageBuffer, Pixel, RgbaImage, Rgba, GenericImageView};
use imageproc::rect::Rect;

pub(crate) fn create_square_image<I>(source_img: &I, size: u32) -> ImageBuffer<I::Pixel, Vec<u8>>
where
    I: ConvertBuffer<ImageBuffer<<I as GenericImageView>::Pixel, Vec<u8>>> + GenericImage,
    <I as GenericImageView>::Pixel: Pixel<Subpixel = u8>
{
    let (width, height) = source_img.dimensions();

    let mut block_image = create_white_image(size, size);

    let delta_x = (size - width) / 2;
    let delta_y = (size - height) / 2;

    for y in 0..height {
        let target_y = y + delta_y;
        if target_y >= size {
            continue;
        }
        for x in 0..width {
            let target_x = x + delta_x;
            if target_x >= size {
                continue;
            }
            let pixel = source_img.get_pixel(x, y);
            block_image.put_pixel(x, y, pixel)
        }
    }

    block_image
}

pub(crate) fn create_white_image<I>(width: u32, height: u32) -> ImageBuffer<I::Pixel, Vec<u8>>
where
    I: ConvertBuffer<ImageBuffer<<I as GenericImageView>::Pixel, Vec<u8>>> + GenericImage,
    <I as GenericImageView>::Pixel: Pixel<Subpixel = u8>
{
    let mut image = ImageBuffer::new(width, height);
    draw_filled_rect_mut(&mut image, Rect::at(0, 0).of_size(width, height), Rgba([255u8, 255, 255, 255]));

    image.convert()
}
```

but it doesn't work because P doesn't have a type constraint `FromColor<Rgba<u8>>`:
```rust
error[E0277]: the trait bound `<I as image::image::GenericImageView>::Pixel: image::color::FromColor<image::color::Rgba<u8>>` is not satisfied
  --> src\main.rs:44:11
   |
39 |     <I as GenericImageView>::Pixel: Pixel<Subpixel = u8>
[...]

PR #2255, opened 2024-06-04
Revamp the `Pixel` trait via the `pixeli` crate

### LINKED ISSUE #1523: Let's talk about `Pixel`
Hey! I've been trying to implement f32 support for a while now, but progress always stops when I try to modify `Pixel` in a backwards-compatible and sane way. While we could probably think of some way to add f32 support in a non-breaking way, I see no possibility that doesn't drive us further and further into technical debt. I think we can't continue adding yet another hack to the pixel trait forever. Adding new features will become impossible eventually. 
I have been thinking about the Pixel type for a whole lot of time now. I've seen #1099, #956, #1464, #1212, #1013, #1499, #1516. That's why I want to call for action right now.

With the upcoming `0.24.0` release planned, which allows us to do breaking changes, we finally have the chance to actually improve the design of the trait in a sustainable manner. I propose we redesign the pixel trait right now! The goal of the redesign would be to maximise the flexibility such that it will last for a long time without major modifications. 

### Limitations of the current design

As far as I understand, these are currently the biggest problems with the pixel trait:
1. __It is not generic enough:__
    (#1511, #956, #1464, #1212). 
    The purpose of a trait is to enable generic code. 
    Implementing the pixel trait is problematic for multiple reasons.

2. __It has too many responsibilities:__
    (#1511, #1099, #1013, #1212)
    Adding multiple responsibilities to a trait is problematic, as the flexibility is reduced further and further. In the `std` library, we can observe multiple traits that are split up atomically. Many times, traits contain only a single method definition.   
    I identified the following responsibilities in the Pixel trait:
   - Field access. Methods like `as_slice`, `from_slice` and friends exist to access the samples in the pixel. Granting access to the sample data should be a separate responsibility, just like image operations and codecs should be separated. 
   - Color Type hints. `COLOR_TYPE: ColorType` is passed to the image encoders.
   - Color conversion. This will be reworked anyways. It should be removed from the pixel trait. This allows us to add color conversion without having to modify the pixel trait, which adds stability to the design.

    All of these could theoretically be separate traits.

Furthermore, there are other minor problems:
- Many algorithms use `channels4` method of the pixel trait. 
  - For example, see `hue_rotate_in_place`. In reality, the hue algorithm will only make sense for Rgb. These algorithms should probably not use the pixel trait in the first place. 
  - `horizontal_sample` should probably use the generic slice methods, and work with every possible number of channels. Even more idiomatic would be to actually treat the pixel as a whole, by not adding up each sample of a pixel individually. Instead, the algorithm should use hypothetical operations such as `blend` or `sum` or `average` on the pixel.

### How to start?
Of course, a redesign is a huge task. And even settling on a design will be challenging. Splitting up the pixel into atomic traits would probably help reduce bike shedding, as color conversion and similar tasks could be added independently. To speed up the design process, it could also help to list the new requirements of the pixel trait(s). 

Maybe we can start with the following list of requirements:
- Generic over Sample Type (works with `u8`, `u16`, `f32`, or even `f16` if the user wants to implement that)
- Single Responsibility: Field Access. No Color Conversion. Helper functions and color type declarations are done with additional traits and extension traits.
- Generic over Alpha Channel.
- Supports RGB and Luma. Maybe even user-defined colors such as YCbCr? Can we have `add` and `blend` for those`?
- Continuous Memory: Internal storage can be assumed to be a primitive array of fixed size.
- Maybe support converting between sample types? `pixel.map(|f32| (f32  255.0) as u8)`?
- Supports different color spaces already? What about sRGB vs LinearRGB? From my point of view, this is the only open question.

I propose we remove the color conversion from the pixel trait into a separate module. We could keep the old conversion algorithms until real color spaces are introduced.

### Example Design
This is by far not a full design. It should rather be understood as an example. It shows an extreme variant of the trait separation. I just typed this off of my head, so it probably doesn't factor in every requirement yet. 

The design focuses on separating concerns and providing default implementations for most methods. The code does not make any assumptions about the sample type, except for it being `Primitive`. The pixel is never assumed to be a specific color space. The only real assumption made, is that there exists an underlying array of samples. Color space models can be introduced to the crate by adding more traits, probably even without changing the existing traits. 

```rust

pub trait PrimitivePixel: Sized + Copy + Clone {

    /// The type of each channel in this pixel.
    type Sample: Primitive;

}

/// Access the samples of the pixel.
pub trait SlicePixel: PrimitivePixel {

    /// A slice of all the samples in this pixel, including alpha.
    fn as_slice(&self) -> &[Self::Sample];

    /// A slice of all the samples in this pixel, including alpha.
    fn as_slice_mut(&mut self) -> &mut [Self::Sample];

    fn apply(&mut self, mapper: impl FnMut(Self::Sample) -> Self::Sample) {
        for value in self.as_slice_mut() {
            *value = mapper(*value);
        }
    }

    fn apply2(&mut self, other: &Self, mapper: impl FnMut(Self::Sample, Self::Sample) -> Self::Sample) {
        for (own, other) in self.as_slice_mut().iter().zip(other.as_slice()) {
            *own = mapper(*own, *other);
        }
    }

    fn map(&self, mapper: impl FnMut(Self::Sample) -> Self::Sample) -> Self {
        let mut cloned = self.clone();
        cloned.apply(mapper);
        cloned
    }

    fn map2(&self, other: &Self, mapper: impl FnMut(Self::Sample, Self::Sample) -> Self::Sample) -> Self {
        let mut cloned = self.clone();
        self.apply2(other, mapper);
        cloned
    }
}

/// Automatically implemented for all types that implement `SlicePixel`.
/// This example design is based on the assumption that
/// all colors with alpha will have alpha as the last channel.
/// The assumption is only needed to provide default implementations for the first two methods.
pub trait AlphaSlicePixel: SlicePixel {
    const HAS_ALPHA: bool;

    fn alpha_channel_count() -> u8 { if Self::HAS_ALPHA { 1 } else { 0 } }

    fn color_channel_count() -> u8 where Self: ArrayPixel {
        Self::CHANNEL_COUNT - Self::alpha_channel_count()
    }

    fn as_color_slice_and_alpha(&self) -> (&[Self::Sample], Option<&Self::Sample>) {
        let slice = self.as_slice();
        debug_assert_ne!(slice.len(), 0);

        if Self::HAS_ALPHA { (&slice[..slice.len()-1], slice.last()) }
        else { (self.as_slice(), None) }
    }

    fn as_color_slice_and_alpha_mut(&mut self) -> (&mut [Self::Sample], Option<&mut Self::Sample>) {
        let slice = self.as_slice_mut();

        if Self::HAS_ALPHA {
            let (last, color) = slice.split_last_mut().expect("zero samples found");
            (color, Some(last))
        }
        else { (self.as_slice_mut(), None) }
    }

    fn as_color_slice(&self) -> &[Self::Sample] {
        self.as_color_slice_and_alpha().0
    }

    fn as_color_slice_mut(&self) -> &mut [Self::Sample] {
        self.as_color_slice_and_alpha_mut().0
    }

    fn apply_without_alpha(&mut self, mapper: impl FnMut(Self::Sample) -> Self::Sample) {
        for value in self.as_slice_without_alpha_mut() {
            *value = mapper(*value);
        }
    }

    fn apply2_without_alpha(&mut self, other: &Self, mapper: impl FnMut(Self::Sample, Self::Sample) -> Self::Sample) {
        for (own, other) in self.as_slice_without_alpha_mut().iter().zip(other.as_slice_without_alpha_mut()) {
            *own = mapper(*own, *other);
        }
    }

    fn map_without_alpha(&self, mapper: impl FnMut(Self::Sample) -> Self::Sample) -> Self {
        let mut cloned = self.clone();
        cloned.apply_without_alpha(mapper);
        cloned
    }

    fn map2_without_alpha(&self, other: &Self, mapper: impl FnMut(Self::Sample, Self::Sample) -> Self::Sample) -> Self {
        let mut cloned = self.clone();
        self.apply2_without_alpha(other, mapper);
        cloned
    }
}

// If we know the count of elements, we can use real arrays. 
// This replaces the runtime-checked `from_slice` method which would panic for invalid input.
/// Implemented by all colors.
/// You can use `Px::from_array(slice.try_into().expect("length mismatch"))` to get a pixel from a slice.
pub trait ArrayPixel: PrimitivePixel{
    const CHANNEL_COUNT: u8;
    fn to_array(&self) -> [Self::Sample; Self::CHANNEL_COUNT as usize];
    fn from_array(&self, array: [Self::Sample; Self::CHANNEL_COUNT as usize]) -> Self;
}

/// Add an alpha channel, for example go from RGB to RGBA, or L to LA. No color conversion is happening here.
/// Implemented by all colors.
pub trait WithAlpha: PrimitivePixel {
    type WithAlpha: PrimitivePixel<Sample=Self::Sample>;
    fn with_alpha(self, alpha: Self::Sample) -> Self::WithAlpha;
}

/// Take away an alpha channel, for example go from RGBA to RGB, or LA to L. No color conversion is happening here.
/// Implemented by all colors.
pub trait WithoutAlpha: PrimitivePixel {
    type WithoutAlpha: PrimitivePixel<Sample=Self::Sample>;
    fn without_alpha(self) -> Self::WithAlpha;
}

/// Used to detect the image type when encoding an image.
pub trait InherentColorType {
    const COLOR_TYPE: ColorType;
}

/// Collect into Pixel by creating a default pixel and filling the values. 
/// Can also be used instead of `from_slice`.
impl<Pixel> FromIterator<Pixel::Sample> for Pixel where Pixel: SlicePixel + Default {
    fn from_iter(iterator: impl Iterator<Item=Pixel::Sample>) -> Self {
        let mut result = Self::default();
        for (target, source) in result.as_slice_mut().iter_mut().zip(iterator) {
            *target = *source;
        }
        result
    }
}

/// Implemented for all colors individually. 
/// Allows converting the sample to a different sample type.
pub trait MapSamples<NewSampleType>: PrimitivePixel {
    type NewPixel: PrimitivePixel<Sample = NewSampleType>;
    fn map_samples(self, mapper: impl FnMut(Self::Sample) -> NewSampleType) -> Self::NewPixel;
}

/// This alias might not be a good idea - it's very monolithic again, just like the old design.
// Most methods should only use the traits they actually need, instead of requiring all these traits.
pub trait Pixel:
    ArrayPixel + SlicePixel +
    Copy + Clone + Default +
    InherentColorType +
    AlphaSlicePixel + AlphaSlicePixelExt
{ }


// ### example color implementation ###

#[derive(Clone, Copy)]
struct Rgb <T> ([T; 3]);

impl<T> PrimitivePixel for Rgb<T> { type Sample = T; }

impl<T> SlicePixel for Rgb<T> {
    fn as_slice(&self) -> &[Self::Sample] { &self.0 }
    fn as_slice_mut(&mut self) -> &mut [Self::Sample] { &mut self.0 }
}

impl<T> AlphaSlicePixel for Rgb<T> {
    const HAS_ALPHA: bool = false;
}

impl<T> ArrayPixel for Rgb<T> {
    const CHANNEL_COUNT: u8 = 3;
    fn to_array(&self) -> [Self::Sample; 3] { self.0 }
    fn from_array(&self, array: [Self::Sample; 3]) -> Self { Self(array) }
}

// non-generic implementations for the color type, as not all sample types may be supported for a color
impl InherentColorType for Rgb<u8> { const COLOR_TYPE: ColorType = ColorType::Rgb8; }
impl InherentColorType for Rgb<u16> { const COLOR_TYPE: ColorType = ColorType::Rgb16; }
impl InherentColorType for Rgb<f32> { const COLOR_TYPE: ColorType = ColorType::Rgb32F; }

impl<T> WithAlpha for Rgb<T> {
    type WithAlpha = Rgba<T>;
    fn with_alpha(self, alpha: Self::Sample) -> Self::WithAlpha {
        let [r,g,b] = self.0;
        Rgba([r,g,b,a])
    }
}

impl<T> WithoutAlpha for Rgb<T> {
    type WithoutAlpha = Self;
    fn without_alpha(self) -> Self::WithoutAlpha { self }
}

impl<T, D> MapSamples<D> for Rgb<T> {
    type NewPixel = Rgb<D>;

    fn map_samples(self, mapper: impl FnMut(Self::Sample) -> D) -> Self::NewPixel {
        self.as_slice().iter().cloned().map(mapper).collect()
    }
}

```

ripytide:
This is an overhaul of the crate to extract out the `Pixel` types and traits and improve them simultaneously into the [`pixeli`](https://docs.rs/pixeli) crate. This would fix #1523 and possibly many of its linked issues.

Summary of Changes:
- Pixel types and traits move to the `pixeli` crate which becomes a dependency
- Pixel types switch representation from `Rgb([T; 3])` to `#[repr(C)] struct Rgb<T> {r: T, g: T, b: T}` which is nicer to work with (`pixel.g` vs `pixel.0[1]`)
- Re-export the `pixeli` dependecy since users will often need to use those types/traits for doing pixel related stuff
- `Luma` is renamed to `Gray` as that is more understandable
- `Pixel::Subpixel` is renamed to `Pixel::Component` since subpixels are used with physical monitors for the rgb components and so the alpha component doesn't fit
- Two new `DynamicImage` variants needed to be added to pass the tests `Gray32F` and `GrayAlpha32F` which removes the edge condition for the `grayscale()` function not actually converting `Rgb32F` image variants to grayscale.
- Refactored away from deprecated `channels4()` legacy methods.
- Deprecated some less generic color conversion functions over a new more flexible version `convert_generic_image()`

I understand this is huge PR, but for this to be an atomic PR there wasn't really a way of splitting it any smaller. I also understand that this probably is a breaking change in many different areas, but I think that's acceptable given the scale of improvement this PR offers.

Btw the `pixeli` crate is pretty cool in by biased opinion, the new `Pixel` trait is wayyy more powerful than the previous one (credit to @johannesvollmer for some of the inspiration (#1523) and @kornelski for the related discussions in the `rgb` crate) and solves lots of previous issues, (like now `map_components()` can map to a different component type!).


ripytide:
The only remaining failing CI check is the `deny(public_private_dependencies)` lints failing for all the `pixeli` trait bounds which seems fine to me since its a public dependency? Can we add an exception for it somehow?

fintelia:
This change is far too large and has way too many breaking changes to review/iterate on. Breaking changes and big API additions need to be discussed and agreed on individually before they can be added. And in any case, we've generally been aiming to go 1-2 years between breaking releases

ripytide:
Why are we so scared of breaking changes that we can't make improvements? That seems very counter-productive and downright unhealthy for the long-term sustainability of the project. Technical debt will only build the longer we postpone important breaking changes.

And most of these changes have been discussed in #1523, with the rest just from a waterfall effect.

ripytide:
Just to be clear here, this PR would solve:
- #1523
- #1099
- #956
- #1464
- #1212
- #1013
- #1499

But if you think not making breaking changes is more important to users, then that is your decision to make I suppose.

HeroicKatora:
I don't think this addresses enough of major points raised by #1523 to warrant the breakage either. This PR changes the representation which the least problem here. Declaring a pre-`0.*` crate as a compatible solution is a non-starter, there already is [`rgb`](https://docs.rs/rgb/latest/rgb/) with several advantages such as its bytemuck integration allowing many opaque (not mentioned in type-bounds) uses.

Designing around the easy goals won't work, as it is the harder goals that break the current design, too. The two key points missing are awareness of color spaces and single responsibility. Instead, this design encodes even more information into the type system and it will run into the same troubles. The `Pixel`, if indeed one even supposes such a trait is needed, needs foremost a clear design boundary, but this offers even less of one.

There are a few interesting pieces here and there:
- Overall, renaming of the variants and types from `*Luma*` to `*Gray*` might be good for consistency.
- `Component` is a better wording than `Subpixel`.
- On the other hand, `ContiguousPixel` suggests a `NonContiguousPixel` which is a pretty obvious non-goal for the simple kinds of buffers that `DynamicImage` tries to offer as variants. There's definitely a conflict in design here. You're not going to be able to semantically encode all existing texel layouts in types (at least not in Rust).
- `FromPixelCommon` raises an interesting point. If we explicitly close the `DynamicImage` design against new variants, we do have a static 'top' element on our pixel conversion lattice and it's not `Rgba<u8>` but rather `Rgba<f64>`. Then `DynamicImage` should probably use that one. But also, if we close the conversion trait anyways we could do type-id dispatch instead of relying on associated methods in traits, which would probably be extremely more performant.

If you take one key point, then let it be the one summary also raised during `gfx`'s lessons learned to `wgpu`: It's not user-friendly to encode all of this in the type system, and algorithms built onto the representation can be (much) more optimal if they get to decide the mode of access (such as: bringing lookup-tables for color conversion instead of `From<_>` traits not having any such parameters). `image`'s own design still pre-dates this by quite a bit and _if_ there be breakage then for the sake of fixing this. (But also, `DynamicImage` tries to represent a simplified view that shouldn't be so affected by the underlying reasons, so no breakage-for-breakage sake).

kornelski:
BTW we've had a similar conversation about the `rgb` crate. I'm interested in adding such improved `Pixel` trait to it, but also want to preserve backwards compatibility with the current structs.

kornelski:
> gfx's lessons learned to wgpu: It's not user-friendly to encode all of this in the type system, and algorithms built onto the representation can be (much) more optimal if they get to decide the mode of access

Do you have more info on this?

fintelia:
Repeating some feedback given offline:

> The breaking changes aspect of the PR was a concern, but there were actually a few bigger factors...
> 
> **Multi-thousand line PRs are too big:** The larger a PR is the more rounds of feedback it takes to review and the longer each of those reviews takes. There's a good chance the reviews needed for a PR of this scale would need 10+ hours of maintainer time. And even if some parts are ready after a couple rounds of back-and-forth, there's still a very high chance of all the work being wasted because the PR stalled out before everyone reached agreement on the remaining details.
> 
> **API changes need to be discussed individually:** A single GitHub PR is a bad place to try to discuss a lot of different changes. Every API change has tradeoffs and questions to address, even when it isn't a breaking change.

ripytide:
> The two key points missing are awareness of color spaces and single responsibility. Instead, this design encodes even more information into the type system and it will run into the same troubles.

Good point, I didn't tackle color space issue at the same time, in that regard the four choices I can think of for attaching color-space to pixels are:
1. Enum in Pixel types:
```rust
enum ColorSpace {
    Unknown,
    Linear,
...
}
struct Rgba<T> {
    r: T,
    g: T,
    b: T,
    a: T,
    color_space: ColorSpace,
}
```
2. Enum in Image Type
```rust
struct ImageBuffer<P> {
    pixels: Vec<P>,
    color_space: ColorSpace,
}
```
3. Same as 1, but using a `PhantomData<U>` with a generic like `eudlid` [`Point`](https://docs.rs/euclid/latest/euclid/struct.Point2D.html) types.
4. Same as 2, but using `PhantomData` tags as well.

As you say the type-system heavy approaches in 3. and 4. are harder to reason with and learn, (is that similar to how `gfx` did things?). 1. Seems terrible for storage space which leaves options 2. as probably the best option given all the tradeoffs I think but I don't know if you had something else in mind?

With option 2. I think `FromPixelCommon` would change to something like:
```rust
trait FromPixelCommon<P> {
    fn from_pixel_common(pixel: P, source_color_space: ColorSpace, destination_color_space: ColorSpace) -> Self;
}
```
Am I understanding correctly?

On the single responsibility aspect, which bit of this design goes did you see going against that principle? Are you suggesting moving the helper functions such as `component_array()` `color_component_array()` and the `map_*()` functions to an extension trait? I think I'd probably be fine with that but I'm not sure as to the advantages/disadvantages of separating them either.

PR #2264, opened 2024-06-23
Map `.apng` files to `ImageFormat::Png`

SludgePhD:
> I license past and future contributions under the dual MIT/Apache-2.0 license,
> allowing licensees to choose either at their option.

`.apng` is a common file extension used for animated PNGs.


HeroicKatora (approved):
[As per IANA](https://www.iana.org/assignments/media-types/image/apng). Thank you.

Issue #1736, opened 2022-06-25
Resize doesn't change image dimensions


RetroRock:
This happens in version `0.24.2`

## Expected

The original image with `6000x4000` pixels and `16MiB` in size gets resized to `600x400`.

## Actual behaviour

The original image with `6000x4000` pixels and `15MiB` in size gets resized to `6000x4000` and `3.6MiB`.

## Reproduction steps

```rust
let img = ImageReader::open("testimage.jpg");
img.resize(600, 400, FilterType::Nearest);
img.save("testimage600x400.jpg")?;
```
I also tried other filter types, but they all gave the same result.
Is this even a bug, or am I not understanding what resizing means in this context?

Help is greatly appreciated!

fintelia:
You are misunderstanding how to use the resize function. It _returns_ a new resized image:
```rust
let img = ImageReader::open("testimage.jpg");
let new_img = img.resize(600, 400, FilterType::Nearest);
new_img.save("testimage600x400.jpg")?;
```

HeroicKatora:
Should we add `#[must_use]` annotations similar to how `std` defines them for `str::replace` etc? I do see how it could be confusing given other methods such as `DynamicImage::invert` take a mutable reference and operate in-place instead.

RetroRock:
> You are misunderstanding how to use the resize function. It _returns_ a new resized image:
> ```rust
> let img = ImageReader::open("testimage.jpg");
> let new_img = img.resize(600, 400, FilterType::Nearest);
> new_img.save("testimage600x400.jpg")?;
> ```

Thank you! I should've probably looked at the implementation before creating this issue. But why does the size decrease when I save the original image to the disk? 🤔

kaj:
> But why does the size decrease when I save the original image to the disk? :thinking:

The compression parameters are probably different from when the original image was saved.  Jpeg is a lossy format that can be saved in different qualitys.

Also, not that the "returning a new image" pattern is nice because you can chain operations like this:

```rust
ImageReader::open("testimage.jpg")?
    .resize(600, 400, FilterType::Nearest)
    .save("testimage600x400.jpg")?;
```

PR #2268, opened 2024-06-28
Add `#[must_use]` attribute to some `DynamicImage` methods

dzfrias:
Hello! Thanks for maintaining such a helpful crate. This PR is a small API improvement to make it clear to users that most of the methods on `DynamicImage` are non-mutating. #1736 along with my own usage of this crate suggest that it would be helpful to make these methods `#[must_use]`.

Let me know if there's some reason that we're not using `#[must_use]` already, and if there are some methods that warrant or don't warrant being marked with this attribute.


HeroicKatora (approved):
That's cool, this change aligns them with string modification in the standard library.

kornelski (approved):

PR #2270, opened 2024-06-28
A bunch of Clippy fixes

kornelski:
size_of will be in prelude (it is in nightlies already).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.25.1"
+version = "0.25.2"
 edition = "2021"
 resolver = "2"
 
--- a/benches/copy_from.rs
+++ b/benches/copy_from.rs
@@ -6,7 +6,7 @@ pub fn bench_copy_from(c: &mut Criterion) {
     let mut dst = ImageBuffer::from_pixel(2048, 2048, Rgba([0u8, 0, 0, 255]));
 
     c.bench_function("copy_from", |b| {
-        b.iter(|| dst.copy_from(black_box(&src), 0, 0))
+        b.iter(|| dst.copy_from(black_box(&src), 0, 0));
     });
 }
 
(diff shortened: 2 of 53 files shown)


HeroicKatora (approved):

HeroicKatora:
Soo many opportunities for `must_use`, nice.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.25.1"
+version = "0.25.2"
```
fintelia, line 3: The PR description says bumping the version field was needed for semver-checks? Might be worth seeing if we can avoid that so an accidental `cargo publish` with the repository checked out doesn't result in a new version upload to crates.io until we're actually ready to release

kornelski: This is because the semver check saw a change that required a minor version bump, and would continue to complain about it until the version has been bumped accordingly. 

I guess the CI could be changed to look only for semver-major changes.

HeroicKatora: There's no harm bumping by two minor versions in a single release, so I think the _risk_ is at least low. At least lower than missing some of these warnings, I imagine. Maybe we can have a look at our release workflow instead, should be simple to compare against the latest published tag and do the version bump conditionally. I think this diff could also serve as one signal for when a release should be done.

fintelia: The failure mode I'm afraid of is me trying to publish a different crate but accidentally running `cargo publish` in the wrong terminal window. That'll push whatever branch I happen to have checked out as the next `image` release even if it just includes some half-baked in progress changes.

fintelia: Created https://github.com/image-rs/image/pull/2277 which seems to fix it

kornelski: Maybe dev version of the crate could have `publish = false` set?

or add `./cargo/config.toml` that somehow breaks the registry configuration?

```
[registry]
default = "…"  
```

PR #2272, opened 2024-06-29
feat: add sccache to the build pipeline

sassman:
This PR builds all on CI with the sccache from Mozilla. It should in theory speed up builds by a significant amount of time.

But let's see.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,6 +8,11 @@ on:
   schedule:
     - cron: '5 16 * * 6'
 
+env:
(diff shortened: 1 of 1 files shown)


HeroicKatora:
I don't think it had much of an effect, yet? In fact seems a little slower on first sight of comparing the different feature configuration tests with the ones from the main branch.

sassman:
yeah, does not look too promising. I guess I'll close the PR then.

HeroicKatora:
Thank you for the experiment in any case :+1:

PR #2291, opened 2024-07-26
Expose zune-jpeg's Exif chunk extraction

Shnatsel:
Carbon copy of the ICC chunk handling, but for Exif.

Required for [`wondermagick`](https://github.com/Shnatsel/wondermagick) so that we could rotate the JPEGs correctly based on metadata.

In the long term we probably want to integrate `kamadak-exif`, since it doesn't have any unsafe code and seems to be in quite a good shape. But this PR seems like a simple and uncontroversial start.

### DIFF
--- a/src/codecs/jpeg/decoder.rs
+++ b/src/codecs/jpeg/decoder.rs
@@ -55,7 +55,7 @@ impl<R: BufRead + Seek> JpegDecoder<R> {
 
     /// Returns the raw [Exif](https://en.wikipedia.org/wiki/Exif) chunk, if it is present.
     /// A third-party crate such as [`kamadak-exif`](https://docs.rs/kamadak-exif/) is required to actually parse it.
-    pub fn exif(&mut self) -> ImageResult<Option<Vec<u8>>> {
+    pub fn exif_metadata(&mut self) -> ImageResult<Option<Vec<u8>>> {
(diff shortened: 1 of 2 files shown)


Shnatsel:
The single failing CI job seems to be unrelated to this change.

Shnatsel:
In theory PNG and WebP images can have Exif metadata too. Do you want me to make this a method on `ImageDecoder` with a default impl that just returns `None`?

Shnatsel:
Oh god, that actually doesn't work because Rust **always calls the default impl** on an `impl Trait`, not the specialized one. So this way we always get `None` on `ImageReader.into_decoder()`.

The workaround is to make the method required and to NOT provide a default impl, but that's semver-breaking because the trait is public and can be implemented by third-party crates.

This means ICC chunk extraction has also been broken the entire time...

Shnatsel:
nvm I figured it out, this works and is ready for review

fintelia (approved):
Looks good to me. The image-webp crate also [already supports this functionality](https://docs.rs/image-webp/latest/image_webp/struct.WebPDecoder.html#method.exif_metadata).

Shnatsel:
Nice! I'll try to add that later, but I'd appreciate a merge of this as-is so that the PR doesn't stall.

Issue #2238, opened 2024-05-18
Moving `imageops` and similar functionality to the `imageprocs` crate


ripytide:
At the moment the `image` crate has several image processing functions in the `imageops` module which seems a bit strange to me as they look like they would better belong in the `imageprocs` crate. Is there a reason why such functionality is in this crate, or is perhaps a holdover from long ago?

I would be happy to do the work migrating/integrating the functionality into `imageproc` if you'd like.

fintelia:
The `imageops` module was added long before my time and has gotten very little attention in the last few years. It contains some methods like `crop` or `resize` that totally make belong in the crate, but also has others like `huerotate` that probably aren't necessary. I'd be open to moving functionality, but we should discuss on a case-by-case basis first

ripytide:
Great, https://github.com/image-rs/imageproc/issues/7 is the corresponding issue on the `imageprocs` crate.

My current action plan is to write up a list of all the functions in `imageops` and then individually integrate each one into `imageprocs`. After each function is integrated and released that would give you the ability to deprecate those individual functions and point users at the corresponding functions in the `imageprocs` crate. Then at some point in the future you would have an easier time deleting the deprecated functions hoping most users will have migrated.

How does that sound?

theotherphil:
As the image crate has a huge number of users and very little activity happens on the `imageops` code, the cost of maintaining duplicated functionality is low but the impact of breaking changes is potentially high.

I'm in favour of:
1. Adding all current `imagops` functionality to `imageproc`, either vs re-exports or re-implementations depending on whether we want to tweak anything.
2. Adding a note in the `imageops` module (via code/readme/PR template/whatever) that new image processing functionality belongs in `imageproc`.

The question of deprecation or retiring existing `imageops` functionality is trickier and thankfully not up to me!

fintelia:
Yeah, this would involve deprecating the methods now and only removing in the next major release (which won't be any time soon given we haven't even started planning it yet)

During this process it is also worth looking at whether these methods are using internal functionality / are tightly tied to the image crate. Like the `invert` method only works because one of the required methods on the `Pixel` trait is `Pixel::invert`. The APIs around writing image processing code that's generic across image type is something I've held off working on because it seemed like there were no elegant options and any changes would just cause a lot of breakage

ripytide:
I've written up the list in a tracking table in the top comment which can be updated as we make progress toward this larger goal. Let me know if you'd like to change anything or if I've missed anything.

[...]

PR #2292, opened 2024-07-27
Expose in place operations on DynamicImage

Shnatsel:
This is a straightforward pass-through of the existing imageops functions.

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -890,13 +890,13 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::rotate90(p))
     }
 
-    /// Rotate this image 180 degrees clockwise.
+    /// Rotate this image 180 degrees.
     #[must_use]
     pub fn rotate180(&self) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::rotate180(p))
     }
 
-    /// Rotate this image 180 degrees clockwise in place.
+    /// Rotate this image 180 degrees in place.
     pub fn rotate180_in_place(&mut self) {
         dynamic_map!(*self, ref mut p, imageops::rotate180_in_place(p))
     }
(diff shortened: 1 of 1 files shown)


ripytide:
I thought we were removing the image ops functionality to the `imageproc` crate, propagating the `imageops` methods seems to be in conflict with this goal #2238 as they would eventually be deprecated and removed

Shnatsel:
Rotations and flips are required to apply JPEG orientation metadata. We will want reading JPEGs with the correct orientation to "just work" in the future without extra dependencies, so rotations and flips specifically have to stay. 

I agree with removing blur, huerotate, etc.

ripytide:
But is that not the responsibility of the decoding jpeg library and not something we should expose as a public method on the image types in this crate? Especially since none of the rotation or flipping functions are used anywhere within the crate as far as I can see.

Shnatsel:
Exif metadata can also appear on PNG and WebP images. And none of the format decoders actually parse Exif themselves. So it is preferable to implement it once on top of the different decoders, instead of trying to plumb it into each low-level format decoder.

ripytide:
Would you be able to use the flip/rotate functions from the `imageproc` library when they eventually get released?

kornelski:
These days EXIF rotation is de-facto required for decoding of JPEG images the way users expect.

For historical reasons it's a bit overcomplicated, but it can be considered a part of the codec now.

fintelia (commented):
I commented [here](https://github.com/image-rs/image/pull/2299#discussion_r1732062083) but I'd actually rather only expose a single rotate-in-place method. The `DynamicImage` type already has so many methods that each additional one we add makes it that much harder for users to find the functionality they're looking for

Shnatsel:
Understood. I'll change the newly added methods to private and add a single function for in-place rotation in #2299. Closing in favor of #2299.

Issue #2294, opened 2024-07-28
Add in-place versions of all rotation operations


Shnatsel:
In v0.25.2, certain operations such as flips and 180 rotations that do not modify the dimensions of the image are performed in-place. But rotations that modify the dimensions [do not have](https://docs.rs/image/0.25.2/image/imageops/index.html) an `_in_place` variant and require copying the image, doubling the peak memory usage.

There is no reason why such operations couldn't be performed in place. The area of the image remains constant, and so the underlying buffer stays exactly the same size. It is totally possible to shuffle pixels around in-place, without making a copy of the entire image.

This is going to be especially important for JPEG orientation handling, because metadata-rotated JPEGs are now common thanks to mobile phones, and take up quite a lot of RAM in their uncompressed form due to the high resolution of the photos.

Shnatsel:
I've done some literature review, and while algorithms for this exist, they are both complex and quite inefficient: 
https://en.wikipedia.org/wiki/In-place_matrix_transposition

So it's probably not practical to implement this.

Issue #2222, opened 2024-04-30
Image metadata handling


fintelia:
This crate currently has limited support for handling image metadata. Taking a bit of time to jot down some thoughts on how we might be able to add support for basic image metadata. This is very much still a draft

## Structure of metadata

Basing on what what metadata can be stored in PNG images (though other formats are similar) we might want to represent metadata something like: 

```rust
/// Orientations from EXIF
pub enum Orientation {
    Normal = 1,
    FlipH = 2,
    ...
}

#[non_exhuastive]
pub enum ColorPrimaries {
    Unknown,
    bt709,
    rec2020,
    ...
}

#[non_exhuastive]
pub enum TransferFunction {
    Unknown,
    Srgb,
    Gamma(f32),
    Linear,
    ...
}
pub enum ColorSpace{
    /// Based on cICP PNG chunk
    Cicp {
        primaries: ColorPrimaries,
        transfer_function: TransferFunction,
        full_range: bool,
    },
    // Would be nice to parse the ICC profile further if we had a library that could do so.
    Icc(Vec<u8>),
}
pub const SRGB_COLOR_SPACE= ColorSpace::Cicp {
    primaries: ColorPrimaries::Srgb,
    transfer_function: TransferFunction::Srgb,
    full_range: true,
}

#[non_exhuastive]
pub struct Metadata {
    // Most (but not all) formats store orientation in the EXIF data. However, often this is the 
    // only metadata people care about from the EXIF chunk. Should be possible to extract this
    // field without implementing full EXIF parsing support.
    pub orientation: Orientation,

    pub color_space: ColorMetadata,

    // Could consider parsing this ourselves, but might be better to just tell users to use the 
    // `kamadak-exif` crate if they care about the contents. 
    pub exif: Option<Vec<u8>>,

    // XML encoded. Not super eager to ship a full XML parser to decode it...
    pub xmp: Option<Vec<u8>>,
}

impl Reader {
    fn decode_with_metadata(self) -> (DynamicImage, Metadata) { ... }
}
```

## Decoding API

```rust
trait ImageDecoder {
    fn icc_profile(&mut self) -> ImageResult<Option<Vec<u8>>>; // already exists
    fn color_metadata(&mut self) -> ImageResult<ColorMetadata>;
    fn orientation(&mut self) -> ImageResult<Orientation>;
    fn exif_metadata(&mut self) -> ImageResult<Option<Vec<u8>>>;
[...]

Issue #793, opened 2018-07-18
Split ImageBuffer and relevant low-level structs and traits for raw images into a separate crate?


newpavlov:
It could be useful for using those structs in other crates which work with raw images and don't want to depend on whole `image` crate. Alternatively we could add additional default feature which will gate functionality not related to raw image representation.

tomaka:
> Alternatively we could add additional default feature which will gate functionality not related to raw image representation.

I would prefer this solution. Extracting crates and types into an external crate will likely cause semver hell in the long term.

newpavlov:
But on the other hand it will allow to stabilize crate (lets call it `image-core` for now) earlier. Plus encoder/decoder crates will be able to depend on `image-core`, thus providing unified API, which will allow to move out shims which you currently have in `image` crate. As kind of prior art you can look at motivation behind `rand_core` split.

fintelia:
I think we should give this some serious consideration. Right now the crates for specific image types (image-png, image-tiff, etc.) each have to implement their own versions of ColorType, ImageDecoder, and so forth which leads to a lot of redundancy.

HeroicKatora:
I found another reason to support this option: There are two separate concerns at play here. The decoder part, and the image manipulation part. For web, such as in `servo`, the decoder part is much more focussed and should undergo extreme scrutiny. However, the focus of this library as is is far to large to generate complete 'input' coverage, as even the input term itself is extremely hard to define.

So, in the long term I think we should aim to narrow down the scope of several libs (bikeshedding names):

* `image-core` that only provides buffers and methods for pixel handling **but no policy** and also no interpretation of pixel values (no color spaces). The best equivalent I can think of is the Linux DRM subsystem.
* `image` which would re-export those for compat and integrate them with codecs decoders, but excluding what is currently in `crate::imageops`. `DynamicImage` may be better left here than in `image-core`, to make is more likely for `image-core` to be possibly `no-std` and because of its auto converting mechanisms being technically policy. And here would be the extension point for color space conversion if we choose to do so.
* `imageproc` for operations on images.

That would give us the possibility of achieving a strong, and well tested foundation with `image-core` so that users who don't want to opt into the wider scope of auto conversion and operations can make use of decoders still.

icefoxen:
I like this but I would suggest that the interpretation of pixel values be part of `image-core`.  Pretty much every image format has its own enum for describing the options here and having them all be convertible into a single one would be a useful point of unification.

fintelia:
[...]

PR #2299, opened 2024-07-30
Add a function to apply Exif rotation

Shnatsel:
This is required to display JPEG photos correctly.

Doesn't wire it up to the JPEG decoding process yet, because that would be a semver-breaking change. But at least provides all the pieces for an API user to apply the correct rotation, together with #2291

Depends on #2292

**Why this API?**

A common case for a JPEG image is not to be rotated at all, i.e. have the orientation value of 1. Therefore we do not want to unconditionally make a copy of the image - often it is just a no-op. We need to do this in place.

Sadly nobody has implemented in-place 90 and 270 degree rotation yet (see #2294), so we are forced to make a copy in some cases. This copying behavior prevents us from implementing this in imageops module, which is generic over the input and so we cannot use the "replace the input with a copy" trick there.

So there are two options:
1. Enshrine the unnecessary copying in the API by returning something like `Result<Option<GenericImage>>` and make the caller deal with it
1. Only implement Exif rotation on DynamicImage until #2294 is fixed

This PR picks the second option.


kornelski (approved):

ripytide:
I still think that rotations and flips are image processing operations and as such belong in `imageproc` and so should not be exposed from the `image` crate. But if they are needed for a few codecs then keeping a small private module of rotation/flipping functions in `image` is fine.

Shnatsel:
@ripytide it doesn't really make sense to move it to `imageproc` if `image` needs to use it anyway. Neither does it make sense to put it in a private module, because API consumers need it, so we'd have to expose it in some way anyway; and copy-pasting it into imageproc and exposing it from there is not going to help anyone, least of all `image` maintainers.

ripytide:
So I think there are two main use-cases in play here:
- Users wanting to rotate/flip an image in their code.
- Codec authors needing to rotate/flip images as part of the codec decode/encode process due to exif and other such metadata.

For the first use-case, It makes the most sense to me for users to look in a single location for operations that process/modify images such as rotating/flipping images, hence I would expect as a user to look in the `imageproc` crate. Otherwise as a user I would be confused, which methods do look for in `image` vs `imageproc`? I think this is called the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle). 

For the second use-case I can see why it is incompatible with the above solution to the first use-case since if the rotate/flip functions were located in the `imageproc` crate then the codec libraries could not depend on `imageproc` since that would create a circular dependency:

```
codec_crate -> imageproc -> image -> codec_crate
```

I think this is a bit of a problem, and the root cause is the grouping of multiple functionalities in the `image` crate:
- Encoding/Decoding images (`ImageDecoder` and similar traits/types)
- Non-Encoded Image representation (`ImageBuffer` and similar traits/types)

The circular dependency above is due to the fact that `imageproc` is depending on `image` but only needs the second functionality. Were the two functionalities split into separate crates then this circular dependency would not be an issue:

```
image-encode-decode -> codec_crate -> imageproc -> image-representation
```

I think this is called [tight/loose coupling](https://en.wikipedia.org/wiki/Coupling_\(computer_programming\))

Perhaps we should treat the root cause of the issue and split the `image` crate?

src/dynimage.rs
```diff
@@ -878,30 +878,81 @@ impl DynamicImage {
...
         dynamic_map!(*self, ref p => imageops::rotate270(p))
     }
 
+    /// Applies the [Exif orientation](https://web.archive.org/web/20200412005226/https://www.impulseadventure.com/photo/exif-orientation.html) to the image.
+    ///
+    /// Orientation is specified in the Exif metadata, and is often written by cameras.
+    /// It is expressed as an integer in the range 1..=8; passing other values will return an error.
+    ///
+    /// Due to an implementation detail, orientations 5..=8 copy the image internally.
+    pub fn apply_exif_orientation_in_place(&mut self, orientation: u8) -> Result<(), ImageError> {
```
fintelia, line 926: I'd rather this function take an `Orientation` enum instead of a `u8`. That would have a couple nice benefits:

- This function would become infallible and thus wouldn't need to return a `Result`.
- We'd no longer need to add the standalone `*_in_place` methods since we could just recommend that users instead call this one with a hard-coded value.
- We could reuse the enum if we added a method to extract the orientation from an EXIF chunk (full EXIF parsing feels out of scope this crate, but adding logic to extract a single field feels pretty manageable)

Shnatsel: Sounds good! I'll update the PR with these changes.

ripytide:
Oh I've just found the #793 issue, that would have saved me some typing regarding the crate splitting :smile:

Shnatsel:
@fintelia I've updated the code based on your feedback. Does that look good?

I wasn't quite sure where to put the `Orientation` so I put it in its own file and exported it as a top-level type.

src/dynimage.rs
```diff
@@ -878,30 +878,76 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::flip_vertical(p))
     }
 
+    /// Flip this image vertically in place
+    pub fn flipv_in_place(&mut self) {
```
fintelia, line 882: Could you leave this private? (Or just inline it into apply_orientation)

src/dynimage.rs
```diff
@@ -878,30 +878,76 @@ impl DynamicImage {
...
+    }
+
     /// Flip this image horizontally
     #[must_use]
     pub fn fliph(&self) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::flip_horizontal(p))
     }
 
+    /// Flip this image horizontally in place
+    pub fn fliph_in_place(&mut self) {
```
fintelia, line 893: See above

src/dynimage.rs
```diff
@@ -878,30 +878,76 @@ impl DynamicImage {
...
 
-    /// Rotate this image 180 degrees clockwise.
+    /// Rotate this image 180 degrees.
     #[must_use]
     pub fn rotate180(&self) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::rotate180(p))
     }
 
+    /// Rotate this image 180 degrees in place.
+    pub fn rotate180_in_place(&mut self) {
```
fintelia, line 910: See above

src/dynimage.rs
```diff
@@ -878,30 +878,76 @@ impl DynamicImage {
...
+        dynamic_map!(*self, ref mut p, imageops::rotate180_in_place(p))
+    }
+
     /// Rotate this image 270 degrees clockwise.
     #[must_use]
     pub fn rotate270(&self) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::rotate270(p))
     }
 
+    /// Applies the specified [Orientation] to the image.
```
fintelia, line 920: Perhaps "Rotates and/or flips the image as indicated by `orientation`" or something to that effect?

src/dynimage.rs
```diff
@@ -878,30 +878,76 @@ impl DynamicImage {
         dynamic_map!(*self, ref p => imageops::flip_vertical(p))
     }
 
+    /// Flip this image vertically in place
+    pub fn flipv_in_place(&mut self) {
+        dynamic_map!(*self, ref mut p, imageops::flip_vertical_in_place(p))
+    }
+
     /// Flip this image horizontally
```
fintelia, line 886: Might be worth suggesting `apply_rotation(Orientation::FlipH)` as an in-place alternative of the same operation

src/orientation.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Do not perform any transformations.
+    NoTransforms,
+    /// Rotate by 90 degrees clockwise.
+    Rotate90,
+    /// Rotate by 180 degrees. Can be performed in-place.
+    Rotate180,
+    /// Rotate by 90 degrees clockwise.
+    Rotate270,
+    /// Flip horizontally. Can be performed in-place.
+    FlipH,
```
fintelia, line 16: Perhaps `FlipHorizonal` or `FlipHorizontally`?

Shnatsel:
@fintelia I've addressed all the feedback. All good points, thank you!

Sorry it took so long. Somehow I was convinced I've already done it, I must have dreamt it or something :sweat_smile:

Shnatsel:
There are some `cargo deny` and `clippy` issues on CI, but they are unrelated to the changes made in this PR.

src/orientation.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Rotate by 180 degrees. Can be performed in-place.
+    Rotate180,
+    /// Rotate by 270 degrees clockwise. Equivalent to rotating by 90 degrees counter-clockwise.
+    Rotate270,
+    /// Flip horizontally. Can be performed in-place.
+    FlipHorizontal,
+    /// Flip vertically. Can be performed in-place.
+    FlipVertical,
+    /// Rotate by 90 degrees clockwise and flip horizontally.
+    Rotate90FlipH,
```
fintelia, line 20: It is a bit long, but perhaps `Rotate90FlipHorizontal` for consistency?

src/orientation.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Rotate by 270 degrees clockwise and flip horizontally.
+    Rotate270FlipH,
```
fintelia, line 22: `Rotate270FlipHorizontal`?

fintelia (approved):
Left two comments on naming. Otherwise looks good to me

fintelia:
Actually, I wonder if we should create a new `metadata` module for the `Orientation` enum. It would be a natural place if we end up adding other types from https://github.com/image-rs/image/issues/2222

src/orientation.rs
Shnatsel: I tried that, and my eyes just glaze over when I try to read it. I genuinely believe this option is more readable. 

And if its meaning is ever unclear, there is always the doc comment.

Shnatsel:
I've renamed the module to `metadata`. I did re-export the `Orientation` type on the top level still. I think that's appropriate for this type; we might revisit that when we add more metadata types.

I've also added a function to convert back into Exif and added a doc link from `Orientation` to `apply_orientation()`.

Issue #986, opened 2019-07-08
`blur` function is too slow


Aloxaf:
`blur` is too slow, `imageproc::filter::gaussian_blur_f32` is faster than it, but still slow.

## Reproduction steps

```Rust
use image::imageops::blur;
use image::{Rgb, RgbImage};
use imageproc::drawing::draw_filled_circle_mut;
use imageproc::filter::gaussian_blur_f32;
use std::time::Instant;

fn main() {
    let mut image = RgbImage::new(1000, 1000);
    draw_filled_circle_mut(&mut image, (500, 500), 500, Rgb([255, 255, 255]));

    let start = Instant::now();
    blur(&image, 100.0).save("rust_1.png").unwrap();
    dbg!(start.elapsed());

    let start = Instant::now();
    gaussian_blur_f32(&image, 100.0).save("rust_2.png").unwrap();
    dbg!(start.elapsed());
}
```

result is 
```
[src/main.rs:13] start.elapsed() = 12.362780732s
[src/main.rs:17] start.elapsed() = 3.335411994s
```

and Pillow
```python
from PIL import Image, ImageDraw, ImageFilter
from time import time

img  = Image.new('RGB', (1000, 1000))
draw = ImageDraw.Draw(img)
draw.ellipse((0, 0, 1000, 1000), fill='#ffffff')

start = time()
img.filter(ImageFilter.GaussianBlur(100)).save('python.png')
print(f'{time() - start}s')
```

result is `0.1433885097503662s`

Here is the final image 

rust_1.png (it's a little strange...
![rust_1](https://user-images.githubusercontent.com/17017672/60798873-994a4280-a1a4-11e9-8db5-78528b2605e1.png)

rust_2.png
![rust_2](https://user-images.githubusercontent.com/17017672/60798882-9e0ef680-a1a4-11e9-9c93-007ca0253153.png)

python.png
![python](https://user-images.githubusercontent.com/17017672/60798894-a36c4100-a1a4-11e9-99c2-79e41f4e87aa.png)

HeroicKatora:
Just to clarify, PIL uses a C version indirectly https://github.com/python-pillow/Pillow/blob/ab9a25d623fdd7f8de3e724b538f5660eac589ae/src/libImaging/BoxBlur.c#L294

It's still somewhat embarrasingly slow.

Aloxaf:
@HeroicKatora Yes，maybe porting it to rust will be a good idea？

theotherphil:
[...]

PR #2302, opened 2024-08-02
Fast Blur

torfmaster:
# PR

This is an attempt to solve #986 as the current solution (using 
`fastblur`) is fragile and requires the consumers to access raw buffers
of the image which is undesirable.
The implementation is inspired by
[this](https://blog.ivank.net/fastest-gaussian-blur.html) article
which is in turn based on Kovesi, P.:  Fast Almost-Gaussian Filtering The Australian Pattern
Recognition Society Conference: DICTA 2010. December 2010. Sydney.

The short story is that the real gaussian blur takes about 30 second for
real-world photos which is undesirable. The algorithm here is quick and 
has pretty good results.

### Is this PR ready?

No, but I want to have some feedback on the implementation so I can move on (in the right direction).

### Are you aware of #2238?

Yes, but I don't know how to deal with it. `blur` is not yet deprecated and
users of the library will search for a fast alternative in this crate,
not in `imageproc`. I am open to move my code there now or later.
The advantage of having it here would be that I do not suffer from
the `non_exhaustive` of `DynamicImage` and can implement blur
for this type (if desired).

### Why didn't you simply c&p the code from the blurslice crate (or the article)?

The `blurslice` crate is distributed under the MIT license and
re-licencing the code under MIT/Apache2 would require the approval
of all contributors of the crate.
On the other hand it doesn't integrate well with the principal
structures `DynamicImage` and `ImageBuffer`. That's why I 
decided to re-implement the algorithm. I have to admit here are some
details in both implementations that I don't understand and where
I couldn't find an explanation. Hence I just did my own implementation
that I think I understand.

### What about the performance?

I didn't check the performance (and didn't optimize it).
But opposed to the `blur` implementation the blurring finishes
in a reasonable amount of time.
As soon as the interface is fixed (i.e. whether to accept `GenericImageView` or
`DynamicImage`) I can start to performance tweaks.

### What about tests?

I can't really imagine how to represent functional tests. However, I plan to

 * write benchmarks as soon I dig into performance tuning
 * write trivial tests (`sigma=1` is a no op)
 * write kani tests that make sure that the potentially panicking slice access doesn't panic
 
### Is the interface fixed?

Currently, I accept an `ImageBuffer` simply because it lets me operate on slices rather
than using another interface to access pixels. However, it might make sense
to also accept an `DynamicImage` (can be easily accomplished) or a
`GenericImageView` (requires a complete rewrite of the algorithm).

### Will the "old" blur stay?

The fast blur algorithm is inaccurate but for enough for many applications. I would recommend 
that the "real" blur stays. It is not much code anyways as it is implemented general convolution
with gaussian kernel.

### License


ripytide:
Wow great work!

> Are you aware of https://github.com/image-rs/image/issues/2238?
Yes, but I don't know how to deal with it. blur is not yet deprecated and
users of the library will search for a fast alternative in this crate,
not in imageproc. I am open to move my code there now or later.
The advantage of having it here would be that I do not suffer from
the non_exhaustive of DynamicImage and can implement blur
for this type (if desired).

Long-term I think this function should belong in `imageproc` since it is a image processing operation, for short-term, since a lot of `imageops` are yet to be deprecated it could be added temporarily in `image` and then copied across later but that seems like extra work to me.

Maybe we could add a note in the docs of the "slow" blur method in this crate that points to this faster method in the `imageproc` crate?

Regarding `DynamicImage` you can use the `dynamic_map` macro, (not sure if it's currently exported from the crate but it's easy enough to redefine yourself):

```rust
use image::{DynamicImage, ImageBuffer, Pixel};

macro_rules! dynamic_map(
        ($dynimage: expr, $image: pat => $action: expr) => ({
            use DynamicImage::*;
            match $dynimage {
                ImageLuma8($image) => ImageLuma8($action),
                ImageLumaA8($image) => ImageLumaA8($action),
                ImageRgb8($image) => ImageRgb8($action),
                ImageRgba8($image) => ImageRgba8($action),
                ImageLuma16($image) => ImageLuma16($action),
                ImageLumaA16($image) => ImageLumaA16($action),
                ImageRgb16($image) => ImageRgb16($action),
                ImageRgba16($image) => ImageRgba16($action),
                ImageRgb32F($image) => ImageRgb32F($action),
                ImageRgba32F($image) => ImageRgba32F($action),
                _ => unreachable!(),
            }
        });

        ($dynimage: expr, $image:pat_param, $action: expr) => (
            match $dynimage {
                DynamicImage::ImageLuma8($image) => $action,
                DynamicImage::ImageLumaA8($image) => $action,
                DynamicImage::ImageRgb8($image) => $action,
                DynamicImage::ImageRgba8($image) => $action,
                DynamicImage::ImageLuma16($image) => $action,
                DynamicImage::ImageLumaA16($image) => $action,
                DynamicImage::ImageRgb16($image) => $action,
                DynamicImage::ImageRgba16($image) => $action,
                DynamicImage::ImageRgb32F($image) => $action,
                DynamicImage::ImageRgba32F($image) => $action,
                _ => unreachable!(),
            }
        );
);

fn main() {
    let image1 = DynamicImage::new_luma8(3, 3);
    let image2 = DynamicImage::new_rgb8(3, 3);

    let image1 = dynamic_map!(image1, i => none(i));
    let image2 = dynamic_map!(image2, i => none(i));
}

fn none<P: Pixel, Container>(
    _image: ImageBuffer<P, Vec<Container>>,
) -> ImageBuffer<P, Vec<Container>> {
    todo!()
}
```

torfmaster:
Thank you for your comments. I added tasks for your (and my points) so we can track progress more easily.

torfmaster:
I have pushed a rather coarse fix for the private dependency checks. However, I don't know how to deal with these checks. I found no easy way of systematically fix this. The main problem is, that I can't use the public dependency property in the regular `Cargo.toml` as the [feature](https://rust-lang.github.io/rfcs/1977-public-private-dependencies.html) is unstable and can't be used in stable Rust. Hence, to use this feature we would have to

 * add the cargo feature in the github action (as it is currently done) using a shell script
 * specify the `rayon` crate as `public` using a shell script (very ugly)

To make this PR mergeable I hacked an "allow" for this llint to `buffer_par` knowing that clippy won't complain about allowing an unknown lint. However, this solution is likely to fail in the future.

torfmaster:
I did some initial benchmarking: on my intel i9 machine the bench (1024x768 rgb) takes 680ms with the true gaussian blur whereas `fast_blur` takes 57 ms. I checked in both benchmarks.
Next I will try to find some bottlenecks in the implementation.

etemesi254:
Hi, some ways to improve noted in my comments on zune-imageprocs (author of the library)

https://github.com/etemesi254/zune-image/blob/734bdc7bfcd58860b0a92dde0a98055a1f94b422/crates/zune-imageprocs/src/box_blur.rs#L197-L220

Never did the loop unrolling to save ilp, but others were implemented, really nice properties like no bounds check in inner loop, may be worth investigating (see https://godbolt.org/z/zexPEd976)

torfmaster:
I invested some time into performance optimization:

* simplify leftmost value by `r * f(0)`
* get rid of bound checks by asserting the length of the samples
* get rid of conditionals checking for lower and upper bounds

None of these measures produced any results. I am leaving the code as it is for now (if no one comes up with another idea, of course).

torfmaster:
Furthermore, I see no need for kani/quickcheck tests. First the bounds are obviously correct. On the other hand there is no point in using unsafe so securing the code using `kani` or `quickcheck` is unnecessary.
Regarding the other tests I added tests for edge cases of `sigma`. If you have more ideas regarding reasonable tests I am happy to implement them.

torfmaster:
@etemesi254:
I now implemented the performance improvements I am able to understand (that is I used transposing to improve usage of CPU caches). The algorithmic tweaks were already in place, though. I total I could speed up the blurring by 15 per cent. I feel I am not able to understand the more complicated improvements as I am lacking deeper assembly language knowledge.

etemesi254:
Nice. 

FYI these were the benchmark results for gaussian blur for big images . (image-rs, libvips,zune-image)

https://etemesi254.github.io/assets/criterion/imageprocs_%20gaussian%20blur/report/index.html

torfmaster:
My work here is done for now and I would be happy about a code review

torfmaster:
@ripytide sorry to bother you (but you already commented on the issue): Do you know what would be the next steps to merge this PR (and get it released)?
Thank you!

ripytide:
You'll need one of the maintainers signoff, of which I am not one.

Also, I don't think this function belongs in this library and should instead be in the `imageproc` library.

torfmaster:
Thank you for your feedback. I created a PR in `imageproc` as well (https://github.com/image-rs/imageproc/pull/683).

src/imageops/fast_blur.rs
```diff
@@ -0,0 +1,144 @@
...
+    channel_num * idx + channel
+}
+
+fn my_fast_horizontal_blur<P: Primitive>(
+    samples: &[P],
+    width: usize,
+    height: usize,
+    r: usize,
+    channel_num: usize,
+) -> Vec<P> {
```
HeroicKatora, line 69: nit: I'd prefer a more descriptive name, `horizantal_fast_blur_half`.

src/imageops/fast_blur.rs
```diff
@@ -0,0 +1,144 @@
...
+                vals[channel] = vals[channel]
+                    - extended_f(
+                        samples,
+                        width,
+                        height,
+                        j as isize - r as isize,
+                        i as isize,
+                        channel,
+                        channel_num,
+                    )
```
HeroicKatora, line 111: nit: more readable with intermediate variable names for this and the mirrored invocation.

examples/fast_blur/lenna.png
HeroicKatora, line 1: Not a good test image. Prefer [Mandril_color](https://raw.githubusercontent.com/JuliaImages/TestImages.jl/images/images/mandril_color.tif).

src/imageops/fast_blur.rs
```diff
@@ -0,0 +1,144 @@
...
+        samples = my_fast_horizontal_blur::<P::Subpixel>(
+            &horizontally_blurred_transposed,
+            height as usize,
+            width as usize,
+            *radius,
+            P::CHANNEL_COUNT as usize,
+        );
+    }
+    ImageBuffer::from_raw(width, height, samples).unwrap()
+}
```
HeroicKatora, line 35: `width = 0` as well as `height = 0` are 'normal' cases for input data. I think this misbehaves in a few of these cases in differing ways. (Overflow in `x - 1` in `extended_f` for instance). A test for that case would be good.

src/imageops/mod.rs
```diff
@@ -483,4 +486,18 @@ mod tests {
         let image = RgbaImage::new(50, 50);
         let _ = super::blur(&image, 0.0);
     }
+
+    #[test]
+    /// Test blur doesn't panick when passed 0.0
+    fn test_fast_blur_zero() {
+        let image = RgbaImage::new(50, 50);
+        let _ = super::fast_blur(&image, -1.0);
```
HeroicKatora, line 494: This is incongruent with the comment, it's passed `-1.0`.

src/imageops/mod.rs
```diff
@@ -483,4 +486,18 @@ mod tests {
...
+    }
+
```
HeroicKatora, line 496:
```suggestion

    #[test]
    /// Test blur doesn't panick when passed 0.0
    fn test_fast_blur_zero() {
        let image = RgbaImage::new(50, 50);
        let _ = super::fast_blur(&image, 0.0);
    }

    #[test]
    /// Test blur doesn't panick when passed an empty image (any direction)
    fn test_fast_blur_empty() {
        let image = RgbaImage::new(0, 0);
        let _ = super::fast_blur(&image, 1.0);
        let image = RgbaImage::new(20, 0);
        let _ = super::fast_blur(&image, 1.0);
        let image = RgbaImage::new(0, 20);
        let _ = super::fast_blur(&image, 1.0);
    }

```

HeroicKatora (commented):
Code review time :)

src/imageops/mod.rs
torfmaster: fixed

src/imageops/fast_blur.rs
torfmaster: fixed

src/imageops/fast_blur.rs
torfmaster: fixed

examples/fast_blur/lenna.png
torfmaster: fixed

src/imageops/fast_blur.rs
torfmaster: See below.

torfmaster:
I just wanted to give a short status report (I also cross post this comment to https://github.com/image-rs/imageproc/pull/683)

 - I added a functional test showing (for an example) that the effect of `fast_blur` is close to the Gaussian blur implementation in this library
 - I added a benchmark that shows that fast blur is much faster than the current blur implementation
 - I think I addressed all code review comments

What would be the next steps?

torfmaster:
@HeroicKatora sorry for pinging you directly but you already started a review. Is there anything I can do to support the review of this PR? Thank you a lot!

src/buffer_par.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(exported_private_dependencies)]
```
Shnatsel, line 1: Please avoid changes unrelated to the blur algorithm. The issue with a false positive lint is known and also happens on the main branch. We will take care of it separately, and CI failing on this issue is not a blocker for merging this PR.

torfmaster: I was not sure about the policy but I simply dropped the commit and the other unrelated clippy failure.

Shnatsel:
I've written the following example code for blur - feel free to include it in the PR under `examples/`:

```rust
use std::env;
use std::error::Error;
use std::path::Path;
use std::process;

    // Collect command-line arguments
    let args: Vec<String> = env::args().collect();

    // Ensure that we have 2 CLI arguments
    if args.len() != 3 {
        eprintln!("Usage: {} <path> <radius>", args[0]);
        process::exit(1);
    }

    let radius: f32 = args[2].parse()?;

    // Load the input image
    let path_str = &args[1];
    let path = Path::new(path_str);
    let input = image::open(path)?;

    // Correct but slow Gaussian blur
    let mut out_path_gauss = path.to_owned();
    out_path_gauss.set_extension("gaussian.png");

    let correct_but_slow = input.blur(radius);
    correct_but_slow.save(out_path_gauss)?;

    // Approximate but fast blur
    let mut out_path_fast = path.to_owned();
    out_path_fast.set_extension("fast.png");

    let correct_but_slow = input.fast_blur(radius);
    correct_but_slow.save(out_path_fast)?;

    Ok(())
}
```

With this test image both blur implementation show artifacts:

[blur-test-input](https://github.com/user-attachments/assets/20244122-fb60-4dcd-bd18-a82cd1aa4ed1)

It's just a white rectangle in the middle, surrounded by fully transparent pixels.

Both Gaussian blur in `image` and the fast blur implementation bleed color from fully transparent pixels into the white rectangle, resulting in the following image:

blur-test-input fast

The fully transparent pixels have R set to 255 while all the other channels are set to 0, and that red color bleeds into the white rectangle.

The expected blur as produced by GIMP is like this - blurred white rectangle without any red color:

[blurred-gimp-20px](https://github.com/user-attachments/assets/7ebf71f6-87de-4af1-a64c-d2d96ddca826)

To avoid such artifacts, the pixel's contribution to the color change should be weighted by its alpha channel value. Fully transparent pixels should get multiplied by 0 and not contribute to the changes in non-alpha channels at all.

But since both blur implementations are incorrect, I'll leave deciding how to handle this up to the maintainers.

HeroicKatora (approved):
I think it's fine to keep the incorrect for now, given that they both are. I think we should bunch it with the known color-space issues, afterall it's not clear whether to treat channels as pre-multiplied or not. Given that ambiguity, this just ignores the alpha channel (edit: or assumes pre-multiplied in contrast to the other parts of the library). I think it'd be best to introduces separate alpha-aware version of both pre-multiplied and non-multiplied channel interpretations at a later point and to keep it consistent for now.

PR #2305, opened 2024-08-13
Mark rayon as a public dependency

kornelski:
Tests on the main branch fail. It seems that rayon is intentionally exposed.

https://github.com/image-rs/image/blob/e176cd414ac6cc73909ff162e5d0b677f9d4fb08/src/lib.rs#L187-L188


HeroicKatora (commented):
Yep, I believe it's intentional as well. And since it's `v1` and optional anyways there's not much wrong doing so.

HeroicKatora (approved):

PR #2312, opened 2024-08-21
Implement DoubleEndedIterator on `enumerate_*()` pixel iterators

### LINKED ISSUE #2298: EnumeratePixels and EnumerateRows should implement DoubleEndedIterator
The [`EnumeratePixels`](https://docs.rs/image/0.25.2/image/struct.ImageBuffer.html#method.enumerate_pixels) and [`EnumerateRows`](https://docs.rs/image/0.25.2/image/struct.ImageBuffer.html#method.enumerate_rows) iterators should implement [`DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html).

This will make it possible to iterate in reverse order (for example right-to-left instead of left-to-right).

The [`Pixels`](https://docs.rs/image/0.25.2/image/struct.ImageBuffer.html#method.pixels) and [`Rows`](https://docs.rs/image/0.25.2/image/struct.ImageBuffer.html#method.rows) iterators already implement `DoubleEndedIterator`.

ripytide:
fixes #2298

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ num-complex = "0.4"
 glob = "0.3"
 quickcheck = "1"
 criterion = "0.5.0"
+itertools = "0.13.0"
 
 [features]
 default = ["rayon", "default-formats"]
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -332,8 +332,10 @@ where
     <P as Pixel>::Subpixel: 'a,
 {
     pixels: Pixels<'a, P>,
-    x: u32,
-    y: u32,
+    x_front: u32,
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -62,6 +62,7 @@ num-complex = "0.4"
 glob = "0.3"
 quickcheck = "1"
 criterion = "0.5.0"
+itertools = "0.13.0"
```
fintelia, line 65: Why is this needed?

ripytide: For the `itertools::assert_equal()` method in the iterators test.

PR #2322, opened 2024-09-13
fix: make set_icc_profile function available

misl-smlz:
We need to set the icc profile for webp. This PR makes the set_icc_profile function available from the outside.


fintelia:
I think this should probably be added as a method on `ImageEncoder`, with a default implementation that returns an unsupported error

misl-smlz:
@fintelia I've changed it the way you suggested

kornelski (dismissed):
It should return `Result<(), …>`. 

I'm not sure if a generic error or a dedicated `UnsupportedError`.

it would be good to also document that it doesn't check ICC for validity at this point, so doesn't guarantee to fail if the profile is invalid.

misl-smlz:
@kornelski Now it returns an Result

fintelia:
Looks like there's a few compile errors:
```
error[E0433]: failed to resolve: use of undeclared type `UnsupportedError`
   --> src/image.rs:798:13
    |
798 |         Err(UnsupportedError::from_format_and_kind(
    |             ^^^^^^^^^^^^^^^^ use of undeclared type `UnsupportedError`
    |
help: consider importing this struct
    |
2   + use crate::error::UnsupportedError;
    |
```

src/image.rs
```diff
@@ -787,6 +787,21 @@ pub trait ImageEncoder {
         height: u32,
         color_type: ExtendedColorType,
     ) -> ImageResult<()>;
+
+    /// Set the ICC profile to use for the image.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the ICC profile is not implemented for the format.
```
fintelia, line 795: Please update

fintelia:
Thanks!

Issue #2265, opened 2024-06-23
Animated PNG is decoded incorrectly


SludgePhD:
## Expected
![bonzi-book](https://github.com/image-rs/image/assets/96552222/7b50019a-7c6f-4625-99c2-f1e69de95b61)

This is how the animation should look (assuming your browser correctly implements APNG decoding; Firefox and Chromium-based browsers work fine in my tests).

## Actual behaviour

[Screencast_20240623_180709.webm](https://github.com/image-rs/image/assets/96552222/090ec233-f1c4-43fd-abd7-f5899783ee9e)

Using `image` to decode the animation and rendering the result to a window produces this flickering output instead. Other images work fine, so I don't think the bug is in my code.

## Reproduction steps

The failing APNG is attached above.

ripytide:
I can reproduce the issue:

```rust
use std::io::Cursor;

use image::codecs::png::PngDecoder;
use image::AnimationDecoder;

fn main() {
    let png_decoder = PngDecoder::new(Cursor::new(std::fs::read("monkey.png").unwrap())).unwrap();

    dbg!(png_decoder.is_apng().unwrap());

    let apng_decoder = png_decoder.apng().unwrap();

    let frames = apng_decoder.into_frames().collect_frames().unwrap();

    for (i, frame) in frames.into_iter().enumerate() {
        frame
            .buffer()
            .save_with_format(format!("frame-{i}.png"), image::ImageFormat::Png)
            .unwrap();
    }
}
```

It seems frames 2, 12 and 13 have missing sections (possibly via alpha=0) and all other frames seem fine.

Shnatsel:
At a glance this looks like an issue with compositing. Those frames do not change the the book, so that area is omitted from the frames. Other decoders fill in that area from the composited result of the previous frames, but apparently `image` does not.

We have correct compositing in the GIF decoder in `image` already. It even handles memory limits. We should generalize it to APNG and WebP animations too.

MocA-Love:
I'm currently running into the same problem and will look into it.

PR #2327, opened 2024-09-14
Fixed the problem of Animated PNGs being decoded incorrectly

higumachan:
related: #2265 

Fixed the update range of dispose_op for Previous and Background in APngDecoder, which was applied to the entire image, to now be within the update range of the previous frame. 

I referred to the following URL https://wiki.mozilla.org/APNG_Specification

I have checked in my environment and it works fine for the image in #2265.

The following is the data saved as an image of the frame.
[frames.zip](https://github.com/user-attachments/files/17002986/frames.zip)

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -18,8 +18,8 @@ use crate::error::{
     ParameterError, ParameterErrorKind, UnsupportedError, UnsupportedErrorKind,
 };
 use crate::image::{AnimationDecoder, ImageDecoder, ImageEncoder, ImageFormat};
-use crate::{GenericImageView, Limits};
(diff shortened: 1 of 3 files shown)


fintelia (commented):
The code looks good to me, but I don't think the test image has a suitable license for us to include in the repository

higumachan:
Thank you for reviewing.

> The code looks good to me, but I don't think the test image has a suitable license for us to include in the repository

Hmm, I’ve removed the data for now, but I don’t have a suitable replacement on hand, so I can’t add the test. What should I do?

fintelia:
If anyone wants to either manually construct a test image or track down a suitably licensed one, that would be appreciated. But I don't think it needs to block landing this

Issue #2340, opened 2024-10-08
Memory Allocation Failed in `image::DynamicImage::resize_exact`


rfuzz:
While performing fuzz testing on the `image::DynamicImage::resize_exact` function, I discovered an input that can cause a program crash. 

## Expected

The program should catch this crash-inducing condition and return an error instead of crashing.

## Actual behaviour

Crash with memory allocation failed.

## Reproduction steps

```rust
use base64::prelude::*;

fn main() {
    if let Ok(data) = BASE64_STANDARD.decode("ZmFyYmZlbGT/UMnJAAAAAAAAAAAAANj/Nw8A") {
        match image::load_from_memory(&data) {
            Ok(img) => {
                let _ = img.resize_exact(1, 1, image::imageops::FilterType::Nearest);
            },
            Err(e) => println!("ERR: {:?}", e),
        }
    }
}
```

When run, the following output is produced:

```shell
❯ cargo build --release && valgrind ./target/release/image_fuzz_target_test 
==422390== Memcheck, a memory error detector
==422390== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==422390== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==422390== Command: ./target/release/image_fuzz_target_test
==422390== 
memory allocation of 68535753872 bytes failed
==422390== 
==422390== Process terminating with default action of signal 6 (SIGABRT)
==422390==    at 0x4A0F9FC: __pthread_kill_implementation (pthread_kill.c:44)
==422390==    by 0x4A0F9FC: __pthread_kill_internal (pthread_kill.c:78)
==422390==    by 0x4A0F9FC: pthread_kill@@GLIBC_2.34 (pthread_kill.c:89)
==422390==    by 0x49BB475: raise (raise.c:26)
==422390==    by 0x49A17F2: abort (abort.c:79)
==422390==    by 0x2D66B9: std::sys::pal::unix::abort_internal (in /home/ubuntu/image_fuzz_target_test/target/release/image_fuzz_target_test)
==422390==    by 0x2D1429: std::process::abort (in /home/ubuntu/image_fuzz_target_test/target/release/image_fuzz_target_test)
==422390==    by 0x2D8320: std::alloc::rust_oom (in /home/ubuntu/image_fuzz_target_test/target/release/image_fuzz_target_test)
[...]

PR #2342, opened 2024-10-10
Optimize sampling for empty images

HeroicKatora:
This is a tip of an iceberg of better sampling, but the most critical case. Up-sampling in general may in the implementation allocate a larger temporary buffer than its input. Of course this makes little semantic sense here: after all, the actual information can not increase by this.

If one dimension increases while the other decreases the unfortunate consequence is that callers may somewhat reasonably expect a small buffer but internally will get a very large buffer. The approach of swapping sampling orders accordingly (first down, then up) might address the memory issue but is lossy.

So instead let's fix the most pressing issue: if no information was present in the input, nothing can be lost and we can pretend to perform everything by a very small intermediate image.

closes: #2340

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -221,6 +221,8 @@ pub(crate) fn box_kernel(_x: f32) -> f32 {
 // ```new_width``` is the desired width of the new image
 // ```filter``` is the filter to use for sampling.
 // ```image``` is not necessarily Rgba and the order of channels is passed through.
+//
+// Note: if an empty image is passed in, panics unless the image is truly empty.
 fn horizontal_sample<P, S>(
     image: &Rgba32FImage,
     new_width: u32,
@@ -231,6 +233,13 @@ where
     S: Primitive + 'static,
 {
     let (width, height) = image.dimensions();
+    // This is protection against a memory usage similar to #2340. See `vertical_sample`.
+    assert!(
+        // Checks the implication: (width == 0) -> (height == 0)
+        width != 0 || height == 0,
+        "Unexpected prior allocation size. This case should have been handled by the caller"
+    );
+
     let mut out = ImageBuffer::new(new_width, height);
     let mut ws = Vec::new();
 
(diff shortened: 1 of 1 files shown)


fintelia:
I think it would be cleaner to add a check at the start of resize to immediately return `ImageBuffer::new(nwidth, nheight)` if any new or old image dimensions are empty. `vertical_sample` is only called from `resize` and `blur`. In the blur case, `height==new_height` so there isn't much difference from this change. And in `resize`, it seems nicer to handle all combinations of input and output sizes being empty at once.

kornelski (approved):

src/imageops/sample.rs
```diff
@@ -463,13 +463,23 @@ pub fn interpolate_bilinear<P: Pixel>(
...
     S: Primitive + 'static,
 {
     let (width, height) = image.dimensions();
+
+    // This is protection against a regression in memory usage such as #2340. Since the strategy to
+    // deal with it depends on the caller it is a precondition of this function.
+    assert!(
+        height != 0 || width == 0,
+        "Unexpected prior allocation size. This case should have been handled by the caller"
+    );
```
okaneco, line 481: It could just be me, but I find this condition hard to understand as written without more explanation of the issue or why a width of 0 is ok. It's strange to see a 0-width not be a signal of some error beforehand.

Should a precondition assert for `horizontal_sample` also be added? It similarly allocates a buffer without checking dimensions.

src/imageops/sample.rs
```diff
@@ -1257,4 +1282,16 @@ mod tests {
         let result = resize(&image, 22, 22, FilterType::Lanczos3);
         assert!(result.into_raw().into_iter().any(|c| c != 0));
     }
+
+    #[test]
+    fn issue_2340() {
+        let empty = crate::GrayImage::from_raw(1 << 31, 0, vec![]).unwrap();
```
okaneco, line 1288: Should this test include a duplicated version with the dimensions swapped?
```rust
let empty = crate::GrayImage::from_raw(0, 1 << 31, vec![]).unwrap();
```

PR #2343, opened 2024-10-11
Impl Hash for FilterType

Keats:

--

I'm not sure if there is a reason it's not implemented but we could use that for Zola.


kornelski (approved):
I think it's fine to add it. Even if we added a variant that can't be hashed (like a `Box<dyn fn>`), we can always return some constant value for it.

PR #2345, opened 2024-10-13
test(fuzz): cover other image types

LeoDog896:

Adds fuzz targets for the QOI, OpenEXR, and Farbfeld image formats

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -67,5 +67,17 @@ name = "fuzzer_script_exr"
 path = "fuzzers/fuzzer_script_exr.rs"
 
 [[bin]]
+name = "fuzzer_script_qoi"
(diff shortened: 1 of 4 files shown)


fintelia (commented):
Have you checked whether these formats pass fuzzing?

We already have more fuzzing crashes than we have contributor bandwidth to fix, so I don't think wiring up more failing fuzzing harnesses will achieve much

LeoDog896:
> Have you checked whether these formats pass fuzzing?

I already checked that those formats pass fuzzing for around 20 minutes.

> We already have more fuzzing crashes than we have contributor bandwidth to fix, so I don't think wiring up more failing fuzzing harnesses will achieve much

That makes sense, though I doubt these added targets will throw out false positives.

Issue #2316, opened 2024-08-28
Should the next major release be 1.0?


fintelia:
...with versions 2.0, 3.0, etc. planned for subsequent releases?

---

The `image` crate has been doing breaking releases only every 1-2 years at this point. We've been holding off on doing a 1.0 release based on the hope that we'd eventually "finish" and then could stick with a single stable API indefinitely. Honestly, I don't think that's likely to happen for the foreseeable future. The public API surface is just too large, and there's too many edge cases. Not to mention that it would raise the stakes on new feature additions if we couldn't change course when flaws are discovered.

At the same time, I think our current strategy poorly communicates the level of stability we do provide. There's plenty of ecosystem crates that have gone from 1.0 to 2.0 and further, in less time than between our consecutive breaking releases.

So I propose we declare the next release to be 1.0. Based on historical pacing, it would probably come out in 2025. At that point, we'd want to make sure that release messaging was clear that subsequent major releases were planned. But I think that makes more sense than just being pre-1.0 forever.

Shnatsel:
It might be worth splitting the core types into a separate crate and making that 1.0, while keeping some other stuff under less strict stability guarantees. There is a long-standing proposal for such a split, see #793

Stability of these basic types is especially important now that `image` deliberately *excludes* support for certain features, such as libwebp encoding (#1984), JPEG XL decoding/encoding (#1979), or fast image resizing that uses unsafe code (`fast_image_resize` crate). This requires third-party crates to be tied to specific versions of `image` in their public API, and bump their semver anytime `image` is bumped. Having a foundational crate with basic types might result in less churn for the ecosystem.

fintelia:
Are you thinking that the crate with core types would be permanently 1.0 and not make any subsequent breaking changes? That was the original thinking in #793, but after 5+ years we aren't much closer to having final designs for `ImageDecoder`, `ImageBuffer`, `GenericImage`, etc. that we'd be comfortable committing to forever.  In fact, skimming the release notes from the last few breaking releases, probably the majority of changes have been to these types/traits.

fintelia:
The consensus seems to be in favor of this, so I'm going to go ahead and close

Issue #2028, opened 2023-10-14
Rayon should not be available as default feature


qarmin:
Currently I use image-rs library in 3 copies in my app.

One dependency is from my app, but two are from other libraries.

I want to disable rayon, because I use it for different calculations and I, but I cannot do this, because it is not currently possible to disable features in libraries that not allows to do this - e.g. imagepipe uses all default features without possibility to remove unnecessary features.

So to "fix" this, this feature should be disabled by default

Looks that ravif also uses rayon

fintelia:
Before we'd be able to consider whether rayon should be one of the default features, we first need to make sure it can be toggled by features at all. Right now, a couple of our dependencies use rayon without placing it behind a feature. If you're interested in changing this, the first step would be going through each place we include rayon from and making making a PR to have it be optional instead

oligamiq:
I too think rayon must not be the default.

oligamiq:
The only crate dependent on rayon is exrs, and all others are already managed by features. I have submitted a pull request to separate the rayon-related parts of exrs into a feature, so let's switch to not having rayon as a default feature.

https://github.com/johannesvollmer/exrs/pull/241

oligamiq:
> The only crate dependent on rayon is exrs, and all others are already managed by features. I have submitted a pull request to separate the rayon-related parts of exrs into a feature, so let's switch to not having rayon as a default feature.
> 
> [johannesvollmer/exrs#241](https://github.com/johannesvollmer/exrs/pull/241)

This pull request has been merged, but based on past examples, it will likely take some time before a new version is released.

197g:
`rayon` enables some integration with `par_enumerate` etc, so unfortunately it's also a major breaking change to turn it off. But I've noted it here: https://github.com/image-rs/image/issues/2245

fintelia:
I think `rayon` should remain a default feature. It doesn't just enable integration with `par_enumerate`, it also enables the use of threading in codecs that expose it. Most users are going to want us to take advantage of optimizations like that where possible. 

A default feature is precisely the mechanism that Cargo exposes for functionality that most users probably want, but which some folks will prefer to directly control. From the initial message, I think the right approach is actually to change the downstream libraries to also expose rayon as a feature (or just have them use `default-features = false` and never enable rayon).

197g:
I don't feel strongly one way or another, I do side with you on that being a problem with those crates that they should fix. So if you will stand on keeping it default, closing the issue as wontfix is okay with me.

[...]

PR #2347, opened 2024-10-15
Fill in changelog and bump version for 0.25.3

Shnatsel:
Let's get the work we did on image orientation into the hands of users!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,25 @@
 # Release Notes
 
 ## Known issues
-- Many decoders will panic on malicous input. In most cases, this is caused by
-  not enforcing memory limits, though other panics have been seen from fuzzing.
+- Many decoders will panic on malicious input.
 - The color space information of pixels is not clearly communicated.
 
 ## Changes
+
+### Version 0.25.3
+
+Features:
+ - Added support for orientation metadata, so that e.g. smartphone camera images could be displayed correctly:
+   - Added `ImageDecoder::orientation()` and implemented orientation metadata extraction for JPEG, WebP and TIFF formats
+   - Added `DynamicImage::apply_orientation()` to apply the orientation to an image
+ - Added support for extracting Exif metadata from images via `ImageDecoder::exif_metadata()`, and implemented it for JPEG and WebP formats
+ - Added `ImageEncoder::set_icc_profile()` and implemented it for WebP format. Pull requests with implementations for other formats are welcome.
+ - Added `DynamicImage::fast_blur()` for a linear-time approximation of Gaussian blur, which is much faster at larger blur radii
+
+Bug fixes:
+ - Fixed some APNG images being decoded incorrectly
+ - Fixed the iterator over animated WebP frames to return `None` instead of an error when the end of the animation is reached
+ - Disabling the `rayon` feature now also disables multi-threading in `ravif` and actually removes `rayon` from the dependency tree
 
 ### Version 0.25.2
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,12 +1,26 @@
...
+   - Added `ImageDecoder::orientation()` and implemented orientation metadata extraction for JPEG, WebP and TIFF formats
+   - Added `DynamicImage::apply_orientation()` to apply the orientation to an image
+ - Added support for extracting Exif metadata from images via `ImageDecoder::exif_metadata()`, and implemented it for JPEG and WebP formats
+ - Added `ImageEncoder::set_icc_profile()` and implemented it for WebP format. Pull requests with implementations for other formats are welcome.
+ - Added `DynamicImage::fast_blur()` for a linear-time approximation of Gaussian blur, which is much faster at larger blur radii
+
+Bug fixes:
+ - Fixed some APNG images being decoded incorrectly
+ - Fixed the iterator over animated WebP frames to return `None` instead of an error when the end of the animation is reached
+ - Disabling the `rayon` feature now also disables multi-threading in `ravif` and actually removes `rayon` from the dependency tree
```
fintelia, line 22: The "actually removes rayon from the dependency tree" part isn't true yet. https://github.com/johannesvollmer/exrs/pull/241 also needs to land first (tracked by https://github.com/image-rs/image/issues/2028)

fintelia: And actually, don't we need to change our Cargo.toml to enable ravif's threading feature when our rayon feature _is_ enabled?

fintelia:
The one thing I've been second guessing is whether `Orientation` should be a top-level type. I've been wanting to reduce the number of top level types, and we probably will be adding a `metadata` module to group all the other metadata related types we add, so I think we it might be good to place `Orientation` there to start with

CHANGES.md
Shnatsel: We do, and in fact the latest `ravif` with `default-features=false`  will not enable rayon by default anymore, so this release is needed just to preserve the previous default behavior when using the latest `ravif`.

I've opened https://github.com/image-rs/image/pull/2348 for it

Shnatsel:
I've opened https://github.com/image-rs/image/pull/2349 to relocate `Orientation`

Shnatsel:
I've rewritten the changelog entry for rayon. I don't think we should wait on the exr change to ship this, since it appears to be a breaking change, the maintainer is busy, and so there is no clear timeline for it.

HeroicKatora (approved):

HeroicKatora:
@fintelia If you agree with preparing `0.26` next, feel free to merge #2344 after the merge of the tagged release commit.

fintelia:
@HeroicKatora Sorry I've been meaning to respond in that other thread. [I wrote up some thoughts here](https://github.com/image-rs/image/issues/2316), but rather than working on 0.26, I actually think that we should old off starting a major release for a little bit and then make a _1.0_ release.

fintelia (approved):
Might be worth adding that this release brings substantially improved lossless WebP decoding performance. I measured it being between 2x and 2.5x faster!

(We usually don't mention codec improvements because they tend not to be tied to image releases, but these specific optimizations were bundled with the `image-webp` 0.2.0 release)

Shnatsel:
The changelog still mentions #2348 which hasn't landed yet. So please don't publish a release just yet!

I'm hoping we could merge https://github.com/kornelski/cavif-rs/pull/85 soon and fix the whole issue around MSRV. If it's not merged in a couple of days, we can pin an older `ravif` and then ship.

Shnatsel:
Ah, nevermind, too late :sweat_smile: 

I've opened https://github.com/image-rs/image/pull/2352 to amend the changelog.

PR #2349, opened 2024-10-15
Move Orientation to metadata module

Shnatsel:

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -941,7 +941,7 @@ impl DynamicImage {
     ///
     /// ```
     /// # fn only_check_if_this_compiles() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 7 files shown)


src/dynimage.rs
```diff
@@ -940,7 +941,7 @@ impl DynamicImage {
     ///
     /// ```
     /// # fn only_check_if_this_compiles() -> Result<(), Box<dyn std::error::Error>> {
-    /// use image::{Orientation, DynamicImage, ImageReader, ImageDecoder};
+    /// use image::{DynamicImage, ImageReader, ImageDecoder, metadata::Orientation};
```
fintelia, line 944: I think we can actually just remove the import of `Orientation` here, since the example just calls `decoder.orientation()` to construct one

Shnatsel: Good call, it's a shame the compiler doesn't surface warnings from doctests

fintelia (approved):

PR #2352, opened 2024-10-16
Drop incorrect changelog entry

Shnatsel:
#2348 didn't make it, so don't mention it in the changelog


fintelia:
@HeroicKatora I think you published 0.25.3 with commit fe94eabb7f7491b9ba9378ea5ece2f8884c30c65 rather than the merged commit d52a194e5c3fa304143cc71d85d551e88fd211d9 which is the one on the main branch, which resulted in #2349 not getting included.

Issue #2363, opened 2024-10-25
Policy for uncommon formats implemented via third-party crates


Shnatsel:
### The problem

There is a number of formats that are implemented as third-party crates, but are not supported by `image`:

JPEG XL: https://crates.io/crates/jxl-oxide
RAW: https://crates.io/crates/rawloader or https://crates.io/crates/quickraw
PSD subset: https://crates.io/crates/zune-psd or https://crates.io/crates/psd
ICNS: https://crates.io/crates/icns
JPEG 2000: https://crates.io/crates/openjp2 (not sure if the port is complete enough just yet)

None of these formats have specifications available publicly and for free. As per #1979, right now they are not supported by `image` at all.

This leads to each API users that needs them reimplementing the integrations into `image`.  [oculante](https://github.com/woelper/oculante) has already done this, and I'm going to have to do this in `wondermagick` too. 

### Possible solutions

Adding `image` integrations to those crates (like it's done for libwebp) is a step in the right direction, but doesn't really solve this because the API is still quite awkward: you cannot just use `reader.with_guessed_format().decode()?`, the API user has to explicitly list every format and also deal with magic numbers for them. 
Also, when `image` ships a new semver-incompatible version, every integration would have to bump their semver as well, and it could take a long while until all the format integrations are upgraded.

A solution that seems promising to me is a tiered format support, where we specifically advertise third-party crates listed above as Tier 2. We clearly state that we do not accept bug reports about them and that any issues should go to the third-party crate's issue tracker. These formats would also be disabled by default, with explicit opt-in required to enable them.
This avoids the issues with the other option, but still adds some (hopefully small) additional maintenance burden for the in-tree integration of the third-party crate.

Thoughts? Any other good options I've missed?

fintelia:
Agreed that having tiers is a helpful frame for different levels of format support. I'd probably break up formats into more categories than just real vs. hobby though. 

* **Major formats** -- Formats like PNG, JPEG, GIF, TIFF, etc.
* **Specialized and obscure formats** -- Less common than the major formats but used because they outperform the major formats on some dimension or because they're the chosen standard in some field.
* **Emerging formats** -- Recently created formats likely to land in one of the prior categories. Frequently accompanied by a large marketing push to pressure software makers to include the format so they'll gain traction.
[...]

PR #2364, opened 2024-10-25
Add PCX decoder support

lupine:
This PR adds support for decoding PCX images. This is a raster format introduced in 1985. Further details on Wikipedia: https://en.wikipedia.org/wiki/PCX; it saw fairly extensive use up to the end of the 20th century, but is definitely obsolete now. However, there is still extant data using the format.

My hobby project is a game engine recreation that happens to need to load a range of assets in the PCX format; using the image crate through bevy to (attempt to) load PCX files is how I ended up writing this PR.

Since it's an obsolete format, I've opted to only implement the decoder side, but the pcx crate does have an encoder as well if we wanted to wire that up. Perhaps "historic format" sounds better!


Cargo.toml
```diff
@@ -47,6 +47,7 @@ exr = { version = "1.5.0", optional = true }
 gif = { version = "0.13", optional = true }
 image-webp = { version = "0.2.0", optional = true }
 mp4parse = { version = "0.17.0", optional = true }
+pcx = { version = "0.2.3", optional = true }
```
lupine, line 50: This dependency is small, and licensed WTFPL. I got the test PCX images from their repository ( https://github.com/kryptan/pcx/tree/master/test-data 

The test suite in this project is rendering them as they appear in that repository, but it's worth noting that there are bugs in the decoder's implementation, particularly visible in cga_bw and cga_tst1.

src/codecs/pcx.rs
```diff
@@ -0,0 +1,157 @@
...
+                        .map_err(ImageError::from_pcx_decode)?;
+                }
+            }
+
+            // We need to convert from the palette colours to RGB values inline,
+            // but the pcx crate can't give us the palette first. Work around it
+            // by taking the paletted image into a buffer, then converting it to
+            // RGB8 after.
+            Some(palette_length) => {
+                let mut pal_buf: Vec<u8> = iter::repeat(0).take(height * width).collect();
```
lupine, line 123: This is annoying, but the borrow checker gets upset if I call `read_palette` first.  Digging into the pcx crate, it has a comment about going to great lengths to avoid seeking to the palette :/.

src/image_reader/free_functions.rs
```diff
@@ -148,6 +148,7 @@ static MAGIC_BYTES: [(&[u8], ImageFormat); 23] = [
     (b"\0\0\0\x1cftypavif", ImageFormat::Avif),
     (&[0x76, 0x2f, 0x31, 0x01], ImageFormat::OpenExr), // = &exr::meta::magic_number::BYTES
     (b"qoif", ImageFormat::Qoi),
+    (&[0x0A], ImageFormat::Pcx),
```
lupine, line 151: Only the first byte is truly a magic byte. Some decoders take the first four bytes, but this can lead to false negatives.

lupine (commented):
Just some initial self-review.

deny.toml
```diff
@@ -22,6 +22,7 @@ allow = [
     "MIT-0",
     "MPL-2.0",
     "Unicode-DFS-2016",
+    "WTFPL",
```
lupine, line 25: Crate pcx is licensed WTFPL: https://en.wikipedia.org/wiki/WTFPL

Certainly for me, it's an acceptable license, so I've changed it here, but this would need project approval, of course.

If there's resistance to adding this dependency, I could look at a from-scratch PCX implementation, which would also give an opportunity to fix the decoder bugs and memory inefficiency around paletted images, perhaps.

tests/reference/pcx/images/cga_bw.pcx.620c687c.png
lupine, line 1: pcx crate bug: the background should be black and the text white. Compare with what is shown by, e.g., `display cga_bw.pcx`

tests/reference/pcx/images/cga_tst1.pcx.2a36d106.png
lupine, line 1: More PCX crate bug ^^

tests/reference/pcx/images/gmarbles.pcx.1359bee5.png
lupine, line 1: I know this one *looks* off, but it's as `display` renders it :shrug:

src/image_reader/free_functions.rs
HeroicKatora: Great self-review. I think we want to weigh the accuracy of magic bytes such that they don't preclude alternatives. For that goal aligning with imagemagick as [they did here](https://github.com/JuliaIO/FileIO.jl/pull/114) is probably a good default. The only better alternative I could see is the functional approach of letting some patterns filter by evaluated code.

lupine: Sure thing; I've just pushed https://github.com/image-rs/image/pull/2364/commits/f163f1449251f0049373b52b05c07974ac5ee188 which changes to checking for those two byte sequences.

Cargo.toml
```diff
@@ -67,7 +68,7 @@ criterion = "0.5.0"
 default = ["rayon", "default-formats"]
 
 # Format features
-default-formats = ["avif", "bmp", "dds", "exr", "ff", "gif", "hdr", "ico", "jpeg", "png", "pnm", "qoi", "tga", "tiff", "webp"]
+default-formats = ["avif", "bmp", "dds", "exr", "ff", "gif", "hdr", "ico", "jpeg", "pcx", "png", "pnm", "qoi", "tga", "tiff", "webp"]
```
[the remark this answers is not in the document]
kornelski: I would definitely prefer this to be off by default. Especially that Cargo's feature merging means it's sometimes pretty difficult to turn off the default features.

This is a truly obsolete format, so for nearly every user of this library it will be nothing but overhead that adds to the compilation time and bloats executables.

fintelia:
I wrote up some thoughts on format support in #2367.

High level, I don't think we should add PCX, but do want to think about making it easier for users to deal with formats not directly supported by this crate.

Cargo.toml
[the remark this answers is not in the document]
lupine: :thumbsup: - https://github.com/image-rs/image/pull/2364/commits/84cd6041098cf74306178f09b481469aea9793e7 disables it by default.

lupine:
@fintelia just so I know, does that decide the fate of this PR or is it still up in the air? I just want bevy to load my files; if `image` doesn't want to support a particular format, that's entirely fair, and I can make it happen via a custom asset loader over there (I think).

FWIW, I was planning on adding `ANI` support as well ^^. In my head, an image crate naturally wants to support all the image formats, but I appreciate the reality is more complicated than that.

A tidy alternative would be a way to register codecs at build time (runtime probably isn't necessary), as go's `image` stdlib package does; I don't quite know offhand how that would look in Rust. Right now, this PR would need to be followed up with a PR to bevy to wire up the new format, which really isn't ideal even for the case of popular new formats.

fintelia:
Yes. If @HeroicKatora feels differently we can discuss further, but otherwise we can go ahead and close this PR. I'll add that from skimming the code quality does look quite good, and I've also just opened #2369 to encourage having discussions _prior_ to doing all the implementation effort.

(As a side note, supporting all image formats is impossible because there's no official list: anybody can invent their own format and people do so pretty often.)

sorairolake:
I think it would be good to implement [`image::ImageDecoder`](https://docs.rs/image/0.25.4/image/trait.ImageDecoder.html) for [`pcx::Reader`](https://docs.rs/pcx/0.2.3/pcx/struct.Reader.html). I think this will help you achieve your goal to some extent. The problem is that `pcx` crate is not being actively developed.

HeroicKatora:
Following the thoughts laid out inof #2367

src/codecs/pcx.rs
HeroicKatora: I'm up for doing something (slightly?) inefficient in favor of doing something right. If there's sufficient problems to motivate a fork of the underlying crate, sure. That shouldn't hold up this PR imo.

deny.toml
HeroicKatora: I'm fine with it, I guess? I've chosen it before for projects. That said, I think we should very explicitly document these in both locations, due to corporate users that do not intent to introduce such dependencies: a) at the feature location that enabling it introduces an unusual `WTFPL` license b) at the allow list that it comes from a feature. If we could have deny check (b) automatically in some way, even better.

Maybe allow _just_ that crate? [As in this example here](https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html#example) — an old `adler32` crate.

HeroicKatora:
First of all, I do agree with code quality :+1.

Following the thoughts raised out in #2367, and in particular I like the tiered view https://github.com/image-rs/image/issues/2363#issuecomment-2439982410 , I think we **should** support this format in some capacity. And in particular, by having the trait implemented but not dispatching onto the file without explicit opt-in (as written in the linked comment, I'd liken it to tier 3). I tend to agree that some `image`-level configuration mechanism will be necessary at some point, but let's design that in an actual issue (#2368 seems appropriate). I wouldn't want to hold up the PR over that, we have enough leverage to fine-tune the integration when it lands and can use having validation targets of all tiers anyways, right?

It would be most beneficial to users, that's the crucial part. My argument would be it doesn't matter so much to me if a format itself is fringe, as rather if `image` would be an appropriate tool to use for the people that need that format. That's purely a maintenance question _for this code_. In terms of lines of code, already _including_ a test suite_ (!), it doesn't seem to put much strain on `image`. Rather on the underlying implementation but really that doesn't need to be our immediate problem?

So :+1: me with the licensing comment resolved.

deny.toml
lupine: Thanks. Is https://github.com/image-rs/image/pull/2364/commits/25cede1450143eb797066d56eca3e66e4d16c249 sufficient, do you think, or should it be called out in the README, or in more detail?

src/codecs/pcx.rs
lupine: I'm happy enough with it, personally - the memory use is bounded at a third of the memory we're checking against limits.

deny.toml
HeroicKatora: LGTM.

HeroicKatora (approved):

fintelia:
Just to double check, what is the source of the test images and what license are they under?

lupine:
@fintelia they're from the pcx crate - https://github.com/image-rs/image/pull/2364#discussion_r1817436912 -> https://github.com/kryptan/pcx/tree/master/test-data . They didn't originate there though. https://github.com/drewnoakes/metadata-extractor-images/tree/main/pcx has them, for instance. 

https://netghost.narod.ru/gff/sample/images/pcx/index.htm has a 1996 imprint and has marbles.pcx and gmarbles.pcx  :thinking:

fintelia:
Yeah, "Copyright © 1996 O'Reilly & Associates, Inc. All Rights Reserved." isn't exactly a suitable license statement... If that's the original source of those images they need to be replaced

lupine:
That's fair. I wrongly assumed that since they were in a WTFPL crate, it'd be fine to reuse them! I'll put together a PR that replaces them.

I tracked down the CGA PCX files to https://samples.ffmpeg.org/image-samples/pcx/cga/cga-pcx.txt - they also appear in the MPlayer samples directory. No explicit copyright grant in that file; ffmpeg itself is lgpl, mplayer gpl. I'll just yoink those out as well.

All the obscure parts of the format are still tested in the pcx crate; I think we just need a single image that allows us to verify that we're up to that crate correctly.

Oh, I also found a "proper" use of PCX in the wild - according to https://convertimage.net/convert-a-picture/to-pcx/ :

> While PCs do not generally use PCX anymore, the worldwide airline industry does. In fact, PCX is still the only valid format for logos and graphics printed on baggage tags and tickets, as defined in the AEA (Association of European Airlines) ATB Technical Specs.

A brief look tells me this was true in 2007, at least.

Issue #2368, opened 2024-10-26
Ergonomics for external image formats


fintelia:
At the moment, it is pretty difficult to work with formats that aren't implemented directly in this crate. Other crates largely haven't chosen to implement our `ImageDecoder` trait, and even if they did, there isn't a way to use them when working with `ImageReader` or our file format detection logic. This has created a lot of pressure for us to expand the list of supported formats.

The question is whether we could add some form of hooks support so that downstream users could inject their own format decoders to use in addition to the built-in ones. It is a bit messy, but perhaps something like this could work:

```rust
trait FormatHook<R> {
    fn file_extensions(&self) -> &[&str];
    fn magic_bytes_match(&self, file_prefix: &[u8]) -> bool;
    fn make_decoder(&self, reader: R) -> Box<dyn ImageDecoder>;
}

impl ImageReader<R> {
    fn with_hooks(&mut self, hooks: &[Box<dyn FormatHook<R>>]) {}
    fn open_with_hooks(path: &Path, hooks: &[Box<dyn FormatHook<R>>]) -> ImageResult<Self> {}
}
```

kornelski:
Maybe the hooks could even have a global registry. Otherwise you still need to coordinate with all the deps using the `image` crate on how to collect and pass down the hooks. If there's a global registry, you can just register a format in your `main()`, and then have the format behave as if it was built-in.

kornelski:
`magic_bytes_match` could cause problems if some implementations weren't good at sniffing them. Decoding first that match would make sense, but could require users to carefully pick the order in which the formats are registered.

What if the `image` crate was still in charge of image format sniffing, and maintain a central list of formats? Contributing and maintaining magic bytes sniffing is less of a burden than maintaining full codecs, so for sniffing `image` crate could be maximally inclusive.

fintelia:
Including all the format selection logic into `image` might be a good option. We'd need some inclusion criteria, but something like only accepting formats supported by ImageMagick would probably be sufficient.

I'm not too worried about file extensions, but if we went this direction we'd need to improve our magic byte format sniffing. For instance, at the moment we assume that any file that starts with RIFF is WebP, even though there's a bunch of other file formats that also use RIFF containers...

[...]

PR #2372, opened 2024-10-28
Add decoding hooks

fintelia:
This is an attempt at implementing the format decoding hooks registry described in #2368. It works out to be less code than I was expecting.

At the moment the hooks aren't especially useful because the only image formats that hooks can be registered for already have built-in decoders, but that could be changed in followup PRs


src/hooks.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+impl Seek for BoxReadSeek<'_> {
+    fn seek(&mut self, pos: std::io::SeekFrom) -> std::io::Result<u64> {
+        self.0.seek(pos)
+    }
+}
+
+/// A function to produce an `ImageDecoder` for a given image format.
+pub type DecodingHook = Box<
+    dyn for<'a> Fn(BufReader<BoxReadSeek<'a>>) -> ImageResult<Box<dyn ImageDecoder + 'a>>
```
fintelia, line 33: What's going on here is a bit tricky. We need a type-erased function that produces `ImageDecoder`s given a reader.  The reader itself needs to be type-erased but we also don't want to have a virtual function indirection for every `read` call. Thus we wrap the reader in a `BufReader`

src/hooks.rs
```diff
@@ -0,0 +1,45 @@
...
+>;
+
+/// Register a new decoding hook or replace an existing one.
+pub fn register_decoding_hook(format: ImageFormat, hook: DecodingHook) {
+    let mut hooks = DECODING_HOOKS.write().unwrap();
+    if hooks.is_none() {
+        *hooks = Some(HashMap::new());
+    }
+    hooks.as_mut().unwrap().insert(format, hook);
+}
```
HeroicKatora, line 45: I like the internal mechanism, but am not completely sure about the interface. It seems primarily designed around the implementation instead of user expectations. For instance, if we consider the case of third-party support for a decoder that isn't enabled by feature switches then we'd expect both decoder and encoder in a single line.

It's possible to register a hook that's ignored, and that ignore can happen by a feature flag enabled via a sibling create. This poses the question, should there be a mechanism for discovering the utilized features and should the internal translation pre-register itself in the map to provide a more uniform view?

kornelski: Registering both encoding and decoding in one go can always be added later as a dedicated function. A 3rd party codec crate can always provide their method to register themselves, like `my_codec_crate::register()`.

But it could be common to want *different* implementations for decoding and encoding. For decoding you may want a decoder that is safe and/or compatible, but for encoding you may prefer either a fast one or heavily-compressing one (safety is usually less of an issue when encoding).

src/hooks.rs
```diff
@@ -0,0 +1,45 @@
...
+
+/// A function to produce an `ImageDecoder` for a given image format.
+pub type DecodingHook = Box<
+    dyn for<'a> Fn(BufReader<BoxReadSeek<'a>>) -> ImageResult<Box<dyn ImageDecoder + 'a>>
+        + Send
+        + Sync,
+>;
+
+/// Register a new decoding hook or replace an existing one.
+pub fn register_decoding_hook(format: ImageFormat, hook: DecodingHook) {
```
kornelski, line 39: Maybe replacing shouldn't be allowed?

The application controls `main`, so it can always be the first to register its preferred codecs. But it could be annoying if some dependency later "helpfully" replaced that with its own preferred codecs. 

Especially if there's no way to check if there's an existing hook already, crates that need a certain format supported won't have much choice except to always register.

src/hooks.rs
kornelski: Maybe make it a concrete type that implements `BufRead` instead of exposing a `BufReader` explicitly? A custom type can always be swapped for something else. 

`BoxReadSeek` is created from `dyn BufRead`, so it's a bit unfortunate that there's double buffering going on. With some self-referential hacks it should be possible to provide a fast read implementation that reads off a slice returned by a `dyn BufRead`.

fintelia: I could add a new `struct GenericReader(BufReader<BoxReadSeek<'a>>)` type so the BufReader isn't part of the public signature. But the important thing here isn't actually the `BufRead` implementation. As you note, the inner reader will already have one. Rather, the thing we need is a `read` implementation that can be inlined everywhere the decoder calls it. It is fine if it involves a slowpath that does a virtual function call into a `Box<dyn Read>`, but decoding an image might involve millions of `read` calls, so you really don't want to be doing that every time.

A fun quirk of `BufReader` is that there _isn't_ actually double buffering going on if both have the same buffer size. That's because the inner `BufReader` [will never fill its internal buffer when fulfilling reads >= to its capacity](https://github.com/rust-lang/rust/blob/3f1be1ec7ec3d8e80beb381ee82164a0aa3ca777/library/std/src/io/buffered/bufreader.rs#L338), but the outer `BufReader` will always make exactly capacity sized reads.

fintelia: Using the buffer from the inner `dyn BufRead` would require a self referential type that could hold by the inner reader alongside a reference to its internal buffer. Might be possible with `Box::pin` and some unsafe?

src/hooks.rs
fintelia: Yeah, I was kind of imagining an external crate would expose its own `register()` function that registered all the encoders/decoders it had. That could be only a single decoder, or several encoders and decoders all defined in the same crate 

I didn't add encoding support to this PR because at the moment `Box<dyn ImageEncoder>` isn't actually usable for encoding images (you can't call `write_image` because that requires `Self: Sized`).

src/hooks.rs
fintelia: Switched this to blocking replacement. Also added a `decoding_hook_registered` to check whether a hook has been registered for a given format

src/hooks.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn stream_position(&mut self) -> std::io::Result<u64> {
+        self.0.stream_position()
+    }
+
+    // TODO: Add `seek_relative` once MSRV is at least 1.80.0
+}
+
+/// A function to produce an `ImageDecoder` for a given image format.
+pub type DecodingHook =
+    Box<dyn for<'a> Fn(GenericReader<'a>) -> ImageResult<Box<dyn ImageDecoder + 'a>> + Send + Sync>;
```
HeroicKatora, line 67: Should this be `Arc<..>` instead of box? The `Fn` type implies shared ownership and in this manner we could have the ability to `clone` parts or the whole of the configured hooks chain.

fintelia: I don't think the public interface needs to have `Arc`. The hook itself has a `Send + Sync` bound so we can always convert `Box` -> `Arc` internally if desired.

Shnatsel:
We can kick the tires of this API by splitting PCX into a separate crate and using these hooks to wire it up. 

Also, [jxl-oxide](https://github.com/tirr-c/jxl-oxide) has shipped `image` integration just days ago, and it would be a good test bed for a more modern format.

fintelia:
@HeroicKatora how would you feel about moving forward with this PR and then spinning out a single `image-extras` crate that uses hooks to support additional non-default format bindings, like the PCX decoder that was just written?

Shnatsel:
That sounds like a good way to dogfood the hooks.

IMHO there's also value in segregating the formats with weird licenses. PCX is under WTFPL, we probably don't want to have that as a dependency in the main crate. And all the RAW crates are under LGPL, it would be nice to signpost that as well. There's precedent in GStreamer, where decoders with weird licenses are segregated as gstreamer-plugins-ugly.

Although the PCX author is still around, let me ask them to relicense and see what comes of it.

Shnatsel:
`pcx` crate has been [relicensed](https://github.com/kryptan/pcx/commit/7ee82e83f8a7ce10cbba70407be40cf76aa6b14c) as `WTFPL OR MIT OR Apache 2.0`, so the license should no longer pose any issues.

HeroicKatora:
The biggest concern being solved with an extra crate is the licensing / static dependency tree situation. However, I can also see users expecting formats to 'just be available' when depending on crates,  similar to the discover mechanism of the traditional c libraries. We can't introduce life-before-main in Rust however.

This combination not possible right now when moving to a separate crate, afaik. To automatically register the decoders from `image-extra`, I think we would need `image` to depend on `image-extra` (optionally). However first of all that is cyclic, so some common `image-core` containing the actual registry is required. Secondly, upstream needs to provide appropriate weak feature selections as appropriate `image/image-extra` must be explicitly enabled for it to be visible in `cfg`.

I think the hooks can address the needs for users _to substitute_ some formats, but they do not appropriately solve our internal needs. And in this implementation, the substitution does not work in all situations as intended as the feature selection is always given priority.

Shnatsel:
On the plus side, splitting out `image-extra` would also let us break semver on those decoders without impacting `image`. I doubt that's going to be relevant to PCX though.

But now I think the licensing argument is not as strong as I made it out to be. AFAIU bundling a bunch of formats into -extra isn't really an improvement from the licensing standpoint, and the licensing motivation for PCX is gone regardless.

HeroicKatora:
Addressing the use-case of substituting the builtin decoder with a personal one is quite valid. I'd be glad to approve on the PR and registry design if the registered hooks ran with _higher_ priority than the builtin format dispatch. The trait boxing of `Read` etc. is neatly wrapped here.

fintelia:
My top-level goal is finding a way to spend less time thinking about, reviewing, and supporting obscure formats.

IMO most of the effort around obscure formats is preparing/preventing them from causing issues for the rest of the crate. Any PR that touches the crate's Cargo.toml or lib.rs might accidentally (or maliciously) break them. Test images bloat the size of the repository. If some dependency yanks their crate, we may have to fork and publish our own copy. Some codec changes like swapping the backend decoder implementation might require a semver break.

Separating out less common formats into a different crate provides a much nicer boundary to be more lax about reviews and concerns like those. If a codec breaks, we can drop support in a new major version. If CI is broken, it doesn't block PRs to the main image crate. If a decoder uses unsafe code, it doesn't prevent the main crate from eventually having `#![forbid(unsafe_code)]`

kornelski (approved):

fintelia:
I created https://github.com/image-rs/image-extras to test out the usage of this API.

In my view, there isn't much point in having hooks if they're going going to be for overriding the handling of built-in formats. That's because the [`ImageReader::make_decoder`](https://github.com/image-rs/image/blob/f95d04058a9981bee4989643f1ee0430694e417e/src/image_reader/image_reader_type.rs#L139) method that actually does the dispatch is quite short and could be replicated by user code. I'm also not aware of other Rust crates that allow overriding behavior for _other_ API users.

Thus the question: __Are we OK with merging this and then going ahead with publishing `image-extras`* to provide first-party support for less common formats?__

<sup>*We've been saying both `image-extra` and `image-extras` in this thread. I'm fine with either, but if you have strong feelings, now is the time to say so! </sup>

kornelski:
One of the motivating use cases is Bevy where devs want to load custom formats through the standard asset loader. 

https://docs.rs/bevy/latest/bevy/image/struct.ImageLoader.html

src/hooks.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.0.read_to_end(buf)
+    }
+    fn read_to_string(&mut self, buf: &mut String) -> std::io::Result<usize> {
+        self.0.read_to_string(buf)
+    }
+    fn read_exact(&mut self, buf: &mut [u8]) -> std::io::Result<()> {
+        self.0.read_exact(buf)
+    }
+}
+impl BufRead for GenericReader<'_> {
```
mstoeckl, line 37: Most of this change looks good to me, but it appears GenericReader implements [`BufRead`](https://doc.rust-lang.org/std/io/trait.BufRead.html) without requiring the input is buffered; anyone using `GenericReader(Box::new(File::new()))` might get surprised by many tiny `read()` syscalls. Should `GenericReader` require `BufRead + Seek` instead?

fintelia: The problem is that `read` calls against a `Box<dyn ReadSeek>` will be slow due to the virtual function indirection even if the underlying reader is a `Cursor`. Wrapping it again in a `BufReader` isn't ideal, but enables the fast-path for read operations to be inlined into callers.

This code is also structured so that the internal details of GenericReader can change without impacting the public API.

mstoeckl: > The problem is that read calls against a Box<dyn ReadSeek> will be slow due to the virtual function indirection even if the underlying reader is a Cursor. Wrapping it again in a BufReader isn't ideal, but enables the fast-path for read operations to be inlined into callers.

OK, that makes sense; I don't see a better solution either.

(This comment and my other one can be resolved. Looking over earlier resolved comments, I see you had explained this before; I apologize for reraising the issue.)

fintelia:
Haven't had much time to work on decoding hooks, but wanted to at least update where things stand...

While trying to use this API from `image-extras` I discovered that the requirement for the `ImageFormat` to be defined in this crate was really inconvenient. Adding new formats there required PRs across both repositories and waiting for new releases of `image` so they could be referenced there. Instead we'll probably want externally implemented formats to be registered by file extension and/or their magic bytes.

fintelia:
Ok, I've switched over the hooks to operate on file extensions rather than `ImageFormat`. There's now format detection hooks which take magic bytes and an optional mask, and map to file extension. I also deprecated the `ImageFormat::Pcx` type so that `image-extras` can implement it (built-in types are now never overridable).

197g:
I think it's great that hooks can be specified in terms of patterns now. But that is somehow at odds with use cases and evolvability.

> That's because the [ImageReader::make_decoder](https://github.com/image-rs/image/blob/f95d04058a9981bee4989643f1ee0430694e417e/src/image_reader/image_reader_type.rs#L139) method that actually does the dispatch is quite short and could be replicated by user code.
> […]
> built-in types are now never overridable.

I really don't see the use case addressed here. If user code is being ran, one could as well construct the decoder by-hand directly. And with that change not even the provided format detection mechanism provides much of a help. Running of your code depends on a feature flag __not__ being unified through third-party sibling dependencies. That seems … odd?

Meanwhile [that other use-case]( https://github.com/image-rs/image/pull/2372#issuecomment-2633870083) is much clearer:

> One of the motivating use cases is Bevy where devs want to load custom formats through the standard asset loader.

The _standard_ path is something you can not currently customize; but Bevy wants to customize. In another use case we actually had they wanted to load lossless jpeg, a functionality we had to cut in `image-jpeg -> zune_jpeg` (well, and for it being buggy). Again this can only be customized if the hooks run _before_ the builtins.

Also consider what happens if an extension crate (not `image-extra`) defines a hook for a format, and later we adopt that format, or just its detection, into `image`. Then the logic will switch from `Format::Extension` to `Format::Builtin` and the whole decoding changes in a potentially minor version bump. (Same situation if a sibling enables a new format feature flag). And there is _no_ remedy that I see they could take apart from downgrading.

I think in total, hooks __must__ override the default behavior. That's the only consistent way to address both the uses cases and the failure modes. (Of course the internal dispatch still requires a newtype like `Format` instead of happening on `ImageFormat` directly, file extension seems like a documentable way of handling that). Since siblings can't just run code, registration of hooks will retain its behavior in upgrade/feature flag scenarios and the same decoders will run. You just sometimes get less errors on previously unrecognized formats if they got built into `image`/`image-extra` for free.

197g (approved):
LGTM. We still maintain a split between format detections internal and external but since they are explicit we have room to refine everything in the future (e.g. adding a hook against an `ImageFormat`). This covers everything for `image-extras` and the jpeg example by registering a separate jpeg detection hook, which should be good for now.

PR #2373, opened 2024-10-29
Avif YUV decoder, drop dcv, high bit depth

### LINKED ISSUE #1504: Better avif support
I would like to be able to decode 10-bit avif images. It appears that cavif-rs outputs 10-bit avif images that cannot be decoded using image-rs.

### LINKED ISSUE #1647: Avif and webp recognition is bugged

### LINKED ISSUE #1930: avif decoder error
this avif [1.avif.zip](https://github.com/image-rs/image/files/11485598/1.avif.zip)

image::guess_format is UnsupportedError

with file header[3] == 24

[src/lib.rs:46] &format = Avif
[src/lib.rs:53] &bin[..12] = [
    0,
    0,
    0,
    24,
    102,
    116,
    121,
    112,
    97,
    118,
    105,
    102,
]
[src/lib.rs:54] b"\0\0\0\x1cftypavif" = [
    0,
    0,
    0,
    28,
    102,
    116,
    121,
    112,
    97,
    118,
    105,
    102,
]

avif decoder error 
image

[src/lib.rs:55] image::guess_format(bin) = Err(
    Unsupported(
        UnsupportedError {
            format: Unknown,
            kind: Format(
                Unknown,
            ),
        },
    ),
)
  ✘ [fail]: img → jpg Rejected promise returned by test
  ─

  img → jpg

  Rejected promise returned by test. Reason:

  Error {
    code: 'GenericFailure',
    message: 'The image format could not be determined',
  }

### LINKED ISSUE #2232: 10-bit avif support
Please support 10-bit avif. 10-bit is the standard for AV1, because it offers significant BD-rate gain over 8-bit at the expense of marginally higher en/decoding complexity. Everything in my library is 10-bit -- both downloaded from other sources and what I've encoded myself. 

A more specific use case I have for this is using [ssimulacra2_rs](https://github.com/rust-av/ssimulacra2_bin/issues/31), which uses image-rs for decoding, to help evaluate encoding quality. 

Another use case is [wpaperd](https://github.com/danyspin97/wpaperd), which uses image-rs for decoding.

#1504 is related, but is more about handling improperly formatted avif files.

awxkee:
This should close #2130, #2232, #1504, #1647.

Here is also bugfix because actually YUV400 just tried to copy Y plane data straight to RGBA what is always fails.
YUV decoding is intentionally was made only to RGBA color format without Rgb, Gray, GrayAlpha and others to not break current `ColorType` resolution strategy and to not make PR even larger.

YUV decoder made as per ITU-R standard. Here is [info](https://www.itu.int/rec/T-REC-H.273/en).

Closes #2130, Closes #2232, Closes #1504, Closes #1647, Closes #1930

### DIFF
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -1,4 +1,9 @@
 //! Decoding of AVIF images.
+use crate::error::{
+    DecodingError, ImageFormatHint, LimitError, LimitErrorKind, UnsupportedError,
+    UnsupportedErrorKind,
+};
+use crate::{ColorType, ImageDecoder, ImageError, ImageFormat, ImageResult};
 ///
 /// The [AVIF] specification defines an image derivative of the AV1 bitstream, an open video codec.
 ///
@@ -7,12 +12,6 @@ use std::error::Error;
 use std::fmt::{Display, Formatter};
 use std::io::Read;
 use std::marker::PhantomData;
-
-use crate::error::{
-    DecodingError, ImageFormatHint, LimitError, LimitErrorKind, UnsupportedError,
-    UnsupportedErrorKind,
-};
-use crate::{ColorType, ImageDecoder, ImageError, ImageFormat, ImageResult};
 
 use crate::codecs::avif::yuv::*;
 use dav1d::{PixelLayout, PlanarImageComponent};
@@ -35,7 +34,6 @@ pub struct AvifDecoder<R> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum AvifDecoderError {
     AlphaPlaneFormat(PixelLayout),
-    MemoryLayout,
     YuvLayoutOnIdentityMatrix(PixelLayout),
 }
 
@@ -48,9 +46,6 @@ impl Display for AvifDecoderError {
                 PixelLayout::I422 => f.write_str("Alpha layout must be 4:0:0 but it was 4:2:2"),
                 PixelLayout::I444 => f.write_str("Alpha layout must be 4:0:0 but it was 4:4:4"),
             },
-            AvifDecoderError::MemoryLayout => {
-                f.write_str("Unexpected data size for current RGBx layout")
-            }
             AvifDecoderError::YuvLayoutOnIdentityMatrix(pixel_layout) => match pixel_layout {
                 PixelLayout::I400 => {
                     f.write_str("YUV layout on 'Identity' matrix must be 4:4:4 but it was 4:0:0")
@@ -125,7 +120,7 @@ fn reshape_plane(source: &[u8], stride: usize, width: usize, height: usize) -> V
         .zip(source.chunks_exact(stride))
     {
         for (dst, src) in shaped_row.iter_mut().zip(src_row.chunks_exact(2)) {
-            *dst = u16::from_le_bytes([src[0], src[1]]);
+            *dst = u16::from_ne_bytes([src[0], src[1]]);
         }
     }
     target_plane
@@ -134,6 +129,15 @@ fn reshape_plane(source: &[u8], stride: usize, width: usize, height: usize) -> V
 struct Plane16View<'a> {
     data: std::borrow::Cow<'a, [u16]>,
     stride: usize,
+}
+
+impl Default for Plane16View<'_> {
+    fn default() -> Self {
+        Plane16View {
+            data: std::borrow::Cow::Owned(vec![]),
+            stride: 0,
+        }
+    }
 }
 
 /// This is correct to transmute FFI data for Y plane and Alpha plane
@@ -230,8 +234,7 @@ fn get_matrix(
     david_matrix: dav1d::pixel::MatrixCoefficients,
 ) -> Result<YuvStandardMatrix, ImageError> {
     match david_matrix {
-        // Identity just a stub here, we'll handle it in different way
-        dav1d::pixel::MatrixCoefficients::Identity => Ok(YuvStandardMatrix::Bt709),
+        dav1d::pixel::MatrixCoefficients::Identity => Ok(YuvStandardMatrix::Identity),
         dav1d::pixel::MatrixCoefficients::BT709 => Ok(YuvStandardMatrix::Bt709),
         // This is arguable, some applications prefer to go with Bt.709 as default,
         // and some applications prefer Bt.601 as default.
@@ -301,60 +304,6 @@ fn get_matrix(
     }
 }
 
-fn check_target_rgba_dimension_preconditions(
-    width: usize,
-    height: usize,
-) -> Result<(), ImageError> {
-    // This is suspicious if this happens, better fail early
-    if width == 0 || height == 0 {
-        return Err(ImageError::Limits(LimitError::from_kind(
-            LimitErrorKind::DimensionError,
-        )));
-    }
-    // Image dimensions must not exceed pointer size
-    let (v_stride, ow) = width.overflowing_mul(4);
-    if ow {
-        return Err(ImageError::Limits(LimitError::from_kind(
-            LimitErrorKind::InsufficientMemory,
-        )));
-    }
-    let (_, ow) = v_stride.overflowing_mul(height);
-    if ow {
-        return Err(ImageError::Limits(LimitError::from_kind(
-            LimitErrorKind::InsufficientMemory,
-        )));
-    }
-    Ok(())
-}
-
-fn check_plane_dimension_preconditions(
-    width: usize,
-    height: usize,
-    target_width: usize,
-    target_height: usize,
-) -> Result<(), ImageError> {
-    // This is suspicious if this happens, better fail early
-    if width == 0 || height == 0 {
-        return Err(ImageError::Limits(LimitError::from_kind(
-            LimitErrorKind::DimensionError,
-        )));
-    }
-    // Plane dimensions must not exceed pointer size
-    let (_, ow) = width.overflowing_mul(height);
-    if ow {
-        return Err(ImageError::Limits(LimitError::from_kind(
-            LimitErrorKind::InsufficientMemory,
-        )));
-    }
-    // This should never happen that plane size differs from target size
-    if target_width != width || target_height != height {
-        return Err(ImageError::Limits(LimitError::from_kind(
-            LimitErrorKind::DimensionError,
-        )));
-    }
-    Ok(())
-}
-
 impl<R: Read> ImageDecoder for AvifDecoder<R> {
     fn dimensions(&self) -> (u32, u32) {
         (self.picture.width(), self.picture.height())
@@ -382,7 +331,12 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
         assert!(bit_depth == 8 || bit_depth == 10 || bit_depth == 12);
 
         let (width, height) = self.dimensions();
-        check_target_rgba_dimension_preconditions(width as usize, height as usize)?;
+        // This is suspicious if this happens, better fail early
+        if width == 0 || height == 0 {
+            return Err(ImageError::Limits(LimitError::from_kind(
+                LimitErrorKind::DimensionError,
+            )));
+        }
 
         let yuv_range = match self.picture.color_range() {
             dav1d::pixel::YUVRange::Limited => YuvIntensityRange::Tv,
@@ -403,46 +357,29 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
         }
 
         if bit_depth == 8 {
-            if self.picture.pixel_layout() != PixelLayout::I400 {
-                let ref_y = self.picture.plane(PlanarImageComponent::Y);
-                let ref_u = self.picture.plane(PlanarImageComponent::U);
-                let ref_v = self.picture.plane(PlanarImageComponent::V);
+            let ref_y = self.picture.plane(PlanarImageComponent::Y);
+            let ref_u = self.picture.plane(PlanarImageComponent::U);
+            let ref_v = self.picture.plane(PlanarImageComponent::V);
 
-                let image = YuvPlanarImage {
-                    y_plane: ref_y.as_ref(),
-                    y_stride: self.picture.stride(PlanarImageComponent::Y) as usize,
-                    u_plane: ref_u.as_ref(),
-                    u_stride: self.picture.stride(PlanarImageComponent::U) as usize,
-                    v_plane: ref_v.as_ref(),
-                    v_stride: self.picture.stride(PlanarImageComponent::V) as usize,
-                    width: width as usize,
-                    height: height as usize,
-                };
+            let image = YuvPlanarImage {
+                y_plane: ref_y.as_ref(),
+                y_stride: self.picture.stride(PlanarImageComponent::Y) as usize,
+                u_plane: ref_u.as_ref(),
+                u_stride: self.picture.stride(PlanarImageComponent::U) as usize,
+                v_plane: ref_v.as_ref(),
+                v_stride: self.picture.stride(PlanarImageComponent::V) as usize,
+                width: width as usize,
+                height: height as usize,
+            };
 
-                if !is_identity {
-                    let worker = match self.picture.pixel_layout() {
-                        PixelLayout::I400 => unreachable!(),
-                        PixelLayout::I420 => yuv420_to_rgba8,
-                        PixelLayout::I422 => yuv422_to_rgba8,
-                        PixelLayout::I444 => yuv444_to_rgba8,
-                    };
+            let worker = match self.picture.pixel_layout() {
+                PixelLayout::I400 => yuv400_to_rgba8,
+                PixelLayout::I420 => yuv420_to_rgba8,
+                PixelLayout::I422 => yuv422_to_rgba8,
+                PixelLayout::I444 => yuv444_to_rgba8,
+            };
 
-                    worker(image, buf, yuv_range, color_matrix)?;
-                } else {
-                    gbr_to_rgba8(image, buf)?;
-                }
-            } else {
-                let plane = self.picture.plane(PlanarImageComponent::Y);
-
-                let gray_image = YuvGrayImage {
-                    y_plane: plane.as_ref(),
-                    y_stride: self.picture.stride(PlanarImageComponent::Y) as usize,
-                    width: width as usize,
-                    height: height as usize,
-                };
-
-                yuv400_to_rgba8(gray_image, buf, yuv_range, color_matrix)?;
-            }
+            worker(image, buf, yuv_range, color_matrix)?;
 
             // Squashing alpha plane into a picture
             if let Some(picture) = self.alpha_picture {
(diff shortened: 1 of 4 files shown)


HeroicKatora (commented):
I like the transform, quite clean and encompassing in terms of what ITU actually specified. I don't see `Unsupported` being used correctly and an error type instead of `String` would be highly preferable.

src/codecs/avif/decoder.rs
```diff
@@ -102,84 +109,369 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
...
+
+            // Squashing alpha plane into a picture
+            if let Some(picture) = self.alpha_picture {
+                if picture.pixel_layout() != PixelLayout::I400 {
+                    return Err(ImageError::Unsupported(
+                        UnsupportedError::from_format_and_kind(
+                            ImageFormat::Avif.into(),
+                            UnsupportedErrorKind::GenericFeature(format!(
+                                "Alpha must be PixelLayout::I400 but was: {:?}",
+                                picture.pixel_layout() // PixelLayout does not implement display
```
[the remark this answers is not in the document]
awxkee: This one actually was before me.
Yes, `DecodingError` would be best fit here because it normally should not happen, and I think better to consider image as invalid.

src/codecs/avif/yuv.rs
```diff
@@ -0,0 +1,886 @@
...
+    kb: f32,
+    precision: u32,
+) -> Result<CbCrInverseTransform<i32>, String> {
+    let range_uv = range_bgra as f32 / range_uv as f32;
+    let y_coef = range_bgra as f32 / range_y as f32;
+    let cr_coeff = (2f32 * (1f32 - kr)) * range_uv;
+    let cb_coeff = (2f32 * (1f32 - kb)) * range_uv;
+    let kg = 1.0f32 - kr - kb;
+    if kg == 0f32 {
+        return Err("1.0f - kr - kg must not be 0".parse().unwrap());
```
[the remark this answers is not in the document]
awxkee: Yes, assert should be here.

src/codecs/avif/decoder.rs
```diff
@@ -102,84 +109,369 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
...
+                        Ok(slice) => slice,
+                        Err(_) => {
+                            shape_a_plane();
+                            _bind_a.as_slice()
+                        }
+                    }
+                } else {
+                    shape_a_plane();
+                    _bind_a.as_slice()
+                };
```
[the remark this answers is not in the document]
awxkee: Nice idea! I did some rework here with extracting this to method, it is still an ugly one, however much more clean.

src/codecs/avif/decoder.rs
```diff
@@ -102,84 +109,369 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
...
+                );
+                let cr = yuv400_to_rgba(
+                    gray_image,
+                    rgba16_buf,
+                    self.picture.bit_depth() as u32,
+                    yuv_range,
+                    color_matrix,
+                );
+                if let Err(err) = cr {
+                    return Err(ImageError::Unsupported(
```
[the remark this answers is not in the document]
awxkee: All errors has been reworked.

awxkee:
I'm changed `Unspecified` to `Bt.601`.

Even here it is marked as Bt.709 (what is a bug I consider )
https://github.com/image-rs/image/blob/cf9c5326d6feba8fc67eb081b6feff3600b74d02/src/codecs/avif/encoder.rs#L35-L40

Actually [ravif](https://github.com/kornelski/cavif-rs/blob/1c7b2b83c8e39ea8974bd43b8dc1daa60be34deb/ravif/src/av1encoder.rs#L285) is using Bt.601 as default, so to not create gaps in implementation I think it is better to go this way.

fintelia (commented):
Left a few comments, but I had a hard time reviewing the PR. I've started reading through like three times and every time I get lost trying to keep all the different cases in my head.

I'd encourage finding ways to consolidate different cases. As one example, if `yuv400_to_rgba8` took a `YuvPlanarImage` with `u_plane` and `v_plane` set to `&[]` then you wouldn't need a completely separate code path to deal with it.

src/codecs/avif/decoder.rs
```diff
@@ -86,13 +117,255 @@ impl<R: Read> AvifDecoder<R> {
...
+}
+
+fn check_plane_dimension_preconditions(
+    width: usize,
+    height: usize,
+    target_width: usize,
+    target_height: usize,
+) -> Result<(), ImageError> {
+    // This is suspicious if this happens, better fail early
+    if width == 0 || height == 0 {
```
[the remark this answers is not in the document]
awxkee: Note, there is two different check. One is checking RGBA dimensions, other for planar dimensions.

src/codecs/avif/decoder.rs
```diff
@@ -102,84 +375,248 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
...
-            for (buf, slice) in Iterator::zip(
-                buf.chunks_exact_mut(width as usize * 4),
-                plane.as_ref().chunks_exact(stride),
-            ) {
-                for i in 0..width as usize {
-                    buf[3 + i * 4] = slice[i];
+
+            // Expand current bit depth to target 16
+            let target_expand_bits = 16u32.saturating_sub(self.picture.bit_depth() as u32);
+            if target_expand_bits > 0 {
```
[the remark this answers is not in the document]
awxkee: I agree. This was before I did assertion.

src/codecs/avif/decoder.rs
```diff
@@ -102,84 +375,248 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
...
+                    plane.as_ref().chunks_exact(stride),
+                ) {
+                    for (rgba, a_src) in buf.chunks_exact_mut(4).zip(slice) {
+                        rgba[3] = *a_src;
+                    }
+                }
+            }
+        } else {
+            // 8+ bit-depth case
+            let rgba16_buf: &mut [u16] = match bytemuck::try_cast_slice_mut(buf) {
```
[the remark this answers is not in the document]
awxkee: Not really good news then. May it have a slope or size at least is exact?

src/codecs/avif/decoder.rs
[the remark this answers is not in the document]
fintelia: But below we check that the RGBA dimensions match the planar dimensions? That check alone seems like it should be sufficient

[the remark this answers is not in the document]
fintelia: But below we check that the RGBA dimensions match the planar dimensions? That check seems like it should be sufficient

awxkee:
> Left a few comments, but I had a hard time reviewing the PR. I've started reading through like three times and every time I get lost trying to keep all the different cases in my head.
> 
> I'd encourage finding ways to consolidate different cases. As one example, if `yuv400_to_rgba8` took a `YuvPlanarImage` with `u_plane` and `v_plane` set to `&[]` then you wouldn't need a completely separate code path to deal with it.

Sorry, I tried to keep it small, but all these takes some space.

src/codecs/avif/decoder.rs
```diff
@@ -86,13 +117,255 @@ impl<R: Read> AvifDecoder<R> {
...
+    }
+}
+
+/// Getting one of prebuilt matrix of fails
+fn get_matrix(
+    david_matrix: dav1d::pixel::MatrixCoefficients,
+) -> Result<YuvStandardMatrix, ImageError> {
+    match david_matrix {
+        // Identity just a stub here, we'll handle it in different way
+        dav1d::pixel::MatrixCoefficients::Identity => Ok(YuvStandardMatrix::Bt709),
```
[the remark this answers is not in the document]
awxkee: I thought about that. However, at the moment it will introduce a few branches in the code, and this won't actually make things easier.
I'll try to think what I can do here.

src/codecs/avif/decoder.rs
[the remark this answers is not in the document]
fintelia: All our decoders assert that the size of `buf` is exactly the number of bytes needed to store the output. Not sure what you mean by slope

awxkee:
Yes, using `YuvPlanarImage` for YUV 4:0:0 will remove some code. However, if you'll decide to expose this method to whole create it will be definitely not clear how to use this.
If it fine to you it is fine by me.

src/codecs/avif/decoder.rs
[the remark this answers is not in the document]
awxkee: I'll move `Identity` from here to YUV 4:4:4 to handle that case this way
```rust
pub(crate) fn yuv444_to_rgba8(
    image: YuvPlanarImage<u8>,
    rgba: &mut [u8],
    range: YuvIntensityRange,
    matrix: YuvStandardMatrix,
) -> Result<(), ImageError> {
    if matrix == YuvStandardMatrix::Identity {
        gbr_to_rgba8(image, rgba)
    } else {
        yuv444_to_rgbx_impl::<u8, 4, 8>(image, rgba, range, matrix)
    }
}
```

This still far away from something good, however will remove some branches in decoder.

awxkee:
This is more clean now.

However, actual complexity just became obscured instead of explicit.
If someone we’ll try to use those methods better to think twice about what he’s doing.

awxkee:
Perhaps, I can do some simple strategy pattern here, for each case GBR, 4:0:0, or other, something like that.
```rust
trait AvifResolutionStrategy {
    fn is_supported(&self, picture: Picture) -> bool;
    fn process<R: Read>(picture: Picture, buf: &mut [u8]) -> ImageResult<()>;
}
```
This will also keep everything clear, but will notably increase amount of code.

Otherwise, if you don't like that one, I think at the moment there is the best code at least in `AvifDecoder`, for a price that YUV converters magically accept or not accept some arguments.

src/codecs/avif/decoder.rs
```diff
@@ -86,13 +112,209 @@ impl<R: Read> AvifDecoder<R> {
...
+            )),
+        ),
+        dav1d::pixel::MatrixCoefficients::ICtCp => Err(ImageError::Unsupported(
+            UnsupportedError::from_format_and_kind(
+                ImageFormat::Avif.into(),
+                UnsupportedErrorKind::GenericFeature(
+                    "ICtCp Derived Luminance matrix is not supported".to_string(),
+                ),
+            ),
+        )),
```
HeroicKatora, line 303: Ah, yes, there's an additional non-linearity to consider in this case. Let's implement this is a seperate PR, should be easier with the infrastructure in place.

src/codecs/avif/yuv.rs
```diff
@@ -0,0 +1,1403 @@
...
+        let y_iter = y_src.chunks_exact(2);
+        let rgb_chunks = rgb.chunks_exact_mut(CHANNELS * 2);
+        for (((y_src, u_src), v_src), rgb_dst) in y_iter.zip(u_src).zip(v_src).zip(rgb_chunks) {
+            let y_value = (y_src[0].as_() - bias_y) * y_coef;
+            let cb_value = u_src.as_() - bias_uv;
+            let cr_value = v_src.as_() - bias_uv;
+
+            let r = ((y_value + cr_coef * cr_value + ROUNDING) >> PRECISION).clamp(0, max_value);
+            let b = ((y_value + cb_coef * cb_value + ROUNDING) >> PRECISION).clamp(0, max_value);
+            let g = ((y_value - g_coef_1 * cr_value - g_coef_2 * cb_value + ROUNDING) >> PRECISION)
```
HeroicKatora, line 732: The re-occurence of this (across all formats) makes me consider if there should be a helper:

```suggestion
#[inline(always)]
fn round<const PRECISION: i32, const MAX: i32>(val: i32) -> i32 {
    let ROUNDING: i32 = 1 << (PRECISION - 1);
    ((val + ROUNDING) >> PRECISION).clamp(0, MAX)
}

            let r = round::<PRECISION, MAX>(y_value + cr_coef * cr_value);
            let b = round::<PRECISION, MAX>(y_value + cb_coef * cb_value);
            let g = round::<PRECISION, MAX>(y_value - g_coef_1 * cr_value - g_coef_2 * cb_value);
```

The control flow / arithmetic order will be more apparent and it should not influence codegen.

HeroicKatora (commented):
At least from the standpoint of review, the complexity has gone *down* drastically—so thank you for the refactorings. The 'odd' case still is an unfortunate fence post in the loop and it looks bulky. But more importantly it is possible to read many sections of code independent from each other now. In terms of numerics, I'm predisposed to the yuv transform which probably helped.

I wouldn't refactor with _more_ type system. That's just going to decrease readability and the dispatch here doesn't need to be dynamic, so please don't. If anything iterate on this structure instead :+1: 

Thanks for contributing these to the `image-rs` project. In my head I'm considering copying to a separate crate but let's land it first. Since it is quite a bulk, I'd let @fintelia weigh in with comments before the merge; but this is quite close to approval from my side.

src/codecs/avif/decoder.rs
awxkee: As to me actual issue is not an implementation itself here, as also for CL 2020, YCgCo and others, is that there is no reference software allows you to do make those image and view them, or at least I don't know anything.

AVIF's with `Bt.2020 CL` CICP exists only in Netflix samples avif repository and [author of libavif](https://github.com/AOMediaCodec/libavif/issues/324)marked that it was highly likely made by mistake.

I never saw intentionally, and correctly encoded image with CICP have `ICtCp` matrix as well as Bt.2020 CL also.
If you have one in your possesion it is interesting to have a look on it, and probably then ask what reference software may do so.

awxkee: As for `YCgCo`, I think I can also do commit with it but after this PR successfully lands, and, perhaps, if you'll make this in separate repo in your organisation it also might be more convenient.
Even `YCgCo` might be blend in current implementation, actually, it was made to make possible encoding using only bit shift operators, and much simpler decoding process, so in terms of performance, different implementation might be preferred. 
Here is in example of `YCgCo` decoding and for limited range it is even simpler.
```rust
    let y_value = (y_plane[0] as i32 - bias_y)
                * range_reduction_y;
            
            let cg_value =
                (cg_plane[0] as i32 - bias_uv) * range_reduction_uv;

            let co_value =
                (co_plane[0] as i32 - bias_uv) * range_reduction_uv;

            let t = y_value - cg_value;

            let r = ((t + co_value + ROUNDING_CONST) >> PRECISION)
                .min(255)
                .max(0);
            let b = ((t - co_value + ROUNDING_CONST) >> PRECISION)
                .min(255)
                .max(0);
            let g = ((y_value + cg_value + ROUNDING_CONST) >> PRECISION)
                .min(255)
                .max(0);
```

src/codecs/avif/yuv.rs
HeroicKatora: @awxkee Sorry if this was unclear, this was not meant as a literal suggestion but a rough draft of a refactor of these rounding conversions. Would you move that function out of the body and apply it to all the other occurences in the various yuv functions where it can be used? Maybe you're already doing so in the background.

src/codecs/avif/decoder.rs
HeroicKatora: Good point on missing (public) reference software. Without any validation cases it'll indeed be quite hard to land it with confidence. This is definitely a separate issue to discuss, let's not trail too far off-topic in this PR.

src/codecs/avif/yuv.rs
awxkee: Yup, sure, no worries. This was clear enough.

awxkee:
I also splat a little 4:2:0 and 4:2:2 processing, I think it'd make quite easier to understand what's going on there.

src/codecs/avif/decoder.rs
kornelski: AVIF has inherited HEIF and a video codec, which have tons of features and variants, many of which exist just because an older standard had them, and are of questionable utility. 

I suggest implementing only features that are supported across web browsers. That's a de-facto practical profile of the format.

kornelski: YCgCo is a neat color space for lossless encoding, but lossless encoding in AV1 is pretty weak.

awxkee: YCgCo [has added to libavif](https://github.com/AOMediaCodec/libavif/issues/2077) so as it exists in reference implementation in some future it will be likely added everywhere ( and I it looks like it is [already added](https://source.chromium.org/chromium/chromium/src/+/main:media/base/video_color_space.h;l=69;drc=4f00803ca25c0d0480ed14844d6406933c21e80e) )
As for others, CL, chromacity derived and others, I'd agree.

If you want to target as browsers do then issue with `Unspecified` still in the air because Chromium [definitely goes always with Bt.709](https://source.chromium.org/chromium/chromium/src/+/main:media/base/video_color_space.h;l=69;drc=4f00803ca25c0d0480ed14844d6406933c21e80e).

awxkee: > YCgCo is a neat color space for lossless encoding, but lossless encoding in AV1 is pretty weak.

I think it is just a question to @HeroicKatora and @fintelia if they want to support that or not rather that encoding quality.

awxkee: And it is looks like [libavif](https://github.com/AOMediaCodec/av1-avif/issues/84) also want to consider unspecified as Bt.709.

src/codecs/avif/decoder.rs
```diff
@@ -102,93 +324,231 @@ impl<R: Read> ImageDecoder for AvifDecoder<R> {
...
                 }
             }
         }
 
-        Ok(())
-    }
+        // Expand current bit depth to target 16
+        let target_expand_bits = 16u32 - self.picture.bit_depth() as u32;
+        for item in target.iter_mut() {
+            *item <<= target_expand_bits;
```
fintelia, line 548: When increasing the bit depth, I think we actually want to replicate the high bits: 0x000 -> 0x0000, 0x123 -> 0x1231, 0xfff -> 0xffff.

Should be something like this:
```rust
*item = (*item << target_expand_bits) | (*item >> (16 - target_expand_bits));
```

fintelia (commented):
Left one more comment. I've only looked at the interface in decoder.rs, so I'll defer to other reviewers about whether this PR is otherwise ready to merge.

One other thing to mention is that image-png just got support for coding-independent code points (aka ITU-REC-H.273) in https://github.com/image-rs/image-png/pull/529. PNG lacks supports for YUV, so it is just the metadata information about what colorspace the RGB channels are in. Might be worth thinking about how to have a unified interface between AVIF and PNG

src/codecs/avif/decoder.rs
awxkee: Thanks for suggestion!

awxkee:
So, as there again was raised a concern about what actually targets are,
I would like to hear what matrix should be set on `Unspecified`.

fintelia:
I'd say we should probably match web browsers. Plus, we tend to assume sRGB throughout the crate, so this wouldn't be a big departure

Shnatsel:
Now that this PR is approved I'm going to hit the merge button, unless @fintelia objects?

I'll also start preparing a point release because we have plenty of fixes and improvements in git already.

HeroicKatora:
@Shnatsel You can press the merge button next time. As far as I understood @fintelia had already deferred the remaining review to me. I just got side tracked between the LGTM and merging with #dayjob work. Thanks for preparing the release notes already :+1:

PR #2378, opened 2024-11-04
Release 0.25.5 with 10-bit AVIF and bugfixes

Shnatsel:


CHANGES.md
```diff
@@ -6,6 +6,18 @@
 
 ## Changes
 
+### Version 0.25.5
+
+Features:
+ - Added support for decoding 10-bit AVIF
```
awxkee, line 12: 12 bit also handled as well.

fintelia:
I'd like to get decisions on #2372 and #2363 before releasing. Particularly the question about whether non-default image formats should live in an external `image-extras` crate. Using a second crate would be my preference. And if we decide to do so, it would be better for PCX to be moved there now, rather than having to wait until the next major release.

Shnatsel:
How about I remove all references to PCX from Cargo.toml, ship a point release and then put them back?

I don't want to block shipping bugfixes on decisions about support for an obscure format from the 80s.

fintelia:
That sounds good. I think cargo now generates warnings for features used in the code but not mentioned in the Cargo.toml, but I think those should only be visible to us, not to downstream users

Shnatsel:
PCX temporarily disabled in #2379, this can now be merged and published.

Shnatsel:
I don't have publishing rights, so I'll need one of you to `cargo publish`. I can add the git tag myself once that's done.

Shnatsel:
Release announcement is up: https://www.reddit.com/r/rust/comments/1gk3v7i/image_v0255_brings_much_improved_avif_decoding/

PR #2379, opened 2024-11-05
Temporarily disable PCX format

Shnatsel:
As discussed in #2378, temporarily disable PCX support so that we could ship bugfixes without being blocked on decisions about support for an obsolete format.

In my tests, Cargo shows warnings about nonexistent features even to people who use the crate as a dependency, so I had to add `#![allow(unexpected_cfgs)]` as well.

I'll revert this PR once 0.25.5 is released.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


fintelia (approved):

fintelia:
Could you also update the CI to remove testing of the pcx feature?

PR #2380, opened 2024-11-05
Revert "Temporarily disable PCX format"

Shnatsel:
Reverts image-rs/image#2379

To be merged once 0.25.5 ships


fintelia:
I'd rather go with #2372 and support the PCX format via `image-extras`

197g:
Since #2372 is merged, we're keeping this in `image-extras` for now.

Issue #2275, opened 2024-07-01
Thread panics on EXR containing NAN


gyk:
Some EXR files might contain NaNs, e.g., [AllHalfValues.exr](https://openexr.com/en/latest/test_images/TestImages/AllHalfValues.html). Calling `to_rgba8` on such images causes a panic:

```
thread 'main' panicked at image-0.25.1\src\color.rs:435:30:
called `Option::unwrap()` on a `None` value
```

fintelia:
Shorter repro:
```rust
let img = Rgba32FImage::from_raw(1, 1, vec![1.0 / 0.]).expect("create nan image");
DynamicImage::ImageRgba32F(img).to_rgba8();
```

FlareFlo:
According to my research, there is no defined operation for how to handle NaN when converting to another format. I can open a PR against my branch if the below behavior is agreed upon.
Firefox render of the linked sample image
![image](https://github.com/user-attachments/assets/47872819-5d67-4dc8-ab39-32a6b2730dde)
PNG conversion with clamp applied, from [the fixed branch](https://github.com/FlareFlo/image/tree/2275-float-nan-conversion)
![out](https://github.com/user-attachments/assets/0e9fec72-a7ba-4fff-8d1a-e885330d4bff)

awxkee:
Num traits does not support Positive infinites, Negative infinites also, only subnormal numbers are supported.
Usually most hardware flushes NaNs towards negative infinity ( zero for unsigned, max possible negative value for signed ).

FlareFlo:
> Num traits does not support Positive infinites, Negative infinites also, only subnormal numbers are supported.
> Usually most hardware flushes NaNs towards negative infinity ( zero for unsigned, max possible negative value for signed ).
> 

Well, in this case the image parses just fine when filtering out NaNs, all other funny values appear to be handled (as this image contains every possible bitrepr)

awxkee:
Not really, here two immediate fails

```rust
    let img1 = Rgba32FImage::from_raw(1, 1, vec![f32::INFINITY]).expect("create +inf image");
    DynamicImage::ImageRgba32F(img1).to_rgba8();

    let img2 = Rgba32FImage::from_raw(1, 1, vec![f32::NEG_INFINITY]).expect("create -inf image");
    DynamicImage::ImageRgba32F(img2).to_rgba8();
```

FlareFlo:
> ```rust
[...]

PR #2381, opened 2024-11-16
Clamp NaN to white when converting to integer

FlareFlo:
Linked issue: https://github.com/image-rs/image/issues/2275

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -399,15 +399,15 @@ fn sanitize_nan(i: f32) -> f32 {
 
 impl FromPrimitive<f32> for u8 {
     fn from_primitive(float: f32) -> Self {
-        let inner = (float.clamp(0.0, 1.0) * u8::MAX as f32).round();
-        NumCast::from(sanitize_nan(inner)).unwrap()
+        let inner = (sanitize_nan(float).clamp(0.0, 1.0) * u8::MAX as f32).round();
+        NumCast::from(inner).unwrap()
     }
 }
 
 impl FromPrimitive<f32> for u16 {
     fn from_primitive(float: f32) -> Self {
-        let inner = (float.clamp(0.0, 1.0) * u16::MAX as f32).round();
-        NumCast::from(sanitize_nan(inner)).unwrap()
+        let inner = (sanitize_nan(float).clamp(0.0, 1.0) * u16::MAX as f32).round();
+        NumCast::from(inner).unwrap()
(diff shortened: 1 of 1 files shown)


FlareFlo:
No clue why clippy throws a fit, this PR does not touch those failing files at all.

FlareFlo:
> There's new clippy complaints every time a new clippy version comes out. Don't worry about them.
> 
> Do you know the performance implications of this change? I'm not sure performance was great to start with, but I don't want to unintentionally add guarantees that will make it unnecessary slow forever.

https://godbolt.org/z/7318KPd4f Not great, but im not sure what else to do

src/color.rs
```diff
@@ -385,17 +385,29 @@ impl<T: Primitive> FromPrimitive<T> for T {
...
+    } else {
+        i
+    }
+}
+
 impl FromPrimitive<f32> for u8 {
     fn from_primitive(float: f32) -> Self {
         let inner = (float.clamp(0.0, 1.0) * u8::MAX as f32).round();
-        NumCast::from(inner).unwrap()
+        NumCast::from(sanitize_nan(inner)).unwrap()
```
fintelia, line 403: Wait, shouldn't the `sanitize_nan` call happen before scaling by `u8::MAX`?

FlareFlo: Good point, ill fix that in a sec

fintelia (approved):

kornelski:
It should be possible to replace `clamp` with custom comparison that handles `NaN` at zero cost:

`if !(x < 1.0) { 1.0 }` handles both `NaN` and `>1.0`.

FlareFlo:
> It should be possible to replace `clamp` with custom comparison that handles `NaN` at zero cost:
> 
> `if !(x < 1.0) { 1.0 }` handles both `NaN` and `>1.0`.

Something like this appears to produce better codegen indeed https://godbolt.org/z/qnbjWhr3c

FlareFlo:
> There's new clippy complaints every time a new clippy version comes out. Don't worry about them.
> 
> Do you know the performance implications of this change? I'm not sure performance was great to start with, but I don't want to unintentionally add guarantees that will make it unnecessary slow forever.

I have a few ideas for speeding things up, too much for this PR though. Ill experiment and open a PR when i have something that might make sense.

fintelia (approved):

PR #2382, opened 2024-11-18
Make avif-native 1.70-compatible

FreezyLemon:
`cargo +1.70 build --no-default-features -F avif-native` doesn't build (0.25.5). The fixes for this are trivial and only language-related, so I hope these can be included so we don't have to bump the MSRV to `1.80` when using avif-native.

- Add `use std::mem::size_of`
- Remove `pub` visibility where unnecessary

Note: Real MSRV could be lower than 1.70, haven't tested <1.70

### DIFF
--- a/src/codecs/avif/yuv.rs
+++ b/src/codecs/avif/yuv.rs
@@ -2,15 +2,16 @@ use crate::error::DecodingError;
 use crate::{ImageError, ImageFormat};
 use num_traits::AsPrimitive;
 use std::fmt::{Display, Formatter};
+use std::mem::size_of;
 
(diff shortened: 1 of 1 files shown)


FreezyLemon:
CI error seems unrelated.

fintelia (approved):

fintelia:
We don't have a specific release schedule. Mostly just try to make a release when a bunch of changes have landed or when someone asks for one. Can try to get one out soon!

FreezyLemon:
I see. Would be nice to have soon, but not critical. Thanks again :+1:

PR #2385, opened 2024-11-26
Fix YUV 4:2:0, 4:2:2 large strides

awxkee:
If the stride is larger than the plane size, it might contain junk data beyond the actual valid region.
To avoid processing artifacts when working with odd-sized images, the buffers is reshaped to their actual size.

### DIFF
--- a/src/codecs/avif/yuv.rs
+++ b/src/codecs/avif/yuv.rs
@@ -521,13 +521,23 @@ fn process_halved_chroma_row<
 
     let max_value = (1 << BIT_DEPTH) - 1;
 
+    // If the stride is larger than the plane size,
+    // it might contain junk data beyond the actual valid region.
+    // To avoid processing artifacts when working with odd-sized images,
+    // the buffer is reshaped to its actual size,
+    // preventing accidental use of invalid values from the trailing region.
+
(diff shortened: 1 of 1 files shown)


awxkee:
Looks like CI errors are unrelated to this PR

src/codecs/avif/yuv.rs
```diff
@@ -521,13 +521,23 @@ fn process_halved_chroma_row<
...
     let max_value = (1 << BIT_DEPTH) - 1;
 
+    // If the stride is larger than the plane size,
+    // it might contain junk data beyond the actual valid region.
+    // To avoid processing artifacts when working with odd-sized images,
+    // the buffer is reshaped to its actual size,
+    // preventing accidental use of invalid values from the trailing region.
+
+    let y_plane = &image.y_plane[0..image.width];
+    let chroma_size = (image.width + 1) / 2;
```
HeroicKatora, line 531:
```suggestion
    let chroma_size = image.width.div_ceil(2);
```

HeroicKatora (commented):
Nit: code style, we can use expressive methods since `1.73` so this shouldn't have an effect on our MSRV.

HeroicKatora (approved):

PR #2387, opened 2024-11-29
YUV in JPEG in fixed point

### LINKED ISSUE #2311: Increased generation loss in JPEG encoder
In image v0.25.2, decoding and re-encoding JPEG 100 times makes the whole image go mostly green.

A showcase and reproduction code can be found here: https://github.com/OetkenPurveyorOfCode/generation_loss

### LINKED ISSUE #2386: Fixed point in JPEG encoder
YUV encoding/deconding usually works well enough in fixed-point arithmetic and it is faster than floating point. Therefore, it may be possible to replace current JPEG encoding pipeline with Q16 multiplication. Notably, [libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/9758c8a74ceb6ee7d4f8350fa2767fdc5b2bad2a/simd/arm/aarch64/jccolext-neon.c#L31) also done in fixed-point arithmetic, so there will be no significant deviation from almost reference implementation.

However, switching to fixed-point is unlikely to yield significant improvements, there is only a few percent gains in end-to-end encoding.

Here is an experimental implementation to try out this approach: https://github.com/awxkee/image/tree/test_yuv_jpeg_fixed_point

```bash
cargo bench --bench fixed_point
```

Tests shows that fixed-point faster something about 20%, but in end-to-end encoding gain are varies from 2% to 7%.

P.S.
Anyway, it seems here `.round()` is missing, so luma and chroma values are truncated on each pass. Over time, this truncation leads to #2311

https://github.com/image-rs/image/blob/5080752fa41640e1a4adeb962bcdf79b3f428134/src/codecs/jpeg/encoder.rs#L773-L789

awxkee:
YUV in JPEG encoding as per #2386

Fixes #2386, Fixes #2311

Here is result of #2311 after 100 encoding iterations

jpg

Actually I might still note sometimes slightly more greener images and over time there is still areas with artifacts that means at some other place is likely also rounding errors occurs.


awxkee:
Seems CI errors unrelated to this PR

awxkee:
Here is also one time proof that it will never overflow on edge cases, because we're not using clamp:

Y:
https://www.wolframalpha.com/input?i=%2819595+*+255+%2B+38469+*+255+%2B+7471+*+255+%2B+%28%281+%3C%3C+15%29+-+1%29%29+%3E%3E+16
U:
https://www.wolframalpha.com/input?i=%28-11059+*+255+-+21709+*+255+%2B+32768+*+0+%2B+%28128+*+%281+%3C%3C+16%29%29+%2B+%28%281+%3C%3C+15%29+-+1%29%29+%3E%3E+16
V:
https://www.wolframalpha.com/input?i=%2832768+*+0+-+27439+*+255+-+5329+*+255+%2B+%28128+*+%281+%3C%3C+16%29%29+%2B+%28%281+%3C%3C+15%29+-+1%29%29+%3E%3E+16

HeroicKatora (approved):
Note to other reviewers: jpeg encoding as implemented runs on an 8-bit image in all cases.

PR #2388, opened 2024-12-03
Gaussian blur

awxkee:
~This is some initial implementation of gaussian blur, at the moment, it is more too have a look rather than merge.~
This implementation is ~x20+ faster than current, faster than _fast blur_ until radius goes reasonable huge (~50+) and will be probably faster than any other blur implementation (stack blur etc ) on low radius, with rayon it should even more effective, parallelization level of this algorithm is very high.

Here is some concerns:

~- Old implementation considers `sigma` something different, more like radius base instead of flattening level~
~- I'm not sure how to handle `GenericImageView` without explicit going _unsafe_; and even with _unsafe_ I'm not sure I know best way to get direct typed buffer without copying.~
- Loops are hardly unrolled what is not common practice in this crate, and this is important.
- This unrolling not really good adopted for x86 with low registers bank.
- Loops unrolled in current implementation with target `u8` storage type in mind, for other types performance variance is higher
- It is named `filter_1d`, but at the moment only kernels with symmetry is handled properly. Are there some point to finalize it?
- This algorithm requires more memory than previous
- Some beautiful types as `ColorGroup` also is not common practice in this crate

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -15,6 +15,7 @@ use crate::error::{ImageError, ImageResult, ParameterError, ParameterErrorKind};
 use crate::flat::FlatSamples;
 use crate::image::{GenericImage, GenericImageView, ImageDecoder, ImageEncoder, ImageFormat};
 use crate::image_reader::free_functions;
+use crate::imageops::gaussian_blur_dyn_image;
 use crate::math::resize_dimensions;
 use crate::metadata::Orientation;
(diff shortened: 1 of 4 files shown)


awxkee:
I think this is ready, at most.
I modified sigma to follow previous implementation, but it was incorrect, thus on higher radii it produces notable ringing and aliasing. This is unavoidable without breaking compatibiltiy with previous implementation.

output

PR #2389, opened 2024-12-06
Implement writing ICC profiles for JPEG and PNG images.

alula:
Notes:
- `png` crate doesn't actually write ICC profiles at all in latest published version as of writing (0.17.14), however it seems that 0.17.15 will be released soon: <https://github.com/image-rs/image-png/commit/10644dbff4d5b5614199cbc81145838abdafd98c>

### DIFF
--- a/src/codecs/jpeg/encoder.rs
+++ b/src/codecs/jpeg/encoder.rs
@@ -30,6 +30,7 @@ static SOS: u8 = 0xDA;
 static DQT: u8 = 0xDB;
 // Application segments start and end
 static APP0: u8 = 0xE0;
+static APP2: u8 = 0xE2;
 
 // section K.1
 // table K.1
@@ -346,6 +347,8 @@ pub struct JpegEncoder<W> {
     chroma_actable: Cow<'static, [(u8, u16); 256]>,
 
     pixel_density: PixelDensity,
+
+    icc_profile: Vec<u8>,
 }
 
 impl<W: Write> JpegEncoder<W> {
@@ -415,6 +418,8 @@ impl<W: Write> JpegEncoder<W> {
             chroma_actable: Cow::Borrowed(&STD_CHROMA_AC_HUFF_LUT),
 
             pixel_density: PixelDensity::default(),
+
+            icc_profile: Vec::new(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/codecs/jpeg/encoder.rs
```diff
@@ -648,6 +656,40 @@ impl<W: Write> JpegEncoder<W> {
...
+            ));
+        }
+
+        let num_chunks = ((self.icc_profile.len() + MAX_CHUNK_SIZE - 1) / MAX_CHUNK_SIZE) as u8;
+
+        for (i, chunk) in self.icc_profile.chunks(MAX_CHUNK_SIZE).enumerate() {
+            let chunk_number = (i + 1) as u8;
+            let length = 14 + chunk.len();
+
+            let mut segment = Vec::with_capacity(length);
```
kornelski, line 682: You could create the `segment` before the loop, and then `segment.clear(); vec.reserve(length)`.

src/codecs/jpeg/encoder.rs
```diff
@@ -648,6 +656,40 @@ impl<W: Write> JpegEncoder<W> {
...
+        const MAX_ICC_PROFILE_SIZE: usize = MAX_CHUNK_SIZE * MAX_CHUNK_COUNT;
+
+        if self.icc_profile.len() > MAX_ICC_PROFILE_SIZE {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "ICC profile too large",
+            ));
+        }
+
+        let num_chunks = ((self.icc_profile.len() + MAX_CHUNK_SIZE - 1) / MAX_CHUNK_SIZE) as u8;
```
kornelski, line 676: If you create `self.icc_profile.chunks(MAX_CHUNK_SIZE)` iterator first, then `iter.len()` will give you the number of chunks (it's efficient).

src/codecs/png.rs
```diff
@@ -669,6 +681,11 @@ impl<W: Write> ImageEncoder for PngEncoder<W> {
             ))),
         }
     }
+
+    fn set_icc_profile(&mut self, icc_profile: Vec<u8>) -> Result<(), UnsupportedError> {
+        self.icc_profile = icc_profile;
```
kornelski, line 686: This is okay, but the png crate will need a fix: https://github.com/image-rs/image-png/issues/546

src/codecs/jpeg/encoder.rs
alula: fixed, good catch

Shnatsel:
It should be possible to implement for WebP as well, although that doesn't necessarily have to happen in this PR: https://docs.rs/image-webp/latest/image_webp/struct.WebPEncoder.html#method.set_icc_profile

fintelia:
@Shnatsel WebP ICC writing was already added in #2322

HeroicKatora (approved):

PR #2408, opened 2025-01-27
Update free_functions.rs

neckaros:
Support guessing AVIF image generated by this library (with native avif)

Exemple of image generated by this library with the magic bytes not yet recognized:
[image.zip](https://github.com/user-attachments/files/18562280/image.zip)

### DIFF
--- a/src/image_reader/free_functions.rs
+++ b/src/image_reader/free_functions.rs
@@ -124,7 +124,7 @@ pub(crate) fn write_buffer_impl<W: std::io::Write + Seek>(
     }
 }
 
-static MAGIC_BYTES: [(&[u8], ImageFormat); 25] = [
+static MAGIC_BYTES: [(&[u8], ImageFormat); 26] = [
     (b"\x89PNG\r\n\x1a\n", ImageFormat::Png),
     (&[0xff, 0xd8, 0xff], ImageFormat::Jpeg),
     (b"GIF89a", ImageFormat::Gif),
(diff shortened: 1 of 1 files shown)


neckaros:
> The array size needs to be updated, but otherwise looks good to me

You are right my bad I will update it

neckaros:
I ran the test locally and the failing test do not seem related to my commit but I could be wrong?

fintelia:
Thanks!

kornelski:
These bytes before `ftyp` are the size of the `ftyp` chunk. This chunk contains a variable-length list, so it could have any size.

Instead of collecting implementation-specific lengths, can the length be allowed to be any bytes? The two most significant bytes could be required to be 0, since the chunk is unlikely to ever exceed 64K.

neckaros:
> These bytes before `ftyp` are the size of the `ftyp` chunk. This chunk contains a variable-length list, so it could have any size.
> 
> Instead of collecting implementation-specific lengths, can the length be allowed to be any bytes? The two most significant bytes could be required to be 0, since the chunk is unlikely to ever exceed 64K.

I wondered it myself but could not see how to do it with the current implementation

kornelski:
A simple solution could be to add a start offset to the table of magic strings, and skip 4 bytes in AVIFs case.

fintelia:
If we included a mask of bytes to ignore, we could also have an actual fix for https://github.com/image-rs/image/issues/660

PR #2414, opened 2025-02-01
Calculate GCD in Ratio::new

krakow10:
Hi, I saw [this comment](https://github.com/image-rs/image/pull/1986#issuecomment-1694402502) and wrote this code on a whim.  If it's completely unnecessary that's fine.

### DIFF
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -64,6 +64,19 @@ impl Clone for Frame {
 }
 
 /// The delay of a frame relative to the previous one.
+///
+/// The ratio is reduced on construction which means equality comparisons is reliable even when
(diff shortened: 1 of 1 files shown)


fintelia:
The code itself looks good to me, but TBH I'm not really sure what the implications of this change are

RunDevelopment:
I think you can derive `(Partial)Eq` now, since equivalent ratios are mapped to the same representation.

197g (approved):
I think I'm on board with the advantage of being able to rely on comparison, no matter the underlying granularity. The loop has at most 32 iterations as well so the effects on any existing code using that constructor should not be drastic. And in any case, I do not expect this to be used as part of a really tight loop anywhere. (We're already deriving Eq though which is .. odd).

Issue #2410, opened 2025-01-30
Add support for decoding XBM and XPM


mstoeckl:
[XBM](https://en.wikipedia.org/wiki/X_BitMap) and [XPM](https://en.wikipedia.org/wiki/X_PixMap) are two obsolete image formats that were used by the X window system.

It would be nice if `image` were able to decode them, because some image viewer programs indirectly use `image`, among them GNOME's [`Loupe`](https://apps.gnome.org/Loupe/). I happen to have 29 XPM and 88 XBM images scattered through `/usr`.
<details>
<summary>Edit: XBM file frequency statistics, if you are interested.</summary>

I wrote a script to count file endings for various image formats in the output of `pacman -Fl` on Arch Linux, which lists all files in all installable packages in the main repositories. The results are of course weighted towards established projects, rather than newer ones, toward images used for or by programs rather than graphics, towards programs like emacs with multiple variant packages, and toward formats for Linux. Counts may also be distorted by symlinks. A † indicates the entry is corrected for false positives (e.g., `.pm` is always Perl, not XPM, `.ff` can be force field data, `.pfm` is a often a font file instead of Portable Float Map, etc.)

| Format  | Number of Paths | Number of packages | Example packages |
| ------------- | ------------- | ------------- | ------------- |
| All  | 7385118 | 14624  | |
| .ani | 9 | 2 | pop-icon-theme, python-scipy |
| .avif | 3 | 1 | vaultwarden-web |
| **.bm, .xbm** | 704 | 29 | swi-prolog, fricas, openbox |
| .bmp | 1352 | 54 | megaglest-data, lazarus, intel-oneapi-basekit |
| .cur | 291 | 16 | pop-icon-theme, reaper, lazarus |
| .dds | 452 | 9 | pioneer, sauerbraten-data, ogre-next |
| .emf, .emz | 1 | 1 | metasploit |
| .exr | 33 | 3 | blender, python-tests, pypy3 |
| .ff, .ff.gz, .ff.bz2, .ff.zstd | 0† | 0† | |
| .fits, .fts, .fit | 64 | 4 | python-astropy, root, root-cuda |
| .gif | 26487 | 277 | pysolfc-cardsets, scratch, openui5 |
| .hdr, .rgbe, .xyze, .rad | 84 | 7 | python-astropy, materialx, enscript |
| .heif | 0 | 0 |  |
| .ibm, .lbm, .bbm, .ilbm | 0 | 0 |  |
[...]

PR #2416, opened 2025-02-01
Add XBM decoder

mstoeckl:
This change adds a decoder for XBM files, covering the easy part of https://github.com/image-rs/image/issues/2410 . 

My motivation for adding this is that, even though the XBM format is old and inefficient, `.xbm` files still stick around on Linux systems and I'd like for image viewers using the `image` crate to be able to interpret them. 

Since XBM is not a very common format, the decoder is disabled by default and is not included in the default feature list. (I only see XBM files used for icons, for which PNG and binary PBM are better choices.)

I've tried to make this implementation as reliable and unlikely to need further improvement as I can:
* Is written in pure Rust, with no `unsafe`.
* Has no dependencies.
* Does not allocate any memory.
* Does not panic or unwrap.
* Makes a single pass over the input, does not require the reader to implement `Seek`, and runs in O(n) time.
* Provides error messages on parse failure indicating where the failure occurred.

Note: this PR was in large part made by looking at how the other image formats are handled, so it is possible I am copying old mistakes. I was not able to test the xbm fuzzer due to a random LLVM symbol error that I am still trying to figure out.

I plan to make a PR for XPM decoding later, depending on what feedback and improvements this gets. (The XPM decoder has already been written and is at https://github.com/mstoeckl/image/commits/xpm/, but I'm holding off on further improvements until I know what is needed.)


Shnatsel:
The decoding hooks for external formats (#2372) have been merged. It should ship in a stable release very soon, see  #2555

I think it is best to proceed by opening a PR against the image-extras repository.

Shnatsel:
Oh, I see a PR for that is already up: https://github.com/image-rs/image-extras/pull/4

PR #2418, opened 2025-02-09
Improve format detection

### LINKED ISSUE #2409: The AVIF generated by this library are not recognized when guessing format on ImageReader
When trying to guess image format from raw data this library fail to recognize some AVIF (mainly those generated by this library)

## Expected
Following image shoud have its type guessed thanks to the magic bytes
[image.zip](https://github.com/user-attachments/files/18562294/image.zip)

## Actual behaviour

unknown format error

## Reproduction steps

Create an image reader with the content of the image in the zip and try to guess the format

Proposed fix: https://github.com/image-rs/image/pull/2408

fintelia:
Allows formats to specify a mask of which bytes to compare against the signature. Particularly useful for WebP and AVIF which place size fields prior to their signatures.

Fixes #2409

### DIFF
--- a/src/image_reader/free_functions.rs
+++ b/src/image_reader/free_functions.rs
@@ -173,15 +173,12 @@ pub(crate) fn guess_format_impl(buffer: &[u8]) -> Option<ImageFormat> {
             if buffer.starts_with(signature) {
                 return Some(format);
             }
-        } else if buffer.len() >= signature.len() {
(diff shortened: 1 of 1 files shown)


kornelski (approved):

kornelski:
The build failure is a bug in clippy: https://github.com/rust-lang/rust-clippy/issues/14189

PR #2421, opened 2025-02-16
pnm: Optimize and improve validation for read_separated_ascii

mstoeckl:
Per commit message:
> This function now rejects numbers that start with a leading +, although to match netpbm behavior it does not reject leading zeros. It also avoids allocating on the non-error path, reducing the time needed to read a large plain PPM test image to 50%. As the function no longer keeps the entire token in memory at once, the error messages no longer specify the bad token.

1. I don't think losing error message specificity here is a big problem, because PNM images are essentially never written by hand so the errors will rarely occur; if parsing does fail, one can easily find non-digit non-whitespace characters in a file with a regexp (e.g. `grep -P '[^0-9 \t]'`)
2. To create a large "plain" (P3) image to test this with, use e.g. `pngtopnm -plain large.png > ~/output.ppm`. Depending on the png bit depth this can create both 8-bit and 16-bit PPM files.

### DIFF
--- a/src/codecs/pnm/decoder.rs
+++ b/src/codecs/pnm/decoder.rs
@@ -3,7 +3,7 @@ use std::fmt::{self, Display};
 use std::io::{self, Read};
 use std::mem::size_of;
 use std::num::ParseIntError;
-use std::str::{self, FromStr};
+use std::str;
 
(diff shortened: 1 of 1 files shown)


fintelia:
Thanks!

PR #2424, opened 2025-02-17
Update resize and blurs doc

awxkee:
Updated doc for `blur` and `resize`.
Clippy error seems to be the error in the clippy itself.


src/imageops/fast_blur.rs
```diff
@@ -5,6 +5,7 @@ use crate::{ImageBuffer, Pixel, Primitive};
 /// Approximation of Gaussian blur after
 /// Kovesi, P.:  Fast Almost-Gaussian Filtering The Australian Pattern
 /// Recognition Society Conference: DICTA 2010. December 2010. Sydney.
+/// This method requires alpha pre-multiplication for images that contain non-constant alpha.
```
fintelia, line 8: Perhaps "assumes" rather than "requires"?

awxkee: Done

fintelia (approved):

Issue #1687, opened 2022-03-17
Fails to write 16 bit image with pnm


Icelk:
When trying to write a 16-bit image with the PnmHeader below, I get the error `Invalid color type for selected PAM color type`

```rust
pnm::ArbitraryHeader {
    depth: 3,
    height: image.height(),
    width: image.width(),
    maxval: u16::MAX as u32,
    tupltype: Some(pnm::ArbitraryTuplType::RGB),
}
```

Taking a look at the source, I see no reason as to why only RGB8 & RGBA8 [are allowed](https://github.com/image-rs/image/blob/183e74ea7540da9ce11e613256f8f74640ea70c5/src/codecs/pnm/encoder.rs#L416-L417).

## Expected

This should work.

## Reproduction steps

Construct an 16-bit RGB image and then pass it to `PnmEncoder::write_image`, with the header above.

197g:
This was fixed in #2431

PR #2431, opened 2025-02-27
Add support for saving 16-bit images to pnm formats

### LINKED ISSUE #2430: Cannot save 16-bit images directly to PNM via save
This problem happens when trying to save a `DynamicImage` of color type `L16` or `Rgb16` to any PNM format (`.pbm`, `.pgm`, `ppm`) via its `save` method.

This is somewhat related with #1687, but it isn't the same because this problem also happens to grayscale images.

## Expected

The encoder should be able to recognize that the data coming from the `DynamicImage` object contains wide samples (2 bytes per sample), and write the image with that in mind. Other encoders such as the PNG encoder can handle this well.

## Actual behaviour

The PNM encoder API is funneled to the method `write_image`, which only accepts a `&[u8]` as the image buffer, and subsequently interprets the data as flat samples of `u8`, regardless of the color type in the original dynamic image object. As a consequence, the header dimension check fails with the error `DimensionMismatch` ("The Image's dimensions are either too small or too large") for finding twice the amount of expected samples.

## Reproduction steps

```rust
fn main() {
    let input = std::env::args().nth(1).unwrap();
    let output = input.clone() + ".pgm";

    let img = image::open(input).expect("failed to open");
    img.save(&output).expect("failed to save");
}
```

```none
❯ file basi0g16.png
basi0g16.png: PNG image data, 32 x 32, 16-bit grayscale, interlaced

❯ cargo run -- basi0g16.png 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running `target\debug\to-pnm.exe basi0g16.png`

thread 'main' panicked at src\main.rs:6:23:
failed to save: Parameter(ParameterError { kind: DimensionMismatch, underlying: None })
error: process didn't exit successfully: `target\debug\to-pnm.exe basi0g16.png` (exit code: 101)
```

Any 16-bit grayscale image should work. [Here is one](https://github.com/lunapaint/pngsuite/blob/main/png/basi0g16.png) from the pngsuite project.

Enet4:
Should hopefully resolve #2430, as well as #1687. It takes the assumption that when an image is 16-bit and the encoder receives a `&[u8]`, it means that the samples were in reality 16-bit samples in native endianness. An aligned copy of the input is made if the slice isn't 2-byte aligned.

### Summary

- add sample adaptation layer which recovers 16-bit samples from the original image buffer depending on the color type provided
   - Once this is done, the number of samples should match the expected number of samples, and the rest of the encoding process should simply work
- allow for RGB and RGB Alpha tuple types in header check
- add test suite `save_pnm.rs` and test sample images from pngsuite:
   - `basi0g16.png`: 16-bit grayscale image
   - `basn2c16.png`: 16-bit RGB image

### Known caveat

- I am not entirely sure whether the assumption can be made in general, or whether it would be possible for someone to request 8-bit samples to be extended and encoded as 16-bit through the encoder.
- While this should also enable the encoder to write PAM files, AFAIK PAM file decoding is not yet supported in the library.


fintelia:
Thanks for this PR! I am in favor of adding 16-bit support here.

One concern is that the `write_image` should not require that users pass aligned slices. If the color type is a 16-bit per pixel format then the user is allowed to pass any native endian pixel data without worrying about alignment. One option to handle potentially misaligned inputs is to allocate a temporary `Vec<u16>` and copy the provided data into it. That adds some memory overhead, but should be pretty fast to implement.

Enet4:
> One option to handle potentially misaligned inputs is to allocate a temporary `Vec<u16>` and copy the provided data into it. That adds some memory overhead, but should be pretty fast to implement.

I'll see what I can do. The encoder implementation cannot switch the lifetime of the samples midway, so the best place to re-adapt the byte buffer (with an allocation if necessary) would be as early as possible in the impl.

Enet4:
This has been rebased and the previous concerns have been addressed, so it is ready for another review. I cannot inspect the logs of the failing CI checks.

fintelia:
Looks good to me. Not quite sure why GitHub didn't rerun the CI tests when you pushed again

PR #2432, opened 2025-03-01
pnm: Decode PAM format variants which have alpha

mstoeckl:
This implements the transparent tupltypes listed in the PAM spec, at https://netpbm.sourceforge.net/doc/pam.html.

* I am submitting this PR to a) complete PAM format support so that image viewers based on `image` can read all
  reasonably constructable PAM images b) ensure that the `image` crate itself can read all the PAM tuple types that it
  writes. The format technically allows non-RGBA tuple types and depths > 4, but these should be rejected as they
  have no clear interpretation as an image.

* I've only added simple tests in-file for PAM, extending the existing ones, but if desired I can make regression test images.

* Other than by using `image` itself, PAM images with alpha channels can be produced by NetPBM and ImageMagick.
For example:
```
pngtopam -alphapam rgb_alpha.png > pam_rgb_alpha.pam
pngtopam -alphapam gray_alpha.png > pam_grayscale_alpha.pam
convert rgb_alpha.png pam_rgb_alpha.pam
convert gray_alpha.png pam_grayscale_alpha.pam
convert blackandwhite_alpha.png -depth 1 pam_blackandwhite_alpha.pam
```
rgb_alphagray_alphablackandwhite_alpha

(Note: `pngtopam` does not produce BLACKANDWHITE_ALPHA; `pamthreshold` can, but isn't general and appears to have bugs; so
ImageMagick's `convert` is the only standard converter which works. On the other hand, going from PAM to PNG,
ImageMagick's `convert` incorrectly flips the alpha channel, while `pamtopng` works correctly.)

* I wasn't able to find detailed documentation on the ColorType memory layout, but assume, based on how the other encoders and decoders handle alpha, that pixel types like `ColorType::Rgba16` are stored unpremultiplied, with native-endian channels in R,G,B,A memory order, and a linear alpha channel.

### DIFF
--- a/src/codecs/pnm/decoder.rs
+++ b/src/codecs/pnm/decoder.rs
@@ -221,10 +221,15 @@ impl Display for ErrorDataSource {
 enum TupleType {
     PbmBit,
     BWBit,
+    BWAlphaBit,
     GrayU8,
+    GrayAlphaU8,
     GrayU16,
+    GrayAlphaU16,
     RGBU8,
+    RGBAlphaU8,
     RGBU16,
+    RGBAlphaU16,
 }
 
 trait Sample {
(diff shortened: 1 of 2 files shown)


197g (approved):

197g:
The encoder also needs a bit of a rewrite to support this. With regards to pre-multiplied or not, iirc PNM is supposed to store it exactly how you hand it over. This is, it does not weigh in on the color space at all; its primary purpose was storing a numeric array, simply, to be transferable over any sort of IPC ("This format does not define the meaning of the data at any particular point in the array.")

Issue #2300, opened 2024-07-30
Implement a `rows()` iterator on `GenericImageView`


Shnatsel:
Right now the [`ImageBuffer`](https://docs.rs/image/0.25.2/image/struct.ImageBuffer.html) type has a `rows()` function to iterate over the rows of the image, but [`GenericImageView`](https://docs.rs/image/0.25.2/image/trait.GenericImageView.html) does not. 

Working with rows rather than individual pixels is often much faster, because it lets you leverage SIMD to process multiple pixels at once. For example, #2295 would be trivial to fix given this API.

fintelia:
What would be the return type of this function? `ImageBuffer::rows()` gets good performance between it is able to return a wrapper on top of `ChunksExact` that works really well with SIMD. But `GenericImageView` currently doesn't make any guarantee pixel layout, so it would presumably have to return something different

Shnatsel:
Ah, you are correct! With GenericImageView we have no guarantee that pixels in a row are laid out in memory contiguously.

We can however guarantee this about the `SubImage` type. I've updated the description to make this about SubImage.

fintelia:
The `SubImage` type is a wrapper around `SubImageInner` which is a wrapper around an `GenericImageView` (the struct itself doesn't have the trait bound, but the methods/impl blocks do have it).

I've gone in circles about this part of the API. First I think that it would be nice to constrain the layout of GenericImage. Then I realize that you can just use ImageBuffer directly if you want that, so I think about deprecating GenericImage. But that would break a lot of downstream code, so might as well just leave it around?

Shnatsel:
I don't think `GenericImageView` being *that* abstract makes a whole lot of sense. Per-pixels accesses will always be slow, so building an API solely around them feels rather pointless.

I think we need a type or a trait for a rectangular sub-image of an existing image, so that we could implement `rows()` on it. This is what's needed for operations like "make a copy of this region" (crop), "draw something in this region", etc. Naturally it can also be the entire image. We could change `GenericImageView` to be that, or add an entirely new type. Personally I'd rather change `GenericImageView`.

ripytide:
How often is image pixel data non-contiguous, and is it worth catering to the potential non-contiguous image use-cases at the cost of heavier genericness? Additionally, do crop and region drawing functions even need a `SubImage` type at all rather than simply taking a `Rect` type?

Perhaps we should simply remove/deprecate all the generic traits since they have dubious usefulness and could be misleading to users.

[...]

PR #2438, opened 2025-03-17
change invert() and add invert_in_place()

Its-Just-Nans:
This PR copy the design of `flip_horizontal` for `invert()`

Invert is the only inplace function

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -736,9 +736,15 @@ impl DynamicImage {
     }
 
     /// Invert the colors of this image.
+    /// use [DynamicImage::invert_in_place()] for an in-place version.
+    pub fn invert(&mut self) -> DynamicImage {
+        dynamic_map!(*self, ref p =>  imageops::invert(p))
+    }
+
+    /// Invert the colors of this image.
     /// This method operates inplace.
-    pub fn invert(&mut self) {
-        dynamic_map!(*self, ref mut p, imageops::invert(p));
+    pub fn invert_in_place(&mut self) {
+        dynamic_map!(*self, ref mut p, imageops::invert_in_place(p));
     }
 
     /// Resize this image using the specified filter algorithm.
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -3,10 +3,11 @@
 use num_traits::NumCast;
 
 use crate::color::{FromColor, IntoColor, Luma, LumaA};
+use crate::error::{ParameterError, ParameterErrorKind};
 use crate::image::{GenericImage, GenericImageView};
 use crate::traits::{Pixel, Primitive};
 use crate::utils::clamp;
-use crate::ImageBuffer;
+use crate::{ImageBuffer, ImageError};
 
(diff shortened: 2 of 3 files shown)


fintelia:
This is a breaking change, so at a minimum it would have to wait for our next major release.

However, I'm also not sure whether this change is worthwhile. Switching to the `*_in_place` version would be consistent with the other methods in the colorops module. But I'm concerned that anyone currently using the `invert` method would have their code continue to compile, but now give different results. Adding `#[must_use]` would at least generate a warning, but fixing it would still be manual effort

Its-Just-Nans:
I changed the design for no breaking change

Still I think that not having the same API for method is not great.

Maybe deprecated `invert()` could do the job

Its-Just-Nans:
I also think that a breaking change is a good option with the version bumped

awxkee:
I believe before adding any new processing methods [iterators](https://github.com/image-rs/image/issues/2300) at the very least need to get sorted out.
As for now we're technically encouraging users to use a highly inefficient implementation, even though we already know it is not efficient.
I think it is reasonable to _discourage_ as much as possible using plainly inefficient implementations.

fintelia:
We've been intermittently discussing a possible overhaul of the `GenericImage` trait [for at least five years](https://github.com/image-rs/image/issues/1118), so that alone doesn't need to block other progress.

What is this PR trying to achieve? My interpretation is that it is about adding consistency to the [colorops](https://docs.rs/image/latest/image/imageops/colorops/index.html) module. Currently, the `brighten`, `contrast`, and `huerotate` methods all have both cloning and `_in_place` versions. While the `dither` and `invert` methods operate in-place but don't have the name suffix, and none of the methods in that module have an `_to` version.

But stepping back, is having two versions of every method really the approach we'd pick starting from scratch? We'd probably just have the in-place versions, right? Which we could do in the next major release by dropping all the `_in_place` suffixes and making those the default. That is much less likely to be silent breakage: the caller would have to both be passing a mutable image _and_ never actually use the return value anywhere.

197g:
In the meantime, we do have _some_ evidence that in-place transformations (or rather in-allocation) are generally desirable: https://github.com/image-rs/image/discussions/2650. I think more people stumbled upon it but simply did not ask and accepted the allocation. That said there is a bit of design space on how to achieve them since it's correctly observed that we do need two separate APIs for aliasing reasons when we take `&mut _` at least. It's also not entirely clear if in-place refers to an output argument or modifying the input itself—which makes for actually three incompatible call interfaces.

----

If we spin this thread of thought further, we could possibly have `Rc`/`Arc` like images (at least, I found that relevant in my own designs) and that would be even more different interfaces. But that is a bit besides the point here, I would explicitly _avoid_ basing an argument around that.

---

So:

- `DynamicImage::invert(&self) -> DynamicImage`
- `DynamicImage::invert_in_place(&mut self)`
- `DynamicImage::invert_into(&self, _: &mut DynamicImage)`: opportunistically reuse the buffer or error, that is discussion for another day

Something as simple as `invert` seems like a comfortable way to at start with that variety without too much of a maintenance problem.

PR #2443, opened 2025-03-21
Add Turbofish convert

Its-Just-Nans:
Hi a little PR to use the turbofish syntax for conversion


fintelia:
Thanks for this PR!

My one concern would be whether this is a backwards compatibility hazard and would prevent us from adding more variants to `DynamicImage`. But after staring at it for a while I'm pretty sure that isn't a problem because `FromColor` isn't publicly exposed. Going to go ahead and merge

fintelia (approved):

PR #2444, opened 2025-03-21
add color type conversion

Its-Just-Nans:
Add a `from_pixel()` method

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -81,23 +81,6 @@ impl ColorType {
         let e: ExtendedColorType = self.into();
         e.channel_count()
     }
-
-    /// Returns the color type for a Pixel
-    pub fn from_pixel<P: Pixel, T>() -> Self {
-        match (P::CHANNEL_COUNT, size_of::<T>()) {
-            (1, 1) => ColorType::L8,
-            (1, 2) => ColorType::L16,
-            (2, 1) => ColorType::La8,
-            (2, 2) => ColorType::La16,
-            (3, 1) => ColorType::Rgb8,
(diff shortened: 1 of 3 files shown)


fintelia:
There's already the `PixelWithColorType` trait that gives access to the `ExtendedColorType` associated with a pixel. Though looks like we might be missing a way to convert from `ExtendedColorType` -> `ColorType`

Its-Just-Nans:
I've updated the code

fintelia:
Sorry for the delay. I think this would be better to expose by adding a method on `ExtendedColorType` that returns an `Option<ColorType>`.

src/color.rs
```diff
@@ -228,13 +231,51 @@ impl ExtendedColorType {
...
+        match *self {
+            ExtendedColorType::L8 => Some(ColorType::L8),
+            ExtendedColorType::La8 => Some(ColorType::La8),
+            ExtendedColorType::Rgb8 => Some(ColorType::Rgb8),
+            ExtendedColorType::Rgba8 => Some(ColorType::Rgba8),
+            ExtendedColorType::L16 => Some(ColorType::L16),
+            ExtendedColorType::La16 => Some(ColorType::La16),
+            ExtendedColorType::Rgb16 => Some(ColorType::Rgb16),
+            ExtendedColorType::Rgba16 => Some(ColorType::Rgba16),
+            _ => None,
```
fintelia, line 269: Don't forget `ExtendedColorType::{Rgb32F, Rgba32F}`

Its-Just-Nans: fixed

197g (approved):

PR #2451, opened 2025-04-15
Expose `enlargeable` trait

Glatzel:
This trait will allow user to restrict type or extend the `T` in `Rgb<T: Primitive Enlargeable>` and `Rgba<T: Primitive Enlargeable>`.

## example1

A simple example for allowing restrict type.

```rust
pub trait  MyPrimitive:
    image::Primitive + image::Enlargeable
{}
impl MyPrimitivefor u8 {}
impl MyPrimitivefor u16 {}
// This function only allow 8bit or 16bit image as input.
pub fn process<T>(src: &image::ImageBuffer<image::Rgb<T>, Vec<T>>)
{
    todo!();
}
```

## example2

This is what I used in camera raw image processing with `libraw`. `libraw` gives a pointer that could convert to u16 rgba (representing R,G,B,G2 pixel in sensor) imagebuffer after `raw2image()` and `subtract_black()`. But in custom raw image post processing for those camera with standard bayer sensor, knowing CFA pattern and a grayscale bayer image is satisfied and could reduce the memory usage.

```rust
pub trait  BayerPrimitive :
    image::Primitive + image::Enlargeable + std::marker::Send + std::marker::Sync
{}
impl BayerPrimitive for u16 {}
impl BayerPrimitive for f32 {}
// This function only allow 16bit or f32 image as output.
// 16bit for general photographic workflow.
// 32bit for calibration and analysis purpose.
// Other types like i32 which have negative values are not prctical for photographic workflow.
pub fn extract_bayer<T>(src: &image::ImageBuffer<image::Rgba<u16>, Vec<u16>>)
->&image::ImageBuffer<image::Rgb<T>, Vec<T>>
{
    let dst = ImageBuffer::from_par_fn(raw_img.width(), raw_img.height(), |x, y| {
            let pixel = raw_img.get_pixel(x, y);
            // Extract the max value in pixel and convert it to `T` type.
            // `T::from()` require the `T` impl `Send` and `Sync` trait.
            let value = T::from(pixel[0].max(pixel[1]).max(pixel[2]).max(pixel[3])).unwrap();
            image::Luma::<T>([value])
        });
    dst
}
```

### DIFF
--- a/src/codecs/avif/encoder.rs
+++ b/src/codecs/avif/encoder.rs
@@ -18,7 +18,7 @@ use crate::{ImageError, ImageResult};
 
 use bytemuck::{try_cast_slice, try_cast_slice_mut, Pod, PodCastError};
 use num_traits::Zero;
-use ravif::{Encoder, Img, RGB8, RGBA8};
+use ravif::{BitDepth, Encoder, Img, RGB8, RGBA8};
 use rgb::AsPixels;
 
 /// AVIF Encoder.
@@ -72,7 +72,7 @@ impl<W: Write> AvifEncoder<W> {
             .with_quality(f32::from(quality))
             .with_alpha_quality(f32::from(quality))
             .with_speed(speed)
-            .with_depth(Some(8));
+            .with_bit_depth(BitDepth::Eight);
 
         AvifEncoder { inner: w, encoder }
     }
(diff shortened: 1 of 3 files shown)


fintelia:
Could you say what this is trying to solve / what impact this has on the overall API?

Glatzel:
Hi, @fintelia . Sorry for missing the description. I have added it above.

fintelia:
Ah, so the goal is calling generically calling methods that are bounded by `Enlargeable` but not actually wanting to directly implement the trait or access its methods.

I created #2453 to try out having a public `PrimitiveExt` trait which lets you generically call methods that need the extra functionality, but not to use `Enlargeable` directly.

Glatzel:
I tried it in my repo, the `PrimitiveExt` does work.

Glatzel:
Close this pr as my requirement will be implemented in #2453

PR #2452, opened 2025-04-16
Update `ravif` to v0.11.12

RunDevelopment:
The latest version of `ravif` was published 2 days ago, and it deprecated certain APIs. Since `image` uses `#![deny(deprecated)]`, this causes CI to fail, as can be seen in recent PRs.

This PR fixes this by updating `ravif` to the latest version and using the new APIs.

---

If updating `ravif` is **not** wanted, I can also make a PR that just `#[allow]`s the now deprecated APIs to pass CI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ image-webp = { version = "0.2.0", optional = true }
 mp4parse = { version = "0.17.0", optional = true }
 png = { version = "0.17.11", optional = true }
 qoi = { version = "0.4", optional = true }
-ravif = { version = "0.11.11", default-features = false, optional = true }
+ravif = { version = "0.11.12", default-features = false, optional = true }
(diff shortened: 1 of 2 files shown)


fintelia (approved):

kornelski:
`deny(deprecated)` is fragile. Deprecations are supposed to help people avoid future breakage, not immediately cause one. 

Maybe make that conditional on `cfg_attr(test, `?

fintelia:
I don't think `deny(deprecated)` actually impacts downstream users? It should just apply when directly compiling this crate. Though our CI already runs (one job) that has `-D warnings` so I don't think it matters much whether deprecated lints are warnings or errors: they should fail CI either way. Which is what we want because without that, it would be infeasible to make sure that PRs didn't introduce new warnings

PR #2459, opened 2025-04-23
Linear light documentation

### LINKED ISSUE #2458: Resize operations do not respect gamma of the passed image
This happens in any `resize` operation

## Expected

Image is converted to linear colorspace before sampling and back converted before output.

## Actual behaviour

Image is sampled as is.

## Reproduction steps

Use any image and resample. In particular, I recommend using 

## See also:
http://www.ericbrasseur.org/gamma.html?i=2

awxkee:
Documentation about scene linear light assumption

Closes #2458

### DIFF
--- a/src/dynimage.rs
+++ b/src/dynimage.rs
@@ -847,6 +847,8 @@ impl DynamicImage {
     /// `sigma` is a measure of how much to blur by.
     /// Use [DynamicImage::fast_blur()] for a faster but less
     /// accurate version.
+    /// This method typically assumes that the input is scene-linear light.
+    /// If it is not, color distortion may occur.
     #[must_use]
     pub fn blur(&self, sigma: f32) -> DynamicImage {
         dynamic_map!(*self, ref p => imageops::blur(p, sigma))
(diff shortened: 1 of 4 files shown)


fintelia:
Could you add some blank lines to the doc comments so they get broken into multiple paragraphs? Right now this is what the `blur` docs look like:


image

It is also good practice to have a one sentence description of the function on the first line so that the module level summary page is more concise.

fintelia:
Also, apparently that "Image sampling" at the start of the description is coming from the doc comment on the pub use statement that exports the function:
https://github.com/image-rs/image/blob/ad962c05c7b4a276982317c5af68d8f408bb19e6/src/imageops/mod.rs#L18-L22

awxkee:
Done. I also did some minor documentation reformatting for the methods I changed.

If you want to build docs locally.
```bash
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features --no-deps --open
``` 

This also produces some warnings about incorrectly formatted docs, can fix that as well if that makes sense.

kornelski (approved):

kornelski (approved):

Issue #2245, opened 2024-05-26
Version 0.26.x


fintelia:
Tracking issue for items to consider for the 0.26 major version. One lesson from the prior release is that it is much easier if we have PRs ready _before_ the release window starts. 

# Planned

(none yet)

# Possible
* [ ] Remove `encode` method on other encoders (in because it is redundant with ImageEncoder::write_image)
* [ ] Pixel access changes:
  - #1853
 * [ ] Remove more deprecated methods:
   - `Pixel::{channels4, from_channels}`
   - `GenericImage::{get_pixel_mut, blend_pixel}`
 * [ ] Remove `GenericImage<Pixel = Rgba<u8>>` implementation for DynamicImage
   - Prevent confusion like in #1952
   - #2136 
* [ ] Remove `PngDecoder::with_limits` and fix `PngDecoder::set_limits`
  - #2084 
* [ ] Rename Reader to ImageReader
  - #2243

ripytide:
In #2243 you mentioned not making breaking changes to the main branch in case patch releases are needed, but leaving them as open PRs risks merge conflicts. How about we make a branch upon every release which can be used for making patch releases and then the default branch can be used as the "dev" branch including breaking changes?

fintelia:
@ripytide I should probably give more context on `image`'s release strategy...

This crate has been around for a very long time. I joined much later, but the crate actually started before Rust 1.0 and even before crates.io itself. Given the age it would make sense for us to have hit 1.0 and completely stabilized the API. However, that never happened. By the time I came on as a maintainer the sprawling scope of the public API and many lingering design questions were pretty daunting. At one point I tried spinning off part of the interface into its own [image-core crate](https://github.com/image-rs/image-core) to get that portion to 1.0 but that didn't pan out because the design didn't actually let us iterate the API.  

Instead, we've adopted the approach of treating each minor release as long-lived. The 0.22 series lasted a year, 0.23 lasted two years, and 0.24 lasted two more years. This strategy lets us make breaking changes when we identify better ways of doing something, while still providing stability. And that stability is evidently important given how slowly users upgrade: about two thirds of users still haven't upgraded to 0.25 several months after release. And we still get over 100,000 downloads/month of the 0.23 series that last released in early 2021!

Accordingly, we're still extremely early into the 0.25 series. Any new feature development should happen there for the foreseeable future

fintelia:
[...]

PR #2462, opened 2025-05-08
Add `no_std` Support

### LINKED ISSUE #1184: no_std support

bushrat011899:
# Objective

- Fixes #1184
- Supersedes #1868
- Supersedes #2436

## Solution

- Increased MSRV to 1.81 for `core::error::Error`
- Added `libm` and `std` features.
  - `std` provides access to the `std` crate and is required for most existing features.
  - `libm` activates `num-traits/libm`, allowing for existing `f32`/`f64` operations to work without `std`.
- Addressed several Clippy lints using `expect` rather than `allow`
  - This ensures stale exceptions don't remain
- Added a `no_std` CI task which checks against `wasm32v1-none`, a typical `no_std` target. This could be replaced with any other official target that does not have `std`.
- Added `#[cfg(feature = "std")]` to items which require `std`
- Added #[cfg(any(feature = "std", feature = "libm"))]` to items which require full `f32`/`f64` support.
- Made `ImageError` `non_exhaustive` to allow feature gating `ImageError::IoError` behind `std`. This isn't strictly required, but it ensures features are truly additive.
- Imported `num_traits::Float` or `num_traits::float::FloatCore` where required for `f32`/`f64` operations.

---

## Notes

- I have not attempted to add `no_std` support to any of the format features, as that will require careful consideration around how to decouple from `std::io`. For example, [`ciborium-io`](https://crates.io/crates/ciborium-io) (or an equivalent trait local to `image`) could be used as a `no_std`-specific feature.
- Increasing MSRV is only required for `core::error::Error`. Instead of the increase, a feature could be added to enable usage of `core::error::Error`. I opted to just increase MSRV for simplicity. 1.70 to 1.81 isn't too large of a jump IMO.
- This is my first time contributing to this project, please let me know if there's anything I can do to help!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -207,3 +207,15 @@ jobs:
         with:
           feature-group: default-features
           release-type: minor
+
+  clippy_no_std:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+      with:
+        components: clippy
+        targets: "wasm32v1-none"
+    - run: cargo clippy --no-default-features --target wasm32v1-none --features serde -- -D warnings
+      env:
+        SYSTEM_DEPS_DAV1D_BUILD_INTERNAL: always
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,14 +38,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ["1.70.0", nightly, beta]
+        rust: ["1.81.0", nightly, beta]
     steps:
     - uses: actions/checkout@v4
 
     - uses: dtolnay/rust-toolchain@nightly
-      if: ${{ matrix.rust == '1.70.0' }}
+      if: ${{ matrix.rust == '1.81.0' }}
     - name: Generate Cargo.lock with minimal-version dependencies
-      if: ${{ matrix.rust == '1.70.0' }}
+      if: ${{ matrix.rust == '1.81.0' }}
       run: cargo -Zminimal-versions generate-lockfile
 
     - uses: dtolnay/rust-toolchain@v1
@@ -58,7 +58,7 @@ jobs:
     - name: build
       run: cargo build -v
     - name: test
-      if: ${{ matrix.rust != '1.70.0' }}
+      if: ${{ matrix.rust != '1.81.0' }}
       run: cargo test -v && cargo doc -v
 
   test_other_archs:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 resolver = "2"
 
 # note: when changed, also update test runner in `.github/workflows/rust.yml`
-rust-version = "1.70.0"
+rust-version = "1.81.0"
 
 license = "MIT OR Apache-2.0"
 description = "Imaging library. Provides basic image processing and encoders/decoders for common image formats."
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ avif-native = ["dep:mp4parse", "dep:dav1d", "std"] # Enable native dependency li
 benchmarks = ["std"] # Build some inline benchmarks. Useful only during development (requires nightly Rust)
 serde = ["dep:serde"]
 std = ["byteorder-lite/std", "num-traits/std"]
+libm = ["num-traits/libm"]
 
 [[bench]]
 path = "benches/decode.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 bytemuck = { version = "1.8.0", features = ["extern_crate_alloc"] } # includes cast_vec
-byteorder-lite = "0.1.0"
-num-traits = { version = "0.2.0" }
+byteorder-lite = { version = "0.1.0", default-features = false }
+num-traits = { version = "0.2.0", default-features = false }
 
 # Optional dependencies
 color_quant = { version = "1.1", optional = true }
@@ -54,7 +54,7 @@ rgb = { version = "0.8.48", default-features = false, optional = true }
 tiff = { version = "0.9.0", optional = true }
 zune-core = { version = "0.4.12", default-features = false, optional = true }
 zune-jpeg = { version = "0.4.13", optional = true }
-serde = { version = "1.0.214", optional = true, features = ["derive"] }
+serde = { version = "1.0.214", default-features = false, optional = true, features = ["derive"] }
 
 [dev-dependencies]
 crc32fast = "1.2.0"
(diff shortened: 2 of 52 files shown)


bushrat011899:
Unsure how to proceed regarding the `cargo-semver-checks` CI failure. I believe increasing the version number from 0.25 to 0.26 should be enough, but that didn't appear to work when I tested it locally.

fintelia:
Thanks for putting this together! Seeing how much of the crate works with no_std (including `ImageBuffer`, `DynamicImage`, `imageops`, etc.) I do think this is worth doing.

Figuring out how to replace `std::io::Write/Read` is really the limiting factor for having encoding/decoding with no_std. I'm not sure there even _is_ a way to do so without either making the feature non-additive or using non-standard traits for both std and no_std. At some point, the few remaining blockers to getting alloc::io::Error into the standard library might end up getting sorted out, which would be the ideal outcome.

As for breaking changes, [the plan is for the next major release to be 1.0](https://github.com/image-rs/image/issues/2316). We're tracking all the possible/planned changes for that release in https://github.com/image-rs/image/issues/2245.

bushrat011899:
> Thanks for putting this together! Seeing how much of the crate works with no_std (including `ImageBuffer`, `DynamicImage`, `imageops`, etc.) I do think this is worth doing.

Exactly my opinion. I've helped add `no_std` support to a few crates now, and it can be surprising just how much functionality exists just within `core` and `alloc`.

> Figuring out how to replace `std::io::Write/Read` is really the limiting factor for having encoding/decoding with no_std. I'm not sure there even _is_ a way to do so without either making the feature non-additive or using non-standard traits for both std and no_std.

I have been experimenting with this issue, and I think there are two paths forward:

1. Update `byteorder-lite` upstream (I understand it's controlled by `image-rs` so this shouldn't be too painful) to abstract the `std::io` traits using the same technique as `ciborium-io`. The trick it uses is an additive feature allowing a single trait to be used by the API (`ciborium_io::Read`/`Write`) which can always be implemented by consumers (and has provided implementations for the relevant `core` and `alloc` types), but will also be blanket implemented for any type implementing the corresponding `std::io` trait.
2. Use a sans-io approach as the internal implementation, and then expose it for external consumers. This would be a larger change to the API, but might have less code overall.

I personally think path 1 is the more viable option as it'd keep both internal and external usage of `image` largely the same. 

> At some point, the few remaining blockers to getting alloc::io::Error into the standard library might end up getting sorted out, which would be the ideal outcome.

This would be the ideal, completely agree. But I'm not sure how far away that future is at this point.

> As for breaking changes, [the plan is for the next major release to be 1.0](https://github.com/image-rs/image/issues/2316). We're tracking all the possible/planned changes for that release in #2245.

Oh well that's good timing! Adding an `std` feature to control this functionality would've been a breaking change, hard to add after 1.0! I would say that if we can't get more functionality working in `no_std` before 1.0, it could be added in 1.1 as long as we're able to keep the public interfaces compatible.

bushrat011899:
[color_quant/#24](https://github.com/image-rs/color_quant/pull/24) adds `no_std` support which is relevant to this PR. I would recommend trying to get this merged.

fintelia:
> > Figuring out how to replace `std::io::Write/Read` is really the limiting factor for having encoding/decoding with no_std. I'm not sure there even _is_ a way to do so without either making the feature non-additive or using non-standard traits for both std and no_std.
> 
> I have been experimenting with this issue, and I think there are two paths forward:
> 
> 1. Update `byteorder-lite` upstream (I understand it's controlled by `image-rs` so this shouldn't be too painful) to abstract the `std::io` traits using the same technique as `ciborium-io`. The trick it uses is an additive feature allowing a single trait to be used by the API (`ciborium_io::Read`/`Write`) which can always be implemented by consumers (and has provided implementations for the relevant `core` and `alloc` types), but will also be blanket implemented for any type implementing the corresponding `std::io` trait.
> 
> 2. Use a sans-io approach as the internal implementation, and then expose it for external consumers. This would be a larger change to the API, but might have less code overall.
> 
> I personally think path 1 is the more viable option as it'd keep both internal and external usage of `image` largely the same.

This is a very big topic so it probably warrants its own thread. Depending on the details, it could require simultaneous breaking changes across a dozen or more crates, many of which aren't controlled by `image-rs`. Additionally, a big benefit of the current API is that the traits from `std::io` are already familiar for most Rust users. That means lower friction and an easier time for beginners to get help if they run into difficulty. We also have to be mindful of performance, and make sure that any changes don't regress that aspect.

HeroicKatora:
Since you've done the std transform in many crates, one central question is how are we to maintain this? As in how does this affect the ability to accept patches and new features.

- It seems (with regards to `libm` and feature flags) that we have to tread somewhat careful not to populate the dependency tree with anything inconvenient as to say. How exactly is inconvenient defined and importantly, will we catch this in CI or manually?
- The fallout from `std::io` was already mentioned. But also importantly, I disagree with the statement that the referenced implementation actually delivers what is claimed. Enabling the `std` feature in that crate adds a blanket implementation. Adding blanket implementations is a major breaking change according to the Rust semver specification. In particular, such a blanket implementation will _conflict_ with any manual implementations on the same type. Thus all such no-std impls must be cfg-restricted to apply *only* when the `ciborium_io/std` feature is disabled. This works in the crate itself, but for reader wrappers outside, hardly. Tooling support for catching this just doesn't exist and it may result in compilation failures from a sibling dependency on `ciborium_io/std` if any such type is defined.
- All kinds of locks must be changed `spin` iirc? Any downsides to this, in particular for the `std` users.
- Will there come a point where we need to tradeoff performance vs. no-std support? If so, is there a good guidance on how  one can achieve this correctly with a seamless experience when `std` is defined? It's already somewhat hard to test performance comprehensively.

bushrat011899:
> Since you've done the std transform in many crates, one central question is how are we to maintain this? As in how does this affect the ability to accept patches and new features.

Excellent question. The aspirational solution here is to make the `no_std` subset of the crate useful enough for enough people. For example, `serde` is `no_std` and a large number of its users are likely unaware that disabling default features and enabling `derive` means they're a `no_std` user.

Since the only way to use `std` in a crate is to either _not_ be `#![no_std]` _or_ include `extern crate std;` explicitly, compilation on a `no_std` target is largely sufficient when combined with `std`-enabled testing. This PR includes that in CI against `wasm32v1-none`, likely to be the most important single `no_std` platform going forward (as it's the only way to have high-compatibility WASM builds).

>     * It seems (with regards to `libm` and feature flags) that we have to tread somewhat careful not to populate the dependency tree with anything inconvenient as to say. How exactly is inconvenient defined and importantly, will we catch this in CI or manually?

Right now, any non-compliant dependencies (including transitively) _will_ be caught at compile time in CI _without_ exception. `no_std` is an explicit declaration that must be evaluated at compile-time, so no "runtime" testing is required to assure this.

>     * The fallout from `std::io` was already mentioned. But also importantly, I disagree with the statement that the referenced implementation actually delivers what is claimed. Enabling the `std` feature in that crate adds a blanket implementation. Adding blanket implementations is a major breaking change according to the Rust semver specification. In particular, such a blanket implementation will _conflict_ with any manual implementations on the same type. Thus all such no-std impls must be cfg-restricted to apply _only_ when the `ciborium_io/std` feature is disabled. This works in the crate itself, but for reader wrappers outside, hardly. Tooling support for catching this just doesn't exist and it may result in compilation failures from a sibling dependency on `ciborium_io/std` if any such type is defined.

After experimenting I do agree that the trait would _probably_ need to be sealed (or potentially marked `unsafe`? Worth considering). The semver-break comes from implementing _both_ `ciborium_io::Write` _and_ `std::io::Write` when `ciborium_io/std` is enabled. I would say that we might not _need_ to poly-fill `std::io`. For example, some methods in `image` don't accept a generic `std::io` type (e.g., `dyn Read`), they instead accept a closure for the particular method they need, similar to a C-style V-Table object. Instead of using this technique ad-hoc, maybe it could be formalised and exposed to end-users as a sort-of "raw" version of the recommended `std::io`-based API.

>     * All kinds of locks must be changed `spin` iirc? Any downsides to this, in particular for the `std` users.

It can be a zero-cost option pretty easily. I created [`bevy_platform`](https://crates.io/crates/bevy_platform) for Bevy as a way to handle `std::sync` (and others). It provides a consistent API for the sync primitives and features control the back-end implementation (`std`, `spin`, and we've even discussed `parking_lot`). Since it's internal to the API there's no semver concern here. This is the same technique used by `wgpu` to provide either `std`, `parking_lot`, or their own custom dead-lock detecting backend, so lots of precedent here.

>     * Will there come a point where we need to tradeoff performance vs. no-std support?

In my experience, no. `no_std` lends itself to simplicity (since you don't have the `std` crate to provide that baseline complexity boost), and away from syscalls/allocation entirely. Since embedded devices (the prime audience for `no_std`) are typically so resource constrained, their libraries are designed for maximum performance. So far, this has been my experience in `wgpu`, `tracing`, and Bevy as some examples.

>       If so, is there a good guidance on how one can achieve this correctly with a seamless experience when `std` is defined? It's already somewhat hard to test performance comprehensively.

I would say that _if_ this is a concern, the normal feature guidelines apply; as long as features are truly additive testing shouldn't be too onerous.

bushrat011899:
As a proof-of-concept for wider `no_std` support, I've updated the `qoi` feature to be as-complete-as-possible in `no_std`, short of much larger changes to the rest of the public API. I want to bring particular attention to the `QoiEncoder` type, as that highlights the way a `no_std` API can be clean without a high maintenance burden or challenging feature flags.

I think the clearest path forward at this point would be to use the same technique employed by `qoi`, with privately implemented [`Write`](https://github.com/aldanor/qoi-rust/blob/56697191e46eb3987a730cfd0822c18ee4c8041f/src/utils.rs#L27-L31), [`Read`](https://github.com/aldanor/qoi-rust/blob/56697191e46eb3987a730cfd0822c18ee4c8041f/src/decode.rs#L228-L231), etc. traits which will either blanket-impl when `std` is enabled, or only be implemented for some subset of types (e.g., `Vec<u8>`) which would already be covered by the blanket-impl.

This is based on `ImageReader::make_decoder`, which I think highlights a critical minimum API for reading images (`BufRead + Seek`).

fintelia:
I think looking at the QOI codec might give a false impression of the work involved in moving encoding/decoding over to no_std. Others are far more complicated. The TIFF and WebP decoders do a bunch of seeking around the file. The png and tiff crates rely on `flate2` which doesn't support no_std. etc.

QOI, by constrast, is a comparatively simple format and already exposes a no_std interface. And if you look into the details, they actually have separate implementations for the [std::Read decoder](https://github.com/aldanor/qoi-rust/blob/master/src/decode.rs#L142-L208) and [byte-slice decoder](https://github.com/aldanor/qoi-rust/blob/master/src/decode.rs#L25-L93).

bushrat011899:
@HeroicKatora and @fintelia, I've opened [image-gif/#200](https://github.com/image-rs/image-gif/pull/200) to not just add `no_std` support to one of `image`'s dependencies, but to also test out a general approach to replacing `std::io` in a zero-compromises way. Would appreciate some eyes there, as the discussions here and there are closely related.

197g:
Since is inherently hard to rebase and we don't plan it in `no_std` (several formats make it impossible anyways plus we'll use some `fs` constructs like `PathBuf`/`OsStr`) I'll close this. Thank you for the effort and the follow-ups in individual decoders. It's still a long-term goal but not before the IO stuff settles a bit and the library goes deeper into operations that are sans-IO.

PR #2463, opened 2025-05-10
AVIF: YCgCo

awxkee:
Here is YCgCo decoder implementation.
ITU-R CICP spec [link](https://www.itu.int/rec/T-REC-H.273/en).

This is YCgCo only implementation. YCgCo-R de-facto is not used in AVIFs.
libavif introduced YCgCo-Re and YCgCo-Ro types (matrix coefficients 16 and 17), but dav1d-rs does not support them at the moment. They are more complex due to bit-depth shifts, so their implementation is postponed for the future.

Note: eq. 27-29, colors range in YCgCo handled differently than in YCbCr.

[Here](https://github.com/awxkee/image/tree/ycgco_samples/samples) is valid YCgCo images to try out. I don't know software that could open them correctly, except Chrome and avifdec ( part of libavif ).

I spent few hours trying to figure out why colors didn't match Chrome after simple decoding those images, because I already did it and everything matched perfectly. I tried a basic re-encoding and noticed visible color shifts. After several hours I realized that issue is not in YCgCo paths, but in any AVIF file. There's almost definitely something wrong in dav1d-rs, or in image encoding path, or both.

### DIFF
--- a/src/codecs/avif/yuv.rs
+++ b/src/codecs/avif/yuv.rs
@@ -706,7 +706,7 @@ fn process_halved_chroma_row_cbcr<
 fn process_halved_chroma_row_cgco<
     V: Copy + AsPrimitive<i32> + 'static + Sized,
     const PRECISION: i32,
-    const CN: usize,
+    const CHANNELS: usize,
     const BIT_DEPTH: usize,
 >(
     image: YuvPlanarImage<V>,
@@ -728,12 +728,12 @@ fn process_halved_chroma_row_cgco<
     let chroma_size = (image.width + 1) / 2;
     let u_plane = &image.u_plane[0..chroma_size];
     let v_plane = &image.v_plane[0..chroma_size];
-    let rgba = &mut rgba[0..image.width * CN];
+    let rgba = &mut rgba[0..image.width * CHANNELS];
 
     let bias_y = range.bias_y as i32;
     let bias_uv = range.bias_uv as i32;
     let y_iter = y_plane.chunks_exact(2);
-    let rgb_chunks = rgba.chunks_exact_mut(CN * 2);
+    let rgb_chunks = rgba.chunks_exact_mut(CHANNELS * 2);
 
     let r_coef = ((max_value as f32 / range.range_y as f32) * (1 << PRECISION) as f32) as i32;
 
@@ -748,12 +748,12 @@ fn process_halved_chroma_row_cgco<
         let b = qrshr::<PRECISION, BIT_DEPTH>((t0 - co_value) * r_coef);
         let g = qrshr::<PRECISION, BIT_DEPTH>((y_value0 + cg_value) * r_coef);
 
-        if CN == 4 {
+        if CHANNELS == 4 {
             rgb_dst[0] = r.as_();
             rgb_dst[1] = g.as_();
             rgb_dst[2] = b.as_();
             rgb_dst[3] = max_value.as_();
-        } else if CN == 3 {
+        } else if CHANNELS == 3 {
             rgb_dst[0] = r.as_();
             rgb_dst[1] = g.as_();
             rgb_dst[2] = b.as_();
@@ -769,12 +769,12 @@ fn process_halved_chroma_row_cgco<
         let b = qrshr::<PRECISION, BIT_DEPTH>((t1 - co_value) * r_coef);
         let g = qrshr::<PRECISION, BIT_DEPTH>((y_value1 + cg_value) * r_coef);
 
-        if CN == 4 {
+        if CHANNELS == 4 {
             rgb_dst[4] = r.as_();
             rgb_dst[5] = g.as_();
             rgb_dst[6] = b.as_();
             rgb_dst[7] = max_value.as_();
-        } else if CN == 3 {
+        } else if CHANNELS == 3 {
             rgb_dst[3] = r.as_();
             rgb_dst[4] = g.as_();
             rgb_dst[5] = b.as_();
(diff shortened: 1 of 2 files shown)


HeroicKatora (commented):
YCgCo is just a color model, not an absolutely defined system. It could be that any part along the way defines a different whitepoint, RGB primiaries or even some transfer functions that are misinterpreted. There are probably some parameters encoded in the avif that are not being utilized by the decoding path.

All that being said; this is going to happen more often with more recent formats until non-rgb representations are better. Doing more conversions to fit everything into the sRGB box is lossy and expensive. With a bit of luck, I'll be able to spend significant amounts of time on aligning that the rest of the year.. So no need to worry about in the PR immediately.

src/codecs/avif/yuv.rs
```diff
@@ -201,6 +201,7 @@ pub(crate) enum YuvStandardMatrix {
     Smpte240,
     Bt470_6,
     Identity,
+    CgCo,
```
HeroicKatora, line 204: I'm really not sure what this is supposed to mean. That enum defines the combined transfer matrix you get when combining a specific primary with the Yuv encoding, so that you have coefficients for directly computing RGB' from YCgCr (see eq 39, 66-71 n ITU H.273). Those coefficients are undefined for YCoCg (Matrix coefficients = 8, eq 51-54)?

 Instead, a new enum of `YCgCoStandardMatrix` might be the way to go if we want to go directly to linear

awxkee:
> There are probably some parameters encoded in the avif that are not being utilized by the decoding path.

Ah, yes, I embedded Display P3 profile.

src/codecs/avif/yuv.rs
awxkee: That's right. I wanted to avoid introducing a new subset of methods or enums.
Adding more methods or enums would make the decoder more complex, and last time, people weren’t very happy with a decoder that had tricky logic.
Instead somewhat vague definition in the standard matrix, it would make the decoder even more complicated to support additional YCgCo methods.
If you think that something like this is better, then I'll do it. However, it won't work for YCgCo-R/YCgCo-Re/YCgCo-Ro, because they cannot fit exactly in the same data size. 
```rust
if matrix == YCgCo {
      let worker = match self.picture.pixel_layout() {
                PixelLayout::I400 => unreachable(),
                PixelLayout::I420 => ycgco420_to_rgba8,
                PixelLayout::I422 => ycgco422_to_rgba8,
                PixelLayout::I444 => ycgco444_to_rgba8,
            };

            worker(image, buf, yuv_range, color_matrix)?;
} else {
      let worker = match self.picture.pixel_layout() {
                PixelLayout::I400 => yuv400_to_rgba8,
                PixelLayout::I420 => yuv420_to_rgba8,
                PixelLayout::I422 => yuv422_to_rgba8,
                PixelLayout::I444 => yuv444_to_rgba8,
            };

            worker(image, buf, yuv_range, color_matrix)?;
}
```

HeroicKatora (commented):
I like that modelling much better :+1:  The document is unfortunately complex, so it isn't exactly surprising that there would be some necessary complexity in processing the data. Integrations like this should hopefully become simpler at some point. My biggest headache in understanding any code at a glance is that it is never _locally_ obvious what the interpretation of some numbers should be i.e. is `r` linear or a signal? etc, obviously complicated by the fact that all is writing to the same target for efficiency which is at odds with encoding such information as a static type.

Imo, this actually does a decent job of abstracting away the processing paths exhaustively.

awxkee:
H.273 defines only matrix coefficients in equations 10 and 13 as operating on linear intensities to compute Y, so any other matrix is not linear. Thus, for any other matrix, transfer characteristics should be applied.

However, the standard does not specify whether it is correct to apply linear transfer characteristics in those cases.

HeroicKatora:
I believe it says that eq 20-22 apply in these cases, and eq 45, 76, 82, 85 give the non-linear E_y signal in all other cases where a luminance is involved as derived form the non-linear rgb/lms signals.

src/codecs/avif/yuv.rs
```diff
@@ -610,6 +703,125 @@ fn process_halved_chroma_row<
     }
 }
 
+fn process_halved_chroma_row_cgco<
+    V: Copy + AsPrimitive<i32> + 'static + Sized,
+    const PRECISION: i32,
+    const CN: usize,
```
HeroicKatora, line 709: I shortly pondered if could make a static assertion here to clarify `CN in {3,4}` but then noticed that all my attempted would skyrocket the wtf-rate-per-minute so let's not. We might do it if bounds on const-expressions get implemented in Rust.

HeroicKatora, line 709: In many other instances this is spelled out:

```suggestion
    const CHANNELS: usize,
```

HeroicKatora (approved):
Actually seems quite straightforward, very similar style to the other impls but using less parameters as the transform is fixed. If anything it's hard to review because everything looks similar (but not quite the same) :+1: We'll see about the color profile issue in the future, hopefully.

awxkee:
I think the issue with color profiles was my fault — my own software correctly recognized the tagged image, then it embedded a valid profile, but software author wasn't able to figure it out :)

As for readability yes, before adding something else, it's likely that this needs to be split into a few modules. IMHO, adding traits to make it dynamic or sophisticated static dispatch will make it even more complex. I'm not sure what else here to do, because transform is more or less straightforward, and it is hard to simplify it further. I even intentionally skipped ycgco420/422 full range simplified implementation to avoid code bloating. 

If anything I did [cms](https://github.com/awxkee/moxcms). I can't say it is full implementation of ICC v4, but it should work for the most of web images, and it is really fast on mobile devices what is the most importance to me. We've integrated it into jxl-oxide, and everything seems to be going well. So if you're planning to implement profile support, you could copy the biggest part of it, except hot paths it is safe Rust.

src/codecs/avif/yuv.rs
awxkee: I'm not sure this is important to make asserts in such deeply private methods. Anyways the method that calls it have assert to ensure that channels are valid.

197g:
Given your comment and the commit after the approval, I wasn't entirely sure either if there may be something you still wanted to cleanup. But I'll take this question as a no.

awxkee:
Ah, no. I'll do separate PR for cleaning up

awxkee:
ycgco420_to_rgba8 invokes invalid executor. I'll do tomorrow PR with refactorings, if it won't this is reminder that this one still needs to be fixed.

PR #2466, opened 2025-05-16
YCgCo refactor

awxkee:
Attempt to refactor YUV to reduce complexity.
This is probably not something you see every day. However, from my perspective, this approach allows methods to be added in a cleaner and more standartized way.

Feel free to reject it, if this is not something you want to see.

### DIFF
--- a/src/codecs/avif/ycgco.rs
+++ b/src/codecs/avif/ycgco.rs
@@ -1,0 +1,419 @@
+use crate::codecs::avif::yuv::{
+    check_rgb_preconditions, check_yuv_plane_preconditions, qrshr, yuv420_to_rgbx_invoker,
+    yuv422_to_rgbx_invoker, CbCrInverseTransform, HalvedRowHandler, PlaneDefinition,
+    YuvChromaRange, YuvIntensityRange, YuvPlanarImage, YuvStandardMatrix,
+};
+use crate::ImageError;
+use num_traits::AsPrimitive;
+
+/// Computes YCgCo inverse in limited range
+/// # Arguments
+/// - `dst` - dest buffer
+/// - `y_value` - Y value with subtracted bias
+/// - `cb` - Cb value with subtracted bias
+/// - `cr` - Cr value with subtracted bias
+#[inline(always)]
+fn ycgco_execute_limited<
+    V: Copy + AsPrimitive<i32> + 'static + Sized,
+    const PRECISION: i32,
+    const CHANNELS: usize,
+    const BIT_DEPTH: usize,
+>(
+    dst: &mut [V; CHANNELS],
+    y_value: i32,
+    cg: i32,
+    co: i32,
+    scale: i32,
+) where
+    i32: AsPrimitive<V>,
+{
+    let t0 = y_value - cg;
+
+    let r = qrshr::<PRECISION, BIT_DEPTH>((t0 + co) * scale);
+    let b = qrshr::<PRECISION, BIT_DEPTH>((t0 - co) * scale);
+    let g = qrshr::<PRECISION, BIT_DEPTH>((y_value + cg) * scale);
+
+    if CHANNELS == 4 {
+        dst[0] = r.as_();
+        dst[1] = g.as_();
+        dst[2] = b.as_();
+        dst[3] = ((1i32 << BIT_DEPTH) - 1).as_();
+    } else if CHANNELS == 3 {
+        dst[0] = r.as_();
+        dst[1] = g.as_();
+        dst[2] = b.as_();
+    } else {
+        unreachable!();
+    }
+}
+
+/// Computes YCgCo inverse in full range
+/// # Arguments
+/// - `dst` - dest buffer
+/// - `y_value` - Y value with subtracted bias
+/// - `cb` - Cb value with subtracted bias
+/// - `cr` - Cr value with subtracted bias
+#[inline(always)]
+fn ycgco_execute_full<
+    V: Copy + AsPrimitive<i32> + 'static + Sized,
+    const PRECISION: i32,
+    const CHANNELS: usize,
+    const BIT_DEPTH: usize,
+>(
+    dst: &mut [V; CHANNELS],
+    y_value: i32,
+    cg: i32,
+    co: i32,
+) where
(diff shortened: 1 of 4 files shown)


src/codecs/avif/yuv.rs
```diff
@@ -1418,7 +1067,7 @@ where
 
     let range = range.get_yuv_range(BIT_DEPTH as u32);
     let kr_kb = matrix.get_kr_kb();
-    const PRECISION: i32 = 11;
+    const PRECISION: i32 = 13;
```
197g, line 1070: This seems like a subtle change. What's happened here? There are a bunch of places that basically appear renamed from `ycbcr` to `ycgco` from the way git does patches, but this isn't one of them. (ycgco444_to_rgbx_impl had defined PRECISION as 13 previously).

awxkee: I changed this to use the same Q2.13 format for all implementations. In practice, we only support YUV formats with up to 12-bit depth, and 2 integer bits are sufficient to represent any YUV matrix. Since i32 can store up to 31 bits, the available space for the fractional part is:
31 (total bits) - 12 (bit depth) - 2 (integer part) = 17.
This means any Q2.N format with a N part less than 17 is safe. (Technically, 17 could also be safe, but rounding during accumulation could cause an overflow.).
13 bits from my view just doing a little better precision without any notable changes. Even if here 16 bit YUV planes will be submitted it is safe because `16 + 13 = 29` bits that still have safe margin from 31.

awxkee: As I already noted, changing storage type could help with easy optimizations, we could use i16 everywhere just to let compiler know that 15 bits is enough. At some common point it will at least double throughput on major platforms. If any plans to keep here performance in lane. However, then it won't be so easy to add 16 bit YUV here.

src/codecs/avif/ycgco.rs
```diff
@@ -0,0 +1,368 @@
...
+
+/// Computes YCgCo inverse in limited range
+///
+/// # Arguments
+/// - `dst` - dest buffer
+/// - `z` - offset in dest buffer
+/// - `y_vl` - Y value with subtracted bias
+/// - `cg` - Cg value with subtracted bias
+/// - `co` - Co value with subtracted bias
+macro_rules! ycgco_execute_limited {
```
fintelia, line 17: I generally prefer to avoid macros where possible because they lose type information and can hurt readability. How feasible would it be to make this a generic function rather than a macro?

awxkee: Generic functions will add unwanted bounds checks. If readability is a goal then okay, but if there is a plan to keep at reasonable fast: only macro, or plain method is possible

awxkee: I actually don’t have a preferences. Will do as you say

fintelia: Don't convert everything over just yet. Want to understand the trade offs first

awxkee: When you're using macros it essentially "part of context" it is even not a real "code", so if compiler knows at some block that bounds checks is unnecessary here, it won't insert bounds checks in all the cases. 
On the methods side context is switched and compiler "knoweldge" about previous block is erased and compiler needs to make decision "are bounds checks needed here?" again. Thus, on at some point it could decide that bounds check are needed, or disable inlining ( if inlining is disabled bounds checks inserted 100% ) by making decision based on opt-level and LTO settings.

awxkee: Thus code below have a change to be succesfully inlined and everything will work completely the same.
And there is a change that compiler will decide that "too many generics will cause code bloat", by opt-level or LTO setting and simply will disable inlining.
So inlined method may work the same, or may not.

```rust
#[inline]
fn ycgco_execute_limited<
    V: Copy + AsPrimitive<i32> + 'static + Sized,
    const PRECISION: i32,
    const CHANNELS: usize,
    const BIT_DEPTH: usize,
    const OFFSET: usize,
>(dst: [V; CHANNELS], y_value: i32, cg: i32, co: i32, scale: i32) {
    let t0 = y_value - cg;

    let r = qrshr::<PRECISION, BIT_DEPTH>((t0 + co) * scale);
    let b = qrshr::<PRECISION, BIT_DEPTH>((t0 - co) * scale);
    let g = qrshr::<PRECISION, BIT_DEPTH>((y_value + cg) * scale);

    if CHANNELS == 4 {
        dst[OFFSET] = r.as_();
        dst[OFFSET + 1] = g.as_();
        dst[OFFSET + 2] = b.as_();
        dst[OFFSET + 3] = ((1i32 << BIT_DEPTH) - 1).as_();
    } else if CHANNELS == 3 {
        dst[OFFSET] = r.as_();
        dst[OFFSET + 1] = g.as_();
        dst[OFFSET + 2] = b.as_();
    } else {
        unreachable!();
    }
}
````

File: src/codecs/avif/ycgco.rs [hunk 3]
```diff
@@ -0,0 +1,368 @@
...
+            let co_value = v_src.as_() - bias_uv;
+
+            ycgco_execute_limited!(rgb_dst, 0, y_value, cg_value, co_value, scale_coef);
+        }
+    }
+}
+
+/// Converts YCgCo 444 planar format to Rgba
+///
+/// Stride here is not supports u16 as it can be in passed from FFI.
```
fintelia, line 143: nit: I think there's a typo here

src/codecs/avif/ycgco.rs
fintelia: I think `#[inline(always)]` should be sufficient to pretty much guarantee inlining. Could you give that a try?

awxkee: Macros bench vs inlined method shows diff about 2%, however asm looks ok. I think it is likely just noise.

src/codecs/avif/yuv.rs
197g: There aren't any concrete plans and it sounds easier when everything is more homogeneous in the usage of the fixed point format. Thank you for explaining.

fintelia (approved):

Issue #2469, opened 2025-05-24
TIFF and EXR decoders slow down compilation


fintelia:
Along the same lines as #2468, the `load_from_memory` method causes all of our encoders to be monomophized regardless of whether users actually need to decode images from an in memory buffer. Unfortunately, due to the method signature involved we cannot use the same trick to avoid it.

Luckily, the compile time impact of building decoders is smaller than for encoders at ~1.5 seconds on my machine. And there may be room to improve it further since it turns out that the bulk of the compile time can be attributed to two encoders. Preventing the TIFF and EXR decoders from being used during `load_from_memory` (but still exposing them via the lower-level `codecs` module) each save ~0.5 seconds from the release build time. Validated by commenting out these lines...

https://github.com/image-rs/image/blob/9e357b0df798480f43c74e04ab4a8bbdce7b5549/src/image_reader/image_reader_type.rs#L160-L161

https://github.com/image-rs/image/blob/9e357b0df798480f43c74e04ab4a8bbdce7b5549/src/image_reader/image_reader_type.rs#L172-L173

PR #2470, opened 2025-05-24
Make `load_from_memory` generic to improve compile times

fintelia:
Currently, the `load_from_memory[_with_format]` methods cause rustc to generate code for them when compiling the image crate, even if downstream code never decodes images from in-memory buffers. This PR changes them to be generic so that they only get monomorphized only if they are actually used. 

Reduces time for release builds by ~1.2 seconds on my machine.

See also: #2468, #2469


197g:
I didn't realize `load_from_memory_with_format` was so heavy on compiletime, 1.2 seconds is *a lot*. Nice find.

PR #2473, opened 2025-05-25
Fixes TGA orientation

### LINKED ISSUE #2465: TGA: right orientation is loaded mirrored
The following image is rendered mirrored

https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/master/tests/test-images/reftests/tga/gtk-logo-rle-24bpp-bottom-right.tga

Left should be correct, right is image-rs

Image

awxkee:
Closes #2465 

Previously orientation was invalid because vertical mirroring was performing horizontal mirroring in-place, which was not necessary.
Also tried to do more advanced orientation checking. Information about this https://www.dca.fee.unicamp.br/~martino/disciplinas/ea978/tgaffs.pdf.

It seems to me correct table is:
| Screen Destination of First Pixel | Bit 5 (Vertical Origin) | Bit 4 (Horizontal Origin) |
|----------------------------------|--------------------------|----------------------------|
| Bottom Left                      | 0                        | 0                          |
| Bottom Right                     | 0                        | 1                          |
| Top Left                         | 1                        | 0                          |
| Top Right                        | 1                        | 1                          |


awxkee:
Just to note:
```rust
    let img = ImageReader::open("./assets/gtk-logo-rle-24bpp-bottom-right.tga")
        .unwrap()
        .decode()
        .unwrap();
    img.save("bottom_left.tga").unwrap();
```
It saves completely black image, png does this correctly. Not sure if this is intented behaviour.

awxkee:
I think it's a bug, but not in the image itself. I encoded it again, just to confirm and tried converting it using ImageMagick instead of Apple Preview, and everything worked. Therefore, it seems that sometimes the encoder produces an image that Apple Preview can't handle correctly. Since Apple Preview isn't the best reference, I think we could ignore it.

PR #2477, opened 2025-06-05
Bump cargo-deny-action

197g:
The previous version has a very outdated docker image by now which will conflict with our MSRV choices in the future (and partially does already for some potential dependencies).


197g:
Since this is changing some of the license enforcement (MPL-2.0 when you want it, which is fine because it is not viral and really opt-in), I'd rather get an explicit okay.

PR #2480, opened 2025-06-09
Reorganize structure

197g:
I'm trying to cleanup the structure similar to what we did with `codecs` a while ago. That is make sure internal module names represent better what's in the file. For instance in `image.rs` we have a whole cluster of different responsibilities under the most generic name we could have in this repo. This is all more or less independent of the names we are exposing the types.


197g:
(failures are from webp matching exact bytes which contradicts the fixes made in 0.2.2)

197g:
I'm only looking for semver checks acknowledging the API interface here.

Issue #2355, opened 2024-10-17
Consider removing the `ImageDecoderRect` trait


Shnatsel:
The `ImageDecoderRect` trait is supposed to help decode only the specified part of the image, as opposed to the entire thing.

As of v0.25, only BMP and Farbfeld decoders implement it: https://docs.rs/image/0.25.4/image/trait.ImageDecoderRect.html

Given the lack of format support, it's probably best to remove it for now, to reduce the API surface and maintenance burden. We could possibly re-introduce it later if there is enough demand.

Shnatsel:
FWIW `imagemagick` does provide this functionality, so a proper implementation of this could potentially be of use to `wondermagick`. But it is only an optimization, ideally and we'd have to generalize it to animations as well, so I don't think it's worth the trouble at this stage.

fintelia:
I think this is worth considering. Part of the problem with `ImageDecoderRect` is there's formats were it is sometimes possible to extract portions of the image, but it depends on the choices made by the encoder.

If we do decide to remove, thing thing to do would be marking the trait as deprecated for a few releases and see if any concerns are raised.

PR #2481, opened 2025-06-10
Reorganize buffers

197g:
Further restructure such that all our buffers are in a submodule. No more awkward renaming of the module `buffer` internally. (In the long run, we should apply the top-level name policy to that module or rethink the policy).

Imo the `image.rs` file should be removed. Its remaining contents are IO traits and the `FileFormat` interaction which both could go to separate files in the `io` module.

Since this restructuring is a little larger than the previous and I'm not certain about the API implications, a draft for the moment.


kornelski:
Maybe `free_functions` could be also split and the functions moved to more topically related modules?

It's a module named after a technicality, but you wouldn't make mod `structs` and `constants`.

197g:
Good point, I think I'll attempt that next. At the very least we could re-export most those functions in the `io` module and consume them from there.

src/io/free_functions.rs
```diff
@@ -198,3 +200,344 @@ pub(crate) fn guess_format_impl(buffer: &[u8]) -> Option<ImageFormat> {
 
     None
 }
+
+/// Decodes a specific region of the image, represented by the rectangle
+/// starting from ```x``` and ```y``` and having ```length``` and ```width```
+#[allow(dead_code)]
+// When no image formats that use it are enabled
+pub(crate) fn load_rect<D, F1, F2, E>(
```
fintelia, line 208: This is a helper function for implementing `ImageDecoderRect`. Might be good to keep it close to the definition of that trait

I think of io/free_functions.rs as being for the top-level public functions in the crate (`image::open`, `image::save_buffer`, etc.) and their implementations.

197g: Mainly I didn't want to associated it too deeply with the trait, considering we were talking [about removing that trait](https://github.com/image-rs/image/issues/2355) and this method might stay beyond that as associated methods on those decoders instead.

But even then this method isn't tied as deeply to the trait for me as it seems for you.  The main part of its implementation is spent on policy working _around_ the underlying decoder not really cropping by itself and still yielding *all* the data. In this sense it is more like the antipole to what the trait really should be. That is of course part of the issue with the trait; none of our implementations provide the real benefit over the naive 'throw-away-data' strategy and that happens by them using what is, really, a _pitfall_ to be quite frank. You really should not be using this except in the circumstance that you can't avoid it. And performant decoding, in general, should not work by 'scanlines' (apart from png where the filter/compression depends on this concept and hence the previous line is always necessary¹).

----

¹Actually, I just realize, in an opportunistic look-ahead in `png` the full scanline is *not* required if the next filter is `None` or `Sub`. Curious.

fintelia: We could also think about making a "utils for codec implementations" file somewhere. But I'm fine with moving this function to io for now

PR #2484, opened 2025-06-12
Alow extracting orientation from an Exif chunk and optionally clearing it in Exif

Shnatsel:
Tools that not only display images but also encode or convert them may want to apply the Exif orientation to the image up front, so that they don't have to rely on other software being Exif-aware to display the images with the correct orientation.

Such tools need to clear the orientation data from the Exif chunk, otherwise the image will be rotated twice and display incorrectly in Exif-aware software.

Modifying Exif data is exceptionally cursed, and general-purpose transformation code is bound to break some offset-dependent Exif chunks : https://github.com/kamadak/exif-rs/issues/1

However, in this particular special case we can rely on the fact that we are not changing the length of the data to reset orientation to `NoTransforms` and keep the rest of the Exif chunk intact.

I need this functionality in `wondermagick` and I believe this is the least cursed way to go about it.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -152,8 +152,7 @@ enum ExifEndian {
     Little,
 }
 
-// The tests module. It's conventional to put it at the end of the file.
-#[cfg(test)] // This attribute tells Rust to compile this code only when running tests.
+#[cfg(all(test, feature = "jpeg"))]
 mod tests {
     use crate::{codecs::jpeg::JpegDecoder, image::ImageDecoder};
 
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -1,6 +1,6 @@
 //! Types describing image metadata
 
-use std::io::{Cursor, Read, Write};
+use std::io::{Cursor, Read};
 
 use byteorder_lite::{BigEndian, LittleEndian, ReadBytesExt, WriteBytesExt};
 
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -63,7 +63,16 @@ impl Orientation {
         }
     }
 
-    pub(crate) fn from_exif_chunk(chunk: &[u8]) -> Option<Self> {
+    /// Extracts the image orientation from a raw Exif chunk.
+    ///
+    /// You can obtain the Exif chunk using
+    /// [ImageDecoder::exif_metadata](crate::ImageDecoder::exif_metadata).
+    ///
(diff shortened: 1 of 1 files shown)


Shnatsel:
I'm going to add tests as well, but I wanted to put this up for review ASAP so that we could agree on the basic direction.

Shnatsel:
Tests added, CI is green. This should be ready to go.

Shnatsel:
Thanks for the quick review!

fintelia:
Had to think about this design for a while. Although I don't necessary like the interface, I think I'm convinced that it is the least-cursed way to handle editing the EXIF orientation.

Shnatsel:
If you have a better idea for the interface, there's still time to change it since it wasn't shipped this to crates.io yet.

Right now the point of wondermagick is to kick the tires of `image` APIs and add the missing ones, and this seemed the most natural to me for wondermagick use.

fintelia:
I've thought about a few other options, but I don't think any are better.

My first thought was to add a `set_orientation` method, but that would have some annoying edge cases. If the EXIF chunk didn't already have an orientation we'd have to zero-out the old IFD and append a new one, which would grow the size of the EXIF chunk and complicate the API (perhaps by taking a mutable `Vec`). And if the the EXIF was fully corrupt we'd have to return an error, which either means reusing `ImageError` or adding a new single-variant enum. None of that would be terrible but seemed like more of a hassle than it is worth.

Another idea would be creating an entire `metadata::exif` module with more functionality. Adding JPEG thumbnail extraction for instance wouldn't be that hard. But it would add a bunch of code. I tried sketching out a design to cut down on the code duplication, which very quickly started to look like a full blown TIFF or EXIF decoder. And there's already other crates for that.

One thing that would fit with the current API and might be good to add would be a `Orientation::to_exif_chunk`. The implementation would be quite simple: just a hard-coded `vec![...]` call where one of the bytes is determined by `self.to_exif()`.

197g:
As detailed in the 'editing exif is hard' the editing is very complicated even though building isn't. So realistically the only better API would somehow wrap _existing_ exif as bytes and offer either in-place edits or rebuilding a completely new exif structure from scratch with just the recognized parts but no in-between. But I don't see the point of adding a new type just to wrap bytes (for now) until we get at least a couple more exif related pieces here that would justify it better.

Shnatsel:
There is a crate specifically dedicated to editing Exif in-place as much as possible, used by Glycin: https://crates.io/crates/gufo-exif

Issue #2179, opened 2024-03-24
Thread panicking in TiffReader::read_image()


RedPixlFox:
https://github.com/image-rs/image/blob/d407e366e7c080d3757f83c4e120be7f9444b8e1/src/codecs/tiff.rs#L244

I don't know, what path this happened on, because [this application](https://github.com/RedPixlFox/thumbnailed.git) iterates over _all_ files in "C:/"

### output:
(with RUST_BACKTRACE=full)
```
thread 'some-thread' panicked at C:\Users\MYNAME\.cargo\registry\src\index.crates.io-6f17d22bba15001f\image-0.25.0\src\codecs\tiff.rs:244:21:
source slice length (24300) does not match destination slice length (97200)
stack backtrace:
   0:     0x7ff676948252 - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\sys_common\backtrace.rs:44
   1:     0x7ff67696d92d - core::fmt::rt::Argument::fmt
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\core\src\fmt\rt.rs:142
   2:     0x7ff67696d92d - core::fmt::write
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\core\src\fmt\mod.rs:1120
   3:     0x7ff6769437c1 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\io\mod.rs:1810
   4:     0x7ff67694807a - std::sys_common::backtrace::_print
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\sys_common\backtrace.rs:47
   5:     0x7ff67694807a - std::sys_common::backtrace::print
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\sys_common\backtrace.rs:34
   6:     0x7ff67694a529 - std::panicking::default_hook::closure$1
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\panicking.rs:272
   7:     0x7ff67694a1e5 - std::panicking::default_hook
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\panicking.rs:292
[...]

PR #2492, opened 2025-06-17
fix: avoid panic multi planar tif images

swarnimarun:
### Motivation

Avoid panics for tif images with multi planar configuration.

### todo

- [x] Fail early on decode for tif images with multi planar support, otherwise it may produce a panic
- [ ] Provide test cases.

I have just made a fix for usage by a company I work at, lmk if you folks would prefer it fixed in some other way.

It's a draft as it hasn't quite been extensively tested, yet. And might require more changes to target only cases which aren't quite supported correctly.

Lmk if you folks want any other changes.

Fixes https://github.com/image-rs/image/issues/2179

### DIFF
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -11,7 +11,6 @@ extern crate tiff;
 use std::io::{self, BufRead, Cursor, Read, Seek, Write};
 use std::marker::PhantomData;
 use std::mem;
-use std::str::FromStr;
 
 use crate::color::{ColorType, ExtendedColorType};
 use crate::error::{
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -60,7 +60,7 @@ where
             .unwrap_or_default();
 
         // Decode not supported for non Chunky Planar Configuration
-        if planar_config <= 1 {
+        if planar_config > 1 {
             Err(ImageError::Unsupported(
                 UnsupportedError::from_format_and_kind(
                     ImageFormat::Tiff.into(),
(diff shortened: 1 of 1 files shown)


src/codecs/tiff.rs
```diff
@@ -11,6 +11,7 @@ extern crate tiff;
 use std::io::{self, BufRead, Cursor, Read, Seek, Write};
 use std::marker::PhantomData;
 use std::mem;
+use std::str::FromStr;
```
swarnimarun:
```suggestion
```

197g (approved):
Looks fine to me, much better than a late error and consistency panics.

197g:
Separately, I'm looking if we can get planar support  within months  not for `ImageBuffer` but a separate target buffer) since planar RGB is decently common across formats. Then an appropriate conversion should be a no-brainer and we could retrofit hopefully.

georgeboot:
@197g will your PR https://github.com/image-rs/image-tiff/pull/313 add the necessary things that the fix in this PR wont be needed anymore, and planar tiffs can be fully supported in this crate?

197g:
@georgeboot Cautious yes. We don't plan for supporting planar layouts for the buffers in `image` (yet?) but that PR adds the necessary interfaces to fully decode the channel data so that the bindings could re-arrange the channels into the expected order.

georgeboot:
@197g got it thanks! Yeah not sure if planar makes sense internally in `image`. We saw some planar tiffs today that errored when loading using `image`.

But will wait for your pr and then manually decode and convert them, before feeding into `image`.

PR #2496, opened 2025-06-18
Gaussian blur

### LINKED ISSUE #986: `blur` function is too slow

awxkee:
Closes #986 

<details>
<summary>Benchmarks</summary>

My working laptop has 10-15% noise, so numbers just to understand the magnitude of the difference. As well as comparison to libblur, because it actually is completely different execution class.

```
fast blur: sigma 3.0    time:   [22.242 ms 22.345 ms 22.460 ms]
                        change: [-6.7313% -5.5870% -4.4410%] (p = 0.00 < 0.05)

fast blur: sigma 7.0    time:   [22.404 ms 22.427 ms 22.453 ms]
                        change: [-9.3688% -8.5638% -7.7808%] (p = 0.00 < 0.05)

fast blur: sigma 50.0   time:   [23.251 ms 23.279 ms 23.313 ms]

gaussian blur: sigma 3.0
                        time:   [8.8081 ms 8.8672 ms 8.9305 ms]

gaussian blur: sigma 7.0
                        time:   [20.795 ms 20.959 ms 21.141 ms]

gaussian blur: sigma 50.0
                        time:   [174.65 ms 175.70 ms 176.80 ms]

libblur gaussian blur: sigma 3.0
                        time:   [1.5107 ms 1.5178 ms 1.5262 ms]

libblur gaussian blur: sigma 7.0
                        time:   [3.2371 ms 3.2688 ms 3.3020 ms]

libblur gaussian blur: sigma 50.0
                        time:   [25.141 ms 25.244 ms 25.373 ms]

libblur fast_blur exact alternative: sigma 3.0
                        time:   [2.9289 ms 2.9360 ms 2.9436 ms]

libblur fast_blur exact alternative: sigma 7.0
                        time:   [3.1053 ms 3.1424 ms 3.1801 ms]

libblur fast_blur exact alternative: sigma 50.0
                        time:   [3.1044 ms 3.1450 ms 3.1875 ms]
```

</details>


src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+    for (x, dst) in dst_row.chunks_exact_mut(4).enumerate() {
+        let v_cx = x * 4;
+        let src = &src[v_cx..];
+
+        let mut k0 = src[half_len * N].as_() * hc;
+        let mut k1 = src[half_len * N + 1].as_() * hc;
+        let mut k2 = src[half_len * N + 2].as_() * hc;
+        let mut k3 = src[half_len * N + 3].as_() * hc;
+
+        for (i, &coeff) in scanned_kernel.iter().take(half_len).enumerate() {
```
awxkee, line 434: Just to clear up any doubts: you might think that windowed operators would be a better fit here. But it's not. Rust has some insane windowed operators.

So this code 
```rust
  for (x, dst) in dst_row.chunks_exact_mut(4).enumerate() {
        let v_cx = x * 4;
        let src = &src[v_cx..(v_cx + (length - 1) * N + 4)];

        let mut k0 = src[half_len * N].as_() * hc;
        let mut k1 = src[half_len * N + 1].as_() * hc;
        let mut k2 = src[half_len * N + 2].as_() * hc;
        let mut k3 = src[half_len * N + 3].as_() * hc;

        for ((&coeff, front), back) in scanned_kernel
            .iter()
            .take(half_len)
            .zip(src.windows(4).step_by(N))
            .zip(src.windows(4).step_by(N).rev())
        {
            k0 = mla(k0, front[0].as_() + back[0].as_(), coeff);
            k1 = mla(k1, front[1].as_() + back[1].as_(), coeff);
            k2 = mla(k2, front[2].as_() + back[2].as_(), coeff);
            k3 = mla(k3, front[3].as_() + back[3].as_(), coeff);
        }

        dst[0] = k0.to_();
        dst[1] = k1.to_();
        dst[2] = k2.to_();
        dst[3] = k3.to_();
    }
```
Changes performance number to ones like so:
```
gaussian blur: sigma 3.0
                        time:   [12.932 ms 13.006 ms 13.085 ms]
                        change: [+36.945% +38.456% +39.751%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 15 outliers among 100 measurements (15.00%)
  13 (13.00%) high mild
  2 (2.00%) high severe

gaussian blur: sigma 7.0
                        time:   [46.195 ms 46.256 ms 46.321 ms]
                        change: [+92.786% +99.983% +105.84%] (p = 0.00 < 0.05)
                        Performance has regressed.
  10 (10.00%) high mild

Benchmarking gaussian blur: sigma 50.0: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 40.3s, or reduce sample count to 10.
gaussian blur: sigma 50.0
                        time:   [400.88 ms 404.82 ms 409.67 ms]
                        change: [+127.78% +130.41% +133.32%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
```

Shnatsel:
Thank you!

I wonder, are the benchmark numbers for libblur from single-threaded or multi-threaded execution?

awxkee:
Single threaded as far as I can tell.

src/images/dynimage.rs
```diff
@@ -854,8 +855,8 @@ impl DynamicImage {
     /// This method typically assumes that the input is scene-linear light.
     /// If it is not, color distortion may occur.
     #[must_use]
-    pub fn blur(&self, sigma: f32) -> DynamicImage {
-        dynamic_map!(*self, ref p => imageops::blur(p, sigma))
+    pub fn blur(&self, kernel_size: usize, sigma: f32) -> DynamicImage {
```
fintelia: We'll need to make a decision on whether we want to do this API change. If so, this PR cannot be merged until we start working on the next major release.

Do other library generally require that you specify the kernel size alongside the blur amount?

awxkee: Pure analytical gaussian filter often assumes that you want full control on it. [libblur](https://docs.rs/libblur/latest/libblur/fn.gaussian_blur.html) and [OpenCV](https://docs.opencv.org/4.x/dc/dd3/tutorial_gausian_median_blur_bilateral_filter.html) require kernel size and also support assymetry.

See [here](https://github.com/image-rs/image/issues/986#issuecomment-2983667862).

I'm ok to remove kernel size, and compute correct kernel size from sigma if you think it's better fit. But I'm not ok to return old behaviour, what is wrong.

awxkee: If it needs to wait something it's better to me to go ahead and close this PR. I don't have any plans to support rotting PRs

Shnatsel: An easy-to-use wrapper that only requires the user to specify the sigma and computes the kernel size by itself would be nice. So it'd be two functions, e.g. `blur()` and `blur_advanced()`.

src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+    let (src_row, rest) = buffer.split_at_mut(full_width);
+    for dst in rest.chunks_exact_mut(full_width).take(half_kernel) {
+        for (dst, src) in dst.iter_mut().zip(src_row.iter()) {
+            *dst = *src;
+        }
+    }
+
+    let mut start_ky = column_kernel_len / 2 + 1;
+
+    start_ky %= column_kernel_len;
```
197g, line 631: This one is odd. It just catches the case of `column_kernel_len == 1` but that rather seems like a very special case on its own which doesn't need column buffers at all to be honest.

src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+
+const RING_QUEUE_CIRCULAR_CUTOFF: usize = 55;
+
+fn filter_1d_ring_queue<T, F, I, const N: usize>(
+    image: &[T],
+    destination: &mut [T],
+    image_size: FilterImageSize,
+    row_kernel: &[F],
+    column_kernel: &[F],
+) -> Result<(), ImageError>
```
197g: A comment here should explain what `ring_queue` is referring to, i.e. the buffer for the column kernel. Overall, this method is complicated enough to deserved documentation regarding the data layout of each of the two buffers and how it relates to the flow below.

src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+
+    scanned_row_kernel = prepare_symmetric_kernel(&scanned_row_kernel);
+    scanned_column_kernel = prepare_symmetric_kernel(&scanned_column_kernel);
+
+    if scanned_row_kernel.is_empty() || scanned_column_kernel.is_empty() {
+        for (dst, src) in destination.iter_mut().zip(image.iter()) {
+            *dst = *src;
+        }
+        return Ok(());
+    }
```
197g: An empty kernel is an error for convolution, the no-op case is a `[1.0]` kernel. So this is a leniency contract, right? Should be documented in the function signature.

src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+    if image.len() != image_size.width * image_size.height * N {
+        return Err(ImageError::Limits(LimitError::from_kind(
+            LimitErrorKind::DimensionError,
+        )));
+    }
+    if destination.len() != image_size.width * image_size.height * N {
+        return Err(ImageError::Limits(LimitError::from_kind(
+            LimitErrorKind::DimensionError,
+        )));
+    }
```
197g: Both of these are overflow sensitive computations. And as a post-condition we have that both buffers are equal length which should be interesting to readers of certain copies below.

src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+        .map(|&x| T::transform(x))
+        .collect::<Vec<I>>();
+
+    scanned_row_kernel = prepare_symmetric_kernel(&scanned_row_kernel);
+    scanned_column_kernel = prepare_symmetric_kernel(&scanned_column_kernel);
+
+    if scanned_row_kernel.is_empty() || scanned_column_kernel.is_empty() {
+        for (dst, src) in destination.iter_mut().zip(image.iter()) {
+            *dst = *src;
+        }
```
197g: Since we have established the buffers are same length _and_ we have that `T: Copy` I'm wondering why this is not a `copy_from_slice` call?

src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+/// Currently implemented only for symmetrical filters
+///
+/// # Arguments
+///
+/// * `image`: Single plane image
+/// * `destination`: Destination image
+/// * `image_size`: Image size see [FilterImageSize]
+/// * `row_kernel`: Row kernel, *size must be odd*!
+/// * `column_kernel`: Column kernel, *size must be odd*!
+fn filter_1d<T, F, I, const N: usize>(
```
197g: I don't like `1d` in the name here, makes me expect _one_ filter and a choice of direction rather than what we have: `filter_separable` may be more appropriate. Applies to the exported versions below.

src/imageops/filter_1d.rs
```diff
@@ -0,0 +1,879 @@
...
+
+    scanned_row_kernel = prepare_symmetric_kernel(&scanned_row_kernel);
+    scanned_column_kernel = prepare_symmetric_kernel(&scanned_column_kernel);
+
+    if scanned_row_kernel.is_empty() || scanned_column_kernel.is_empty() {
+        for (dst, src) in destination.iter_mut().zip(image.iter()) {
+            *dst = *src;
+        }
+        return Ok(());
+    }
```
197g: This setup is repeated in the ring_queue version. We should pass it or do the setup after the check, I think I prefer the latter in terms of readability.

src/images/dynimage.rs
197g: We're already preparing a major version bump on the main branch so I'd do the change now. However, another idea is given by your question regarding kernel sizes. Small kernel sizes are faster (just see our cutoff point for ring queue in this). Considering most of our users we should at least _suggest_ the common choices. 

What if we were to introduce a 'BlurKernel' type that wraps those choices and gives a few constructors / static constants of common cases? Then indeed `blur_{by,with}(BlurKernel)` may make sense with `blur(f32)` yielding a somewhat common default.

197g (commented):
>  Do you want fixed point implementations for u8 or u16 instead of f32? Or just pure f32 convolution is needed?

You matched the style in as discussed in the codec case and to me that is neat enough.

But also this seems to be an implementation choice, rather than an API choice. At least when we cast back to the underlying storage type I'd expect that what you're referring to as fixed point is actually exact with regards to rounding ("as-if floating point"). Then it should not matter to the user and the choice should be as fast as possible. However, that is then also a discussion point for the future instead as we can dispatch on those specific types (I::Pixel: 'static allows TypeId).

>  If special cases is wanted then do you expect them to be bit exact?

Not sure. It doesn't seem necessary but appealing for the special casing of _types_ / fixed-point. Less so for special casing for kernel sizes but also see discussion on those below.

src/imageops/filter_1d.rs
197g: Please refer to this Github comment in a code comment in those lines, it's very curious and nicely explained.

src/images/dynimage.rs
awxkee: From a convenience perspective, I'm confident that in most cases this API is expected to accept either a _kernel size_ masked as a radius or a direct _kernel size_.

While everyone in this thread might understand what sigma is, I'm fairly sure that for most users, sigma is just a meaningless number that produces some kind of magic.

Accepting sigma only raises a problem: if a caller wants to perform very soft smoothing, which is more than common for purely analytical gaussian filter using kernel sizes 3/5/7. It is about to impossible just using sigma. Now at least ChatGPT may give a suggestion how to get ideal sigma for kernel size and you have to solve equations a little bit to get this in reverse order.

But I'm ok with any changes except those that obviously wrong.

Shnatsel: I was actually confused by the sigma vs kernel size distinction. 

I just want a high-level API that I can call that roughly matches what I'd get from "blur radius" in GIMP, but I also recognize that there are use cases for more direct control of the parameters. So I'd prefer an easy-to-use version as `blur()` and a more advanced API for people who need it.

src/imageops/filter_1d.rs
awxkee: There is a mistake here — this is a no-op only when both kernels are [1.0]. Currently, the implementation assumes that anisotropy is an acceptable condition.

src/imageops/filter_1d.rs
awxkee: Currently, the implementation assumes that anisotropy is an acceptable condition, and there is no special implementation case that handles column as identity and row as convolution. So implementation assumes that `column_kernel_len == 1` and `row_kernel_len == 5` is just fine.

 I technically could drop any possibility of anisotropy if you see a better fit.

src/images/dynimage.rs
awxkee: I did in libblur [such parameters builder](https://docs.rs/libblur/latest/libblur/struct.GaussianBlurParams.html). However, it might still be overkill, but I don't expect that someone plugging in 70K lanes of SIMD code wouldn't at least do a bit of investigation into how to use the API.

For more general purpose implementation I agree that GIMP style "blur radius" is preferred.

fintelia: I'm in favor of a single argument that behaves like blur radius. That's what I've always assumed that sigma did

src/imageops/filter_1d.rs
awxkee: I did my best to explain how it works.

src/images/dynimage.rs
awxkee: > What if we were to introduce a 'BlurKernel' type that wraps those choices and gives a few constructors / static constants of common cases? Then indeed blur_{by,with}(BlurKernel) may make sense with blur(f32) yielding a somewhat common default.

I think providing something `like image.blur(3)` is enough for general-purpose use.
Adding methods like `image.blur3`, `image.blur5`, `image.blur7` gives a strong impression of API bloat.
It might make some sense if we were doing something truly interesting in these implementations, just to mark them that they  may yield results are different from what you'd expect. But for now all implementations are the same, even when you hit a `ring queue` path it gives you the same result, but using a different way to get it.

awxkee:
A ton of WebP tests started failing today. I'm not sure if this PR is related to that.

Shnatsel:
That should be the fixes shipped in https://crates.io/crates/image-webp v0.2.3 altering the enshrined hashes.

In the long run something like image-webp's pixel difference threshold should be implemented, or https://github.com/image-rs/image-webp/issues/146 should be fixed at which point we could enshrine hashes again. 

But for now they should probably just be regenerated. I'll open a PR with that against the main branch.

fintelia:
WebP failures are probably because of a bugfix included in the `image-webp` release yesterday

awxkee:
I realized that I benchmarked the `GenericImageView` implementation instead of `DynamicImage`, so here are the updated numbers.

<details>
<summary>Benchmark</summary>

```

fast blur: sigma 3.0    time:   [24.121 ms 24.321 ms 24.525 ms]
                        change: [+4.5411% +5.6756% +6.8759%] (p = 0.00 < 0.05)
                        Performance has regressed.

fast blur: sigma 7.0    time:   [23.648 ms 23.834 ms 24.030 ms]
                        change: [-12.527% -5.4979% -0.7447%] (p = 0.07 > 0.05)
                        No change in performance detected.

fast blur: sigma 50.0   time:   [25.422 ms 26.036 ms 26.734 ms]
                        change: [+6.1714% +8.8062% +12.002%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) high mild
  8 (8.00%) high severe

gaussian blur: sigma 3.0
                        time:   [9.1108 ms 9.1780 ms 9.2486 ms]
                        change: [-2.8948% -1.7866% -0.6687%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

gaussian blur: sigma 7.0
                        time:   [21.107 ms 21.351 ms 21.645 ms]
                        change: [-2.4164% -1.0423% +0.5379%] (p = 0.18 > 0.05)
                        No change in performance detected.
  13 (13.00%) high mild
  1 (1.00%) high severe

Benchmarking gaussian blur: sigma 50.0: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 17.7s, or reduce sample count to 20.
gaussian blur: sigma 50.0
                        time:   [172.45 ms 173.46 ms 174.53 ms]
                        change: [-3.2896% -2.4175% -1.5235%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe

gaussian blur (dynamic image): sigma 3.0
                        time:   [5.2648 ms 5.3129 ms 5.3641 ms]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

gaussian blur (dynamic image): sigma 7.0
                        time:   [12.699 ms 12.860 ms 13.099 ms]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe

Benchmarking gaussian blur (dynamic image): sigma 50.0: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.7s, or reduce sample count to 50.
gaussian blur (dynamic image): sigma 50.0
                        time:   [99.216 ms 100.11 ms 101.01 ms]
```
</details>

src/images/dynimage.rs
197g: The argument radius/sigma makes perfect sense to me, too. So agreed, the simple function might as well accept a simple integer argument. I'm still not entirely sure we should remove the interaction entirely though for an advanced function.. Deriving sigma from the radius doesn't make much sense either. For small values (`3`, `5`, `7`) the influence of normalization is quite high, which people will have different preferences for.

But also importantly, imagmagick has it as an option(https://imagemagick.org/script/command-line-options.php#blur). I'm thinking if we had a struct and dispatched internally:

```
struct BlurKernel {
    size: (u32, u32),
    sigma: (f32, f32),
}

impl BlurKernel {
    // Corresponds to calling the simple function with (3).
    pub const THREE: … // Due to float-const this must be filled manually..

    /// The isotropic case.
    pub fn from_radius(sz: u32) -> Self { … }

    /// Document the (1.0, 1.0) default and what anisotropy refers to.
    pub fn with_sigma(self, (x, y): (f32, 32)) -> Self { … }
}
```

I think that is straightforward enough, but please do argue if you think this is API bloat. The anisotropy case in particular, I _think_ there's enough reason to support it if it is one additional parameter to an intermediate/expert-level function call.

src/imageops/filter_1d.rs
197g: I was just confused by the way those lines are written. It suggests that `column_kernel_len == 1` is an even more special case than anisotropy itself since no other case required the modulos operation and the length of 1 does not really require any intermediate buffers.

197g: That said, a comment is a fine resolution to this to me. There are bigger fish to fry.

src/imageops/filter_1d.rs
197g: Neat!

src/imageops/filter_1d.rs
awxkee: I've added comment also to make sure it's clear.

awxkee:
I believe all the feedback has been addressed.

Let me know what do you think about new APIs.

I preserved API to just change argument name, what is good and bad at the same time. It's better for the radius to be as a float because it allows to make a smoother step on computing sigma despite rounding it to the nearest odd integer. Rounding to nearest odd is not very common rounding mode, but not to worry I can handle such an arithmetic.

All the other changes I think is pretty straightforward.

fintelia:
I haven't reviewed most of this (I don't review PRs this large anymore). But if I'm following correctly, then we could split it into one change with the new blur implementation and then a second tiny, but API breaking, PR that just switched from sigma to blur radius as the argument for the blur function?

That would be my preference if possible. It seems to work better to add features / improve functionality in non-breaking releases. That lets us roll out features gradually as they're ready, rather than all at once in a new major version. And if something goes wrong we can yank the release while working on debugging and fixing it. You can yank major versions, but then anyone who already upgraded is stuck.

awxkee:
> PR that just switched from sigma to blur radius as the argument for the blur function?

That's right. PR just switched sigma to radius without changing anything actually.

> But if I'm following correctly, then we could split it into one change with the new blur implementation and then a second tiny, but API breaking

Yes, I'm ok with this one.

awxkee:
@fintelia

It's done.

src/imageops/sample.rs
```diff
@@ -1012,25 +1025,506 @@ pub fn blur<I: GenericImageView>(
...
+#[derive(Copy, Clone, PartialOrd, PartialEq)]
+pub struct GaussianBlurParameters {
+    /// X-axis kernel, must be odd
+    pub x_axis_kernel_size: u32,
+    /// X-axis sigma, must > 0, not subnormal, and not NaN
+    pub x_axis_sigma: f32,
+    /// Y-axis kernel, must be odd
+    pub y_axis_kernel_size: u32,
+    /// Y-axis sigma, must > 0, not subnormal, and not NaN
+    pub y_axis_sigma: f32,
```
197g: Nit: having `check_constraints` conflicts with fully public fields. I'd start with private fields and those below constructors and open them up in future PRs over time. There's little cost to this.

awxkee: Done.

Shnatsel:
I'm happy to see this merged, thanks a lot to everyone involved!

awxkee:
Clarification about [benchmark](https://github.com/image-rs/image/pull/2496#issue-3157453203), particulaly regarding fast_blur ( better late than never, isn't it? ):

The SIMD implementation of box blur in libblur is approximately 2.5 times faster than scalar, so I'd expect the execution time of `fast_blur` roughly matches `libblur_fast_blur_alternative_time * 2.5 ~= 7ms`. However, the actual time is drastically higher. 
Even if account for some overhead that it's dispatched on `GenericImageView` which might add ~30% the expected time would be still much lower than what the current implementation shows.

I'd say there something likely wrong.

Issue #2498, opened 2025-06-18
Add an API that automatically converts color formats when saving


Shnatsel:
### The problem

[DynamicImage::save()`](https://docs.rs/image/latest/image/enum.DynamicImage.html#method.save) is a high-level convenience method for saving an image to a specified format.

Sadly not really usable. If you load e.g. an 8-bit PNG image with alpha and try to save it as a JPEG, you get an error. Because JPEG does not support alpha channel.

So to be able to actually use the API, you need to research what each format supports and manually write a bunch of logic for each format that would coerce the `DynamicImage` into something the format can encode. And at that point you're hand-writing a massive nested `match` statement that lists every combination of every . 

This completely defeats the purpose of the high-level API; you might as well just dispatch to each individual encoder instead of bothering with `DynamicImage::save()` at all.

### Possible solutions

The high-level `DynamicImage::save()` should coerce the `DynamicImage` into a pixel format it can actually encode. Punting this work to the API user makes for an exceedingly unpleasant API.

Since this mutates the `DynamicImage`, the function should accept a `&mut self`, or even `Self if that allows for more optimizations. If preserving the original pixel format is important, the API user can clone the DynamicImage.

The old behavior can be preserved in a separate `try_save()` function. Alternatively, `save()` can remain as it is and a new `coerce_and_save()` can be added, but I thing the coercing behavior should be the default.

Shnatsel:
The same applies to `save_with_format`, `write_to` and `write_with_encoder` methods on DynamicImage.

fintelia:
I think we should be thoughtful about what sorts of conversions we do. It is better to return an error than silently produce broken output.

With suitable documentation, stripping the alpha channel for formats that don't support one seems totally reasonable to me. Automatically converting 16-bit per channel to 8-bit per channel doesn't seem too likely to cause problems either. I get more nervous about converting floating point images to integer. The main options are to either tone-map or clamp values outside the 0-1 range, but there isn't a good way to pick.

Implementation-wise we could add methods like `ImageEncoder::opaque_only` (with return false as the default impl) to indicate that saving should strip out the alpha channel before calling encode.

Shnatsel:
On f32: I understand tonemapping will require additional metadata, and in absence of that we can only clamp? The image+metadata type is probably something we should add regardless, `wondermagick` experience indicates that it is an abstraction that is indeed needed.

awxkee:
[...]

PR #2501, opened 2025-06-20
Encode conversion sequence

197g:
Closes: #2498 

In the draft only for JPEG but it should be simple to expand. See the issue for the basic motivation but suffice to say we want to mirror `decode` which does those conversions, even when they are lossy. Details from the commits:

---

> \[From the design\], we also get that any color conversion we may attempt to
perform for a specific format is delayed until the check of our encoder
list indicates support for that format.

---

> In this design, the path to the underlying encoder is responsible for
choosing whether to degrade or prefer an error based on actual support.
For instance, the raw methods (with `Into<ExtendedColorType>`) still run
without negotiation or automatic conversion. Similarly, in the planned
advanced by-based `Image` type we might choose an expert level and
negotiate only when specifically asked, preferring an error over silent
degradation. But TBD as it will probably be implemented by an
`ImageEncoder` type to mirror similar choices on the decoding side.


fintelia (approved):

src/io/free_functions.rs
```diff
@@ -54,88 +52,118 @@ pub fn save_buffer_with_format(
...
-        #[cfg(feature = "jpeg")]
-        ImageFormat::Jpeg => {
-            jpeg::JpegEncoder::new(buffered_write).write_image(buf, width, height, color)
+
+        fn upcast(&mut self) -> &mut dyn ImageEncoder {
+            self
         }
+    }
+
+    let mut encoder: Box<dyn WriteConsume> = match format {
```
fintelia, line 121: We should think about making `ImageEncoder` dyn compatible in the next major release, which would avoid needing an adapter here

Shnatsel:
After this PR I still get the following error when trying to encode an RGBA image as a JPEG:

> The encoder or decoder for Jpeg does not support the color type `Rgba8`

I am calling `DynamicImage::write_with_encoder`.

Is this intentional? And if yes, how do I get the transparent conversion to work with the supplied encoder?

fintelia:
The problem is that `write_with_encoder` doesn't use `write_buffer_impl`, but instead directly calls methods on the encoder. Didn't realize that we'd end up with inconsistent behavior like that when reviewing the PR. Probably need to think more about exactly which methods should / shouldn't do conversion.

One option would be to say that anything on `DynamicImage` does conversion, while methods that directly take an `ExtendedColorType` or are called on `ImageBuffer` should not do any conversion?

197g:
:+1: for the resolution of making `DynamicImage` consistently perform the conversion. The interaction with the `ImageEncoder` to make that work is already implemented in that module, too.

Shnatsel:
That sounds good to me as well.

PR #2507, opened 2025-06-21
Upgrade tiff crate

fintelia:


197g:
>   process didn't exit successfully: `/linux-runner powerpc /target/powerpc-unknown-linux-gnu/debug/deps/image-740b6c4edc27a95c` (signal: 4, SIGILL: illegal instruction)

:eyes:

fintelia:
Was able to reproduce the powerpc illegal instruction issue locally. Happens on stable but 1.86.0 is fine. If anyone wants to investigate, this command triggers it:
```
cross test --target powerpc64-unknown-linux-gnu openexr::test::compare_cropped --no-default-features --features exr
```
I've switched to 64-bit powerpc as a workaround

Issue #2442, opened 2025-03-20
Animated Gif decoding keep looping when encounter bad frames.


rayguo17:
This happens in when i tried to decode Gif image with below code:
```Rust
let file_path = Path::new("image-blob.gif");
    let image = File::open(file_path).unwrap();
    let buf_reader = BufReader::new(image);
    let decoder = GifDecoder::new(buf_reader).unwrap(); 
    let frames:Vec<ImageFrame> = decoder.into_frames().into_iter().filter_map(|frame|{
        match frame{
            Ok(frame)=>{
                Some(frame)
            },
            Err(e)=>{
                println!("Error decoding frame: {:?}",e);
                None
            }   
        }
    }).collect();
    println!("{}",frames.len());
```

## Expected

the filter map loop should stop when there are no more valid frame.

## Actual behaviour

The filter_map keep looping with Error. See Output:

![Image](https://github.com/user-attachments/assets/206b9d1a-63a6-4213-847d-a0b1301c061e)

## Reproduction steps

The gif image that i try to decode is as attached:
![Image](https://github.com/user-attachments/assets/1f2d1e4c-de60-4628-986a-abf322e8552d)

rayguo17:
I dig in a little bit, this is because the `fn next(&mut self) -> Option<ImageResult<animation::Frame>>` of `GifFrameIterator` will not early exit when it encounter any non-revertable error. Maybe consider adding a state tracker in GifFrameIterator?

197g:
I'm under the impression this API was modeled after the behavior of IO iterators, e.g [Lines](https://doc.rust-lang.org/stable/std/io/struct.Lines.html). These also just retry the underlying action instead of storing a stop flag. If the error is transient (i.e. network attached storage) and the decoder supported it (which gif doesn't but PNG does for WouldBlock) then it may resolve to another image at a later point.

Try `filter_map(<your closure>).fuse()` to stop iterating at an error _or_ end of frames.

rayguo17:
Try to dig in `filter_map` a bit. It will always return `Some(x)`, with its `next()` called, so `fuse()` after `filter_map()` would not stop iterating. But i agree that `UnexpectedEOF` could be recoverable in the scenario of `network storage`.

kornelski:
`.map_while(|frame| frame.ok())` should work around this.

I don't think errors in `gif` are recoverable. In theory it could skip over broken frames, but AFAIR it doesn't do that.

PR #2512, opened 2025-06-27
fix gif decoder infinite looping when getting unexpectedEOF error

rayguo17:
Fix `GifDecoder` infinite loop when encountering non-recoverable error like io::ErrorKind::UnexpectedEOF.
This PR fix: https://github.com/image-rs/image/issues/2442


fintelia:
We should make sure that the behavior for all the `AnimationDecoder` implementations stays in sync

Issue #2513, opened 2025-06-28
Multiple DoS vulnerabilities in HDR, TIFF, and PNG decoders (RustSec advisory pending)


AnubhavDash:
# Multiple DoS Vulnerabilities in image crate v0.25.6

## Summary

The `image` crate version 0.25.6 is vulnerable to multiple denial-of-service attacks triggered by small crafted inputs across HDR, TIFF, and PNG decoders. These vulnerabilities can cause memory exhaustion, panics, or process aborts when processing malicious image files.

## Expected Behavior

Image decoders should validate input dimensions and chunk sizes properly to prevent unbounded memory allocations and gracefully handle malformed data without crashing.

## Actual Behavior

- **HDR Decoder**: Headers containing `+X 4294967295` cause out-of-memory conditions or process aborts due to excessive allocation attempts
- **TIFF Decoder**: Setting `ImageWidth = 0x7FFFFFFF` in the IFD triggers allocator panics and crashes
- **PNG Decoder**: Malformed `IDAT` or `zTXt` chunks cause decoder panics or excessive memory consumption

All vulnerabilities reproduce consistently in release builds under memory pressure (tested with `ulimit -v`). Attack payloads are minimal, ranging from 30-40 bytes in size.

### HDR DoS Proof-of-Concept

```rust
use image::codecs::hdr::HdrDecoder;
use std::io::Cursor;

fn main() {
    let malicious_hdr = b"+X 4294967295\n+Y 1\n";
    let decoder = HdrDecoder::new(Cursor::new(malicious_hdr)).unwrap();
    let _result = decoder.read_image_hdr(); // Triggers OOM/abort
}
```

**Result**: Process crashes due to memory allocation failure or `Vec::with_capacity()` panic.

### TIFF DoS Proof-of-Concept

Minimal TIFF payload with malicious width field (little-endian format, `ImageWidth = 0x7FFFFFFF`):

```rust
use image::load_from_memory;

fn main() {
    let malicious_tiff = [
        // TIFF header (little-endian)
        0x49, 0x49, 0x2A, 0x00, 
        // IFD offset
        0x08, 0x00, 0x00, 0x00,
        // Number of directory entries
        0x01, 0x00,
        // ImageWidth tag (256)
        0x00, 0x01,
        // Field type (LONG)
        0x04, 0x00,
        // Count
        0x01, 0x00, 0x00, 0x00,
        // Value: 0x7FFFFFFF (maximum signed 32-bit integer)
[...]

PR #2514, opened 2025-06-28
Use try_reserve

kornelski:
libstd doesn't handle OOM well (Linux with overcommit is not the only platform in the world), so to improve reliability when working with large images, and to harden against broken ones #2513, I've added `try_reserve(_exact)`.

`.map_err(|_| io::ErrorKind::OutOfMemory)?;` is not required since Rust 1.78.


src/codecs/tga/decoder.rs
```diff
@@ -234,33 +236,38 @@ impl<R: Read> TgaDecoder<R> {
...
 
         if self.bytes_per_pixel == 0 {
             return Err(io::ErrorKind::Other.into());
         }
 
-        let color_map = self
-            .color_map
-            .as_ref()
-            .ok_or_else(|| io::Error::from(io::ErrorKind::Other))?;
+        let color_map = self.color_map.as_ref().ok_or(io::ErrorKind::Other)?;
```
fintelia, line 246: I suspect this should actually be an unwrap/expect call. If `expand_color_map` is getting called without there being a color map, then it seems like there was probably a bug somewhere in the decoder

src/codecs/tga/decoder.rs
```diff
@@ -234,33 +236,38 @@ impl<R: Read> TgaDecoder<R> {
...
     /// Reads a run length encoded data for given number of bytes
     fn read_encoded_data(&mut self, num_bytes: usize) -> io::Result<Vec<u8>> {
-        let mut pixel_data = Vec::with_capacity(num_bytes);
-        let mut repeat_buf = Vec::with_capacity(self.bytes_per_pixel);
+        let mut pixel_data = Vec::new();
+        pixel_data
+            .try_reserve_exact(num_bytes)
+            .map_err(|_| io::ErrorKind::OutOfMemory)?;
+
+        if self.bytes_per_pixel > 16 {
```
fintelia, line 264: Given these two, I think we can just assert that bytes_per_pixel <= 4 and not return an error:

 https://github.com/image-rs/image/blob/c3505ab891836c3f48c2262f8ad8e7acbee7429c/src/codecs/tga/decoder.rs#L142-L145
https://github.com/image-rs/image/blob/c3505ab891836c3f48c2262f8ad8e7acbee7429c/src/codecs/tga/decoder.rs#L122

fintelia (commented):
I support bumping the MSRV to 1.78 to avoid the `map_err`.

Could you also add a `vec_try_with_capacity` method instead of using `Vec::new` + `try_reserve`? That seems a bit cleaner and would make it easier to find all the places to change once `Vec::try_with_capacity` is stable.

197g:
Nice. Even though the report was… I do like the proactive approach of making more allocations fallible. Would it help to lint these occurrences with clippy?

```toml
# clippy.toml
disallowed-methods = [
    # fully qualified function/method name:
    "std::vec::Vec::with_capacity",
]
```

I'd be a fan of enforcing the fallibility in CI to a large degree and work towards actionable and automatable security / robustness guidances for downstream crates.

fintelia:
Also while it doesn't have to (and probably shouldn't) be part of this PR, several of these allocations can probably be eliminated entirely by decoding directly into the output buffer

src/codecs/tga/decoder.rs
kornelski: I've used `self.color_type.bytes_per_pixel()` as a reference, which goes up to 16. 

Currently 32-bit max is supported elsewhere, but this function doesn't rely on that, so I don't see a need to limit it more. The rest of the code could end up supporting bigger pixels in the future.

src/codecs/dxt.rs
```diff
@@ -102,7 +103,7 @@ impl<R: Read> DxtDecoder<R> {
         self.variant.decoded_bytes_per_block() as u64 * u64::from(self.width_blocks)
     }
 
-    fn read_scanline(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+    fn read_scanline(&mut self, buf: &mut [u8], tmp: &mut Vec<u8>) -> io::Result<usize> {
```
fintelia: I'd rather not do this sort of micro-optimization as part of a larger PR. The only way to know whether it helps performance is to benchmark it, and that works better as a standalone PR where we can compare and discuss only one change.

kornelski: I'd generally agree, but this one is pretty basic.

src/codecs/tga/decoder.rs
kornelski: I've added a debug_asset. I try to avoid a panic per pixel in release code, since it usually bloats the executable and optimizes poorly.

src/codecs/tga/decoder.rs
fintelia: This is a private method, so if it is called with `bytes_per_pixel` exceeding the size it supports, that represents a bug and should result in a `panic!`, not an invalid input error.

Separately, we should keep the pixel size limit for this method in sync with the header parsing code. Having a larger (but somewhat arbitrary) limit here makes it confusing for no benefit.

kornelski: I disagree. It has limit that matches the file format. The other code has smaller arbitrary limits, and it's confusing to spread smaller arbitrary limits elsewhere. It would be laborious to undo the smaller arbitrary limit once the header parsing code is properly generalized.

src/codecs/dxt.rs
fintelia: The problem is that it makes the code a bit more complicated and I'm not sure has any performance benefit at all.

kornelski: It's not clear to me which bit you find too complicated, and what aspect of performance you have in mind.

To me allocating once is obviously faster than allocating again on each scanline. Do you find the `tmp: &mut Vec` objectionable regardless of other changes?

Are you concerned about complexity or performance of `clear(); read_to_end()` vs just `read_exact`? If I allocated once, but had `tmp: &mut [u8]` and kept `read_exact`, would that be better?

fintelia: I think that trying to reuse the same allocation is more complicated than simply letting the memory allocator handle it. Allocators tend to have fast-paths for repeatedly allocating and freeing the same size piece of memory. So doing an allocation/free every four rows of image data vs. once for the image might not make any measurable difference.

But the reason I'm pushing back on this is because every incremental bit of complexity makes the code just a bit harder to understand and a bit harder to refactor. As an example, we might be able to get better cache behavior out of this function if we alternated reading each 4x4 block into a stack-allocated buffer and decoding it from there. Now that change requires editing `read_image` and changing the signature on `read_scanline` in addition to everything else.

kornelski: I don't disagree that growth in complexity can eventually cause problems, but pushing against any incremental increase in complexity is a slippery slope argument.

This change is tiny, local, and can be easily undone. 

When you extrapolate any tiny change into an eventual snowball of complexity, it becomes an extreme argument demanding literally zero increase in complexity. This makes most code changes impossible, and prevents the codebase from escaping its existing local maximum. 

The goal of keeping complexity in check is very good, but it needs to be applied with a broader perspective. Stonewalling very simple changes, not for their actual immediate cost, but for extrapolated eventual imagined cost, is not helpful.

kornelski: Anyway, I've dropped the commits that removed redundant reallocations.

fintelia (approved):

PR #2515, opened 2025-06-29
Improving Fast Blur

### LINKED ISSUE #2511: Fast blur is actually slow
[`DynamicImage::fast_blur`](https://docs.rs/image/latest/image/enum.DynamicImage.html#method.fast_blur) which implements a radius-independent blur algorithm is a lot slower than it could be. The break-even point for it compared to gaussian blur is above sigma 7: https://github.com/image-rs/image/pull/2496#issuecomment-2991002448

Tested on `image` v0.25.6 and also in latest git as of 1176e088736949723d9fea52d9487f388208f592

We should check if https://github.com/etemesi254/zune-image/blob/dev/crates/zune-imageprocs/src/box_blur.rs performs better and if so, adopt it.

awxkee:
Closes #2511 

It’s possible to push it even further using a ring buffer and fixed point arithmetic, but that would at least double the amount of code, and this version already gives a nice performance boost.

<details>
<summary>Benchmark</summary>

```
fast blur: sigma 3.0    time:   [10.750 ms 10.758 ms 10.766 ms]
                        change: [-52.185% -52.000% -51.832%] (p = 0.00 < 0.05)

fast blur: sigma 7.0    time:   [10.851 ms 10.873 ms 10.896 ms]
                        change: [-51.702% -51.598% -51.494%] (p = 0.00 < 0.05)

fast blur: sigma 50.0   time:   [11.371 ms 11.426 ms 11.486 ms]
                        change: [-50.958% -50.681% -50.399%] (p = 0.00 < 0.05)

gaussian blur: sigma 3.0
                        time:   [9.4690 ms 9.7881 ms 10.273 ms]

gaussian blur: sigma 7.0
                        time:   [20.815 ms 20.922 ms 21.042 ms]

gaussian blur: sigma 50.0
                        time:   [173.84 ms 174.37 ms 174.92 ms]

gaussian blur (dynamic image): sigma 3.0
                        time:   [5.2525 ms 5.2611 ms 5.2702 ms]

gaussian blur (dynamic image): sigma 7.0
                        time:   [13.056 ms 13.084 ms 13.118 ms]

gaussian blur (dynamic image): sigma 50.0
                        time:   [94.538 ms 94.824 ms 95.138 ms]
```

</details>

### DIFF
--- a/src/imageops/fast_blur.rs
+++ b/src/imageops/fast_blur.rs
@@ -454,7 +454,7 @@ fn box_blur_vertical_pass_impl<T: Primitive, const CN: usize>(
 
 #[cfg(test)]
 mod tests {
-    use crate::{DynamicImage, GrayAlphaImage, GrayImage, RgbImage};
+    use crate::{DynamicImage, GrayAlphaImage, GrayImage, RgbImage, RgbaImage};
     use std::time::{SystemTime, UNIX_EPOCH};
 
     struct Rng {
@@ -483,9 +483,7 @@ mod tests {
 
     #[test]
     fn test_box_blur() {
-        let now = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("Time went backwards");
+        let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
         let mut rng = Rng::new((now.as_millis() & 0xffff_ffff_ffff_ffff) as u64);
         for _ in 0..35 {
             let width = rng.next_u8();
@@ -530,7 +528,7 @@ mod tests {
                 3 => {
                     let vc = vec![px; width as usize * height as usize * 4];
                     let image = DynamicImage::from(
-                        GrayAlphaImage::from_vec(width as u32, height as u32, vc).unwrap(),
+                        RgbaImage::from_vec(width as u32, height as u32, vc).unwrap(),
                     );
                     let res = image.fast_blur(sigma);
                     for clr in res.as_bytes().iter() {
--- a/src/imageops/fast_blur.rs
+++ b/src/imageops/fast_blur.rs
@@ -1,4 +1,4 @@
-use num_traits::clamp;
+use num_traits::Bounded;
 
 use crate::{ImageBuffer, Pixel, Primitive};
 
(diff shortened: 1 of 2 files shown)


awxkee:
PTAL?

kornelski:
It looks okay, but I'm not capable of verifying it is correct from just looking at the code. Last time I implemented such thing, I've had a bunch of off-by-ones or mixed-up offsets when radius was greater than image size.

The current tests only check pixels of an easy case, and the rest just checks for panics.

Maybe you could add more tests?

awxkee:
Done. 
I believe it's not truly feasible to track all errors mentally in such algorithms, so I fuzz such algorithms, just as I fuzzed this one.

src/imageops/fast_blur.rs
```diff
@@ -64,101 +102,441 @@ fn boxes_for_gauss(sigma: f32, n: usize) -> Vec<usize> {
...
+        x + 1
+    } else {
+        x
+    }
+}
+
+#[inline]
+#[allow(clippy::manual_clamp)]
+fn rounding_saturated_mul<T: Primitive>(v: f32, w: f32) -> T {
+    if T::DEFAULT_MAX_VALUE.to_f32().unwrap() != 1.0 {
```
197g, line 140: Nit: I think there should be a note here that this is utilized as 'is-float' in a way that, hopefully and based on optimization probably, const-folds away.

src/imageops/fast_blur.rs
```diff
@@ -64,101 +102,441 @@ fn boxes_for_gauss(sigma: f32, n: usize) -> Vec<usize> {
...
+        }
+        if CN > 2 {
+            weight2 = chunk0[2].to_f32().unwrap() * edge_count;
+        }
+        if CN == 4 {
+            weight3 = chunk0[3].to_f32().unwrap() * edge_count;
+        }
+
+        for x in 1..=half_kernel as usize {
+            let px = x.min(width as usize - 1) * CN;
```
197g: nit: I'd sleep a little more sound if `width as usize - 1` were a variable calculated once in this function (i.e. `width_bound`) and a debug assertion that this conversion and subtraction succeeds. Or since it would be done once it could be `try_from`, we're relying on the caller never passing an empty width though.

src/imageops/fast_blur.rs
```diff
@@ -64,101 +102,441 @@ fn boxes_for_gauss(sigma: f32, n: usize) -> Vec<usize> {
...
+            if CN == 4 {
+                weight3 -= previous_chunk[3].to_f32().unwrap();
             }
         }
     }
+}
+
+fn box_blur_vertical_pass_impl<T: Primitive, const CN: usize>(
+    src: &[T],
+    src_stride: u32,
```
197g: Any stride should be passed as a `usize` when they refer to another slice parameter. There's no necessary relation to our `u32` bound of pixel length. Indeed, `width * P::CHANNEL_COUNT as u32` can overflow where it can not in `usize` since the buffer already holds at least that many elements—with the empty case covered elsewhere of course.

197g (changes requested):
Algorithmically I follow. But there are some style nits, the potential overflows would otherwise keep me up at night :slightly_smiling_face:

src/imageops/fast_blur.rs
awxkee: I'm not quite sure how do you expect that overflow/underflow will happen here. At this very moment width and height is already sanitized at `fast_blur lane 26-26` that they are not zero. The cast u32 to usize must never fail because you it should not be possible to allocate vector with size where `width*height*CN` exceeds pointer size. That means the width should never be less than 1 and always be `height*CN` times less than pointer size.

197g: I don't expect it to, but I have to reason about it (as you just did)—`width > 0` being a crucial precondition that isn't exactly local ( and `height > 0` for the symmetric ones in vertical). Since it occurs a couple of times in the code it was a little mental stumble each time that'd be easily solved by extracting that common subexpression from the loop to the function start.

There's no need to extrapolate to other cast, just those `_ as usize - 1` would already improve readability quite a bit for me here.

src/imageops/fast_blur.rs
awxkee: Done.

awxkee:
I’ve added a few asserts to make sure everything will go according to plan.

197g (approved):

PR #2525, opened 2025-07-11
Update `ravif` to v0.12

Expyron:


kornelski:
rav1e bumped MSRV to 1.83

Expyron:
Indeed, I noticed the MSRV change afterwards.

I'm not sure what's the MSRV policy of `image`.
Should I increase the MSRV to 1.83 in this PR? Or do you believe it's too soon?

kornelski:
https://lib.rs/stats#rustc-usage

* 8% of all crates require 1.83
* about half of recently updated crates requires 1.83+
* 90% of crate downloads are from Cargo 1.83 or later.

Expyron:
I see. No point merging that now then.

197g:
Practically speaking we don't _need_ to bump MSRV to `1.83`. Without using `ravif` the crate still works for older compilers. We should set `rust-version` according to code _in this repository_ in my opinion, or to the version required with minimal features. But admittedly testing is quite a bit more complicated then.

Also, setting `resolver = 3` (and eventually any more recent version than `1.84` on edition `2024`) should ensure that even with the use `ravif` a downstream user just can't update but still has their dependency probably working, I think. Now the question is how to test that to verify our promise in the `Cargo.toml` file.

fintelia:
Yeah, a dependency with both an API breaking release + a MSRV bump makes us pick to either bump MSRV ourselves (for features that require the dependency) or to stick on the old version for now.

Are there particular features/changes in `ravif` that are worth upgrading for?

kornelski:
No, there are no substantial changes in rav1e. It's just a maintenance release that happened to bump semver-major.

atouchet:
Can this be reopened now that image MSRV is 1.85?

Expyron:
Re-opening as the MSRV is compatible now

PR #2527, opened 2025-07-11
Fail CI quicker?

kornelski:
CI immediately launches a ton of jobs, but may fail due to trivial issues like Clippy. 

Here I'm trying to run some basic tests first before spawning the rest.


197g:
Looks good in the workflow graph: https://github.com/image-rs/image/actions/runs/16220960796

PR #2528, opened 2025-07-11
Cache rustdoc builds

kornelski:
The clippy build is slow, trying again without losing so much concurrency


kornelski:
Much faster now

197g:
Nice, clippy now runs under a minute, too!

197g:
It's a shame that Github's support to build containers with the shared `apt-install` setup is rather rudimentary. At least, there does not seem to be a way to specific a `job.container` referencing an image built in a previous step (that could then be cached and shared).

PR #2529, opened 2025-07-11
Reuse build for feature tests

kornelski:


kornelski:
Ugly and not faster

197g:
Good try though, thank you

Issue #2493, opened 2025-06-18
ICC profile read/write support


Shnatsel:
ICC profiles are important for color management, and being able to read them is crucial to be able to display them correctly, while being able to them is crucial to be able to roundtrip images without visual loss for most formats.

Among other things, this is needed by `wondermagick`.

#### Read support

- [x] JPEG
- [x] PNG
- [x] WebP
- [x] TIFF
- [x] AVIF
- [x] GIF https://github.com/image-rs/image/pull/2644
- [X] BMP https://github.com/image-rs/image/pull/2675 

#### Write support

- [x] JPEG
- [x] PNG
- [x] WebP (built-in fast lossless, lossy support [in progress](https://github.com/image-rs/image-webp/pull/155))
- [x] TIFF https://github.com/image-rs/image/pull/2746
- [ ] AVIF blocked on https://github.com/kornelski/cavif-rs/issues/15

#### Third-party read support

- [x] JPEG XL via `jxl-oxide`

#### Third-party write support

- [ ] `webp` crate (currently needed for lossy WebP compression): https://github.com/jaredforth/webp/issues/41

fintelia:
I split the checkbox for `webp` crate support of ICC profiles into a new section, because I kept seeing a "WebP" checkbox unchecked and clicking it (because [`WebPEncoder::set_icc_profile`](https://docs.rs/image/latest/image/trait.ImageEncoder.html#method.set_icc_profile) is available).

mina86:
What’s the scope and status of this issue? Currently, even if decoder reads ICC profiles, `DynamicImage::from_decoder` ignores it.  As far as I can tell, to use `image` user has to do:

```rust
pub fn read(
    path: &std::path::Path,
) -> image::ImageResult<image::DynamicImage> {
    let mut decoder = image::ImageReader::open(path)?
        .with_guessed_format()?
        .into_decoder()?;
    let icc = decoder.icc_profile()?;
    let img = image::DynamicImage::from_decoder(decoder)?;
    match icc {
        Some(icc) => apply_icc_profile(img, icc),
        None => Ok(img),
    }
}

fn apply_icc_profile(
    img: image::DynamicImage,
    icc: Vec<u8>,
) -> image::ImageResult<image::DynamicImage> {
    todo!()
}
```

I assume the end goal is for `DynamicImage::from_decoder` to take the profile into consideration?

fintelia:
[...]

Issue #1554, opened 2021-09-19
Better document luma conversion color space assumptions / consider providing alternatives


naisuuuu:
The `image::imageops::colorops::grayscale` function [currently uses coefficients](https://github.com/image-rs/image/blob/master/src/color.rs#L435) that [assume sRGB color space of the source image](https://en.wikipedia.org/wiki/SRGB#From_sRGB_to_CIE_XYZ). While this is hinted at in the actual source code with a comment and the coefficient constant name, [it is not mentioned in user-facing docs](https://docs.rs/image/0.23.14/image/imageops/colorops/fn.grayscale.html). The assumption of sRGB (or rec 709) color space should be documented as it may not be obvious and will provide significantly different results to rec 601.

It's also worth noting that the image crate may be unusual in it's color space assumption compared to some high-level image manipulation libraries in other languages. Examples include [go's stdlib `image/color`](https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/image/color/color.go;l=244;drc=refs%2Ftags%2Fgo1.17) or [python's pillow](https://pillow.readthedocs.io/en/stable/reference/Image.html?highlight=Image.convert#PIL.Image.Image.convert), which both use [rec 601 coefficients](https://en.wikipedia.org/wiki/Luma_(video)#Rec._601_luma_versus_Rec._709_luma_coefficients). This tripped me up when testing a project of mine.

I'm unsure if there's a good solution to the lack of one good default color space among different libraries. Maybe we could add a more robust grayscale function to the `imageproc` crate and apply different coefficients based on user input (similar to [imagemagick](www.imagemagick.org/script/command-line-options.php#intensity) or [opencv](https://docs.opencv.org/4.5.3/de/d25/imgproc_color_conversions.html))?

micahsnyder:
I've also run into this. In a prototype program for my project I found I needed two changes to closely match results from Python's `PIL.Image.convert("L")` grayscaling feature:
- I needed RGB->Luma coefficients from the ITU-R 601-2 luma transform: `L = R * 299/1000 + G * 587/1000 + B * 114/1000` used by `PIL.Image.convert("L")`.
- I needed to round Luma floating point values to the nearest integer for conversion back to `u8`'s, rather than truncating.

Here's my customized `grayscale()` function: https://github.com/micahsnyder/clam-rust-experiments/blob/07b4fbf685624447ccafa25fb16bb06632a12113/rs-phash/src/main.rs#L220-L260
[...]

Issue #1030, opened 2019-09-05
[Feature Request] Ability to apply color profiles to an ImageBuffer/DynamicImage


davidarmstronglewis:
I would like to be able to apply an `.icc` color profile to an `ImageBuffer`

My specific use case for this functionality is to enable converting an image to CMYK colorspace for printing.

This is more generally applicable to printed documents.

## Draft

```rust
let cmyk_profile = include_bytes!("../GRACoL 2013 CRPC6.icc");
let mut my_image = DynamicImage::ImageRgb8(...);
my_image.apply_color_profile(cmyk_profile);
```

fintelia:
This is related to #786. I'd like to have proper color space support, but there are a bunch of other higher priority items for the crate so it probably won't be coming any time soon (unless you or someone else want to try implementing it!)

Shnatsel:
There are still no built-in methods for this, but you should be able to accomplish this with https://crates.io/crates/moxcms or maybe https://crates.io/crates/qcms

PR #2531, opened 2025-07-18
Implement Cicp for ImageBuffer,DynamicImage and provide ColorTransform on moxcms

197g:
The goal here is to bring CICP profiles into `ImageBuffer` / `DynamicImage`. CICP is a perfect fit since, similar to PNG, we only have an rgb-ish pixel canvas in these types (and the sentiment is to have more complex representations, if at all, in separate types).

The first consequence of performance is that we should not generalize over the subpixel type. Handling combinations with sample conversion is only possible when we have an enumeration of possibilities (as in `DynamicImage`) and requires matches with each individual combination handled. Instead, we here convert through `f32` when encountering mismatch to trade a little bit of performance with somewhat terse code.

This uses the internal sealed trait to ensure we only need to handle the `PixelWithColorType` types in the public interface¹. The dispatch that makes this work is a little bit cursed (see comment in `color::private`) but fully hidden away. Also the code is building all tables eagerly for 16 combinations × 3 color depths. This should not be a huge burden in practice. We can patch it later if users want to convert small images infrequently. The whole conversion apparatus allows caching of the `CicpTransformer`, which is also `Clone` and comprised of mainly `Arc<dyn Fn()>` for the individual paths.

Closes: #786
Relates: #1554, #1030, #2493

¹For historical reasons, `Dynamicimage::to_luma_f32` exists without having `ColorType::Luma32F`


197g:
@awxkee I'd be glad to know if there's any gross misuse of your cms in the integration. Obviously, exercising the transformers on single pixels at a time isn't the final state here, it should take slices of pixels on both sides and forward their channels to the cms transformer in large chunks. (ImageBuffer can make use of the packed layout without any row strides to do everything at once).

src/color/cicp.rs
```diff
@@ -0,0 +1,445 @@
...
+    Primitive,
+};
+
+/// Reference: <https://www.itu.int/rec/T-REC-H.273-202407-I/en> (V4)
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Cicp {
+    primaries: CicpColorPrimaries,
+    transfer: CicpTransferFunction,
+    matrix: CicpMatrixCoefficients,
+    full_range: CicpVideoFullRangeFlag,
```
fintelia: We'll have to see whether these two fields end up being needed.

In PNG, only the matrix coefficients representing "RGB" are allowed. And images without the full range [flag set are expected to be extremely rare](https://github.com/w3c/png/issues/312#issuecomment-2327349614) to the point that even browsers aren't planning on supporting them.

Cargo.toml
```diff
@@ -37,6 +37,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 bytemuck = { version = "1.8.0", features = ["extern_crate_alloc"] } # includes cast_vec
 byteorder-lite = "0.1.0"
+moxcms = { version = "0.7" }
```
awxkee: Since image prefers safe code, you can disable all unsafe optimizations in transform by turning off all intrinsics features in moxcms. This will cost 450% of performance though. I put some efforts to make "this code may not be in SIMD" something fast.

src/color/cicp.rs
```diff
@@ -0,0 +1,445 @@
...
+            // I'd use record update syntax but there's a private field.
+            let mut p = moxcms::ColorProfile::new_gray_with_gamma(1.0);
+            p.profile_class = moxcms::ProfileClass::DisplayDevice;
+            p.rendering_intent = moxcms::RenderingIntent::Perceptual;
+            p.pcs = moxcms::DataColorSpace::Xyz;
+            // Using red; but assuming all channels have the same TRC.
+            p.gray_trc = rgb.red_trc.clone();
+            // I think this is a bug? The build_gamma_table uses (moxcms =0.7.1)
+            //
+            // let gray_gamma = dst_pr.build_gamma_table =:<T; 65536, GAMMA_CAP, BIT_DEPTH>(
```
awxkee: Yes, will do the fix.

src/color/cicp.rs
```diff
@@ -0,0 +1,445 @@
...
+            // and the selected implementation library is checked for suitability in use.
+            return None;
+        }
+
+        let from = from.to_moxcms_profile()?;
+        let into = into.to_moxcms_profile()?;
+
+        let opt = moxcms::TransformOptions::default();
+
+        // TODO: really these should be lazy, eh?
```
awxkee: Creating matrix shaper profiles until you're doing this at scale is cheap. 
Typical 8-bit matrix shaper creation costs 65KB memory and ~4k power function execution what is cheap.

src/color/cicp.rs
```diff
@@ -0,0 +1,445 @@
...
+            // To avoid regressions, we do not support all kinds of transforms from the start.
+            // Instead, a selected list will be gradually enlarged as more in-depth tests are done
+            // and the selected implementation library is checked for suitability in use.
+            return None;
+        }
+
+        let from = from.to_moxcms_profile()?;
+        let into = into.to_moxcms_profile()?;
+
+        let opt = moxcms::TransformOptions::default();
```
awxkee: Here are two caveats:

- Since it uses fixed point transforms as a default it immediately introduces 0.05% additional error to any transformation. Considering the scale of the image it might be better to use as a default more generalized approach. OTH Adobe CMM and Apple ColorSync also do transforms in fixed point as a default.
- It also uses an extended range for f32 by default, so it may unexpectedly return negative values or above 1.0. This also typical behavior of lcms2, but might be not really something what you'd expect.

src/color/cicp.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// Corresponds to value 22 but
+    ///
+    /// > No corresponding industry specification identified
+    ///
+    /// But moxcms identifies it as EBU Tech 3213-E: <https://tech.ebu.ch/docs/tech/tech3213.pdf>
+    ///
+    /// However, there are some differences in the second digit of red's CIE 1931 and the precision
+    /// is only 2 digits whereas CICP names three; so unsure if this is fully accurate as the
+    /// actual source material.
+    Industry22 = 22,
```
awxkee: I suspect the roots of this comes from [FFMPEG](https://github.com/ubitux/FFmpeg/blob/3abb80f65b44a803dda498e111b8c96a80f77d10/libavformat/avisynth.c#L590). I don't have any concrete information on this side, but almost any software I know identifies this as so.

awxkee (commented):
Nice to see it here! I don’t think there are any misuses, but I’ve left some thoughts.

src/color/cicp.rs
awxkee: I believe the fields exist purely for full compatibility. From my view, range and matrix should be handled at the YUV conversion stage and don't make much sense on the colorimetry side.

Cargo.toml
197g: That seem too much cost. I'd rather invest in reducing the amount of unsafe over time as more SIMD is available safely or with appropriately vetted wrappers such as the [`safe_unaligned_simd`](https://docs.rs/safe_unaligned_simd/latest/safe_unaligned_simd/) project.

src/color/cicp.rs
fintelia: Makes sense. Probably means that something like `ImageBuffer` should carry only primaries and transfer function

src/color/cicp.rs
```diff
@@ -0,0 +1,445 @@
...
+            transfer_characteristics: self.transfer.to_moxcms(),
+            matrix_coefficients: self.matrix.to_moxcms()?,
+            full_range: match self.full_range {
+                CicpVideoFullRangeFlag::NarrowRange => false,
+                CicpVideoFullRangeFlag::FullRange => true,
+            },
+        });
+
+        let gray = {
+            // I'd use record update syntax but there's a private field.
```
awxkee: AFAIK ```ColorProfile::default()``` or ```ColorProfile { foo: bar, ..Default::default() }``` should work.

197g: Does not seem so. At least not outside the crate defining it as non-exhaustive due to potential future private fields (those semantics have not been clearly defined). https://github.com/rust-lang/rust/issues/70564

src/color/cicp.rs
```diff
@@ -0,0 +1,445 @@
...
+        assert!(CicpTransform::new(Cicp::DISPLAY_P3, Cicp::SRGB).is_some());
+        assert!(CicpTransform::new(Cicp::DISPLAY_P3, Cicp::DISPLAY_P3).is_some());
+    }
+
+    #[test]
+    fn transform_rgb() {
+        let tr = CicpTransform::new(Cicp::SRGB, Cicp::DISPLAY_P3).unwrap();
+        let p3 = Rgba::<u8>([255, 0, 0, 255]).to_rgba_with(&tr);
+        assert_eq!(p3.0, [234, 51, 35, 255]);
+    }
```
197g: I've expanded test coverage a little but got stumped. Here's what the test looks like when it succeeds:

```rust
    #[test]
    fn transform_luma() {
        let tr = CicpTransform::new(Cicp::SRGB, Cicp::SRGB).unwrap();
        // _, Y, _ = colour.RGB_to_XYZ(
        //   [1.0, 0.0, 0.0],
        //   colour.RGB_COLOURSPACES['sRGB'],
        //   apply_cctf_decoding=True)
        // colour.RGB_COLOURSPACES['sRGB']._cctf_encoding(Y)*255
        let luma = Rgba::<u8>([255, 0, 0, 255]).to_luma_with(&tr);
        assert_eq!(luma.0, [130u8]); // reference: 127.1021805160301

        let luma = Rgba::<u8>([0, 255, 0, 255]).to_luma_with(&tr);
        assert_eq!(luma.0, [220u8]); // reference: 219.93274897493274

        let luma = Rgba::<u8>([0, 0, 255, 255]).to_luma_with(&tr);
        assert_eq!(luma.0, [70u8]); // reference: 75.962697358369013
    }
```

@awxkee I can see floating point being imprecise but that seems a surprisingly large difference to the reference values. And sometimes too large, some too small. Could you verify my approach? In my implementations in `image-canvas` Ive always achieved accurate-to-rounding test vectors.

src/color/cicp.rs
197g: With all the different transform tables this is building, we get about `2 << 19` power function executions in total, which is no longer cheap. (16 tables for u8, u16, f32 respectively). The tests take a very noticeable time even with `opt-level = 2` for `moxcms`.

src/color/cicp.rs
awxkee: Your software likely assumes D65 white point colorimetry. moxcms was originally intended to be an ICC converter, and ICC profiles use a fixed D50 white point until ICC MAX (v5) is released. If you want to use D65, you can't use `update_rgb_colorimetry_from_cicp`, because it performs chromatic adaptation to D50.

At least for D50, it matches the tests with reasonable accuracy. I'd even argue which results may be more accurate, but that's not guaranteed.

If you want to perform reference testing, it's better to avoid using `u8`. This is ICC software and is heavily optimized, so `u8`, `u16`, and `f32` (if extended range is disabled) will use LUT tables, which introduce additional quantization.

awxkee: To be clear: all primaries defined in ICC profiles are assumed to be under D50 illumination. As a result, all input is adopted to D50, and from that point the conversion becomes straightforward ( and no other adaptation is required ), since everything is already in a D50 illumination.

This works well for ICC profiles (since ICC profiles should assume D50 white point ). However, it might work not such well if you want different white point or conversion between them. That's not handled out of the box.

197g: While that clears up the difference, technically CICP primaries _also_ specify a reference whitepoint (https://www.itu.int/rec/T-REC-H.273-202407-I/en, Table 2). So it is a bit surprising that the update to colorimetry would not introduce an adaptation matrix accordingly. I suppose for the purpose of further testing, I could set a von-Kries `chromatic_adaptation` by copying over some code from `image-canvas`.

awxkee: > While that clears up the difference, technically CICP primaries also specify a reference whitepoint (https://www.itu.int/rec/T-REC-H.273-202407-I/en, Table 2)

Those are two slightly different things. The primaries being used are correct, but it's how the primaries are defined. The actual transformation takes place under the D50 white point, as required by the ICC standard. The ```ColorProfile.chromatic_adaptation``` should be set to Bradford ( it's set by default ), since chromatic adaptation is always applied.

You've encountered a small edge case here. It's usually not a big deal for RGB -> RGB conversions, because the adaptation matrix will also be inverted through M^(-1) ( (D65->D50)^(-1) = (D50 -> D65) ), wheareas the final result should still be in D65 sRGB. However, in the case of grayscale, only the D65 -> D50 adaptation is applied, so the result ends up in XYZ values which scaled to D50.

awxkee: I’ll also note if you want to make those images Xyz D65 then you can’t export them directly with embedded ICC profile, since all ICC software will consider such gray images as they are already Xyz scaled to D50.

I’m not sure how it should be handled if only CICP tag is set. I think it’s not really specification defined and rather it will be implementation defined. In most cases lcms2 will be used and it will assume those image as Xyz D50.

I’ll try to check later how Chrome will handle this.

awxkee: [Chrome](https://chromium.googlesource.com/chromium/src/+/lkgr/ui/gfx/color_space.cc) also makes connection on Xyz D50, and I see nothing specific to possibility of Xyz D65. Therefore, I believe it will also consider all gray images as they are D50.

197g: Hm, that still seems odd. The code for luma _does_ copy the chromatic adaptation matrix and `white_point` from the connected rgb color space, hence I assumed that would define the XYZ used to derive the output (i.e. adaptation to define the pcs->luma and white point/adaptation to define luma->pcs) So while the pcs is XYZ-D50, should setting the matrix for the gray color space not also result in the inverse being applied when used as a target? Sure, Luma is just extracting the Y component yet that matrix transform is still relevant. It should extract Y of XYZ/D65 when a bradford transform XYZ/D65 -> XYZ/D50 is set to define the color spaces relation to the PCS.

If applying a chad isn't supported for luma then maybe an error should be produced in `create_color_transform`.

awxkee: See https://www.color.org/specification/ICC.1-2022-05.pdf page 101 point F.2. The specification clearly states that `connection = grayTRC[device] ` and `device = grayTRC^(-1) [connection] `.

I understand that it is quite unexpectable, but for ICC software I believe "just following specification" is preferred.

> Sure, Luma is just extracting the Y component yet that matrix transform is still relevant. It should extract Y of XYZ/D65 when a bradford transform XYZ/D65 -> XYZ/D50 is set to define the color spaces relation to the PCS.

I believe it's better to understand ICC chromatic adaptation as "How did you get to D50?", not "What should the output illumination be?'".

In fact, anything you set in chad is almost always ignored. It's more of an informational tag, since everything is already adapted to D50 ( page 35 p 9.2.15 ). All correct inversions are derived from algebraic identities, not by computing the adaptation inverse.

197g: As a computation model this on its own makes sense. The function ends up being number-to-number either way. When we whitepoint adapt a luma value all that happens is applying a coefficient, as the whole thing is linear. So we can always write a `curveType` tag that models the combined transformation if we wanted to represent gray values from another whitepoint.

So I guess I agree with following specification in moxcms as far as the implementation for `ColorProfile` is concerned. However, I still thing we need to represent those CICP's gray spaces as if they had the whitepoint given in the primaries. We'd then need to encode a different greyTRC not just copying the redTRC. As it also says in 9.2.30 "The tone reproduction curve provides the necessary information to convert between a single device channel and the PCSXYZ or PCSLAB encoding" and that is definitely _not_ what it is achieved when the gray intends to use a different whitepoint.

197g: Of course, that means we can not expect `sRGB -> grey/D65` to be correct. The encoding of the PCS's Y through greyTRC would be lossy without an inverse whitepoint adaptation, whereas luma->luma is _not_.

197g: There is of course yet another alternative.

> [cicpTag] This tag may be present when the data colour space in the profile header is RGB, YCbCr, or XYZ, and the profile class in the profile header is Input or Display. The tag shall not be present forother data colour spaces or profile classes indicated in the profile header.

What models my intent _best_ is that `Luma` represents the `Y` of an YCbCr data color space. Is that usually encoded as a multiProcessElementsType or is a lutAToBType with M -> matrix -> B curves sufficient? Well, and to do that efficiently a `create_masked_transform` is needed that only computes some of the color components (index 0 in this case) of an output data space..

awxkee: I think idea with `greyTRC` which fused with XYZ D50 -> XYZ D65 is good enough.

```multiProcessElementsType``` is quite complicated thing, it require to create not obvious huge LUT which usually in LAB and those are for DeviceLink profiles ( moxcms unfortunately don't support DeviceLink profiles ). I don't think it is the good way.

More common way to link it through `BToA0Tag` and `A0ToBTag` ( for Input Device and Display Device profiles ) but those also require to create much more complicated LUTs.

> Is that usually encoded as a multiProcessElementsType?

Through ICC LUTs, it should be possible to represent almost anything (moxcms can handle any LUTs with up to 16 inks, though I only personally tested up to CMYKOGV, which one I have). However, I think that might be a bit overkill for this case.

More typically ICC LUTs are used for CMYKs, tonemapping, in PDFs some images are not extremely rarely use LAB.

I'll try to make an API and see what the accuracy cost is. However, for common 8-bit transforms I predict that a typical 4096 point curve will be more than enough for good accuracy.

Comment (197g) [reply src/color/cicp.rs]: lutAToBType with no CLUT, M curves as the original TRC, a matrix, and B-curves identity,  seems on paper to also have the intended effect for an YCbCr (though I don't see if the adaptation matrix is used in that path, it can as well be fused with the YCbCr/matrix of that tag manually). So the hinderance would be 'masked' create_transform methods which discards CbCr components in the output and assumes them `0` in the input.

Comment (awxkee) [reply src/color/cicp.rs]: That appeared more complicated than expected :)

So my prediction that tone curve is enough to represent tone reproduction curve with XYZ scaling successfully failed. Non linearity is too high, it's not even close.

16-bit lut (b to a) works nice, accuracy loss is low, thus for reference value ```75.94789574522055``` it produce ```75.93887075264664``` what is acceptable. 8-bit LUT makes ```74.02004223358418``` what is not really accurate but not that bad.

> with no CLUT

That's not possible to create transform with different input and output channel count without CLUT. Only through CLUT you can remap PCS XYZ -> 1 output channel and vice verca.

I haven't used 1 channel LUTs before so a few bugs were fixed along the way.

Perhaps I’ll do some API rework in a few days with a minor version update. This will allow to bypass at least for computing D50 white point requirement, since it's still inconvinient if you just make conversion with different white point to do all this stuff. However, it's unclear how to prevent user from saving such profiles, so this probably will be user responsibility not to export "computing profiles".

Comment (197g) [reply src/color/cicp.rs]: I wasn't thinking of a _profile_ with different output channel count. The missing piece would indeed be purely API that acts on a profile with an output count of `3` (YCbCr) while only writing the Y channel, i.e. taking a slice of outputs with 1 channel each. For YCbCr that even seems sort-of interesting in the general case (to separate the components into planes); but it's your crate so judge the complexity this would add accordingly.

Comment (awxkee) [reply src/color/cicp.rs]: TLDR It’s a nice feature to have, but I probably won’t add it.

Having planar APIs like lcms2 does is actually nice to have.
But if we want to use this in safe Rust, it means no pointer aliasing is available, and that immediately makes things more complicated, since it requires duplicating all the transforms, checks, and other things.
This is a case where C/C++ still wins, as pointer aliasing can be done easily (with known consequences).

This also relates to in-place transforms. I know that many transforms can be done in-place nicely, but this is trade-offs between complexity and safety.

Those two cases will cost about ~20-30K lanes of quite complex code for each, where deinterleaving or allocating additional memory on user side costs 1-3 lanes of very simple code. Therefore, unless I start to work on something with extremely limited resources where such constraints actually matter it's unlikely to happen.

Comment (197g) [reply src/color/cicp.rs]: Sounds reasonable. Would you ping this thread if/when lutAToBType would be available for use so that I can prototype YCbCr in its full 3-component form? The integration here might then have the conversion to/from luma with an additional allocations, that's fine.

---

For eventual planning, with regards to pointer aliasing and in-place making things more complicated: I've of course encountered similar underlying issues for `image-canvas` but am experimenting to solve for it. Rust _allows_ being expressive in terms of aliasing and it would be a shame if that were to restrict it instead of making it more powerful. A lot of `image-texel` is spent on `unsafe` abstractions so that this aliasing can at least be done correctly, this could probably be re-used. I've also pondered if it would be worth moving some of these abstractions into an `image-core`. Based on my experience, 20k-30k seems like an overestimation but I can't speak on whether the code is already structured to make full use of it so it would definitely be a large rework. I've opened a discussion on `safe-unaligned-simd` to deal with this issues, it'd be amazing if that code could simultaneously be completely safe.

Comment (awxkee) [reply src/color/cicp.rs]: Could you try [this](https://github.com/awxkee/moxcms/pull/92) branch? You now have possibility to bypass D50 connection rule, thus getting output in Xyz D65 is easily doable by using ```ColorProfile::new_from_cicp_with_no_d50_adoption``` (also here is also a few new API explicitly allowing bypassing rules), but to create transform  that ignore D50 white point ```TransformOptions.bypass_icc_d50_white_point``` must be also set. 

[Here](https://github.com/awxkee/moxcms/blob/ba763b3e865e47f8a0d0ea19eb79706395a61fe7/app/src/encode_gray_lut.rs#L65) is an example how to encode Xyz D50->Gray D65 LUT. Nothing specific is required for YCbCr it should work in previous versions as well in this one, if you want to go this way.

Let me know if you have any issues.

Comment (awxkee) [reply src/color/cicp.rs]: @197g Have you seen the post above?

File: src/images/dynimage.rs [hunk 9]
```diff
@@ -1043,6 +1071,144 @@ impl DynamicImage {
...
+        &mut self,
+        other: &DynamicImage,
+        options: ConvertColorOptions,
+    ) -> ImageResult<()> {
+        // If any of these images are luma we are only correct when they are in Linear sRGB. Half
+        // of that is the color convert trait which can't pass information and thus only uses
+        // coefficients which are for the D65 whitepoint in sRGB primaries. The other half is that
+        // moxcms support for expressing an YCbCr space with transfer is not there yet.
+        //
+        // Implementation ideas: Maybe we'll get that by buffering into a full YCbCr image
```
awxkee: Just to make sure nothing missed link to post [here](https://github.com/image-rs/image/pull/2531#discussion_r2219970008).

src/color/cicp.rs
197g: Yes, but having a hard time figuring out if the approach is just hard to get working or if I was thinking about it wrong. So in the meantime I'm considering just leaving out the ability for Luma transforms. (Other than in linear sRGB primary color space which is also what is implemented already within `image` in the pixel convesion, so using that).

It seems like there is necessary complexity; it needs to live in a foundational crate so that moxcms can provide the masked YCbCr interface as mention above with bindings; or it lives in moxcms directly; or it lives in `image` to build a different pipeline for that sample layout (i.e. intermediate allocation). And even the branch does not remove much of that complexity. And I'm all the intricacies of bypassing _some_ conversion _sometimes_ will be hard to test. It also violates one of the principles that the CICP implementation is supposed to follow, that we only use an approach that can be expressed precisely enough to be replicable with other types.

197g: So just to clarify: the `create_gray_to_xyz_samples` and `create_xyz_to_gray_samples` and the utilities to create the right LutMultidimensionalType's make a lot of sense. They are somewhat clearly defined and we can explain them in terms of an actual ICC profile; so that's just using what the specification allows if I see correctly. But the other hacks are rough to use as well. 

So can those kinds of profiles be used to get the right result with little hacks? That route is also encoding the whitepoint adaptation so we can use the D50 PCS as intended. Would it make sense to build those into the create instead of the app binary?

awxkee: > And I'm all the intricacies of bypassing some conversion sometimes will be hard to test. It also violates one of the principles that the CICP implementation is supposed to follow, that we only use an approach that can be expressed precisely enough to be replicable with other types.

Agreed.

> It seems like there is necessary complexity; it needs to live in a foundational crate so that moxcms can provide the masked YCbCr interface as mention above with bindings; or it lives in moxcms directly; 

> So just to clarify: the create_gray_to_xyz_samples and create_xyz_to_gray_samples and the utilities to create the right LutMultidimensionalType's make a lot of sense.

I'm OK to provide convenience methods for this in moxcms.

> So can those kinds of profiles be used to get the right result with little hacks?

Yes. I believe there is no need to consider these as hacks, they are legit profiles, just more complicated than typically expected :)

The issue with exporting images that use such profiles still persists. To handle single channel LUTs ( or YCbCr it doesn't really matter), customers will need a fully featured CMS.
Aside from Chrome, lcms2, moxcms, Apple ColorSync, and Adobe CMM ( or also some other old and huge CMS ), most other systems will likely fail. For example, Firefox’s qcms doesn’t support LUT profiles beyond CMYK, or RGB color spaces.
If Linux desktop image viewers aren't backed by lcms2, they will likely fail as well.

However, this seems to be a necessary trade off.

> Would it make sense to build those into the create instead of the app binary?

Do you mean that we could encode those profiles at runtime? If so, then yes we can.

197g: Yes, having those profiles available at runtime just for the conversion between color spaces would already be awesome. While exporting images with full fidelity would also be nice I agree we should go with the more compatible option here (at least on ImageBuffer/DynamicImage).

However, I'm also consider whether it would be okay to use a greyLut for exports. The issue with using it in conversion is that it doesn't roundtrip into RGB—but it does not need to for pure viewing. That is hacky but I could see `image` have a `encodeLutForD50` that turns the `Y` color space into a mangled EOTF displaying the _intended_ luminance when each value is reinterpreted as an input in ICC's D50 gray. Not correct for further calculation but correct-ish (except for the small chromatic tint from a whitepoint mismatch) for _display_.

197g:
In light of the full transformation complexity, I propose we settle for this:
- `ImageBuffer` (and `DynamicImage`by extension) allow specifying their primaries and transfer function.
- `copy_from_color` assigns color data of another color space to an `ImageBuffer` or `DynamicImage`. In the former case we only allow copies we can transform directly (in particular, no Luma) while the latter supports Luma with Linear-sRGB-primaries which is consistent with the coefficients we use in `DynamicImage::to_luma`'s conversion which is being used internally.
- `dynimage_from_decoder` set the buffer's color space to `sRGB` which is correct for the behavior via jpeg and a good default for most other decoders. The protocol to indicate an CICP chunk or another color space entirely will be part of future PRs. So we do not decide whether to convert color spaces in decoding or encoding for now.
- The `Pixel` extensions have been removed as they would not be bulk-optimized and the error for `luma` would make the interface way too confusing, there would be full methods that are not usable in their current implementation.

src/color/cicp.rs
```diff
@@ -86,24 +102,59 @@ impl CicpColorPrimaries {
     }
 }
 
+/// The transfer characteristics, expressing relation between encoded values and linear color
+/// values.
+///
+/// Refer to Rec H.273 Table 3.
 #[repr(u8)]
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub enum CicpTransferFunction {
```
fintelia: nit: add `non_exhaustive` in case more values are added to Rec H.273 (also applies to the color primaries and matrix coefficients)

Cargo.toml
```diff
@@ -2,10 +2,10 @@
 name = "image"
 version = "0.25.6"
 edition = "2021"
-resolver = "2"
+resolver = "3"
```
fintelia: The v3 resolver doesn't really work well for library crates and would cause both local and CI builds to run with outdated dependencies even when not passing `-Z minimal-versions`

What happens is that the MSRV of 1.85.0 gets treated as the maximum MSRV for any direct or transitive dependency, and Cargo pretends that any versions of dependencies that have a higher MSRV don't exist. It can be worked around with the right flags/env variables, but it is a hassle.

src/lib.rs
```diff
@@ -138,13 +138,19 @@ pub use crate::color::{ColorType, ExtendedColorType};
 
 pub use crate::color::{Luma, LumaA, Rgb, Rgba};
 
+pub use crate::color::cicp::{
+    Cicp, CicpColorPrimaries, CicpMatrixCoefficients, CicpTransferFunction, CicpTransform,
+    CicpVideoFullRangeFlag,
+};
```
fintelia: This is a lot of types to add to the crate root. Could we add them to the `metadata` module instead?

src/images/buffer.rs
```diff
@@ -983,6 +991,32 @@ where
     }
 }
 
+impl<P: Pixel, Container> ImageBuffer<P, Container> {
+    /// Define the color space for the image.
+    ///
+    /// Reinterprets the existing red, blue, green channels as points in the new set of primary
+    /// colors, potentially changing the apparent shade of pixels.
```
fintelia: Probably worth saying something like "the pixel values themselves are not changed"

src/images/buffer.rs
```diff
@@ -1406,6 +1443,136 @@ where
...
+    }
+
+    /// Convert this buffer into a newly allocated buffer, changing the color representation.
+    ///
+    /// This will avoid an allocation if the target layout or the color conversion is not supported
+    /// (yet).
+    ///
+    /// See [`ImageBuffer::copy_from_color`] if you intend to assign to an existing buffer,
+    /// swapping the argument with `self`.
+    pub fn to_color<IntoType>(
```
fintelia: Perhaps `apply_color_space` or `apply_color_space_transform`? That would parallel the `apply_orientation` method.

fintelia (commented):
I like the focus of doing color space conversions on `ImageBuffer` / `DynamicImage`.

How would you feel about making the conversions happen in-place? I think it could significantly simplify the public API. It would remove the potential to convert RGB <-> RGBA or change the sample format at the same time as a color space conversion, but those could always be done in two steps.

The `moxcms` crate doesn't directly support in-place conversions but we could implement it ourselves by iterating over the backing slice with `chunks_mut` and copying each chunk into a stack (or heap) allocated buffer.

Cargo.toml
197g: Good hint, I was under the impression it would choose based on the rust-version of the toolchain compiling the package but it instead uses the `rust-version` field in the crate declaring the dependency. I'll undo this then but am continuing to be surprised. Well it makes sense considering `.lock` file generation.

197g:
Allowing two different color types to take place in one operation is a pretty large _feature_.  I'd even like to have different sample types for Input and Output but moxcms does not yet have that capability. The conversion is not *accurate* in all combinations (hence also the use `32F` where it can't use a direct conversion), and (much) more performant directly too—as in allocation-free. I really don't see a good way of teaching or providing that as separate steps. Consequently, I was trying to follow the common designs for BigInt and matrix libraries, with the similarity of requiring some separate in and out buffers for operations. In-place operations are at best a special case.

Following that logic, of course `apply_*` are all in-place methods whereas these methods are not. `to_color` specifically follows the naming of `<[T]>::to_vec` (where `copy_from_color` is conceptually related to `copy_from_slice`). So `DynamicImage::to_color` is still missing but it's target colortype should arguably be a simple argument instead of a type parameter so I'll hold that for a separate PR. There we could also discuss whether to add the special cases of `apply_colorspace` which does _not_

awxkee:
I tried to implement in-place transforms in moxcms, but eventually I stopped, realized it's neither easy nor particularly useful.
It appeared that it requires a significant amount of work, and it provides no measurable performance benefit (performance gains were below the noise level). So, I gave up on implementing it. Supporting in-place transforms would mean rewriting around 50 transforms manually, either in duplicate or using `UnsafeCell`, not counting auxiliary code to make checks and prepare them. You may argue it is doable just to make an abstract transform, but in that case it would be 450% slower, what is also doesn't seem to be particularly useful.

Unless you're in a sandboxed environment where allocations are restricted or too expensive, I see no compelling reason to maintain in-place transforms. Doing a lot of work just to avoid writing a single extra line of code doesn't seem like strong motivation to me.

fintelia:
All the existing methods on `ImageBuffer` and `DynamicImage` currently either operate in-place, or they return a newly allocated image. None of them take in a separate image to completely overwrite with the result of the operation. I strongly feel that these new methods should continue the pattern and either be in-place or allocate+return.

If we're going to add the ability to switch the sample format while doing the transform, that to me is a strong reason to pick allocate+return. In particular, something like converting 8-bit sRGB to 16-bit Rec. 2020 makes a lot of sense to me.

But I'm still not sure about being able to add/remove an alpha channel at the same time as a color space transform. Having it part of the API while banning RGB <-> Luma conversions adds an unfortunate gotcha. And I'd guess that it wouldn't take meaningfully longer to change color space and alpha channel in two steps versus one?

---

@awxkee In place transforms wouldn't require any changes to `moxcms`. The `image` crate could use a small scratch buffer to target the existing API via something like:
```rust
let mut scratch = [P::Subpixel::DEFAULT_MIN_VALUE; 1024 * P::CHANNELS];
for chunk in (&*self.data).chunks_mut(scratch.len()) {
    let scratch = &mut scratch[..chunk.len()];
    scratch.copy_from_slice(chunk);
    transform(scratch, chunk);
}
```

197g:
> But I'm still not sure about being able to add/remove an alpha channel at the same time as a color space transform. Having it part of the API while banning RGB <-> Luma conversions adds an unfortunate gotcha. And I'd guess that it wouldn't take meaningfully longer to change color space and alpha channel in two steps versus one?

But converting the bit-depth at the same time is problematic; and addressed by the very same API. 8-bit Luma to 16-bit or 32-float RGBA makes a lot of sense to avoid some of the banding; as does using another target depth when converting color spaces. That is routine for something like an editor that has an internal surface in lin-rgb RGBA but gets from the outside arbitrary other depths and colors. You can pretty much only do this wrong outside `image` in two-steps, also considering that `DynamicImage` is still marked `#[non_exhaustive]`.

Regarding the claim of consistency, it is the analogue of [copy_from](https://docs.rs/image/latest/image/trait.GenericImage.html#method.copy_from) but correctly modelled with efficiency in mind first, not doing opaque type dispatch that leads to us having to give the problematic advice of writing out the loop pixel-by-pixel. (`SubImage<&mut DynamicImage>` should also get `copy_from_color` but not in this PR).

fintelia:
I was thinking that it would be possible to bound the method with `P: Pixel<Subpixel=T>` and `Q: Pixel<Subpixel=U>` where `P` and `Q` are both `Rgb`, both `Rgba`, etc. But I'm realizing that's not a thing Rust lets you do since `Rgb<u8>` and `Rgb<u16>` are independent types. So yeah, I think we do need the complexity of the API having arbitrary source and destination pixel types. 

For the method signatures, having `DynamicImage` implement something like having both of these seems more natural to me:

```rust
/// Perform the transform in-place (using a scratch buffer)
pub fn apply_color_space_transform(
    &mut self,
    color_space: Cicp,
    options: ConvertColorOptions) -> ImageResult<()>;

/// Check if the transform is supported, then allocate an output image and apply it.
pub fn convert_color_space(
    &self, 
    color_space: Cicp,
    color_type: ColorType, 
    options: ConvertColorOptions) -> ImageResult<DynamicImage>;
```
To me, color space conversion feels similar to operations like `DynamicImage::{blur, grayscale, or apply_orientation}`: you feed in an image and maybe some additional data about how the pixel values should be changed, and you either have the transformation applied in place (where possible) or get back a newly allocated image.

I also think it is important to use a more expressive suffix than `_color` for the methods. I'm not sure if `_color_space` is the best option (or even accurate?) but given that we already refer to colors in a lot of places, I think it would be easier to explain if we're careful to always distinguish "color type" and "color space".

197g:
I think the best option would be to ensure that luma works too. This should be possible by interpreting Luma as if we had set the `matrix` to `ChromaDerivedNonConstant`. This ensures we can very easily transform it into its related RGB space without having to deal with transfer functions. Really all that's required is calculating the Kr and Kb constants per [Rec H.273 (39) and (40)](https://www.itu.int/rec/T-REC-H.273-202407-I/en) from the whitepoint and primary coordinates. That's cheap for `ImageBuffer`.

With that we can resolve the mismatch between the APIs and just provide both: a way to copy into and a way to apply, and change the color space. So all of it. That's probably the most consistent way to deal with it for now. I still wouldn't provide it for pixels because it is _not_ cheap per-pixel transformation. So that interface should stew for a bit..

Then we can add `ChromaDeriveConstant` later where the Luminance is calculated from _linear_ rgb (and the other way), which would use the moxcms utilities discussed above.

197g:
@fintelia Would you have time for re-review?

The scratch buffer approach works somewhat but is unfortunately slightly more complicated than the sketch. If we would be operating on the most generic `Pixel` there's no good way to 'promote' it to a uniform `Rgbaf<32>`. For Luma we must already do a conversion within that promotion (which depends on the color space) and even for the rgb(a) types the numerics are finicky to express generically. I've thus implemented that transform only for the subset necessary for `PixelWithColor` on `CicpTransform` with the caveat that it won't be available in the public API to transfer its implementation to other kinds of buffers.

src/metadata/cicp.rs
```diff
@@ -0,0 +1,1346 @@
...
+}
+
+/// The used encoded value range.
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[non_exhaustive]
+pub enum CicpVideoFullRangeFlag {
+    /// The color components are encoded in a limited range, e.g., 16-235 for 8-bit. This was used
+    /// for some in-band auxiliary data in the past (overswing of noisy filters to be clipped by
+    /// the analog decoder), but is generally not used anymore.
```
awxkee: Actually, limited range YUV still dominates over full-range. If something is not made by high end production camera and is YUV-encoded it’s likely 99% of the time in limited range. 

These settings are generally related to the YUV color model only, not RGB. I’m not really sure it’s a good idea to expose these as range and matrix options if there are no plans to work directly with YUV.

src/metadata/cicp.rs
```diff
@@ -0,0 +1,1346 @@
...
+            full_range: match self.full_range {
+                CicpVideoFullRangeFlag::NarrowRange => false,
+                CicpVideoFullRangeFlag::FullRange => true,
+            },
+        });
+
+        let gray = {
+            // I'd use record update syntax but there's a private field.
+            let mut p = rgb.clone();
+            p.color_space = moxcms::DataColorSpace::YCbr;
```
awxkee:
```
p.color_space = moxcms::DataColorSpace::Gray;
```
Is better fit. It's possible to directly model RGB -> Gray through LUT as you writing here, dealing additionally with YCbCr is not necessary.

src/metadata/cicp.rs
```diff
@@ -0,0 +1,1346 @@
...
+
+pub(crate) trait ColorComponentForCicp: Copy {
+    fn expand_to_f32(self) -> f32;
+
+    fn clamp_from_f32(val: f32) -> Self;
+}
+
+impl ColorComponentForCicp for u8 {
+    fn expand_to_f32(self) -> f32 {
+        self as f32 / Self::MAX as f32
```
awxkee: LLVM rules without fast math don't allow to perform directly reciprocal optimizations https://godbolt.org/z/9dzoMWG53 since in IEEE numbers a / b != a * (1./ b) without error tracking.
If there are any performance plans```const R: f32 = 1./ Self::MAX as f32; (self as f32 * R)``` is preferred.

src/metadata/cicp.rs
```diff
@@ -0,0 +1,1346 @@
...
+        }
+    }
+
+    pub(crate) fn clamp_rgba_luma<P: ColorComponentForCicp>(
+        input: &[f32],
+        output: &mut [P],
+        coef: [f32; 3],
+    ) {
+        for (rgba, pix) in input.chunks_exact(4).zip(output.chunks_exact_mut(2)) {
+            let luma = rgba[0] * coef[0] + rgba[1] * coef[1] + rgba[2] * coef[2];
```
awxkee: You can reuse mla from the blur module, as it’s a good case for using FMA.

src/metadata/cicp.rs
197g: The _compute_ model for these is the YCbCr given by matrix=ChromaticityDerivedNonConstant though, so in terms of color conversions this is correct. But reminder that this is no longer needed, gray conversion does not go through moxcms at this point the whole Y->Rgb conversion/expansion is implemented hhere.

src/metadata/cicp.rs
197g: Huh. Alright, did not know and will remove the opinion from the comment. Although talking about prominance, jpeg's yuv is always full range (ITU REC-T.871, Section 7) and that's mostly what I was thinking of writing that section.

197g: Quick google has me learning that [NVIDIA's MMPU considers yuv to be narrow range](https://forums.developer.nvidia.com/t/hardware-jpeg-encoder-yuv-ycbcr-studio-range-or-full-range/194413) with no good guidance given on how, or if, to configure that differently. Fun fun fun. Well that certainly is a kind of mismatch given they are doing jpeg encoding as an example :clown_face:

awxkee: Yeah, this is a bit controversial, but de facto YUV is almost always in limited range. There are some exceptions with JPEGs, some people encode AVIFs in full range, and high end cameras like Arri Alexa encode full range. Cameras that can write ProRes YUV 444 usually write ProRes YUV 444 full range as well. Otherwise, if not specified it’s definitely limited range. Might be some additional exceptions but anyway it's almost always limited range unless explicitely stated as full range.

src/metadata/cicp.rs
197g: Hm, actually I'm considering undoing that part of the change (should have done it as two commits) since Rust provides [`f32::mul_add`](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.mul_add) as a standard library method, and there's nothing generic in this context.

awxkee: `f32::mul_add` will be replaced with libc call where FMA usually software emulated ( even if it's not in some hypothetical libm indirect calling on each pixel makes performance horrible ) if feature is not activated. That makes performance actually horrible. Therefore, some kind of guard is still needed for cases where the FMA feature is not available.

https://godbolt.org/z/jMxzr33nv

src/images/buffer.rs
```diff
@@ -1406,6 +1443,167 @@ where
     }
 }
 
+/// Inputs to [`ImageBuffer::copy_color`].
+#[non_exhaustive]
+#[derive(Default)]
+pub struct ConvertColorOptions {
+    /// A pre-calculated transform. This is only used when the actual colors of the input and
+    /// output image match the color spaces with which the was constructed.
+    pub transform: Option<CicpTransform>,
```
fintelia: Would it make sense to take an `Option<&mut CicpTransform>` so the transform could be populated lazily and reused between calls?

src/images/buffer.rs
```diff
@@ -1275,6 +1311,7 @@ impl<P: Pixel> ImageBuffer<P, Vec<P::Subpixel>> {
             data: vec![Zero::zero(); size],
             width,
             height,
+            color: Cicp::SRGB_LINEAR.into_rgb(),
```
fintelia: Would `Cicp::SRGB` be a better default to assume?

src/images/buffer.rs
197g: The transform already uses `Arc` for shared data under the hood, if we want to populate any of that lazily that is a `OnceLock` either way. So really it would not be necessary to take anything by mutable reference here; an owned value works pretty much precisely the same for our usage.

src/metadata/cicp.rs
197g: We should at least document our wrapper than to note it is meant for cases where only relative accuracy to some known magnitude is required and the precision of fma is merely a side effect.

src/images/buffer.rs
197g: Probably. With regards to the idea of interpreting the unknown values (`CicpPrimaries::Unspecified = 2`, `CicpTransferFunction::Unspecified = 2`) as sRGB for conversion but not for metadata I'm torn. I think it's fine to assume unwitting _users_ of `new` intend sRGB; and if we do that for compute then arguably we're already saying that the expected perceptual color is sRGB so it's not more wrong to write that to file as is. Users who _do_ know the difference should be able to set the appropriate color space on their buffers before writing them, I'm not very concerned about incorrect usage here.

However for loaded images it's more complicated of course.. The two options should definitely exist as enum variants so they can be set by decoders. The whole decoder path is a separate issue report though. I don't expect to fix any of that in this PR. We should set up the loading path mostly correctly before thinking about the storage path. Also I would argue that the current state, omitting a profile when one was present, is the same error as writing an incorrect one except not only present in files which makes me more easy about potential incorrectness for the unknown case as well.

There is one thing to worry about though, that is copy operations from buffer to buffer. `copy_from_color_space` is explicitly documented as preserving each buffer's attributes and the simple `copy_from` should probably do too (i.e. reinterpret the sample values without conversion) but for other copies it's not entirely clear what semantics _should_ be. I'll make another pass over the buffers for that.

197g: We need to decide how all the `to_*` conversion of `DynamicImage` should behave. For mutable, fallible, edits this is simpler as the caller can control the target (and source if any) and even do-nothing-on-failure would be okay. But many of these are infallibly creating new values.

- Note that `DynamicImage::to_luma` is ill-defined under these circumstances. This is a consequence of the types serving two duties: holding information for io; but also being  buffer to calculate with—the type can not enforce an invariant of being a calculable color space. So there's two problems here:
  - [x] For supported color spaces, these methods currently always use `sRGB` but should arguably use the color space conversion when given an `rgb->luma` conversion. That might regress some performance though (there needs to be a fast path in `CicpTransform` when both color spaces are the same).

    Resolution in [568c976](https://github.com/image-rs/image/pull/2531/commits/568c976e964915dae89e21a72851a94cd6a4284f): use the non-constant luminance color coefficients
  - [x] For impossible color spaces (`Unspecified`) we need to fallback to something, likely `sRGB`. These might occur in the future when decoding such declared images; and users of that method won't want panics in that path. It's less clear what the output's color space field should be. (Anything other than the input would make `DynamicImage::ImageRgba8(_).into_rgba8()` a confusingly non-noop operation). How we handle this should also guide us how to annotate non-RGB images in decoders when they can not be transformed into any valid CICP rgb `DynamicImage`.

    Resolution in [568c976](https://github.com/image-rs/image/pull/2531/commits/568c976e964915dae89e21a72851a94cd6a4284f): fallback to `sRGB` coefficients equivalent to `FromColor`
  - [x] For unsupported color spaces, the output should definitely be the input's color space declaration; but can we accurately do the luma->rgb? We could handle them as non-constant luma, a simple transform that does not need the transfer characteristics, even when they are not 'supported'. This might get hairy if we ever support _constant_ luma in `ImageBuffer` but maybe we just lifted the supported qualification for `CicpTransform` by then..

    Resolution in [b429283](https://github.com/image-rs/image/pull/2531/commits/b42928342c73ecbf91e6d958cb217f1fba1a00ef): calculating Luma coefficients only depends on the CicpColorPrimary and is somewhat simple (~50 lines + 16 in moxcms + known constants) so we support this for all primaries. This allows using non-constant color everywhere even if we do not support the conversion between such color spaces (ie. the transfer characteristics). Exception: Cicp::Xyz would need a special case for converting `xy -> XYZ` so we do not provide it (yet?).
- All the color functions are wrong:
  - [x] `invert` like thumbnail, resizes, blur it does _not_ linearize itself. This is a known defect though and I think we decided on it being fine for now, as the internal complexity is relatively large with slightly smaller gains. (However, it could actually become feasible after this PR. For resizing and blur, we have intermediate copies of pixel values. These copies could perform the conversions to/from lin-RGB as in the dynamic conversion implementation. I'll take a look afterwards.) There is an argument to be made that _current_ users are probably using non-sRGB pixel data when calling these methods as their own workaround—so maybe we should not do anything to these methods and look into more explicit alternatives.

    Resolution in [de62833](https://github.com/image-rs/image/pull/2531/commits/de62833dc96b0795d7abfd8f62f4781eb3cecd59): always use the current rgb values without taking into account the color space; new family of methods should opt-in to the additional performance considerations of conversion and we'll want to add those to `DynamicImage` instead (as per argument of supportable pixels like `::to`). Can generalize later.

    Resolution in [c7d1cb0](https://github.com/image-rs/image/pull/2531/commits/c7d1cb02b0ea4c60895ec69e0ba8b4a95b18b97c): test cases for all the above
  - [x] huerotate is bogus but so is HSV (imo) in the first place, so I don't mind. The documented reference to webkit is obscured by the fact that CSS _does_ define a precise color space in which the arithmetic should take place whereas here it would vary.
  - [x] `flip` et.al should _not_ affect the color space but they do. We construct new `ImageBuffer`'s in `imageops`—but since that code is generic over `I: GenericImageView` it can not access color information. That could be fixed by adding a defaulted method to the trait so that `ImageBuffer` can copy its color space.

    Resulution in [de62833](https://github.com/image-rs/image/pull/2531/commits/de62833dc96b0795d7abfd8f62f4781eb3cecd59): `GenericImageView::buffer_with_dimensions` and `GenericImageView::buffer_like`.
- `apply_` methods are fine as far as I can tell, hooray for in-place.
- `GenericImage::put_pixel` uses `into_luma` which is not color space aware.

197g: We have two odd functions `to_luma_f32` and its alpha variant. Both are _not_ `PixelWithColorSpace` outputs and (in other issues) there was a discussions if we should even add those to `DynamicImage`. For consistency of the implementation it would be nice to have them, but I might get away with using the private bound for now and keeping all that conversion internal.

197g:
Supporting color spaces with edit operations and `into_` ballooned the code size by ~800 lines but that should be easily worth the consistency. The only tragedy is that we can not do it in `DynamicImage::to`. The output pixel type is _too broad_ so it can not be avoided that it calls either `FromColor<Luma<_>>` or `FromColor<Rgba<_>>` incorrectly. There is just no mechanism available to detect the type. Not even `TypeId` since it is not constrained as `'static`.

197g:
Cleaned up documentation, added the bells-and-whistles to get luma/rgb conversion with all primaries (almost never need the fallback) @fintelia: Final review time? Decision log of color space treatment here: https://github.com/image-rs/image/pull/2531#discussion_r2265928720

197g:
Rebased and updated the code to derive luminance coefficients to defer to the `moxcms` implementation in `0.7.4`. Please take a look?

src/images/buffer.rs
fintelia: Thinking about this more, perhaps an `Option<Arc<CicpTransform>>`? That would make it more clear from the type signature what was going on.

src/images/buffer.rs
```diff
@@ -1399,13 +1480,204 @@ where
...
         buffer
     }
 }
 
+/// Inputs to [`ImageBuffer::copy_color`].
+#[non_exhaustive]
+#[derive(Default)]
+pub struct ConvertColorOptions {
+    /// A pre-calculated transform. This is only used when the actual colors of the input and
+    /// output image match the color spaces with which the was constructed.
```
fintelia, line 1496: If the color spaces don't match, that results in an error rather than just not using the precomputed transform, right?

197g: The code treats that as if None was provided, I wanted to handle the value itself like a cache—because really there is no mutability, just lookup precomputation.

A future direction would be having just-in-time computed tables with `OnceLock` as well as transformations of different color spaces stored within the same value with a memory usage policy (`new` certifies infallibility, but we do not need to create or hold on to the tables to retain that), and merging two caches. It's just that the design space of the internals is large yet the basic cache _interface_ is simple.

197g: Might be a bad idea though.. We have `#[non_exhaustive]` so we could as well have a separate attribute for said future directions.

197g: Actually I've reconsidering this point in light of the argument that we can make a second attribute. Arguably a dedicated 'cache' instance can be better designed separately. And the performance costs of missing that cache are, for small images, high enough that there should be an attribute that says: "Yes, I'm sure we've already done the work, now please be fast".

src/images/buffer.rs
197g: That would be double indirection. The sharing happens internally, `moxcms` gives (more-or-less) a `Box<dyn Fn(…)>` and some of these transformers are used be multiple different sample types. Concretely, the `f32` ones are used as fallbacks for all sample types. Hence some of them are already inherently shared and it is easiest to just share all of them, since nothing about the transformers is mutated in any case. If we switched to another implementation it would be very similar.

197g: I've marked the option struct `!Send + !Sync` by a `PhantomData` attribute to ensure that even local options can be added. Except if they have lifetimes, but if we don't tie that to the lifetime of `self` in all the calls its usefulness would always be limited. In any case with `!Send + !Sync` we could add an implementation-detail `Rc<RefCell<_>>` in the worst case. I think we'll be limited by available throughput for quite a while before we can actually explore the implications.

This should be as flexible as we can make it without impacting usability of values.

fintelia: I'd like to find a way to communicate that the object is a cheaply clonable handle to a cache of transform functions. Neither the doc comments nor the struct naming currently convey that point.

If you'd like, there is the option of merging this PR without any public fields in `ConvertColorOptions`, and then figuring out the API for transform caching separately. I think the rest of the PR is ready to merge.

fintelia (approved):

src/images/buffer.rs
197g: It's rather necessary for performance, so I'll currently hide the field as private which makes it useful in our tests. (The full coverage table will take far longer if we were to recompute the luts each time).

Shnatsel:
I'm not sure I understand this PR, so I don't feel qualified to update the changelog in #2555, but I feel this should be included.

Also note that this pulls in `chrono` which is never actually used but pulls in a bunch of transitive dependencies in turn:

      Adding android-tzdata v0.1.1
      Adding android_system_properties v0.1.5
      Adding chrono v0.4.41
      Adding core-foundation-sys v0.8.7
      Adding iana-time-zone v0.1.63
      Adding iana-time-zone-haiku v0.1.2
      Adding js-sys v0.3.77
      Adding windows-core v0.61.2
      Adding windows-implement v0.60.0
      Adding windows-interface v0.59.1
      Adding windows-link v0.1.3
      Adding windows-result v0.3.4
      Adding windows-strings v0.4.2

Many of these are platform-specific, so nobody gets all of this at once, but the battle with dependency sprawl is lost by a thousand cuts. I've opened https://github.com/awxkee/moxcms/issues/97 to track this upstream.

fintelia:
This PR also causes a nearly 3x slowdown in release clean build times when compiling with `--no-default-features`. On my machine it goes from from 3.15s to 8.99s.

It touches a large number of places, but I'm really wondering whether this warrants having a "cms" feature to gives users the choice of whether to pull in all the color conversion code.

197g:
#97 is already merged upstream :tada: With regards to the feature, can you try again with `0.7.5` when published? I believe the slowdown is almost fully related to iana timezones in chrono. It would be surprising to me if the pure _code_ was responsible. We surely want the primaries and luma coefficients even if we were to make the `Transform` part optional; but those also come from moxcms. If there is still significant slowdown then we might split that crate into a much more minimal `cicp` (the crate name is free).

awxkee:
I believe moxcms itself contributes a lot to the compilation time. A huge part of the codegen goes into LUT transforms, which are not used at all here. In fact the code for LUT transforms takes about 70%+ of the crate. If there are no plans to work with ICC directly you end up compiling 70% of code that won’t be used. 

Additionally, it has support for LUT + matrix shaper transforms in `any->any` fashion for exotic color spaces such as CMYKOGV and spurios/non-optimizable ICC profiles which is likely beyond what an image crate can support.

197g:
True, it seems to spend about twice as much time in codegen for `moxcms` as for `image` itself and this is sequential to codegen for `image`. That's unexpected but thank you for the insight. So sure, I'm up for extracting the cicp tables into a separate crate to turn `moxcms` into something optional but it's not pressing enough to warrant immediate attention either? It's still okay-ish with regards to most users compiling in formats for me and the sprawling number of issues that are the result of lacking color space support. Should got into the issue tracker.

@awxkee Here's a data point from `cargo-llvm-lines` for consideration:

```
  746473                9652                (TOTAL)
   82608 (11.1%, 11.1%)   48 (0.5%,  0.5%)  moxcms::conversions::avx::rgb_xyz::TransformShaperRgbAvx<T,_,_,_,_>::transform_impl
   82128 (11.0%, 22.1%)   48 (0.5%,  1.0%)  moxcms::conversions::avx::rgb_xyz_opt::TransformShaperRgbOptAvx<T,_,_,_,_>::transform_impl
   34920 (4.7%, 26.7%)    20 (0.2%,  1.2%)  moxcms::conversions::avx::rgb_xyz_q2_13::TransformShaperRgbQ2_13Avx<T,_,_,_,_,_>::transform_avx2
   34180 (4.6%, 31.3%)    20 (0.2%,  1.4%)  moxcms::conversions::avx::rgb_xyz_q2_13_opt::TransformShaperRgbQ2_13OptAvx<T,_,_,_,_,_>::transform_avx2
   14693 (2.0%, 33.3%)   641 (6.6%,  8.1%)  alloc::boxed::Box<T>::new
   14609 (2.0%, 35.3%)   172 (1.8%,  9.8%)  <alloc::vec::Vec<T> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<T,I>>::from_iter
   14312 (1.9%, 37.2%)   157 (1.6%, 11.5%)  alloc::vec::Vec<T,A>::extend_trusted
   14172 (1.9%, 39.1%)     6 (0.1%, 11.5%)  moxcms::conversions::lut_transforms::make_lut_transform
   10067 (1.3%, 40.4%)   153 (1.6%, 13.1%)  core::iter::traits::iterator::Iterator::fold

```

src/images/dynimage.rs
```diff
@@ -43,6 +45,27 @@ use crate::{
...
+/// Conversion into different pixels types ([`ColorType`][`crate::ColorType`]) _generally_ take the
+/// color space into account, with the exception of [`DynamicImage::to`] due to historical design
+/// baggage.
+///
+/// The imageops functions operate in _encoded_ space, directly on the channel values, and do _not_
+/// linearize colors internally as you might be used to from GPU shader programming. Their return
+/// values however copy the color space annotation of the source.
+///
+/// The IO functions do _not yet_ write ICC or CICP indications into the result formats. We're
+/// aware of this problem, it is tracked in [#2493] and [#1460].
```
Shnatsel, line 62: #2493 lists it as implemented for 3 formats. Should we note that it is implemented for some but not all formats? Or is this a different issue about it the CICP never being written as an ICC profile through these facilities?

197g: The issue here is about writing the CICP from the buffer into the profile of the file, so that is separate from supporting it in the trait interface.

Shnatsel: Ah, I see. Could you open a new issue about that? I don't think the existing ones explain that adequately.

197g: Now tracked in #2562.

PR #2535, opened 2025-07-26
add const `HAS_ALPHA` to trait `Pixel`

var4yn:
### Problem: How to determine has_alpha from a Pixel type?

While working on a bug fix https://github.com/image-rs/imageproc/pull/719 in the `imageproc` crate, I needed to check `Pixel` type has an alpha channel.

The only way I found to do this was by comparing the `COLOR_MODEL: &str` value in the `Pixel` trait, which feels a bit indirect and error-prone.

I think it would be much clearer and easier to use if the `Pixel` trait had a `HAS_ALPHA: bool` constant to indicate this directly.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -286,6 +286,8 @@ impl<T: $($bound+)*> Pixel for $ident<T> {
 
     const COLOR_MODEL: &'static str = $interpretation;
 
+    const HAS_ALPHA: bool = $alphas > 0;
+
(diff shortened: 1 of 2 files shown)


197g (approved):
Sweet, having semverver doing the review for us here. Seems to reasonably complement `_with_alpha` mapping methods and definitely works for everything we define internally.

PR #2536, opened 2025-07-29
Support single channel f32 tiff files 

maxall41:
Adds support for single channel f32 tiff files. See #2251.

### DIFF
--- a/src/codecs/jpeg/encoder.rs
+++ b/src/codecs/jpeg/encoder.rs
@@ -723,7 +723,7 @@ impl<W: Write> ImageEncoder for JpegEncoder<W> {
         use ColorType::*;
         match img.color() {
             L8 | Rgb8 => None,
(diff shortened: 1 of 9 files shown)


fintelia:
Thanks for your interest!

Sorry for not getting into the details in the linked PR, but we intentionally limit the number of variants of `ColorType` and `DynamicImage` because users have to be able to handle getting _any_ possible variant when they decode an image file. On the other hand, the TIFF format allows a nearly unbounded number of pixel formats including {u8, u16, u32, ..., i8, i16, ..., f16, f32, f64, ...} x {1, 2, 3, 4, ...} samples per pixel, so we have to draw a line.

A big factor to guide that decision has been whether the pixels of a given type would generally represent colors. My understanding is that single channel float TIFFs tend to be used for GIS data or other non-color information, and thus better not to include in `ColorType` / `DynamicImage`.

maxall41:
Ah ok Ill just maintain a fork then.

PR #2537, opened 2025-07-30
Add EXIF block support

spoutn1k:
Hey all,

Following up on the great work in the TIFF repo I am making a quick PR to add support for EXIF data in the JPEG codec. This uses the `APP1` marker and I reused the same template as the ICC profile setting. Let me know if you want me to rework some parts !

### DIFF
--- a/src/codecs/jpeg/encoder.rs
+++ b/src/codecs/jpeg/encoder.rs
@@ -33,6 +33,7 @@ static SOS: u8 = 0xDA;
 static DQT: u8 = 0xDB;
 // Application segments start and end
 static APP0: u8 = 0xE0;
+static APP1: u8 = 0xE1;
 static APP2: u8 = 0xE2;
 
 // section K.1
@@ -152,6 +153,10 @@ static UNZIGZAG: [u8; 64] = [
     58, 59, 52, 45, 38, 31, 39, 46,
     53, 60, 61, 54, 47, 55, 62, 63,
 ];
+
+// E x i f \0 \0
+/// The header for an EXIF APP1 segment
+static EXIF_HEADER: [u8; 6] = [0x45, 0x78, 0x69, 0x66, 0x00, 0x00];
 
 /// A representation of a JPEG component
 #[derive(Copy, Clone)]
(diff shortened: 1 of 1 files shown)


src/codecs/jpeg/encoder.rs
```diff
@@ -478,6 +489,17 @@ impl<W: Write> JpegEncoder<W> {
         }
     }
 
+    fn write_exif(&mut self) -> ImageResult<()> {
+        if !self.exif.is_empty() {
+            // E x i f \0 \0
+            let mut formatted = vec![0x45, 0x78, 0x69, 0x66, 0, 0];
+            formatted.extend(self.exif.clone());
```
kornelski: extend_from_slice avoids cloning it one more time.
You could also create an empty `Vec`, and use `try_reserve()` to preallocate the full size needed.

kornelski: #2538

fintelia:
I think this should be added to the `ImageEncoder` trait, analogously to [`set_icc_profile`](https://docs.rs/image/latest/image/trait.ImageEncoder.html#method.set_icc_profile).

src/codecs/jpeg/encoder.rs
spoutn1k: I went the extend route. Do you want me to look at implementing #2538 or should I just make a note ?

spoutn1k:
@fintelia that would be amazing but I am not an image format expert so I do not have the knowledge (or the time) to add it to every codec right now

src/codecs/jpeg/encoder.rs
```diff
@@ -360,6 +366,14 @@ impl<W: Write> JpegEncoder<W> {
         JpegEncoder::new_with_quality(w, 75)
     }
 
+    /// Add an EXIF segment to the JPEG file
+    ///
+    /// Expects a valid TIFF-formatted EXIF block as a byte vector.
+    pub fn with_exif(mut self, exif: Vec<u8>) -> JpegEncoder<W> {
```
197g:
```suggestion
    pub fn with_exif_metadata(mut self, exif: Vec<u8>) -> JpegEncoder<W> {
```

Nit: would align it to `ImageDecoder::exif_metadata`.

src/codecs/jpeg/encoder.rs
```diff
@@ -478,6 +493,16 @@ impl<W: Write> JpegEncoder<W> {
         }
     }
 
+    fn write_exif(&mut self) -> ImageResult<()> {
+        if !self.exif.is_empty() {
+            let mut formatted = EXIF_HEADER.to_vec();
+            formatted.copy_from_slice(&self.exif);
```
197g:
```suggestion
            formatted.extend_from_slice(&self.exif);
```

Can we test this? In the current code, I think it is calling `<&mut [_]>::copy_from_slice` which will panic because the two lengths differ, hardly the intended effect.

197g (changes requested):

src/codecs/jpeg/encoder.rs
kornelski: I've only made a note, since that's a separate issue

src/codecs/jpeg/encoder.rs
spoutn1k: Added

src/codecs/jpeg/encoder.rs
spoutn1k: It was a mistake indeed and is back to `extend_from_slice`

spoutn1k: Also tested and working

197g (approved):

Issue #226, opened 2015-01-03
Update to latest Rust


bvssvni:
I got it working, but want to fix all the warnings...

Issue #2266, opened 2024-06-24
Bug-Triage: Should image codec bug reports be moved to their upstream repos?


ripytide:
I've been looking through the currently open issues and a lot of the bug reports are issues with a particular encoder/decoder not working as expected with a particular image when the decoder/encoder used is provided by another crate with it's own repository.

Should we move these issues to their respective upstream repositories since that is likely where the code and pull requests will need to be made in order to resolve the issues?

If so, two ways I can think of improving this for the future cases is to add a `CONTRIBUTING.md` file that outlines such an please-report-bugs-upstream policy. Secondly we could add a note/checkbox to the bug report issue template.

As for all the already open issues that fit this description, we could either leave them open until if/when their respective issues get resolved upstream or we could manually migrate them to their upstream repositories ourselves.

197g:
Some of these are reasonable. For instance, on the avif we have and had some unsoundness bugs related to downstream dependencies. Seeing them open in bunch could motivate moving from one dependency to another. Also, there often is context already added that may not exist upstream. I've redone the tags to make this distinction clear: `kind: dependency-tracked` can now be used to signal that a bug is (only) pending resolution in a dependency and open for tracking.

fintelia:
There's also the element that it isn't always obvious what's a bug in the upstream repository versus a limitation of the wrapper code that lives in this repository. And sometimes changes are needed to both. So I guess my first reaction is "please don't move the issue unless you know which it is".

197g:
So what should be the outcome of this issue? Do we want to record it into the `.github` bug template?

fintelia:
How about something like: If you know the bug is against an upstream repo then file it there. If you aren't sure or the fix will also require downstream changes, then file it against the crate you were directly using (usually `image`) and a maintainer can move it if necessary.

PR #2541, opened 2025-08-19
Add an issue template for 'Upstream' issues

197g:
This is the summary of: <https://github.com/image-rs/image/issues/2266>

Closes: #226


197g:
I don't know how to validate the form before merging. As far as I know, Github does not provide a schema or preview page? Yet YAML seems to be [the recommended way](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) hence this is is not consistent with the other templates yet.

Issue #2456, opened 2025-04-17
What error variant are encoders supposed to return when they don't support the given color type?


RunDevelopment:
I'm working on integrating the new DDS de/encoder, so I have to handle the case where the encoder doesn't support the `ExtendedColorType` given by the user. I have to return an error, but which one?

My question is: what error variant am I supposed to return?

---

I tried looking at what other encoders are doing, but it's a colorful mix. There are 3 main categories:

1. `avif`, `jpeg`, `tga`, `webp`, `farbfeld`, `gif`, and `tiff` return `ImageError::Unsupported` with `UnsupportedErrorKind::Color`.
2. `hdr`, `openexr`, and `qoi` return `ImageError::Encoding` with a simple string error (message).
3. And the weird ones...
    - `bmp` returns `ImageError::IoError` with `io::ErrorKind::InvalidInput`. (It even creates a new `std::io::Error` for it...)
    - `pnm` returns either `ImageError::Unsupported` with `UnsupportedErrorKind::Color` OR `ImageError::Parameter`.
    - `png` returns either  `ImageError::Unsupported` with `UnsupportedErrorKind::Color` OR `ImageError::Encoding` with its custom `BadPngRepresentation::ColorType` error.

(As for the formats not included above: `ico` uses the PNG encoder under the hood, and `dds` and `pcx` didn't have encoders.)

If I had to guess how this came to be, I'd say that `ImageError::Unsupported` with `UnsupportedErrorKind::Color` was the indented error, but then PNG came along with its `BadPngRepresentation` and that influenced the formats that came after it to use `ImageError::Encoding`. Oh, and BMP is just weird.

If `ImageError::Unsupported` with `UnsupportedErrorKind::Color` really is the intended error, I'm willing to make a PR that (1) documents this fact and (2) changes all encoders to use this error. Just let me know.

PR #2543, opened 2025-08-20
Uniform error for unsupported encoder colors

197g:
Closes: #2456 

Addresses everything on the list, _except_ one case in pnm. This is a collection of sub-formats, it is also the only format that allows the caller to use an _exact_ pre-configured header. In that code path the error is kept as `ParameterError` since it verifies if the caller matched their header to the color, both under the callers control, and not if the color in principle can work with the format. The expected use case here is a roundtrip of extended options including the color type, hence the fault is indicative of a bug on the caller side.

All paths reachable through `ImageEncoder` should result in `ImageError::Unsupported` now (exact for the color type negotiation introduced earlier).


197g:
I'm pretty confident in this with CI alone. There is a note about improving the error message but that gets easier with this fixed, and should be something separate.

PR #2544, opened 2025-08-20
Add alpha mask utilities to ImageBuffer

### LINKED ISSUE #1786: Add possibility to extract an alpha channel as a separate luma image
I would like to be able to extract an alpha channel as a separate luma image.

My specific use case for this functionality is that some formats like PDF need alpha as a separate image called "softmask".

More generally it could be possible to extract any combination of channels like just get RGB from RGBA or maybe R and B channels together or individual channels like just R channels as luma image. It is usable that this is part of this crate as outside code would need to guard every time a new image format is added into DynamicImage enum which is non exhaustive.

197g:
The respective methods on `DynamicImage` are blocked on the type representation. We can't use any specific pixel type if the goal is preserving the type but `DynamicImage` has no Luma32F variant. For applying a mask we would have to take the parameter type as a non-`DynamicImage` somehow? Feels weird, we should just resolve to add at least 32-bit Luma to the crate, even if we do not add 32-bit LumaAlpha.

Update: since #2934 this is all possible.

Closes: #1786

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -778,6 +778,30 @@ where
             .into_view_mut()
             .expect("buffer always uses a non-overlapping strided layout")
     }
+
+    /// Extract the alpha channel as a Luma image.
+    ///
+    /// If the pixel does not have an alpha channel, the value is filled with a fully opaque mask
+    /// using the maximum value of the corresponding subpixel type.
+    pub fn to_alpha_mask(&self) -> ImageBuffer<Luma<P::Subpixel>, Vec<P::Subpixel>> {
+        let pixels = self.subpixels().chunks_exact(P::CHANNEL_COUNT.into());
+        let mut mask = vec![<P::Subpixel as Primitive>::DEFAULT_MAX_VALUE; pixels.len()];
+
+        if P::HAS_ALPHA {
+            assert!(
+                P::CHANNEL_COUNT > 0,
+                "Pixel with zero channels indicated an alpha channel"
+            );
+
+            for (p, alpha) in pixels.zip(mask.iter_mut()) {
+                // If the pixel has an alpha channel, use it.
+                *alpha = *p.last().unwrap();
+            }
+        }
+
+        ImageBuffer::from_vec(self.width, self.height, mask)
+            .expect("used the right pixel and channel count")
+    }
 }
 
 impl<P, Container> ImageBuffer<P, Container>
@@ -941,6 +965,47 @@ where
 
         self.width = selection.width;
         self.height = selection.height;
+    }
+
+    /// Fill the alpha channel of this image from a Luma mask.
+    ///
+    /// Returns an [`ImageError::Parameter`] if the mask dimensions do not match the image
+    /// dimensions. Otherwise, if the pixel type does not have an alpha channel this is a no-op.
+    pub fn apply_alpha_mask<RhsContainer>(
+        &mut self,
+        mask: &ImageBuffer<Luma<P::Subpixel>, RhsContainer>,
+    ) -> ImageResult<()>
+    where
+        RhsContainer: Deref<Target = [P::Subpixel]>,
+    {
+        if (self.width, self.height) != (mask.width(), mask.height()) {
+            return Err(ImageError::Parameter(ParameterError::from_kind(
+                ParameterErrorKind::DimensionMismatch,
+            )));
+        }
+
+        if !P::HAS_ALPHA {
+            return Err(ImageError::Parameter(ParameterError::from_kind(
+                ParameterErrorKind::NoAlphaChannel,
+            )));
+        }
+
+        assert!(
+            P::CHANNEL_COUNT > 0,
+            "Pixel with zero channels indicated an alpha channel"
+        );
+
+        let pixels = self
+            .subpixels_mut()
+            .chunks_exact_mut(P::CHANNEL_COUNT.into());
+
+        let mask = mask.subpixels();
+        for (p, alpha) in pixels.zip(mask.iter()) {
+            // If the pixel has an alpha channel, use it.
+            *p.last_mut().unwrap() = *alpha;
+        }
+
+        Ok(())
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/images/buffer.rs
```diff
@@ -981,6 +1000,42 @@ where
     pub fn put_pixel(&mut self, x: u32, y: u32, pixel: P) {
         *self.get_pixel_mut(x, y) = pixel;
     }
+
+    /// Fill the alpha channel of this image from a Luma mask.
+    ///
+    /// Returns an [`ImageError::Parameter`] if the mask dimensions do not match the image
+    /// dimensions. Otherwise, if the pixel type does not have an alpha channel this is a no-op.
+    pub fn apply_alpha_mask<RhsContainer>(
```
fintelia: Maybe `set_alpha_channel`? To me, applying an alpha mask sounds like doing some sort of blending

RunDevelopment:
Since it's now possible to implement this functionality for `DynamicImage`, we should probably revisit this.

197g:
Sorry for the little bit of drive-by change to the import of `ColorType`, we're usually importing the public path now and this was apparently left over. That was noticable when writing the test, also the naming inconsistency of `fn color` instead of `color_type` as we have established in the `PixelWithColor` trait and elsewhere. To not cause churn, just added a doc alias here for now.

Applying an alpha mask to a `DynamicImage` needs another decision: how should we handle non-luma image parameters? While we pretty clearly want to convert mismatching sample types it's not obvious if a conversion into gray-scale via the color conversion is expected. (That would also be open to optimization; requiring an intermediate buffer of either the whole image or a customized version of the application loop so we can convert pixels to Luma in fixed chunks).

RunDevelopment:
> While we pretty clearly want to convert mismatching sample types

Do we? Returning an error seems like the most reasonable choice to me. It's not like assigning an RGB image to an alpha channel makes any sense to begin with. Coversion also opens ambiguity. For example, one might reasonably expect that `img.assign_alpha(rgba_image)` would assign the alpha channel of the image parameter (ignoring its RGB channels).

Plus, we can still allow (some) conversion semantics in the future, no?

> little bit of drive-by change

Love the doc alias. I actually ran into that recently.

src/images/buffer.rs
```diff
@@ -942,6 +966,47 @@ where
         self.width = selection.width;
         self.height = selection.height;
     }
+
+    /// Fill the alpha channel of this image from a Luma mask.
+    ///
+    /// Returns an [`ImageError::Parameter`] if the mask dimensions do not match the image
+    /// dimensions. Otherwise, if the pixel type does not have an alpha channel this is a no-op.
```
RunDevelopment: > Otherwise, if the pixel type does not have an alpha channel this is a no-op.

This is not what the function does.

src/images/buffer.rs
```diff
@@ -942,6 +966,47 @@ where
...
+    pub fn apply_alpha_channel<RhsContainer>(
```
RunDevelopment: May I suggest the name `replace_alpha_channel` (or just `replace_alpha`)? Apply still sounds like blending.

src/images/buffer.rs
```diff
@@ -778,6 +778,30 @@ where
...
+            assert!(
+                P::CHANNEL_COUNT > 0,
+                "Pixel with zero channels indicated an alpha channel"
+            );
+
+            for (p, alpha) in pixels.zip(mask.iter_mut()) {
+                // If the pixel has an alpha channel, use it.
+                *alpha = *p.last().unwrap();
+            }
+        }
```
RunDevelopment: I think this could be simpler.

```suggestion
        let mask: Vec<P::Subpixel> = self.pixels().iter().map(|p| p.alpha()).collect();
```

src/images/buffer.rs
```diff
@@ -942,6 +966,47 @@ where
...
+
+        let pixels = self
+            .subpixels_mut()
+            .chunks_exact_mut(P::CHANNEL_COUNT.into());
+
+        let mask = mask.subpixels();
+        for (p, alpha) in pixels.zip(mask.iter()) {
+            // If the pixel has an alpha channel, use it.
+            *p.last_mut().unwrap() = *alpha;
+        }
```
RunDevelopment, line 1019: This is subtly incorrect. You are assuming that the alpha channel is the last channel. This is correct for all our pixel types, but not all `Pixel` types. Could you please document this assumption?

It would be possible to get rid of this assumption (`Pixel::alpha` can read the alpha channel, so we can determine the alpha channel index by trails and error. Set all channels to zero and then one channel to max and see if `Pixel::alpha` is max) but not practical. Alternatively, we could verify that the last channel is indeed alpha and return an error if it's not.

src/images/buffer.rs
197g: Oh, that must have gotten lost in the rebase somewhere. We agreed on `set_alpha_mask` previously, yes.

197g:
> Do we? Returning an error seems like the most reasonable choice to me. 

I meant that we should accept `Luma8` to set the alpha channel of an `Rgba32F` image and `Luma32F` for an `Rgba16` etc. That is, channel conversion but no color conversion.

RunDevelopment:
> I meant that we should accept `Luma8` to set the alpha channel of an `Rgba32F` image and `Luma32F` for an `Rgba16` etc. That is, channel conversion but no color conversion.

Ah, I see. I typically use the term "precision" for that, so I misunderstood. I agree that precision conversion is expected. And I still think that non-luma mask parameters should result in an error.

src/images/dynimage.rs
```diff
@@ -964,6 +966,92 @@ impl DynamicImage {
...
+            ImageRgba32F(image_buffer) => ImageLuma32F(image_buffer.to_alpha_mask()),
+        }
+    }
+
+    /// Fill the alpha channel of this image from a Luma mask.
+    ///
+    /// Returns an [`ImageError::Parameter`] if the mask dimensions do not match the image
+    /// dimensions or if there is no alpha channel in the pixel's color type or if the input image
+    /// is not a `Luma` image.
+    pub fn set_alpha_channel<RhsContainer>(&mut self, mask: &DynamicImage) -> ImageResult<()> {
```
RunDevelopment: We probably don't need that.
```suggestion
    pub fn set_alpha_channel(&mut self, mask: &DynamicImage) -> ImageResult<()> {
```

src/images/dynimage.rs
```diff
@@ -964,6 +966,92 @@ impl DynamicImage {
...
+            ImageLumaA32F(image_buffer) => ImageLuma32F(image_buffer.to_alpha_mask()),
+            ImageRgb32F(image_buffer) => ImageLuma32F(image_buffer.to_alpha_mask()),
+            ImageRgba32F(image_buffer) => ImageLuma32F(image_buffer.to_alpha_mask()),
+        }
+    }
+
+    /// Fill the alpha channel of this image from a Luma mask.
+    ///
+    /// Returns an [`ImageError::Parameter`] if the mask dimensions do not match the image
+    /// dimensions or if there is no alpha channel in the pixel's color type or if the input image
```
RunDevelopment: 1. "Pixel's color type" is less obvious here IMO, because `DynamicImage` doesn't directly have a `Pixel` type.
2. "input image" seems not super clear to me. I think "mask image" is clearer here.

```suggestion
    /// dimensions or if this image's color type has no alpha channel or if the mask image
```

197g: An alternative implementation could be:

> If this image has no alpha channel it is augmented with the mask's alpha channel, changing this image's color type.

This would be consistent with interpreting a missing alpha as opaque in `to_alpha_mask`.

RunDevelopment: Basically, it would have set or add semantics. That would be useful. I like the idea.

PR #2548, opened 2025-08-24
BMP: use ChunksExactMut instead of ChunksMut

Shnatsel:
You'd think it would help codegen, ~~but on my machine it makes no difference on benchmarks whatsoever.~~  
I was measuring it wrong, it brings down the decoding time of an indexed RLE-encoded BMP from 4 ms to 3.6ms, so a 10% perf improvement.

Reduces binary size by 1.5KB (stripped), which is completely inconsequential.

I just like knowing that each row is genuinely row-sized and pixels are genuinely pixel-sized when thinking through the rest of the code.

All of the affected functions and types are private so this is not a semver hazard.


Shnatsel:
I was measuring the wrong thing. This does provide a 10% end-to-end perf improvement, I've updated description with the correct figures.

PR #2549, opened 2025-08-24
BMP: correctly decode invalid BMP RLE images produced by imagemagick

### LINKED ISSUE #2321: “Corrupt RLE data” on simple BMP
Load the following image using `image::load_from_memory` :  [image.zip](https://github.com/user-attachments/files/16946090/image.zip)

The image was generated from this PNG 
simple
using ImageMagick: `magick simple.png simple.bmp`. It loads successfully in a different image viewer I tried, so I have reason to believe it’s a valid image. Example code:

```rs
fn main() -> anyhow::Result<()> {
    let bytes = fs::read("simple.bmp")?;
    image::load_from_memory(&bytes)?;
    Ok(())
}
use std::fs;
```

## Expected

Loading succeeds.

## Actual behaviour

```
Error: Format error decoding Bmp: Corrupt RLE data

Caused by:
    Corrupt RLE data
```

(as a side note, it looks like there’s a bug in the error printing that causes (a) the error message to start with an uppercase letter, which [goes against convention](https://doc.rust-lang.org/stable/core/error/trait.Error.html#:~:text=Error%20messages%20are%20typically%20concise%20lowercase%20sentences), and (b) the error message to be duplicated)

Shnatsel:
Fixes #2321 and makes the sample image provided in the issue decode correctly.

Thanks to @mstoeckl for investigating the cause of the error, it helped a great deal!

This PR only touches the 8-bit codepath. I managed to reproduce the imagemagick bug in 8-bit mode but not in 4-bit mode, so this PR should be enough to fix the issues that crop up in practice.

I can also change the 4-bit codepath for consistency if you prefer.

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -1262,19 +1262,18 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
                         x += length as usize;
                     }
                     RLEInsn::PixelRun(n_pixels, palette_index) => {
-                        // A pixel run isn't allowed to span rows, but we
-                        // simply continue on to the next row if we run
-                        // out of pixels to set.
+                        // A pixel run isn't allowed to span rows.
+                        // imagemagic produces invalid images where n_pixels exceeds row length,
+                        // so we clamp n_pixels to the row length to display them properly:
+                        // https://github.com/image-rs/image/issues/2321
                         match image_type {
                             ImageType::RLE8 => {
-                                if !set_8bit_pixel_run(
-                                    &mut pixel_iter,
-                                    p,
-                                    repeat(&palette_index),
-                                    n_pixels as usize,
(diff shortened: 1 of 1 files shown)


fintelia (approved):

fintelia:
Could you also add a test image so we can make sure this doesn't regress?

I think we should keep the 4-bit codepath as-is. I'm hesitant to make changes without spec language or a test image to demonstrate that it is correct. We've introduced decoding bugs in BMP before by trying to fix a failing image but causing other problems in the process.

Shnatsel:
The 4-bit fix would be to just keep going instead of failing. This PR contains a reverted commit to show that would look like. But sure, I've left it untouched. 

I've added a test image too and verified that simply adding it causes tests to fail on main. This should be good to go. Thanks for the quick review!

PR #2552, opened 2025-08-24
Fix BMP decoder seeking to account for post-header bitfield masks

### LINKED ISSUE #2547: BMP decoder seeking doesn't account for post-header bitfield masks
When decoding a Bitmap V5 from the Windows clipboard (`CF_DIBV5`), `BmpDecoder::new_without_file_header` doesn't properly account for the post-header bitfield masks for `BI_BITFIELD` bitmaps. As a result the decoded image data (rewritten to a `PNG` for convenience) has all of its data shifted over to the right slightly.

The [Win32 documentation](https://learn.microsoft.com/en-us/windows/win32/gdi/bitmap-header-types) for Bitmap headers states:
> The red, green, and blue bitfield masks for BI_BITFIELD bitmaps immediately follow the [BITMAPINFOHEADER](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapinfoheader), [BITMAPV4HEADER](https://learn.microsoft.com/en-us/windows/desktop/api/Wingdi/ns-wingdi-bitmapv4header), and [BITMAPV5HEADER](https://learn.microsoft.com/en-us/windows/desktop/api/Wingdi/ns-wingdi-bitmapv5header) structures. The BITMAPV4HEADER and BITMAPV5HEADER structures contain additional members for red, green, and blue masks as follows.

While `image` does parse these values via `read_bitmasks`, it then seeks backwards by 12 bytes before reading the actual image data, essentially undoing the forward seeking already done and starting too early:
https://github.com/image-rs/image/blob/0d028a63a31c6730f23a6f249c689afcda51f162/src/codecs/bmp/decoder.rs#L824-L829

If I make a small change to `image` to change the seek to use this position instead, the image data is read correctly:
```rust
let bitmask_size = match &self.image_type {
    ImageType::Bitfields16 | ImageType::Bitfields32 => {
       self.read_bitmasks()?;
       12
    },
    _ => 0
};

self.reader.seek(SeekFrom::Start(bmp_header_end + bitmask_size))?;
```

## Expected

An image read from Windows 10's snipping tool should look like this:

Image

## Actual behaviour

The data read is offset slightly, resulting in pixels from the right side of the image "wrapping around" to the left, like this:

Image

## Reproduction steps

If you have access to a Windows 10/11 system, this can be reproduced by doing the following:
1. Run `git clone https://github.com/1Password/arboard`
2. Apply this diff to the `get_image.rs` example:
```diff
diff --git a/examples/get_image.rs b/examples/get_image.rs
index 2871906..b595f40 100644
--- a/examples/get_image.rs
+++ b/examples/get_image.rs
@@ -5,5 +5,7 @@ fn main() {
 
 	let img = ctx.get_image().unwrap();
 
-	println!("Image data is:\n{:?}", img.bytes);
+	println!("Image data is:\n{:#?}", img.height);
+
+	image::save_buffer_with_format("./out.png", img.bytes.as_ref(), img.width as u32, img.height as u32, image::ColorType::Rgba8, image::ImageFormat::Png).unwrap();
 }
```
3. Using snipping tool, take a screenshot of something with multiple colors. Use `CTRL+C` to copy it.
4. Run `cargo run --example get_image` to read the image from the clipboard and dump it to `out.png`
5. Open `out.png` in an image viewer and observe that the pixels are shifted.

complexspaces:
This PR comprehensively fixes #2547, including a regression test and exact compliance with the Microsoft documentation stating which bitmap values contain the post-header bitmask data.

The regression test data is a bit opaque because its missing a file header (so it can't just be opened in a viewer locally) so I left some in-test explanation of what it contains and what parts should look like what. For visual reference during review, this is the correct image:
image

This is the incorrect image that the regression test catches by noticing the yellow pixels in the wrong place:
image

Follows up/improves upon #2550. cc @Shnatsel


Shnatsel (approved):

complexspaces:
Thank you for the fast approval and merge!

PR #2557, opened 2025-08-26
Add missing method on ImageDecoder impl for Box

fintelia:
Without this PR, calling `ImageDecoder::orientation` on a `Box<dyn ImageDecoder>` would resolve to the default implementation rather than the one on the inner decoder.


197g (approved):
Would that be caught by: https://rust-lang.github.io/rust-clippy/stable/index.html#missing_trait_methods

fintelia:
It would have. Added!

Issue #2494, opened 2025-06-18
Exif metadata read/write support


Shnatsel:
Exif metadata can contain information about the orientation in which a photo was taken, which is crucial to be able to display images correctly. Being able to write Exif metadata is important to keep the metadata after processing an image.

Among other things, this is needed by `wondermagick`.

#### Read support

- [x] JPEG
- [ ] PNG - implemented in `png` crate but not wired up to `image`
- [x] WebP
- [ ] TIFF
- [ ] AVIF - is Exif common in AVIF?

#### Write support

- [ ] JPEG
- [ ] PNG
- [ ] WebP
- [ ] TIFF
- [ ] AVIF blocked on https://github.com/kornelski/cavif-rs/issues/76

Arvamer:
I think AVIF writer may not be 100% spec compliant. As `image` lacks read support, I tried to use `kamadak-exif` and `little_exif` and neither could extract exif data from AVIF produced by `image`; both interpret TIFF header as offset. `exiftool` works, but warns about "Missing Exif header". Exif data came from jpeg, if this matter.

Shnatsel:
@Arvamer thanks for letting us know! Could you file a separate issue about that and include the steps to reproduce it?

Shnatsel:
The AVIF Exif issue is fixed in avif-serialize v0.8.9 which incorporated https://github.com/kornelski/avif-serialize/pull/18

PR #2555, opened 2025-08-24
Populate changelog for the upcoming 0.25.7 release

Shnatsel:
Doesn't list the improvements made to subcrates such as image-webp.

PR #2566, opened 2025-08-28
Extract exif metadata from pngs

1c3t3a:
The png crate supports extracting exif metadata as of recently, so this is wiring this up to the `ImageDecoder`  type.

This is related to #2494.

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -178,6 +178,15 @@ impl<R: BufRead + Seek> ImageDecoder for PngDecoder<R> {
         Ok(self.reader.info().icc_profile.as_ref().map(|x| x.to_vec()))
     }
 
+    fn exif_metadata(&mut self) -> ImageResult<Option<Vec<u8>>> {
+        Ok(self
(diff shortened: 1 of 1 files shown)


Shnatsel:
Note that while the field already exists in released PNG versions, actually parsing the Exif chunk was only added in 0.18.x which still isn't released: https://github.com/image-rs/image-png/pull/568

So this code will not actually do anything until `png` 0.18 ships and we bump the minimum required version of it in `image` to 0.18

1c3t3a:
Ohh oke I see! I don't know much about the release cadence here, but it looks like png 0.18 [is close](https://github.com/image-rs/image-png/pull/631)? And this crate is [also preparing for 0.25.7](https://github.com/image-rs/image/pull/2555). Would it be possible to align the two and have the EXIF metadata in 0.25.7?

197g:
Sure, should be doable. We just merge this before the bump, it's not really wrong :smile:

1c3t3a:
Thank you so much!

PR #2567, opened 2025-08-28
Provide a way to extract XMP metadata (png & webp & tiff only for now)

1c3t3a:
XMP is a common metadata format and multiple image formats supported by this crate are able to extract XMP metadata. Similiar to the icc profile and exif metadata, we extend the ImageDecoder trait to provide this functionality.

For now this is only implemented for png and webp and tiff.

This is related to #2568.


src/codecs/png.rs
```diff
@@ -45,7 +46,7 @@ impl<R: BufRead + Seek> PngDecoder<R> {
 
         let max_bytes = usize::try_from(limits.max_alloc.unwrap_or(u64::MAX)).unwrap_or(usize::MAX);
         let mut decoder = png::Decoder::new_with_limits(r, png::Limits { bytes: max_bytes });
-        decoder.set_ignore_text_chunk(true);
+        decoder.set_ignore_text_chunk(false);
```
1c3t3a, line 49: #2027 made it unavailable to read the text chunk unfortunately. I am wondering if you have any guidance on how to move forward with this given that enabling that flag makes it impossible for us to ever read XMP metadata (which is inside the text chunk).

I see multiple solutions:
1. Change the behavior: likely unwanted as it will break existing users.
2. Hack it by spinning up a separate decoder just for reading XMP metadata.
3. Change the interface in a way that the user can configure whether they want to ignore invalid text chunks. If they set invalid text chunks, they also don't get metadata (this is what PNG is doing).

197g: > 1. Change the behavior: likely unwanted as it will break existing users.

Why would that break users? We're not passing the text chunks forward, they are disabled due to returning a strong validation error in `png` during decoding which stops the decoding. We'd need to enable the chunks in such a way that invalid chunks are ignored, instead. (Which I believe is recently already implemented in png: https://github.com/image-rs/image-png/commit/5fef8f209acb8145b04860bde42ec283166e6041#diff-ef73ce77825d3cecdd3c7c8f44f94c4be566bd325c5d0294603a2d65cd211f9fR1050-R1062)

1c3t3a: > Why would that break users?

What I meant is just that it breaks users as-is (see the failing test suite). But thanks a lot for the pointer to the png version! I integrated the current main of the png crate locally into image for me and this worked indeed 😃  So I'll guess I can just wait until the 0.18 of png is released and integrated into image!

1c3t3a:
Yehaa, I rebased after https://github.com/image-rs/image/pull/2574, looks like this is ready to go from my side? Any further points to do @197g?

src/codecs/png.rs
1c3t3a: This seems to be fixed with what you mentioned!

1c3t3a:
Sorry for the heck-meck, I decided to add Tiff now as well, since the roll pulled in the fix for processing the Bytes in Tiff :)

197g (approved):
Actually having more decoders support it, that set in particular is really core, makes the addition much more convincing as common behavior.

Shnatsel:
`zune-jpeg` crate also supports XMP, but only in the 0.5.x series which is yet to see a stable release, while `image` is still on 0.4.x.

It would be great to coordinate with the zune-jpeg author to clear the remaining issues and cut a release.

fintelia:
This PR unfortunately makes it harder to fix the memory limit handling for the PNG decoder because it requires `PngDecoder::new` to read all the text chunks into memory before we know what the memory limit should be.  Not impossible to fix, but potentially quite tricky

Shnatsel:
Isn't that what we added the `Seek` bound for?

1c3t3a:
I am also happy to move the metadata extraction to the png crate if it fits better there. That way we could just peek the keyword and otherwise seek over the contents?

197g:
> Isn't that what we added the Seek bound for?

Yes, since we've moved to ignore ancillary chunks that are broken we might also partially ignore them when they exceed memory limits. And by partially I mean keep a list of offsets where they occur so that their contents can be selectively retrieved. That should combine well with interfaces to read such chunks without fully allocating them in memory (also planned for chunks we want to ignore). We could add a flag to do so preemptively while keeping only a prefix in memory, enough to determine if they are relevant for XMP or Adobe embeds. I think there are a lot of potential variants that won't break functionality (with minor coordination to use those features in `image` as they are released maybe).

fintelia:
Thought about this a bit more, and I think there's a way we can implement this without needing invasive changes to the png crate chunk parsing state machine. The main point would be storing the positions of text chunks during initial parsing, and only seeking to each one and reading its label if the `xmp_metadata` method is called. libpng defaults to only storing 1000 text chunks, so we could similarly bound the worst-case behavior.

PR #2571, opened 2025-08-30
Improve the Issue templates for format requests

197g:
As discussed, linking to `image-extras` and a discussion page added. The checkboxes are required to be ticked actively to allow opening an issue, I've also added a field to refer to an existing crate that should implement the format.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,7 +1,7 @@
 ---
 name: Feature request
 about: Enabling something that is not yet possible
-labels: enhancement
+labels: new-feature
 title: ''
 assignees: ''
 
--- a/.github/ISSUE_TEMPLATE/format_request.yml
+++ b/.github/ISSUE_TEMPLATE/format_request.yml
@@ -1,0 +1,44 @@
+name: "File format request"
+description: "An encoded file format that you need"
+title: "Supporting format …"
+labels: ["topic: formats"]
+body:
+- type: markdown
+  attributes:
+    value: |
+      Please understand that maintainer time is limited and many image formats
+      are complex, even the seemingly simple ones. We do not generally start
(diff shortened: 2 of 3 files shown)


fintelia:
I think we might need to separate discussion of adding formats to `image` and to `image-extras`. As written, the template sort of sounds like "if you make a codec crate we'll add it".

My goal for `image-extras` was to accept pretty much any format that meets the inclusion criteria. Though for supply chain security reasons, I think it _would_ be preferable to have decoder implementations directly in that repository rather than as external crates. Support expectation would be along the lines of "we vetted that this code isn't actively malicious, but maintainers generally don't add features or fix bugs".

There should be a far higher bar for `image` itself. While I don't want to fully rule it out, adding new formats here would be very much the exception rather than the rule. Once the release is done, I actually think we should discuss moving a few of the formats _out_ of `image` in 1.0.

For codecs used by `image`, the ideal position is as separate crates under the _image-rs_ org. The process of adding `zune-jpeg` ended up being messy because they suddenly got a huge influx of users they weren't prepared for, and we weren't positioned to help out dealing with the discovered bugs. There's always the last resort of forking a dependency (and we shouldn't take any dependency we'd be incapable of maintaining/replacing ourselves) but it is something to avoid if at all possible.

197g:
Thanks for the input, I will clarify the split accordingly (probably mostly on the discussion page to avoid duplication). Listing some specific contention points is probably the best way to get people to realize it's not as simple as 'cargo add'. Just checking, can you edit my top-level discussion comment? Explaining each criteria of maintainership, vetting, bandwidth can be done incrementally and with some flexibility I guess.

Also, just to be sure we're on the same page, is the discussion format alright with you? Should I begin to close the linked issues pointing to the discussion page?

fintelia:
Yeah, I think the discussion format is worth trying. I confirm that I'm able to edit the top-level discussion post.

fintelia (approved):

Issue #1182, opened 2020-03-29
Plan for improving the documentation structure


HeroicKatora:
A recent PR (#1180) raised awareness that the first page for many new library users is its `docs.rs` page, generated from the crate root documentation and not the extensive Readme. While moving all of the documentation, as proposed, felt like too broad, nevertheless the following discussion was quite constructive in how the overall structure could be improved. I'll quote relevant portions on what we agreed on:

## On the general structure

> So in short - I didn't see any need to change the documentation, it just felt totally wasted hidden away on github where it's hard to find :)

> Another advantage of the docs being on docs.rs [and within the generated documentation] rather than on github is, of course, the ability to do the intra-crate links. I know this is obvious, but I wanted to explicitly state it's really useful to see a brief code example using (for instance) GenericImage and then be able to click a link to drill down into the documentation and find out about all the detailed methods. I suspect this is the workflow most newbies would initially follow when working out how to use this crate.

## Code examples and general usage:

> Specifically, the code examples are the right level of complexity and I think are really helpful for newbies to the crate. (I'd definitely want to keep at least one simple and one complex example in the top-level documentation, e.g. the open and save example and the fractal example).

## On section 4, the `Images`

> The traits and structures are probably better served by linking their documentation, it's technical information and doesn't add much for judging general applicability.

> I think the explanation of the roles of GenericImage/ImageBuffer/DynamicImage is good and belongs in the top level documentation too. Yes, it could be shorter, but I don't see this as urgent.

>  [It's] just that the code definitions is basically a duplicate of what rustdoc generates for the actual items and linking would be more apt and informative as it also allows browsing all of the related items. A short overview of their role is definitely relevant to the main documentation.

## The `Pixel` trait

> The Pixel section is plainly outdated and as well better represented if the impl section of the trait were simply linked.

> * Drop the Pixel stuff

## On the table of supported formats:

> I think the table of formats is actually beneficial and probably belongs in the top-level documentation. I think it shows the number one feature of this library which is to support multiple image formats. (I'm not sure if you'd agree that's the number one feature but I assume it's why a lot of people want to use image).

> Although it really should be the module documentation that expands on this in-depth, for example an entry all obviously depends on the format. (Not that this is directly relevant for the PR but the first column were to actually link to the modules..?)

## On the fractal example

> I too really like the fractal :)

:heavy_plus_sign: 

## Actionable items

[...]

PR #2576, opened 2025-08-31
Restructure the Readme

197g:
Closes: #1182

This tries to make everything a little less redundant. To no-ones surprise, `image` is an image library but we should not focus on processing, hence that title is removed. The pixels fit better with the buffer types and the high-level section really only gives IO operations (which is fine, the other operations are already explained in other sections).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,33 +8,6 @@ Maintainers: [@197g](https://github.com/197g), [@fintelia](https://github.com/fi
 
 [How to contribute](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md)
 
-## An Image Processing Library
-
-This crate provides basic image processing functions and methods for converting to and from various image formats.
-
-All image processing functions provided operate on types that implement the `GenericImageView` and `GenericImage` traits and return an `ImageBuffer`.
-
-## High level API
-
-Load images using [`ImageReader`]:
-
-```rust,ignore
-use std::io::Cursor;
-use image::ImageReader;
-
-let img = ImageReader::open("myimage.png")?.decode()?;
-let img2 = ImageReader::new(Cursor::new(bytes)).with_guessed_format()?.decode()?;
-```
-
-And save them using [`save`] or [`write_to`] methods:
-
-```rust,ignore
-img.save("empty.jpg")?;
-
-let mut bytes: Vec<u8> = Vec::new();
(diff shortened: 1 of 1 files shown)


fintelia:
I think we should consider what information lands "above the fold" at the start of the readme. Right now that's the couple pointless sentences that "`image` is an image library", but they are followed by a short code sample showing what the library can do.

The next thing (which this PR places first) is the long table listing out all the image formats and with only a few exceptions saying that "yes we can encode and decode this format, and the feature flag is what you'd expect". Maybe we could just add `default-features` to the list of features, have the description list out all the formats, and then link to https://docs.rs/image/latest/image/codecs/index.html for the rest of the details?

197g:
I like that structure you sketch, I mainly moved the code examples because they were both _just_ showing the IO part. So do they fit? It was not easy to think of a replacement that went beyond that.

I've also questioned if the `Readme` is really the right place for it. The generated documentation, with a dedicated how-to section, could be far more extensive and structured at the same time. [For instance Pillow is very minimal](https://github.com/python-pillow/Pillow), despite [having vast docs beyond the API](https://pillow.readthedocs.io/en/stable/?badge=latest).

The question really is: who is the Readme aimed at? We serve it both as the landing page in `crates.io` and as the Github one, and maybe that's a little too much. At least myself, when I go find out what a crate can _do_, I almost instantly go over to `docs.rs` instead of `crates.io`. No readme can provide a depth of understanding of capabalities as the linked, searchable `cargo doc` output. Also specifically regarding the code examples they are stuck in a bind that we can't make use of docs's full formatting (i.e. hiding lines) and they are also not executable from that place but we're currently checking them for being executable. Involving `Cursor` was a solution to make them runnable without literring when running `cargo test`, but it's not really a mininmal beginner example. I'm not sure if adding `no_run` etc. interferes with Github's rendering.

The feature / format table is a very good fit for the Readme in that feature flags are harder to parse from the documentation.

197g:
Updated to reflect the already completed changes of consolidating the table and undoing the removal of the initial example (until I've come with a better criteria for deciding on the functionality to put there instead). But, I've taken out the `ImageDecoder` and `ImageDecoderRect` trait references. I don't think these are relevant for most immediate users and especially the latter is in talks of being hard deprecated. They don't really complete the API listing either as some other traits such as animations are missing. We could add them back once that is sufficiently rounded. The buffer list also omits `FlatSamples` and `View` for instance.

PR #2579, opened 2025-09-01
Move format support table to only be in docs

fintelia:
This shortens the crate README by removing a big table that was mostly just filled with the word "Yes". Also rewords the section about supported formats.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,46 +37,25 @@ img2.write_to(&mut Cursor::new(&mut bytes), image::ImageFormat::Png)?;
 
 ## Supported Image Formats
 
-With default features enabled, `image` provides implementations of many common
-image format encoders and decoders.
+With default features enabled, `image` provides implementations of [many common
+image format encoders and decoders.](https://docs.rs/image/latest/image/codecs/index.html#supported-formats)
 
-<!--- NOTE: Make sure to keep this table in sync with the one in src/lib.rs -->
+Decoding support for additional image formats is provided by
+[`image-extras`](https://github.com/image-rs/image-extras), and the same plugin
+interface lets third-party crates act as format implementations for `image`. If
+the format you need to handle some other image format, check crates.io for
+crates that implement it.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,46 +37,25 @@ img2.write_to(&mut Cursor::new(&mut bytes), image::ImageFormat::Png)?;
...
-Other feature flags for the `image` crate:
-
-| Feature       | Description
-| ------------- | -----------
-| `rayon`       | Enables multi-threading with rayon context in some dependencies
-| `nasm`        | Enables the build-time use of `nasm` for `ravif`, requires `nasm` installed
-| `color_quant` | Includes `color_quant` as an implementation of `imageops::ColorMap`
-| `avif-native` | Enables non-Rust dependencies of `avif` (`mp4parse` and `dav1d`)
-| `serde`       | Enables `serde` integration for various structs and options
-| `benchmarks`  | This feature is used (internally) for benchmark dependencies
```
fintelia, line 79: The internal `benchmarks` feature we don't want users to enable seems like precisely the thing that shouldn't be in the readme :smile:

197g (approved):

README.md
```diff
@@ -37,46 +37,25 @@ img2.write_to(&mut Cursor::new(&mut bytes), image::ImageFormat::Png)?;
...
+With default features enabled, `image` provides implementations of [many common
+image format encoders and decoders.](https://docs.rs/image/latest/image/codecs/index.html#supported-formats)
+
+Decoding support for additional image formats is provided by
+[`image-extras`](https://github.com/image-rs/image-extras), and the same plugin
+interface lets third-party crates act as format implementations for `image`. If
+the format you need to handle some other image format, check crates.io for
```
Shnatsel, line 46: > If the format you need to handle some other image format,

probably leftover from editing

fintelia: https://github.com/image-rs/image/pull/2580

PR #2580, opened 2025-09-01
Fix readme typo

fintelia:


197g (approved):
Wow, I glossed right over that as if it was a repeated the the.

PR #2582, opened 2025-09-01
Allow setting DEFLATE compression level in the PNG encoder

Shnatsel:
Not the most elegant API, but semver-compatible. 

My ideal API would be two functions: `new_with_compression(CompressionType)` and `new_with_compression_advanced(DeflateCompression, FilterType)` but that would require an API break.

This is badly needed for `wondermagick`, but it will keep tracking `image` from git for the foreseeable future anyway. So if you'd rather skip straight to the semver-breaking version, I'm fine with that.

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -8,7 +8,7 @@
 use std::borrow::Cow;
 use std::io::{BufRead, Seek, Write};
 
-use png::{BlendOp, DisposeOp};
+use png::{BlendOp, DeflateCompression, DisposeOp};
 
 use crate::animation::{Delay, Frame, Frames, Ratio};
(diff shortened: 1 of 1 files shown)


Shnatsel:
This is what usage of this API in wondermagick looks like: https://github.com/Shnatsel/wondermagick/commit/0739f358c45c858b7d507c585ad7dd73bfbcf62c

fintelia:
The `CompressionType` enum is marked non-exhaustive. I think we could add variants there rather than making a new enum?

Shnatsel:
Oh, that is so much nicer! Great idea!

Shnatsel:
Closing in favor of #2583

PR #2583, opened 2025-09-01
Allow setting a specific PNG DEFLATE compression level

Shnatsel:
#2582 with much nicer API thanks to @fintelia 's suggestion

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -8,7 +8,7 @@
 use std::borrow::Cow;
 use std::io::{BufRead, Seek, Write};
 
-use png::{BlendOp, DisposeOp};
+use png::{BlendOp, DeflateCompression, DisposeOp};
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
Well, `cargo-semver-checks` complains that the enum can no longer be cast to an integer. I don't know if that was ever a reasonable expectation for an enum that is marked non-exhaustive.

Shnatsel:
This is what the usage looks like in practice: https://github.com/Shnatsel/wondermagick/compare/png-compression-level?expand=1

Really quite succinct, I like it

fintelia:
I'm not too worried about the semver break. Do you know if there's a way to disable just that specific lint from cargo-semver-checks? Otherwise I think it defaults to comparing the current contents to what's on crates.io and thus _every_ subsequent PR would be flagged with the same violation

Shnatsel:
I've downgraded that lint to a warning in configuration.

PR #2588, opened 2025-09-01
Add support for TIFF in CMYK-16bit

aki-cat:
- Made it using the same conversion the package already does for CMYK-8bit.
- Fixed number of bytes per pixel returned for CMKY-8bit as well (not sure if that was intentional or not).

Tested with real files.

### DIFF
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -82,6 +82,7 @@ where
             tiff::ColorType::RGBA(8) => ColorType::Rgba8,
             tiff::ColorType::RGBA(16) => ColorType::Rgba16,
             tiff::ColorType::CMYK(8) => ColorType::Rgb8,
+            tiff::ColorType::CMYK(16) => ColorType::Rgb16,
             tiff::ColorType::RGB(32) => ColorType::Rgb32F,
             tiff::ColorType::RGBA(32) => ColorType::Rgba32F,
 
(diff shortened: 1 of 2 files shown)


src/codecs/tiff.rs
```diff
@@ -399,6 +408,18 @@ fn cmyk_to_rgb(cmyk: &[u8]) -> [u8; 3] {
     ]
 }
 
+fn cmyk_to_rgb16(cmyk: &[u16]) -> [u16; 3] {
```
awxkee, line 411: This "quick&dirty" method is fine for debugging or testing purposes, but it is unsuitable for production code.
CMYK is a subtractive color model that can represent multiple print colorspaces, and RGB is an additive color model that can represent multiple display colorspaces. You need to know both the RGB colorspace (e.g. sRGB, Adobe RGB) and the CMYK colorspace (e.g. US Web Coated SWOP v2) in order to properly convert between the two. Without that information, the conversion is mostly meaningless.

aki-cat: This is the same function as a few lines above that. I do agree that this is an innaccurate conversion, but as it is now CMYK-8bit is being converted exactly like this. The question is whether this is a separate issue or included in this one.

Shnatsel (approved):
I'm OK with this as a starting point. We can evolve this and add more accurate color handling later.

Issue #2590, opened 2025-09-02
Weird change in behavior for `to_rgba32f()` in v0.25.7


tomaka:
Here is a reproduction example:

```rust
fn main() {
    let image = image::load_from_memory_with_format(
        &include_bytes!("../SNP_Grass_A_roughness.tga")[..],
        image::ImageFormat::Tga,
    )
    .unwrap();

    let image = image.to_rgba32f();

    let dbg = 
        image
            .pixels()
            .take(512)
            .flat_map(|px| px.0)
            .map(|v| v.to_string())
            .collect::<Vec<_>>()
            .join("\n");
    std::fs::write("out.txt", dbg).unwrap();
}
```

When I run this code with image v0.25.6, I get this:

```
0.8235294
0.8235294
0.8235294
1
0.8235294
0.8235294
0.8235294
1
0.827451
0.827451
0.827451
1
0.8235294
0.8235294
0.8235294
1
0.827451
0.827451
0.827451
1
0.827451
0.827451
0.827451
1
...
```

When I run the same code with image v0.25.7, I get this:

```
0.17512177
0.58897114
0.059436623
1
0.17512177
0.58897114
0.059436623
1
0.17595568
0.5917758
0.059719656
1
0.17512177
0.58897114
0.059436623
1
0.17595568
0.5917758
0.059719656
1
0.17595568
0.5917758
0.059719656
1
...
```

You can download the `SNP_Grass_A_roughness.tga` file here:
[SNP_Grass_A_roughness.tga.zip](https://github.com/user-attachments/files/22092353/SNP_Grass_A_roughness.tga.zip)

A probably very important point is that the file is grayscale, so only one color component.
I would expect the color to be broadcasted to the red, green, and blue channels, like is the case in image v0.25.6, whereas it seems that image v0.25.7 tries to create an image whose total "brightness" is equal to the original grayscale component.

I've noticed this morning that image v0.25.7 has been yanked, but it seems to be due to an API breaking change reason, and I assume that the issue I'm opening is still present in the main branch, though I haven't tried.

197g:
Oh, yikes. That's definitely not how those coefficients should be applied, only to rgb->luma and not the inverse. Good to catch it now, phew. The PR got a bit too big and apparently did not have enough _correct_ test vectors despite having a bunch..

PR #2591, opened 2025-09-02
Luma to rgb transform is broadcast

197g:
A bit of a walk of shame for _having had_ tests but those tests didn't make sense. Not that this was easy to decipher, having the expected coefficients for each color space appear probably shaded over the fact that the test was doing the _opposite_ of the transform where they are relevant.

The derived_luminance coefficients are in the Rgb->YCbCr matrix while for Luma->Rgb we need the first column of the YCbCr->Rgb matrix (since we assume luma to represent a `luma·(1 0 0)` vector in YCbCr). That column is always `1 1 1` as the first row of its inverse, i.e. the luminance coefficients, are constructed in such a way that they sum to `1`. So yeah, expansion is in fact alway just broadcasting in all non-constant and constant luminance chromaticity derived color spaces.

@awxkee Would you give this another pair of eyes? and I feel like we should ship it with `0.25.8` since color spaces are supposed to be flagship feature for it.

Closes: #2590

### DIFF
--- a/src/metadata/cicp.rs
+++ b/src/metadata/cicp.rs
@@ -1249,7 +1249,8 @@ impl ColorComponentForCicp for u8 {
 
     #[inline]
     fn clamp_from_f32(val: f32) -> Self {
-        (val * Self::MAX as f32 + 0.5).clamp(0., Self::MAX as f32) as u8
+        // Note: saturating conversion does the clamp for us
+        (val * Self::MAX as f32).round() as u8
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/metadata/cicp.rs
```diff
@@ -1603,7 +1585,7 @@ mod tests {
         let data = [u16::MAX];
         let color = Cicp::SRGB.into_rgb();
         let rgba = color.cast_pixels::<Luma<u16>, Rgb<u8>>(&data, &no_coefficient_fallback);
-        assert_eq!(rgba, [54, 182, 18]);
+        assert_eq!(rgba, [u8::MAX; 3]);
```
awxkee: It’s better not to test with edge values alone, since you almost always clamping. If something goes wrong in the computation you'll still just get 0 or 255, which doesn't really help.

PR #2596, opened 2025-09-04
Add utility from_raw_bgr{,a} for ImageBuffer

197g:
It is one of the most common channel orders other than Rgb and the construction involves nothing complicated, just swapping the channels around. We don't support that layout during computation, where it would complicate significant parts of `Pixel`, but as a helper for transferring a buffer we can easily have both on the `Rgb{,A}` typed containers.

See: https://github.com/image-rs/image/discussions/2595 which seems like a paper cut that can be easily and confidently solved.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1406,6 +1406,60 @@ impl<P: Pixel> ImageBuffer<P, Vec<P::Subpixel>> {
     }
 }
 
+impl<S, Container> ImageBuffer<Rgb<S>, Container>
+where
+    S: crate::Primitive + crate::traits::Enlargeable,
+    Rgb<S>: Pixel,
+    Container: DerefMut<Target = [<Rgb<S> as Pixel>::Subpixel]>,
+{
+    /// Construct an image by swapping `Bgr` channels into an `Rgb` order.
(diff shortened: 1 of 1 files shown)


fintelia:
Do you know whether the trait bounds would be suitable if we wanted to replace the implementation with one using `std::simd` once that's stable? IIRC swizzles like this are something that LLVM tends not to autovectorize well, and it would be nice to make sure we're not ruling out a faster implementation later.

197g:
Hm, alright, likely not. `std::simd::Simd` checks the trait bound validities on the type itself not only for operations. So we could not use them here and the best alternative could only do `as_chunks_mut`. I suppose using the sealed `PixelWithColorType` would work and support most relevant buffer types.

fintelia:
Actually, if we made `Subpixel` a sealed trait (which I've been meaning to propose anyway) would that be sufficient?

197g:
From what I understand, yes. The bound would be extremely weird but it _seems_ feasible under the current interface. It does not seem very controversial to even expose `Subpixel: SimdElement` publicly but if we wanted to avoid it, we'd need hacky code. 

I'm not entirely sold on sealing the `Subpixel` trait as a mechanism of ensuring we only need to handle a finite number of them; but sure, open for a sketch that can demonstrate any of the imageops being nicer to implement as-if in a minor version.

PR #2597, opened 2025-09-09
Mention MSRV change in 0.25.7 changelog entry

simonrepp:
Probably worth mentioning (I would have appreciated an up-front warning :)).

Thank you for image <3

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,9 @@ Features:
   - Serde support for more types (#2445)
   - PNM encoder now supports writing 16-bit images (#2431)
 
+Structural changes:
+  - Increased MSRV to 1.85.0 (from 1.78.0)
+
 API improvements:
   - `save`, `save_with_format`, `write_to` and `write_with_encoder` methods on `DynamicImage` now automatically convert the pixel format when necessary instead of returning an error (#2501)
   - Added `DynamicImage::has_alpha()` convenience method


197g (approved):
Sure, why not. Note: ignore clippy, new version new lints apparently.

197g:
Thank you. We track MSRV in `Cargo.toml:[package]rust-version`. Usually we avoid bumping it too quickly, or on a last stretch to a release. I'm not sure what warning besides the release information and the indicated version in `main` you'd expected to see. Staying behind rust-version too far is unfortunately holding back many possible changes, the choice is a balance of maintenance effort from churn and maintenace effort from workarounds / backports / …. I don't see us staying behind _more_ without a plan on _who_ would manage the extra requirements.

simonrepp:
Thank you for the swift merge!

> I'm not sure what warning besides the release information and the indicated version in main you'd expected to see.

By "up-front warning" I meant no more than a mention in the release information (which I carefully read). That wasn't there, hence the PR.

Also no hard feelings about the increase itself, 1.85.0 is rather on the fast-moving track, but it still seems reasonable to me. I understand that compromises need to be made, this is really just about communication of the change.

fintelia:
FYI: the MSRV of each version should be visible on crates.io. The [lib.rs versions page](https://lib.rs/crates/image/versions) even shows guessed MSRV values for earlier versions before it started being included in the Cargo.toml

simonrepp:
@fintelia If it's there I don't see it (and my full text search doesn't find it, not on any of the obvious pages around https://crates.io/crates/image/0.25.8)

197g:
Right column "Metadata" -> third entry below the package URL link and the publish date. On mobile it is below the Readme which, considering its length, definitely obscures things.

simonrepp:
@197g Oof, indeed :) thanks a lot for the pointer! (Being on desktop I really should have seen it, I was just too fixated on finding the magic words "MSRV" or "Minimum" and mistook the v1.85.0 for the crate version [which in hindsight makes zero sense except for a vague similarity in terms of the digits - 0.25.8]  :D My bad. Thanks again also @fintelia)

PR #2606, opened 2025-09-13
Remove some useless tuples from `crate::imageops::sample::filter3x3()`

LikeLakers2:
While reviewing my previous PR for errors, I happened to notice that these tuples are somewhat useless. They seem to be an artifact of https://github.com/image-rs/image/pull/384, one that wasn't cleaned up later on.

Since removing these tuples cleans up the code, and may (if rustc isn't already smart enough to detect that these tuples are useless) improve performance slightly, all without changing the behavior of the code, I went ahead and removed them.

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -898,7 +898,6 @@ where
         x if x == 0.0 => 1.0,
         sum => sum,
     };
-    let sum = (sum, sum, sum, sum);
 
     for y in 1..height - 1 {
         for x in 1..width - 1 {
@@ -908,7 +907,6 @@ where
             // Only a subtract and addition is needed for pixels after the first
             // in each row.
             for (&k, &(a, b)) in kernel.iter().zip(taps.iter()) {
-                let k = (k, k, k, k);
                 let x0 = x as isize + a;
                 let y0 = y as isize + b;
 
@@ -924,13 +922,13 @@ where
                     NumCast::from(k4).unwrap(),
                 );
 
-                t.0 += vec.0 * k.0;
-                t.1 += vec.1 * k.1;
-                t.2 += vec.2 * k.2;
-                t.3 += vec.3 * k.3;
+                t.0 += vec.0 * k;
+                t.1 += vec.1 * k;
+                t.2 += vec.2 * k;
+                t.3 += vec.3 * k;
             }
 
-            let (t1, t2, t3, t4) = (t.0 / sum.0, t.1 / sum.1, t.2 / sum.2, t.3 / sum.3);
+            let (t1, t2, t3, t4) = (t.0 / sum, t.1 / sum, t.2 / sum, t.3 / sum);
 
             #[allow(deprecated)]
             let t = Pixel::from_channels(


src/imageops/sample.rs
```diff
@@ -898,7 +898,6 @@ where
         x if x == 0.0 => 1.0,
         sum => sum,
     };
-    let sum = (sum, sum, sum, sum);
```
197g, line 901: Interesting tid bit: f32 mutliplication is usually faster than division. It would make sense to calculate `1.0 / sum` once upfront and use that coefficient instead of dividing by `sum` later on.

197g (commented):
Nice find of a rough spot, I haven't had a look at the implementation for a while.

LikeLakers2:
As an aside, while it is slightly unrelated to this PR's main purpose, I noted this:

https://github.com/image-rs/image/blob/ceb71e59496a32dbe2a56599ff60d09cb0b8cb20/src/imageops/sample.rs#L896-L900

and that these five lines could be rewritten as such:

```rust
    let sum = match kernel.iter().sum() {
        0.0 => 1.0,
        sum => sum,
    };
```

Should I include that in this PR, or open a different PR for this small refactor?

src/imageops/sample.rs
LikeLakers2: I'm unsure I understand what you're suggesting. I don't see anywhere that `1.0 / sum` is being used here?

(Side note: All of my PRs allow edits by maintainers. So if you ever want to include such changes in my PRs, go ahead.)

src/imageops/sample.rs
```diff
@@ -893,12 +893,10 @@ where
     let max = S::DEFAULT_MAX_VALUE;
     let max: f32 = NumCast::from(max).unwrap();
 
-    #[allow(clippy::redundant_guards)]
-    let sum = match kernel.iter().fold(0.0, |s, &item| s + item) {
+    let inverse_sum = match kernel.iter().fold(0.0, |s, &item| s + item) {
         x if x == 0.0 => 1.0,
-        sum => sum,
+        sum => 1.0 / sum,
     };
```
LikeLakers2: @197g, May I rewrite this to the following?
```rust
    let inverse_sum = match kernel.iter().sum() {
        0.0 => 1.0,
        sum => 1.0 / sum,
    };
```

This removes the redundant guard. It also uses `Iterator::sum()` (which is just a nicer way of saying `iter.fold(0.0, |x, &y| x + y)`) for clarity.

197g: Interesting that this works out with type inference. Also interesting that we have `matches!(-0.0, 0.0)`. Of course, as that matches the comparison traits and not `total_cmp` but still a curios case where I was not aware of detailed semantics.

197g: Sure, go ahead with the sum rewrite.

LikeLakers2: > Interesting that this works out with type inference.

Not gonna lie, I was surprised too! I was totally expecting to have to tell Rust "Hey, we want an `f32` here." But it seems Rust is smart enough to pick up that we want an `f32`, from the `match` statement.

src/imageops/sample.rs
LikeLakers2: Just thought I should mention: I took another look at the commit you made.

After thinking about it for a couple moments, it suddenly clicked for me how your suggestion works. And I'm surprised it didn't click for me before.

My apologies for the confusion - but hey, I get it now. :)

LikeLakers2:
Right. Unless anybody spots anything else, I think this PR should be good for merging - or at least, ready for final review. :)

LikeLakers2:
Is there anything blocking this from being merged?

fintelia (approved):

PR #2608, opened 2025-09-14
tga: adjust colormap lookup to match tga 2.0 spec

mstoeckl:
This changes the colormap lookup function to match what I think the TGA specification requires; it also adds a test image (which I created, and license MIT+Apache 2.0 to match `image`) on which `image` agrees with ImageMagick.

Quoting the TGA 2.0 spec (found e.g. at https://www.ludorg.net/amnesia/TGA_File_Format_Spec.html):
> _First Entry Index:_ Index of the first color map entry. Index refers to the starting entry in
loading the color map.
> Example: If you would have 1024 entries in the entire color map but you only need to store 72 of those entries, this field allows you to start in the middle of the color-map (e.g., position 342).
>
> _Color map Length:_ Total number of color map entries included.

In the above example, the color map length ("number ... included") is distinguished from the total number of entries in the color map; the example mentions storing (in the file) only 72 of 1024 entries (in memory/graphic card). For the example, I'd expect the color map length field to be 72, the offset 342, and the indices in the image between 342 and 342+72-1. Having a 1024-sized color map in the file would waste expensive disk space and leave the number 72 in the example unused. 

(The TGA format appears to be tightly aligned with the hardware, with the color map modes corresponding to "Pseudo-Color" graphics card modes -- which I expect perform the color lookup stage in hardware using a power of two sized map; then displaying a TGA image would amount to 1. copying the file's color map into the lookup table at the right offset 2. writing the uncompressed image data to the graphics card, which does the lookups.)

Edit: I found a [TGA reader from 1989](https://groups.google.com/g/comp.graphics/c/GmqbRTSIgIU/m/2Kn7ezq5eJsJ) that also interprets the "First Entry Index" like Imagemagick.


okaneco:
Should all the arithmetic be `checked_` here to avoid possible overflow?

mstoeckl:
> Should all the arithmetic be `checked_` here to avoid possible overflow?

I don't think this is necessary _here_, because `ColorMap` is not public API. When it is used internally, by `TGADecoder.expand_color_map`, the value of `index` is at most `u16::MAX`. The value of `ColorMap::entry_size` is derived from the the TGA header's `map_entry_size` field (a `u8`), and is furthermore checked in `TgaDecoder::new` to be between 2 and 4. With inputs like these overflow will not occur, assuming usize is at least 32 bit. (There may exist targets with smaller usize (see [documentation](https://doc.rust-lang.org/reference/types/numeric.html#r-type.numeric.int.size.minimum), but that's a more extreme restriction than `no_std`.)

fintelia:
This matches my suspicion from #2520, thanks for tracking down authoritative sources

Issue #559, opened 2016-07-12
Support for 15-bit and 16-bit TGA file formats


HarryLovesCode:
Hello,
I noticed recently that the TGA decoder didn't support 15-bit and 16-bit formats. This led me to ask the question, does this crate handle non-8-bit components? From what I understand, it doesn't (see [`dynimage.rs`](https://github.com/PistonDevelopers/image/blob/master/src/dynimage.rs#L45-L57)).

So, if I were to implement support for TGA file formats that use 5-bits per component, how would I need to represent them? Should we add support in the `DynamicImage` file I linked, or should they be scaled and cast to `u8`s?

nwin:
There is limited support for 16-bit formats. DecodingResult::U16 exists but it is not used  in https://github.com/PistonDevelopers/image/blob/master/src/dynimage.rs#L450. The most sensible approach would be to just scale it down. :)

When implementing reading `uN, (N%8 != 0)` image types you have  two options. Scale it to the next bigger uint and lie about the `ColorType`. Or just read it into an `u8` and modify https://github.com/PistonDevelopers/image/blob/master/src/dynimage.rs#L450 in order to scale it, which would be the better approach. See at https://github.com/PistonDevelopers/image/blob/master/src/dynimage.rs#L473 how 1, 2 and 4 bit grayscale images are handled.


Contribution are very welcome. :)

philipc:
The BMP decoder is currently scaling to `u8` and lying about the `ColorType`. It can handle any non-8-bit components. (It's also scaling down in some cases which isn't ideal, but those are unusual BMP formats.) @nwin Is it worth moving that scaling out of the BMP decoder? My main concern with that is it might be slower, but not sure how much that matters.

nwin:
You’re right it’s not worth it.

PR #2609, opened 2025-09-17
tga: support 5-bit-per-primary images

mstoeckl:
This is very loosely based on another PR to implement decoding of 5-bit-per-primary TGA images, https://github.com/image-rs/image/pull/2303. That PR (also licensed `MIT OR Apache-2.0`) has not been updated in a bit more than a year, in which time there have been refactorings and other changes to the TGA decoder which would require additional work to resolve. As a result, I expect it is unlikely that it will be updated this year, and am making this PR to try to have the feature implemented more quickly.

Notes:
* The decoder implementation ignores the "attribute bit" and does not interpret it as alpha, as recommended by the comment https://github.com/image-rs/image/pull/2303#issuecomment-3194350886. I've found that old TGA images often set the attribute bit to zero, and found one which uses it for something other than an alpha channel. 
* This adds an `ExtendedColorType::Rgb5` enum entry to indicate the "type" of image data being provided. A precise description of the pixel format used by _uncompressed_ true-color 5 bit-per-primary TGA images is [DRM_FORMAT_XRGB1555](https://pixfmtdb.emersion.fr/DRM_FORMAT_XRGB1555), but I don't think going into such precision is appropriate for how ExtendedColorType is used.
* I'm taking as minimal of an approach as I can for this PR, not implementing _encoding_ of TGA images with this format, and trying to avoid refactoring any more than necessary. (One thing I've been thinking of is rewriting the core parts of the TGA decoder to use a slice based framework, as in the [mstoeckl/tga-slice](https://github.com/mstoeckl/image/commits/tga-slice/) branch, but that may a) require trait bound changes b) refactoring everything anyway; so I'd rather not do things twice.)
* I made the test images myself, and license them (and this PR) to match the `image` crate, specifically as `MIT OR Apache-2.0`. You may also use them under [CC0](https://creativecommons.org/public-domain/cc0/).
* This should resolve https://github.com/image-rs/image/issues/559.


src/color.rs
```diff
@@ -125,6 +125,8 @@ pub enum ExtendedColorType {
     Rgb4,
     /// Pixel is 4-bit RGB with an alpha channel
     Rgba4,
+    /// Pixel contains 5-bit R, G and B channels
+    Rgb5,
```
fintelia: I think this might be better as `Rgb5x1` or similar to indicate that each pixel is actually 16 bits rather than 15.

src/codecs/tga/decoder.rs
```diff
@@ -389,6 +415,13 @@ impl<R: Read> ImageDecoder for TgaDecoder<R> {
             rawbuf.extend_from_slice(&buf[..num_raw_bytes]);
 
             self.expand_color_map(&rawbuf, buf, color_map)?;
+        } else if self.original_color_type == Some(ExtendedColorType::Rgb5) {
```
fintelia: Might be worth adding a comment here about needing this only if there's no color map, because of the pre-expand in the constructor

mstoeckl: Yes, done.

src/color.rs
mstoeckl: Done, but I should warn you that I think ExtendedColorType is starting to get stuck between two concepts: the in-memory layout of uncompressed arrays of pixel data (needed to extend ColorType), and the logical content of each numeric pixel value (which is just how many bits of float/uint each channel has, without stating an order or binary representation). The former is useful for direct access to uncompressed images; the latter for lossless conversions between image types (assuming compatible color representation information). I'd recommend splitting off the latter into a new type eventually.

Issue #2610, opened 2025-09-22
IPTC metadata support


1c3t3a:
[IPTC](https://en.wikipedia.org/wiki/IPTC_Information_Interchange_Model) is a common metadata format supported by a bunch of the formats that are available in this crate. We can extend the ImageDecoder trait to make it available to extract this metadata.

This is what I would like to work towards:

## Read support for IPTC metadata in:

- [x] JPEG - [implemented in zune-jpeg already](https://github.com/etemesi254/zune-image/pull/287), released in 0.4.21.

- [x] PNG - implemented in png crate but not wired up to image

- [ ] TIFF - implemented in tiff crate but not wired up to image

Is IPTC common in other formats?

Shnatsel:
> implemented in tiff crate but not wired up to image

@1c3t3a could you clarify? I cannot find any references to `tiff` crate supporting IPTC metadata.

1c3t3a:
Sorry, we should maybe open a bug there too. The problem is lined out in the discussion [here](https://github.com/image-rs/image/pull/2611#discussion_r2471909033). The Tiff crate does not need to support this, instead IPTC can be found in certain IFDs. The problem we have here is, that the Tiff crate has no (sane) interface to read a value as a `&[u8]` and we would need to add that to Tiff first.

fintelia:
Depending on the format of the data you're trying to get [`Decoder::get_tag_u8_vec`](https://docs.rs/tiff/latest/tiff/decoder/struct.Decoder.html#method.get_tag_u8_vec) might be what you want?

_Edit: though skimming that other thread it looks like the data isn't already in u8 format, so yeah that makes it more challenging_

197g:
I think we'd want a new buffer type that contains some tag's data as a `Vec<u8>` together with attributes for the byte order and tag-type it was stored in. Then said buffer can be converted to the appropriate endianess (or no conversion, for ascii etc.). We could then also use such a buffer as the output buffer argument, deriving allocation limits from the decoder, and probably supports partial reads in multiples of the type's size. Then a `as_bytes()` method to complete the interpretation. With appropriate constructors it would be powerful for the encoder, too.

fintelia:
I could also imagine an API analogous to our current handling of pixel data. For querying tags, there'd be a way to get the number of bytes and a way to pass in a `&mut [u8]` to write the contents into (converting to native endian if applicable). The higher level API could then allocate a `Vec<u32>` or whatever and rely on the lower level API for the actual reading.

PR #2611, opened 2025-09-22
Expose IPTC metadata for PNG and JPG

1c3t3a:
This change adds an `iptc_metadata` method (similar to `xmp_metadata` and `exif_metadata`) to the `ImageDecoder` and implements it for PNG and JPG.

This is related to #2610. If my research was correct and IPTC metadata only exists in PNG and JPG, then this should fix the issue.


1c3t3a:
The CI failure seems to be due to a spurious network failure in a setup step. Is there a way to retrigger the run?

src/codecs/tiff.rs
```diff
@@ -510,3 +536,42 @@ impl<W: Write + Seek> ImageEncoder for TiffEncoder<W> {
         self.encode(buf, width, height, color_type)
     }
 }
+
+/// This method converts a `Value` to a vector of bytes. A `Value` in Tiff can have different
+/// types, e.g. a byte, a short or a float. This method intents to convert all these types to
+/// a vector of bytes (e.g. a u32 can be represented as a [u8; 4]). However, since this is only
+/// intended to parse values stored in XMP, IPTC and EXIF metadata sections, we ignore / return
+/// an error on a few Values that don't make sense in this context (e.g. Value::Ifd).
+fn value_to_bytes(value: Value, bytes: &mut Vec<u8>) -> ImageResult<()> {
```
1c3t3a: I know that this is most likely not ideal. There is a discussion between me an @197g [here](https://github.com/image-rs/image-tiff/pull/293#issuecomment-3241685229), where Aurelia points out how to archieve this properly as part of the Tiff crate.

I put this change into a separate commit and I can drop it here and get to first do it properly in tiff, if that is requested.

1c3t3a:
The CI failure seems to be due to a spurious network failure in a setup step. Is there a way to retrigger the run?

1c3t3a:
Friendly ping on this PR, would anyone have a moment to take a look? :)

src/codecs/tiff.rs
197g: The one thing I'd worry about here is architectural difference. Since this is doing little-endian conversion it makes me suspicious it may only work on little endian targets? Then again, to quote Linus Torvalds, "Oh Christ. Is somebody seriously working on BE support in 2025?" so I think this may be fine if we first [check `target_endianess`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_endian) and always error for non-little targets.

1c3t3a: Actually I came up with this piece of code during the CI run which I think was running this on BE? I might misunderstood sth there :) Alternatively we could also just handle endianess correctly? I am also fine with erroring out, though!

1c3t3a: Actually this is a bit tricky. The tiff crate exposes a `byte_order` method, but the returned type is not marked as `pub`. The code I have handles the endianess of the target, but ignores the endianess of the file as I cannot get the endianess of the file easily... 

Wdyt about dropping this one for now? As a way forward, can we expose the endianess properly in the tiff crate?

1c3t3a: I drafted something in the latest commit. I know that this is most likely not what we want to go with and instead put it in the tiff crate.

I wanted to ask for some guidance on how to proceed:
1. Merge as is and fix it later in tiff by properly exposing metadata as a `Vec<u8>`.
2. Drop the IPTC Tiff part and first expose the ByteOrder in Tiff. Keep `value_to_bytes` in image.
3. Drop the IPTC Tiff part and expose a `Value` to `Vec<u8>` conversion in the tiff crate.

197g: Ah, right, it makes sense that it is also affected by the file endianess. I would go with (2) or (3). It's simpler to support it at a later point than undo some broken part with people having head scratches figuring out how it is broken in the first place.

Ideally we could have both (2) and (3), exposing endianess seems rather necessary for some value interpretation anyways. The more complex part is exposing the `Value` to `Vec<u8>` conversion or exposing a reader.

1c3t3a: Alright, sounds reasonable! Thanks for the guidance! :)

1c3t3a:
> Except for the tiff aspect this LGTM

As discussed I removed the TIFF part! :) This is now ready.

The test failures come from `codecs::jpeg::encoder::tests::grayscale_roundtrip_sanity_check` being broken at HEAD, I don't really understand why.

197g:
Test failure is due to zune's default output color space, when using `DecodeOptions`, being an `Rgb` and it expands luma to rgb in that setup since `0.5.5` (2025-10-27).

Shnatsel:
The rest failure has been fixed in #2635, please merge main and this should be good to go.

197g:
In case CI fails the Revert button will help us along, merging this should be fine.

PR #2612, opened 2025-09-24
Fix `PngEncoder` returning wrong `ImageFormat` in `ImageError`

alion02:


197g (approved):

197g:
Ah nice, simple copy & paste. Good find.

PR #2613, opened 2025-09-29
Implement `exif_metadata` for TIFF

1c3t3a:
Exif metadata for TIFF is funny, as the Tiff file itself is stored in the exif directory structure. So the trivial answer to what is Exif metadata is just the bytes of the file itself.

Note: This is not the most efficient way of obtaining the metadata as it also copies the image bytes, which we technically don't need. The alternative here is to add this to the tiff create and copy the correct metadata sections together. We didn't pursue this route as there were ideas to rework the entire `tiff::Value` type and this would likely add a huge dependency on the current behavior. We're open for feedback if we should to go down that route, though.

### DIFF
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -30,6 +30,7 @@ where
     dimensions: (u32, u32),
     color_type: ColorType,
     original_color_type: ExtendedColorType,
+    start_pos: u64,
 
     // We only use an Option here so we can call with_limits on the decoder without moving.
     inner: Option<Decoder<R>>,
@@ -40,7 +41,8 @@ where
     R: BufRead + Seek,
 {
     /// Create a new `TiffDecoder`.
-    pub fn new(r: R) -> Result<TiffDecoder<R>, ImageError> {
+    pub fn new(mut r: R) -> Result<TiffDecoder<R>, ImageError> {
+        let start_pos = r.stream_position()?;
         let mut inner = Decoder::new(r).map_err(ImageError::from_tiff_decode)?;
 
         let dimensions = inner.dimensions().map_err(ImageError::from_tiff_decode)?;
@@ -115,6 +117,7 @@ where
             dimensions,
             color_type,
             original_color_type,
+            start_pos,
             inner: Some(inner),
         })
     }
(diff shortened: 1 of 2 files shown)


Shnatsel:
I'm not a fan of this approach because if the application handles Exif like a black box and forwards it to the output (like most image conversion programs would do, including my own `wondermagick`) then this approach would embed the entire TIFF image into the converted image as metadata, resulting in extreme bloat.

1c3t3a:
> then this approach would embed the entire TIFF image into the converted image as metadata, resulting in extreme bloat.

Thanks for the feedback, that makes sense! I think the alternative for sure results in less pixel data, but is also annoying for the image crate:
1. We can read all IFD tags that are not `StripOffsets` or `TileOffsets`.
2. We reencode them to Exif using e.g. kamadak-exif.
3. We return those bytes.

This will result in smaller exif blocks for sure, but currently has the problem that we cannot really write exif blocks easily in Rust. The only thing I found is the experimental Writer in kamadak-exif: https://docs.rs/kamadak-exif/0.6.1/exif/experimental/struct.Writer.html.

197g:
Writing exif is the same as writing `tiff` IFDs with a different but overlapping set of keys (`u16`) used in the kv-map. Several documents gives us an almost exhaustive list of tags and their values which should be used for Exif, GPS, camera specific, private tags. In regards other than missing well-defined expected types all of these behave the same as the tiff structure itself.

We have a choice here:
1. For comprehensive support: Then we need to offer the complete file. Unfortunately, some rather ill-conceived vendor extensions used an obscured IFD pointer, that is byte data that denotes an offset but that we do not rewrite into its new offset if we copy it over.
2. For a correct lower-bound support: We can iterate the IFD structure with the methods `tiff` offers and copy them over into a new file. Now, `tiff::encoder::TiffEncoder` is still a bit rough and does not expose all the methods that you'd want, notably missing a method of adding a raw `Entry` structure with another backing `Read` / `IfdDecoder<'_>` for the data. It's also complex to add a subdirectory. So `kamadak-exif` might be the most convenient for now.
3. Maybe I'm overlooking a middle ground, maybe my interpretation of vendor extensions isn't quite up to date? https://exiftool.org/TagNames/EXIF.html is somewhat detailed but not machine readable and includes explicit unknown items

I tend towards (2). Currently, in `image` we only read the first of a tiff's chain of images so that returning all the metadata is even misleading. I also concur that the overhead is just too high for most expected uses. If we returned a more complex `Exif` type, not `Vec<u8>`, then maybe it would be more justified but this is a pitfall if you're handling images without looking at their format.

1c3t3a:
Yeah I agree and thanks for the summary @197g! Then let me take a stab at (2). Would it be fine to add a dependency on kamadak-exif and its experimental `Writer` for now until the `TiffEncoder` is ready?

197g:
Definitely for sketch purposes and it is lightweight enough. The unsafe dependency needs some muster, I wonder what it is for. But should not hold up the implementation in principle as it does not leak immediately to the API as far as I can tell.

197g:
Update: the main branch of `tiff` is now ready to at least interact with EXIF data. I don't claim read since it doesn't implement the UTF-8 data type extension among other things. But I've successfully tested against the metadata.

```diff
diff --git a/src/codecs/jpeg/decoder.rs b/src/codecs/jpeg/decoder.rs
index 96c433f6..70cb1310 100644
--- a/src/codecs/jpeg/decoder.rs
+++ b/src/codecs/jpeg/decoder.rs
@@ -251,4 +251,16 @@ mod tests {
         let mut decoder = JpegDecoder::new(Cursor::new(data)).unwrap();
         assert_eq!(decoder.orientation().unwrap(), Orientation::FlipHorizontal);
     }
+    #[test]
+    fn test_exif_parsing() {
+        let data = fs::read("tests/images/jpg/portrait_2.jpg").unwrap();
+        let mut decoder = JpegDecoder::new(Cursor::new(data)).unwrap();
+        let exif = decoder.exif_metadata().unwrap().unwrap();
+
+        let exif = Cursor::new(exif);
+        let mut exif = tiff::decoder::Decoder::open(exif).unwrap();
+        exif.next_directory().unwrap();
+        let orientation = exif.find_tag(tiff::tags::Tag::Orientation).unwrap();
+    }
```

PR #2620, opened 2025-10-18
pnm: parse integers in PBM/PGM/PPM headers without allocations

mstoeckl:
This is a small step toward eliminating the variable size allocations from the PNM decoder. 

Specifically, for the P1-P6-type PNM formats, the width, height, and maxval fields were previously being parsed by reading their digits into a `String` and then calling `u32::from_str_radix`. However, the fields could be as almost long as the input file and would still be loaded into RAM even if invalid. (For example, the image `P3 1000 1 255 1 1 1` where `000` is replaced by a very large number of 0s.) This usually isn't a big issue since image files are often small compared to available RAM, but there are a few exceptions (externally compressed `.ppm.gz` images, which some image viewers like ImageMagick's `display` automatically uncompress; images stored as sparse files or on compressed or networked filesystems).

Since I am already modifying the integer parsing, this change also rejects integers in the header that start with a leading + (like in `P6 5 +7 255\n`); Netpbm and most other implementations that I've tried also reject them.

Also, now that the `DecoderError::Overflow` error can come from multiple places, I've added an `ErrorDataSource` to it.

(I have a similar patch planned to avoid the unbounded allocations in the PAM header decoding, but am holding off on it to make this one easier to review and to not have too many PRs active.)

Note: The current CI failure is unrelated (clippy, about src/codecs/avif/decoder.rs).

### DIFF
--- a/src/codecs/pnm/decoder.rs
+++ b/src/codecs/pnm/decoder.rs
@@ -377,10 +377,9 @@ trait HeaderReader: Read {
         Ok(magic)
     }
 
-    /// Reads a string as well as a single whitespace after it, ignoring comments
-    fn read_next_string(&mut self) -> ImageResult<String> {
-        let mut bytes = Vec::new();
-
+    /// Reads an integer as well as a single whitespace after it, ignoring comments
+    /// and leading whitespace
+    fn read_next_u32(&mut self) -> ImageResult<u32> {
         // pair input bytes with a bool mask to remove comments
         #[allow(clippy::unbuffered_bytes)]
         let mark_comments = self.bytes().scan(true, |partof, read| {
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks for the PR! By and large this looks good.

Could you rebase on top of current main? That'll fix CI. And also now that the parsing code is not complete trivial, please add a couple tests for the success cases, including the leading zeroes.

mstoeckl:
> Could you rebase on top of current main? That'll fix CI. And also now that the parsing code is not complete trivial, please add a couple tests for the success cases, including the leading zeroes.

Rebased. I added two small tests covering OK headers, and two more for invalid headers.

Shnatsel (approved):

PR #2624, opened 2025-10-22
jpeg: do not encode empty images

mstoeckl:
JPEG images with a width or height of 0 are invalid according to the spec (see ITU-T Rec. T.81,  Section B.2.2, Table B.2, and Table B.10.) `libjpeg-turbo` also does not read them. This is a minimal change to error instead of encoding them.

(Since I'm already modifying the area, let me know if you'd also prefer the error message TODO resolved.)

### DIFF
--- a/src/codecs/jpeg/encoder.rs
+++ b/src/codecs/jpeg/encoder.rs
@@ -1,10 +1,10 @@
 #![allow(clippy::too_many_arguments)]
 use std::borrow::Cow;
 use std::io::{self, Write};
+use std::{error, fmt};
 
 use crate::error::{
-    ImageError, ImageResult, ParameterError, ParameterErrorKind, UnsupportedError,
-    UnsupportedErrorKind,
+    EncodingError, ImageError, ImageResult, UnsupportedError, UnsupportedErrorKind,
 };
 use crate::traits::PixelWithColorType;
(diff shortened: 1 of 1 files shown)


src/codecs/jpeg/encoder.rs
```diff
@@ -512,6 +512,16 @@ impl<W: Write> JpegEncoder<W> {
         let color_type = I::Pixel::COLOR_TYPE;
         let num_components = if n == 1 || n == 2 { 1 } else { 3 };
 
+        // TODO: not idiomatic yet. Should be an EncodingError and mention jpg.
+        let (width, height) = match (u16::try_from(image.width()), u16::try_from(image.height())) {
+            (Ok(w), Ok(h)) if w > 0 && h > 0 => (w, h),
```
197g: nit: I'd prefer pattern matching here but not by much.

```suggestion
            (Ok(w @ 1..), Ok(h @ 1..)) => (w, h),
```

And yes, addressing the TODO would be much appreciated given the error has become even more JPEG specific.

mstoeckl: Done; I've also addressed the TODO by adding an EncoderError type for JPEG, similar to what TGA has and the various codec DecoderError types.

PR #2625, opened 2025-10-22
feat(bmp): 4 bpp encoding support

Ddystopia:
This PR adds support for 4 bit images for BMP encoder. I tested it with different images (with even/odd number of rows etc) with `tinybmp` and images look exactly right.

As far as I understand the API, if user passes `L4` then we expect bytes user gives to already be packed. In my own code I used this helper to create those bufs (this is not a part of PR):

```rust
fn rgb_to_4bit_packed(image: &image::RgbImage, palette: &[[u8; 3]]) -> Vec<u8> {
    let stride = image.width() + (image.width() % 2); // pad to even width
    let buffer_size = (stride * 4).div_ceil(8) * image.height();
    let mut output = vec![0u8; buffer_size as usize];

    assert!(palette.len() <= 16, "palette must have at most 16 colors");

    for (x, y, pixel) in image.enumerate_pixels() {
        let idx = palette
            .iter()
            .position(|&c| c == pixel.0)
            .expect("Pixel color not in palette") as u8;

        let i = x as usize + y as usize * stride as usize;

        if i % 2 == 0 {
            output[i / 2] = idx << 4;
        } else {
            output[i / 2] |= idx;
        }
    }

    output
}
```

### DIFF
--- a/src/codecs/bmp/encoder.rs
+++ b/src/codecs/bmp/encoder.rs
@@ -56,10 +56,12 @@ impl<'a, W: Write + 'a> BmpEncoder<'a, W> {
         if palette.is_some()
             && color_type != ExtendedColorType::L8
             && color_type != ExtendedColorType::La8
+            && color_type != ExtendedColorType::L4
         {
             return Err(ImageError::Parameter(ParameterError::from_kind(
                 ParameterErrorKind::Generic(
-                    "Palette given which must only be used with L8 or La8 color types".to_string(),
+                    "Palette given which must only be used with L8, La8 or L4 color types"
+                        .to_string(),
                 ),
             )));
         }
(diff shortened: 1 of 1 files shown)


Ddystopia:
CI failure is not related to the pr, avif decoder was not touched:

```rust
error: there is no need to manually implement bit rotation
   --> src/codecs/avif/decoder.rs:652:21
    |
652 |             *item = (*item << target_expand_bits) | (*item >> (16 - target_expand_bits));
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: this expression can be rewritten as: `(*item).rotate_left(target_expand_bits)`
```

Shnatsel:
What is the motivation for this change?

Paletted mode mostly provides size savings, but BMP is an awful format if you want a small image size, so I'm not sure it's worth supporting in 2025.

Ddystopia:
I don't know any better image format for 4bpp images that is fast to decompress too 
 Tga? 8bpp minimum. Qoi? No less then rgb888. In the end 4bpp images with rle are quite good, for embedded use cases.

Ddystopia:
But at the end of the day I just vendored that encoder inside, if anyone would want that stuff they may dig out that pr and use it. I did pr just to share work.

Shnatsel:
Oh, you needed to write a paletted image with 16 colors for embedded use where it's cheap to decode? That explains the motivation, thank you.

197g:
>  I did pr just to share work.

Even though I think we won't merge this (interop with `ExtendedColorType` is not super well defined, having an implied buffer layout visible in the public interface would be questionable), thank you. I really appreciate it and crawled through some closed PRs in other crates already when trying to get specialized things to work. Even just to get a first idea what the problem is about, code is quite precise of a specification.

We had previously debated to spin `image-bmp` into a separate crate (and now got more interest in BMP), there it'll be easier to add more specialized methods. The PR contents can get reused there, too.

fintelia:
I'm been wanting to add support to `image-webp` for automatically detecting when an images uses <=16 colors and having it convert to using a 4-bit pallette. That's an optimization that libwebp does.

Issue #2618, opened 2025-10-06
Upgrade zune-jpeg to 0.5


jbicha:
zune-core and zune-jpeg 0.5.0 were released with a breaking change:

https://docs.rs/zune-jpeg/0.5.0/zune_jpeg/#migrating-from-version-04

PR #2630, opened 2025-10-25
Upgrade zune-core and zune-jpeg to 0.5

jakubadamw:
Upgrades zune-jpeg to 0.5. See the relevant [upgrade guideline](https://docs.rs/zune-jpeg/latest/zune_jpeg/index.html#migrating-from-version-04) for reference.

https://github.com/image-rs/image-tiff/pull/304 bumped the dependency for `tiff`. Until a version of that gets released, `image` will have both 0.4 and 0.5 in its dependency tree.

Closes https://github.com/image-rs/image/issues/2618.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,6 +121,3 @@ harness = false
 # TODO: remove when shipping the next major release after 0.25
 [package.metadata.cargo-semver-checks.lints]
 enum_discriminants_undefined_non_unit_variant = { level = "warn" }
-
-[patch.crates-io]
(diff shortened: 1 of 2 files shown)


fintelia:
Could you add an exception to deny.toml rather than adding a patch entry to the Cargo.toml?

jakubadamw:
> Could you add an exception to deny.toml rather than adding a patch entry to the Cargo.toml?

Done. 🙂

fintelia (approved):

fintelia:
Thanks!

PR #2634, opened 2025-10-29
Implement xmp_metadata for JpegDecoder

1c3t3a:
This is possible after the bump of zune_jpeg to 0.5


Shnatsel:
The code is a straightforward copy of the other metadata paths, so it looks good. But could you include a smaller image for testing and force-push? Something small and trivial that doesn't create copyright concerns would be best.

CI failures are unrelated and should be resolved if you rebase onto current main.

1c3t3a:
> But could you include a smaller image for testing and force-push?

I am not sure what you mean, but this PR already has a test image and test in metadata.rs :) I am rebasing now and force push again to have it properly run by CI and not fail due to the previous zune_jpeg error.

Shnatsel:
The included image is 91kb and shows potentially copyrighted content. Debian will complain about unclear legal status of the included image eventually. It would be better to include a small, trivial image, like a 5x5 gradient.

Shnatsel:
Here's a small, non-copyrightable image that contains both Exif and XMP metadata and can be used for testing both:

[exif-xmp-metadata.jpg.zip](https://github.com/user-attachments/files/23293932/exif-xmp-metadata.jpg.zip)

1c3t3a:
Thanks a lot! I made a copy of another test image and injected the XMP myself, but this looks the easiest in terms of copyright then. I used your image and updated the PR!

Issue #2500, opened 2025-06-20
Expose more JPEG encoding options


Shnatsel:
The built-in JPEG encoder always encodes with 4:2:2 subsampling, while the most useful ones are 4:4:4 and 4:2:0, and does not allow controlling other parameters such as progressive JPEG encoding, etc.

After #1885 gets us all these features, they need to be exposed in the public API.

This is needed for `wondermagick`.

Shnatsel:
Fixed in #2636 

Technically there's more parameters we could expose, but I think the current state strikes the right balance between flexibility and ease of use.

PR #2636, opened 2025-10-30
Replace in-tree JPEG encoder with jpeg-encoder crate

### LINKED ISSUE #1885: Use the `jpeg-encoder` crate for a 2x speedup
The [`jpeg-encoder`](https://crates.io/crates/jpeg-encoder) crate implements a pure-Rust JPEG encoder, with `unsafe` only used for SIMD. 

In my tests encoding [a large image](https://commons.wikimedia.org/wiki/File:Sun_over_Lake_Hawea,_New_Zealand.jpg) with it is over 2x faster than the current encoder (332 ms vs the current 785ms on x86). 

Correctness seems good too - there were no issues on my [web-scraped JPEG corpus](https://mega.nz/file/05snVB7L#KK8rECA5oHGN-dWwdcvDztoqIysEfKxkLmNWlYApHQ8) (23,000 images). The crate also seems to be extensively fuzzed.

Shnatsel:
This PR replaces the in-tree JPEG encoder with the jpeg-encoder crate. Closes #1885 and #2500.

### Background

[jpeg-encoder](https://crates.io/crates/jpeg-encoder) is a fork of our in-tree JPEG encoder with multiple enhancements (optimized huffman tables, more subsampling options, AVX optimizations). It is already used in production by Glycin (GNOME) instead of our built-in encoder.

### Correctness

This encoder used by Glycin so it is pretty well tested. I've added tests for roundtripping Exif and ICC.

### Performance

The end-to-end time for converting from PNG to JPEG with `wondermagick` goes down 2x, so the speedup for JPEG encoding alone has to be **more than 2x.** On x86 with AVX2 anyway; ARM doesn't gain as much because jpeg-encoder doesn't have a NEON implementation.

### Safety

Thanks to Rust 1.86 making most SIMD intrinsics safe, I've removed the vast majority of unsafe code in https://github.com/vstroebel/jpeg-encoder/pull/17 and https://github.com/vstroebel/jpeg-encoder/pull/18. The remaining small amount of `unsafe` is trivial.

### New functionality

This PR also exposes functions to control chroma subsampling factor. Our old encoder only supported 4:2:2, while jpeg-encoder supports 4:4:4, 4:2:2, 4:2:0, and a bunch of more obscure configurations. This should improve compression ratio.

This PR also lets the user opt into optimizing the Huffman tables of the generated JPEG file, further reducing file size.

### Breaking changes

A casualty of this migration is a JPEG-exclusive option to generate an image on the fly via a `GenericImageView`, but with the planned changes to image views this would have to be removed anyway. `pub fn encode()` is also removed because its signature would have to change and there is no reason to keep it around when it just duplicates `write_image()` exactly.

### Future work

TODO: handle `PixelDensityUnit::PixelAspectRatio` variant: https://github.com/vstroebel/jpeg-encoder/issues/21

jpeg-encoder supports more in-depth customization of the encoding process such as encoding progressive JPEGs, restart markets, etc. Exposing that functionality is not part of this PR.

### DIFF
--- a/src/codecs/jpeg/encoder.rs
+++ b/src/codecs/jpeg/encoder.rs
@@ -166,6 +166,13 @@ impl<W: Write> JpegEncoder<W> {
         self.encoder.set_sampling_factor(sampling.to_encoder_repr());
     }
 
+    /// Spend extra time optimizing Huffman tables. Slightly reduces file size at the cost of encoding speed.
+    ///
+    /// Defaults to **false**.
+    pub fn set_optimize_huffman_tables(&mut self, optimize: bool) {
+        self.encoder.set_optimized_huffman_tables(optimize);
+    }
+
     /// Set the pixel density of the images the encoder will encode.
     /// If this method is not called, then a default pixel aspect ratio of 1x1 will be applied,
     /// and no DPI information will be stored in the image.


Shnatsel:
Note that while jpeg-encoder reports IJG license due to SIMD code ported from libjpeg-turbo, it's actually ZLIB licensed according to https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/LICENSE.md

Both licenses are permissive so it shouldn't be a problem regardless ~~but I could change it to ZLIB upstream if you prefer your `cargo deny` to be more straightforward.~~ I think they included a bit of scalar code under IJG as well, but again it's entirely permissive so that shouldn't be a problem.

Shnatsel:
Now that https://github.com/image-rs/image/pull/2624 is merged I've incorporated its changes into this PR.

PR #2637, opened 2025-10-31
chore: Replace `doc_auto_cfg` with `doc_cfg`

sorairolake:
`doc_auto_cfg` was merged into `doc_cfg` in rust-lang/rust#138907.

When `docsrs` is enabled, the old `doc_auto_cfg` results in errors on the latest nightlies:

The new `doc_cfg` retains the automatic `cfg` generation that `doc_auto_cfg` used to provide.


197g (approved):
Makes sense, thanks.

Issue #2422, opened 2025-02-17
Strange white overlay, when resizing some images with alpha channel


qarmin:
When testing differences of resizing in image-rs and fast_image_resize, I found that with some images wtih alpha channel, image-rs output is looking strange with additional whitish border

Left - original
Middle - fast_image_resize (5.1.2)
Right - image-rs (0.25.5)

![Image](https://github.com/user-attachments/assets/0ead8a58-40a9-4553-b0e9-68a1c6775326)
![Image](https://github.com/user-attachments/assets/5ab708e4-db2a-47a9-ac86-d8404048caf8)

Files - [F.zip](https://github.com/user-attachments/files/18821218/F.zip)

I used this test project to test differences 

https://github.com/qarmin/img_hash/blob/master/tools/test_resize_diff/src/main.rs

awxkee:
You have to associate (premultiply) alpha before resampling and unassociate alpha after the operation if needed, this is required for correct blending.
Alpha association/unassociation currently is not supported in resizing.
To avoid artifacts in high-frequency regions of the image do not associate alpha if this is not necessary, or fill background with constant color ( and this one have some drawbacks ).

awxkee:
Also, I'd like to note that because hashing is involved, this implementation is not deterministic, just like in fast-image-resize. The results are implementation-defined, so you can't reliably use them for hashing.
To get deterministic results across all platforms, you need to use deterministic methods only. For example, using only power-of-2 scaling allows deterministic halving or multiplying by a power of 2.
Neither fast-image-resize nor the image crate (or others) do so.

kornelski:
At very least the `resize` docs should warn about scaling alpha without premultiplication. 

It's unfortunate that `Pixel`/`GenericImage` can't define whether the alpha is already premultiplied or not. It would be nice if `resize` could handle this properly by default.

Shnatsel:
I'll open a PR to uplift the wondermagick solution to this into `image` once development on the next semver-breaking release starts, presumably in September.

PR #2638, opened 2025-11-08
Allow extracting the alpha channel of a Pixel in a generic way

Shnatsel:
This is a straightforward and semver-compatible change that enables multiple other improvements.

This is necessary for being able to multiply and un-multiply pixels by the alpha channel, which is in turn necessary for properly resizing images with alpha channel: #2422

This is also beneficial for implementing pixel format optimization: https://github.com/Shnatsel/wondermagick/pull/54


197g (approved):
Seems straightforward to me as well.

Issue #2324, opened 2024-09-13
Blur for images with alpha is incorrect, bleeds color from transparent pixels


Shnatsel:
With this test image the blur implementation shows artifacts:

[blur-test-input](https://github.com/user-attachments/assets/20244122-fb60-4dcd-bd18-a82cd1aa4ed1)

It's just a white rectangle in the middle, surrounded by fully transparent pixels.

Gaussian blur in `image` bleeds color from fully transparent pixels into the white rectangle, resulting in the following image:

![blur-test-input fast](https://github.com/user-attachments/assets/99a7f785-1cb6-4e9e-9e85-00d6210a0a9c)

The fully transparent pixels have R set to 255 while all the other channels are set to 0, and that red color bleeds into the white rectangle.

The expected blur as produced by GIMP is like this - blurred white rectangle without any red color:

[blurred-gimp-20px](https://github.com/user-attachments/assets/7ebf71f6-87de-4af1-a64c-d2d96ddca826)

To avoid such artifacts, the pixel's contribution to the color change should be weighted by its alpha channel value. Fully transparent pixels should get multiplied by 0 and not contribute to the changes in non-alpha channels at all.

Code used for testing:

```rust
use std::env;
use std::error::Error;
use std::path::Path;
use std::process;

fn main() -> Result<(), Box<dyn Error>> {
    // Collect command-line arguments
    let args: Vec<String> = env::args().collect();

    // Ensure that we have 2 CLI arguments
    if args.len() != 3 {
        eprintln!("Usage: {} <path> <radius>", args[0]);
        process::exit(1);
    }

    let radius: f32 = args[2].parse()?;

    // Load the input image
    let path_str = &args[1];
    let path = Path::new(path_str);
    let input = image::open(path)?;

    // Correct but slow Gaussian blur
    let mut out_path_gauss = path.to_owned();
    out_path_gauss.set_extension("gaussian.png");

    let blurred = input.blur(radius);
    blurred.save(out_path_gauss)?;

    Ok(())
}
```

Originally found in https://github.com/image-rs/image/pull/2302#issuecomment-2346108041, moving it to the issue tracker so that it doesn't get lost.

Tested on `image` from git on commit 98ceb7197f5d8469b9b2364fbf9d671ce161ae97

torfmaster:
[...]

Issue #2626, opened 2025-10-23
Feature gate new dependencies


sophie-h:
It would be great if the introduction of new dependencies like `moxcms` could be feature gated if introduced in patch releases. That way, it is easier for Linux distributions to get the bugfixes in image-rs within a stable release, since they usually don't accept new dependencies.

PR #2639, opened 2025-11-09
Fast and correct resizing for DynamicImage

### LINKED ISSUE #2223: Resizing images is slow
The image resizing routines in `image` seem to be really slow compared to the competition.

For example, [`fast_image_resize`](https://github.com/cykooz/fast_image_resize) crate [benchmarks](https://github.com/cykooz/fast_image_resize?tab=readme-ov-file#some-benchmarks-for-x86_64) show that `image` is 5x slower than `libvips` and `fast_image_resize`'s fallback implementation, and 20x slower than `fast_image_resize`'s AVX2 implementation. It also does runtime CPU feature detection.

I have run the RGB benchmarks locally against `image` v0.25.1 to reproduce the results, and I can confirm that they are accurate on my machine - the performance gap is real:

```
|            | Nearest |  Box  | Bilinear | Bicubic | Lanczos3 |
|------------|:-------:|:-----:|:--------:|:-------:|:--------:|
| image      |  21.10  |   -   |  61.47   | 103.41  |  145.23  |
| resize     |    -    | 21.92 |  43.90   |  81.38  |  120.40  |
| libvips    |  5.64   | 46.89 |  15.42   |  23.99  |  31.24   |
| fir rust   |  0.20   | 7.27  |  11.57   |  20.53  |  29.96   |
| fir sse4.1 |  0.20   | 3.23  |   4.52   |  7.91   |  11.45   |
| fir avx2   |  0.21   | 2.32  |   2.90   |  5.04   |   8.25   |
```

### LINKED ISSUE #2422: Strange white overlay, when resizing some images with alpha channel

Shnatsel:
This PR migrates resizing DynamicImage from the built-in resizing impl that's hamstrung by the limitations of our generic APIs to the `pic-scale-safe` crate by @awxkee. This code is ported from the one I had in `wondermagick` for months now.

## Performance

This implementation is **~5x** faster than the previous one when running single-threaded. It also has optional parallelism gated by the `rayon` feature that takes performance even further.

I had to do an unnecessary copy to adhere to the existing API; we can eliminate it in the next semver-breaking release.

Fixes #2223 

## Correctness

Unlike the previous implementation, this one automatically premultiplies the pixels by alpha before resizing, but this is done **only when necessary** to avoid precision loss. Premultiplication is then reversed after resizing. This happens entirely transparently for the user. This API just automatically does the right thing for images with alpha.

Fixes #2422

## Future work

Not part of this PR to make it easier to review:

1. Wire up resizing via `pic-scale-safe` to `ImageBuffer` wrapping concrete types such as `Rgb`, `Rgba` and even some types that aren't in `DynamicImage` like `Brg` or `LumaF32`
1. Change the `thumbnail` methods to produce high-quality thumbnails by resizing down to 5x the using nearest-neighbor and then downscaling properly from there, like `imagemagick` does
1. Use the same alpha premultiplication trick to address #2324

Gamma and CICP are not taken into account. But the only reasonable way to do that is convert to a 16-bit linear colorspace and do the resizing there, which is too expensive to do transparently.

It is theoretically possible to make `pic-scale-safe` code entirely generic and fold it into `image` codebase, but it will require making nontrivial and semver-breaking improvements to our generic APIs first, and I don't think they should block this work.


Cargo.toml
```diff
@@ -86,7 +87,7 @@ tiff = ["dep:tiff"]
 webp = ["dep:image-webp"]
 
 # Other features
-rayon = ["dep:rayon", "ravif?/threading"] # Enables multi-threading
+rayon = ["dep:rayon", "ravif?/threading", "pic-scale-safe/rayon"] # Enables multi-threading
```
awxkee: I want to share an important point about using Rayon for heavy workloads.

Rayon uses some `sort of global scheduler` see [docs](https://docs.rs/rayon-core/1.12.1/src/rayon_core/thread_pool/mod.rs.html#83) that shared between all tasks.

Consider the following pseudocode:
```rust
[array_of_many_images_about_80-100].par_iter()
.for_each(|image| {
do_resize_with_rayoun_enabled()
});
```

Due to this rayon rules this code is vulnerable to DDoS attacks since it quickly exhaust RAM because rayon schedules all images at once where other ones are still not finished. I discovered this surprising behavior in libblur on my Mac M3, processing around 80–100 images is enough to knock the system out. That's a fairly small number. Image resizing is a bit less greedy, but it still shows the same characteristics. For some typical shared, or small server with 1-2GB RAM and 10GB of disk and 1-2 vCPU this amount I think the value is 10 times lower.

Even to my low security standards I had to remove rayon from pic-scale, libblur etc and make more classic solution (https://crates.io/crates/novtb), that's not counting it is actually 10% because it's planning heavy jobs better.

Even if we won't actually fix it here, I think it worth document that this API is subject to vulnerability to DDoS attacks with rayon enabled and require some additional attention to it.

Shnatsel: Thanks for letting me know! I've disabled rayon to sidestep this issue.

We can flip it on and update the documentation in the next semver-breaking release.

Separately, it is indeed very odd that rayon doesn't let us limit the amount of items concurrently in flight.

awxkee:
I thought it would end up being upstreamed rather than included :)

What do you think about transferring pic-scale-safe to image-rs org? I’m not quite sure how it would work on crates.io though, since the naming rules may require the crate to remain under my namespace 🤔

Shnatsel:
That sounds good to me but I'd like to hear from the other maintainers first. I don't think I have the required Github permissions in any case.

src/imageops/resize.rs
```diff
@@ -0,0 +1,206 @@
...
+    if premultiplied_by_alpha {
+        unpremultiply_alpha(&mut resized);
+    }
+    Ok(resized)
+}
+
+fn convert_filter_type(filter: FilterType) -> ResamplingFunction {
+    match filter {
+        FilterType::Nearest => ResamplingFunction::Nearest,
+        FilterType::Triangle => ResamplingFunction::Bartlett,
```
awxkee: FilterType::Triangle maps to ResamplingFunction::Bilinear. Probably not the best naming choice.

fintelia:
We received feedback that adding new default dependencies in patch releases [complicates things for distros](https://github.com/image-rs/image/issues/2626). Unfortunately, we can't make minor non-breaking changes until doing a 1.0 release.

On a somewhat related note, I've really been wanting to fix the inconsistency between `DynamicImage::resize` and `imageops::resize` (the former keeps the aspect ratio while the latter behaves like `DynamicImage::resize_exact`). Probably requires deprecating+removing `DynamicImage::resize` so that there isn't a silent behavior change

Shnatsel:
> On a somewhat related note, I've really been wanting to fix the inconsistency between `DynamicImage::resize` and `imageops::resize`

Yes, I think `imageops::resize` needs to be deprecated and replaced with `resize_exact` and `resize_proportional` implemented on buffer types that pic-scale-safe supports.

> We received feedback that adding new default dependencies in patch releases https://github.com/image-rs/image/issues/2626. Unfortunately, we can't make minor non-breaking changes until doing a 1.0 release

Distros also take issue with us upgrading major versions of crates, so us bumping `tiff` crate was [also a problem](https://matrix.to/#/!kcfNijNyPsBwWlheUx:gitter.im/$zET_qP4FUeT_Gzggtwk38FX63cuwZ_VhuxFa2OFO3O0?via=gitter.im&via=matrix.org&via=gnome.org). And we've already bumped zune-jpeg from 0.4 to 0.5 and ravif from 0.11 to 0.12, so this next release isn't making its way into distros anyway, with or without pic-scale-safe.

But I do believe it's time to start merging breaking changes, now that we know 0.25.8 doesn't need any urgent fixes. I'd love to make this into a breaking change and eliminate the extra copy, and also start merging things like https://github.com/image-rs/image/pull/2636

Shnatsel:
Now that API breaks are allowed, I've eliminated the mandatory copy. The API user can now choose to explicitly clone the image to hold on to an older version of it, or let it be overwritten.

I'm not sure if resizing should operate in-place or if it should accept an owned `DynamicImage` as input and consume it. I went with the in-place variant for now but I'm open to changing that, now or after this PR.

awxkee:
Since there is an API break anyway, I'd suggest renaming the `Triangle` filter to `Bilinear` or `Linear`. While `Triangle` is mathematically correct definition, it's much less commonly used, especially in image processing. I believe most users don’t actually understand what it means. This likely leads to higher use of Lanczos, which is more expensive and in practice not needed for most applications.

Perhaps Mitchell-Netravali filter could be added as well, since it's the most commonly used filter nowadays.

fintelia:
If I'm understanding correctly, the reason for the API change is that `pic-scale-safe` uses the input as a scratch buffer, while the old implementation allocates a dedicated scratch buffer?

My instinct is that in the future we may want to again change around how scratch buffers are handled. For instance, nearest sampling shouldn't need a scratch buffer and in other cases we might be able to use a scratch buffer that's much smaller than the input image. In which case, perhaps it would be better to keep the existing API?

Shnatsel:
No, `pic-scale-safe` does not touch the original buffer.

The original buffer is premultiplied by alpha channel in place, which is done by the code added in this PR. This is necessary to handle resizing images with alpha correctly, see #2422.

Nearest neighbour is indeed a special case that does not need premultiplying by alpha. I think it can be exposed separately as `sample()` or some such and avoid the complications of buffer management.

Whether to keep the original buffer around and make a copy internally (but only for images with alpha), or whether to always consume the buffer and have the API user make a copy explicitly (wasteful for images without alpha), comes down to what we expect the common case to be. In `wondermagick` I need to resize the image and don't need to hold on to the original, so I went with that. If we expect holding on to the original to be the default, then we can accept `&DynamicImage` instead and make a copy internally.

Or I guess there's also a third option of making a more complicated API that surfaces these differences.

fintelia:
Ah, using the original input to apply premultiplication makes sense. In that case I'm in favor of taking `&mut self`. The other option is consuming `self`, but there doesn't seem to be any precedent for that on the other image processing methods of `DynamicImage`.

I also noticed that there's two `thumbnail` methods on `DynamicImage` based on the doc comment seem to be shorthand for calling resize with nearest filtering? If so, perhaps we should remove them to avoid confusion over the differences

Shnatsel:
I'm going to replace `thumbnail` with fast and clever thumbnailing that `imagemagick` does: first use nearest-neighbor to scale down to 5x the target size, then use high-quality scaling the rest of the way.

I was going to do that in a follow-up PR, but if you'd like me to include it here I'm open to that.

fintelia:
That sounds good for a separate PR!

Shnatsel:
So this one's good to go?

I'd like to get this wrapped up before I move on to thumbnailing, etc.

fintelia (approved):

Shnatsel:
I'm going to go ahead and merge this, and then build other changes outlined in the PR description on top of this.

laundmo:
FYI: I think the "implementation is ~5x faster" estimate is underselling it *a bit*.

For my usecase, a small GrayImage, i benchmarked 1.4787 ms for Lanczos3 using the old image-rs resize, and 68.273 µs for Lanczos3 using the pic_scale_safe resize.

Heres some benchmarks on various sizes of GrayImage:
lines

Shnatsel:
You probably have the `rayon` feature enabled, so it runs on all cores instead of just one like the old implementation did.

laundmo:
Nope! No rayon feature, during the benchmark only 1 core at a time is maxed. It truly is just that much better. Oh, no `target-cpu=native` either.

PR #2640, opened 2025-11-10
Release v0.25.9

Shnatsel:
We gave 0.25.8 a couple of months to stew and it didn't really need any urgent fixes. So let's ship one last minor release in 0.25.x series and then start merging breaking changes to main, as [previously discussed](https://matrix.to/#/!kcfNijNyPsBwWlheUx:gitter.im/$QrnQm2KBPYP38vI8D3emG2aqYh_EkrwmvAfqb8GENHM?via=gitter.im&via=matrix.org&via=gnome.org).

~~Blocked on #2641~~

I also want to get GIF ICC and XMP metadata merged before this release, so marking this as draft until that's done. That depends on https://github.com/image-rs/image-gif/pull/222 and #2644

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,15 +9,16 @@
 ### Version 0.25.9
 
 Features:
- - Support extracting XMP metadata from PNG, JPEG, WebP and TIFF files (#2567, #2634)
+ - Support extracting XMP metadata from PNG, JPEG, GIF, WebP and TIFF files (#2567, #2634, #2644)
  - Support reading IPTC metadata from PNG and JPG files (#2611)
+ - Support reading ICC profile from GIF files (#2644)
  - Allow setting a specific DEFLATE compression level when writing PNG (#2583)
  - Initial support for 16-bit CMYK TIFF files (#2588)
  - Allow extracting the alpha channel of a `Pixel` in a generic way (#2638)
 
 Structural changes:
  - EXR format decoding now only uses multi-threading via Rayon when the `rayon` feature is enabled (#2643)
- - Upgraded zune-jpeg to 0.5.x, ravif to 0.12.x
+ - Upgraded zune-jpeg to 0.5.x, ravif to 0.12.x, gif to 0.14.x
  - pnm: parse integers in PBM/PGM/PPM headers without allocations (#2620)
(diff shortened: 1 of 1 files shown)


fintelia:
It would be good to make sure that any items we want to mark as `#[deprecated]` get included with this release. At a minimum, I think that's `DynamicImage::resize` (in favor of `resize_proportional`)

197g:
I'm not sure how sensible it is to mark those deprecated without having a replacement within the same compatible version release train ready. What we'd really want is a one-time signal on upgrading to the next major version, to check for those uses that accidentally still work but the deprecation is instead a constant hassle for everyone _not_ upgrade–which is somehow the opposite effect of the intention.

Shnatsel:
`resize_proportional` isn't implemented yet. I'd rather not show deprecation warnings when people can't actually fix them.

And I'd really like to change the signature of all `resize_*` functions in the next semver-breaking release anyway, so fixing it today wouldn't even avoid breaking changes when migrating to 1.0 or whatever we call it.

Figuring out the shape of the API changes for the next breaking release is going to take a while, so I'd rather not block shipping v0.25.9 on that.

fintelia:
I'm also OK with shipping 0.25.9 now and then doing a 0.25.10 release later once we have a complete list of all deprecations we want to make. It is a little extra effort because we'd need a 0.25.x series branch to make the release off of, but not a big deal.

We can discuss the `resize` methods more in #2645

fintelia:
On the topic of starting to merge breaking changes ahead of the 1.0 release: I think now is a good time to start the process, though I personally don't have time at the moment to drive the effort so other folks will have to take the lead.

197g:
That sounds like a good plan to me, too. So:

- Release `0.25.9`
- Keep a `version-0.25` branch around without breaking changes
- Prepare `1.0` (I'm much less reluctant to go with that as a new version now that color spaces are in place, the `Pixel` and `GenericImage` clarifications, and knowing from several sketches that different buffers are possible to fit in).
- Release `0.25.10` with deprecation that points to a `1` migration once in place.

Sure, can drive that until over the new year.

Shnatsel:
Sounds like v0.25.9 is good to go?

PR #2642, opened 2025-11-10
Try running emulated tests in release mode to speed up CI

Shnatsel:
The compilation time hit will likely be offset by the performance gains inside the slow emulator


Shnatsel:
In the emulated jobs all dependencies seem to be compiled from scratch every time, even in my own fork where this change is already on main and should have pre-warmed the caches: https://github.com/Shnatsel/image/pull/2

We need to make caching work with `cross` before we try tricks like `--release`.

PR #2643, opened 2025-11-11
Make rayon optional when exr is enabled

Shnatsel:
Disabling the `rayon` feature now actually removes the `rayon` dependency.

Previously `exr` would always pull in rayon and enable multithreading unconditionally.

Upstream https://github.com/johannesvollmer/exrs/issues/242 was fixed in git a long time ago but only shipped to crates.io now.

It is debatable whether this constitutes a semver break, so I'm OK with merging it both before and after #2640.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ num-traits = { version = "0.2.0" }
 # Optional dependencies
 color_quant = { version = "1.1", optional = true }
 dav1d = { version = "0.10.3", optional = true }
-exr = { version = "1.5.0", optional = true }
+exr = { version = "1.74.0", default-features = false, optional = true }
 gif = { version = "0.13.1", optional = true }
 image-webp = { version = "0.2.0", optional = true }
 mp4parse = { version = "0.17.0", optional = true }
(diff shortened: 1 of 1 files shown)


fintelia:
I don't think this is semver breaking? We gate some items behind the `rayon` _feature_, but won't be enabled just because the `rayon` _dependency_ is included in the build

Shnatsel:
Previously multi-threading for `exr` was always available, but now it will only be available if the `rayon` feature is enabled on the `image` crate.

That will not break any builds but it is a change in behavior.

197g:
It's probably fine in terms of SemVer, I'm more worried about an exr silently introducing rayon and breaking targets like wasm32 than the inverse. We should probably add it to the release notes in #2640 just as an explicit documentation of the change in behavior, afterall we also have very clear guidance in just enabling the feature.

johannesvollmer:
exr previously ran on wasm32 - even with rayon as dependency. it will fallback to single threaded on wasm. we have a pipeline that build the library and runs tests on wasm successfully :)

Issue #2568, opened 2025-08-28
XMP metadata read support


1c3t3a:
XMP is a common metadata format supported by a bunch of the formats that are available in this crate. We can extend the `ImageDecoder` trait to make it available to extract this metadata.

This is what I would like to work towards:

### Read support for XMP metadata in:

- [x] JPEG - ~~implemented in zune-jpeg already, needs to be released.~~ done in  #2634
- [x] PNG - implemented in png crate but not wired up to image
- [x] WebP - implemented in the image-webp crate but not wired up to image
- [x] TIFF - implemented in tiff crate but not wired up to image
- [ ] GIF - implemented in the gif crate (PR: https://github.com/image-rs/image-gif/pull/215) but not wired up to image

Is XMP common in other formats?

fintelia:
The `image-webp` crate [should already support](https://docs.rs/image-webp/latest/image_webp/struct.WebPDecoder.html#method.xmp_metadata) extracting XMP metadata.

PR #2644, opened 2025-11-12
Bump Gif to 0.14, introducing XMP and ICC

197g:
This brings XMP and ICC metadata along with a small amount of API polish that is unfortunately breaking. But turns out to be a rather small fix.

#2568

### DIFF
--- a/src/codecs/gif.rs
+++ b/src/codecs/gif.rs
@@ -214,6 +214,16 @@ impl<R: BufRead + Seek> ImageDecoder for GifDecoder<R> {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


197g:
@Shnatsel You wanted to be pinged for testing, this looks promising.

Shnatsel:
> along with a small amount of API polish that is unfortunately breaking

Can we split that off into its own PR so that we can merge most of this for 0.25.9 and leave the rest for the semver-breaking follow-up?

197g:
Oh, I meant breaking for `gif` not for `image`. According to the public-private dependencies file we should be fine bumping the version and integrating all of the changes.

Shnatsel:
All good, no regressions on my 17k images.

PR #2645, opened 2025-11-13
Deprecate `DynamicImage::resize` method in favor of `resize_to_fit`

fintelia:
The `resize` method is confusing because the dimensions of the output image won't necessarily match requested dimensions. This is in contrast to `imageops::resize` which returns an image of precisely the requested size. Deprecating the `resize` method and replacing it with an alternative makes the distinction clearer, though I'm open to alternative naming ideas.

I'll also say that if we didn't already have two methods to preserve aspect ratio while resizing an image, I probably won't support adding them. Given an exact resize method, building these other operations is a bit tricky but not something that necessarily needs to be in this crate. However since folks are already using them, the churn/disruption probably isn't worth it.

This PR also reorders some methods on `DynamicImage` so that the resize methods appear together in the docs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,28 @@
 
 ## Changes
 
+### Version 0.25.9
+
+Features:
+ - Support extracting XMP metadata from PNG, JPEG, GIF, WebP and TIFF files (#2567, #2634, #2644)
+ - Support reading IPTC metadata from PNG and JPG files (#2611)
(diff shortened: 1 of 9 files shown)


Shnatsel:
I went ahead and resolved the conflicts I caused by merging #2639

I've outright deleted `resize` because the signature changed and people would have to edit the code anyway.

This should be good to go after a `cargo fmt`

PR #2646, opened 2025-11-14
Clarify default features in README and add usage note

oligamiq:
Documentation improvements.

- Clarify default features in README
- Add usage note when using this library in a library you intend to publish

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ implement it.
 | `avif-native`     | Enables non-Rust dependencies of `avif` (`mp4parse` and `dav1d`)
 | `serde`           | Enables `serde` integration for various structs and options
 
-Note: When using this library in a library you intend to publish, it is recommended to set `default-features = false`. This library enables multithreading by default, which may cause issues when compiling for `no_std`, `Wasm`, or similar environments.
+Note: When using `image` in a library you intend to publish, it is recommended to set `default-features = false` and then explicitly enable format features that are absolutely necessary. This ensures a smaller dependency tree and faster iteration time. The default feature configuration would also enable multithreading which may cause unexpected behavior when running for inherently single-threaded environments such as `wasm` targets.
 
 ## Image Types
 


README.md
```diff
@@ -50,13 +50,15 @@ implement it.
...
-| `default-formats` | Format support for common image formats: AVIF, BMP, DDS, EXR, FF, GIF, HDR, ICO, JPEG, PNG, PNM, QOI, TGA, TIFF, and WebP
-| `rayon`           | Enables multi-threading with rayon context in some dependencies
+| `default-formats` | **Default**<br>Format support for common image formats: AVIF, BMP, DDS, EXR, FF, GIF, HDR, ICO, JPEG, PNG, PNM, QOI, TGA, TIFF, and WebP
+| `rayon`           | **Default**<br>Enables multi-threading with rayon context in some dependencies
 | `nasm`            | Enables the build-time use of `nasm` for `ravif`, requires `nasm` installed
 | `color_quant`     | Includes `color_quant` as an implementation of `imageops::ColorMap`
 | `avif-native`     | Enables non-Rust dependencies of `avif` (`mp4parse` and `dav1d`)
 | `serde`           | Enables `serde` integration for various structs and options
 
+Note: When using this library in a library you intend to publish, it is recommended to set `default-features = false`. This library enables multithreading by default, which may cause issues when compiling for `no_std`, `Wasm`, or similar environments.
```
Shnatsel:
```suggestion
Note: When using `image` in a library you intend to publish, it is recommended to set `default-features = false`. This library enables multithreading by default, which may cause issues when compiling for `no_std`, `Wasm`, or similar environments.
```

I think this wording makes it a bit more clear. Otherwise LGTM.

fintelia:
This library doesn't support `no_std` even with `default-features=false` so we shouldn't mention that part. I also want to confirm what issues are likely when using `rayon` + WASM? My understanding is that `rayon` now has a [fallback implementation for when threading is unsupported](https://docs.rs/rayon-core/1.13.0/rayon_core/#global-fallback-when-threading-is-unsupported) that should be used automatically when targeting WASM

oligamiq:
I want to reduce unnecessary dependencies mainly. Increasing the dependency libraries cause increasing linker's linking time. Linking duration is determined by the quantity and size of link objects. Consequently, compilation time increases substantially, slowing down the development cycle. When the number of dependency libraries exceeds 100, it becomes quite noticeable.

However, the idea of wanting to reduce dependencies isn't understood sometimes, so I put a different plausible reason in the README.

197g:
> I also want to confirm what issues are likely when using rayon + WASM? 

It's also my understanding that there is no issue with the combination per-se if the target detection works properly. But it will still behave very oddly, some APIs will still deterministically error and I don't know how the users ensure that the fallback is always usable. So I propose the wording should be changed to:

> Note: When using `image` in a library you intend to publish, it is recommended to set `default-features = false` and then explicitly enable format features that are absolutely necessary. This ensures a smaller dependency tree and faster iteration time. The default feature configuration would also enable multithreading which may cause unexpected behavior when running for inherently single-threaded environments such as `wasm` targets.

fintelia (approved):

PR #2648, opened 2025-11-18
Fix cargo-semver-checks CI

Shnatsel:
I broke it. Sorry :sweat_smile:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.26.0-alpha.1"
+version = "1.0.0-alpha.1"
 edition = "2021"
 resolver = "2"
 # guard against accidentally publushing alpha releases:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "image"
-version = "0.25.9"
+version = "0.26.0-alpha.1"
 edition = "2021"
 resolver = "2"
```
fintelia, line 9: Please also set `publish = false` so I don't accidentally trigger a publish of this alpha version...

Shnatsel: Since this is a pre-release, we can publish to crates.io without any negative consequences. People would have to explicitly opt in to depending on a pre-release to use it, and it provides no semver guarantees.

But I have no problem with adding `publish = false` if you're sure.

fintelia: Thanks. I'm mostly paranoid about running `cargo publish` from the wrong terminal window and publishing some messy in-progress PR when I thought I was publishing a completely different crate. Making sure that the Cargo.toml will block the publish either because the version has already been used or `publish = false` is set provides an extra safeguard

Shnatsel:
CI is green, should be good to go.

Cargo.toml
```diff
@@ -1,8 +1,11 @@
 [package]
 name = "image"
-version = "0.25.9"
+version = "0.26.0-alpha.1"
```
fintelia: Per #2316...

```suggestion
version = "1.0.0-alpha.1"
```

Shnatsel:
CI is failing because some Github URLs are returning error 500.

Shnatsel:
All green now.

197g (approved):

PR #2653, opened 2025-11-21
Rework PNG metadata extraction

1c3t3a:
This change implements gaps we observed in the current Metadata extraction for PNG. There are multiple significant changes:
- For all metadata types, also search for metadata in the zTXt, tEXT and iTXt chunk.
- If we find metadata in the zTXt chunk, handle it like ImageMagick metadata and interpret it as a hex string prefixed by a header, that is separated by newlines.

We did significant testing on these methods internally and ran it against all test files we could find. We also ran it against all test images in libpng and pngsuite, and so far did not find a single image that had metadata in the zTXt chunk that was not compressed like the ImageMagick hex string.


tests/metadata.rs
```diff
@@ -66,44 +100,16 @@ fn test_read_xmp_tiff() -> Result<(), image::ImageError> {
...
-        50, 53, 97, 48, 48, 48, 56, 52, 98, 54, 57, 54, 101, 54, 55, 55, 51, 55, 52, 54, 102, 54,
-        101, 49, 99, 48, 50, 48, 48, 48, 48, 48, 50, 48, 48, 48, 52, 48, 48, 51, 56, 52, 50, 52,
-        57, 52, 100, 48, 52, 50, 53, 48, 48, 48, 48, 48, 48, 48, 48, 10, 48, 48, 49, 48, 52, 51,
-        53, 57, 99, 52, 52, 54, 99, 101, 101, 97, 49, 48, 48, 52, 51, 50, 53, 57, 101, 54, 55, 100,
-        57, 51, 98, 102, 101, 54, 53, 49, 10,
+        56, 66, 73, 77, 4, 4, 0, 0, 0, 0, 0, 49, 28, 2, 110, 0, 24, 65, 73, 45, 71, 101, 110, 101,
+        114, 97, 116, 101, 100, 32, 119, 105, 116, 104, 32, 71, 111, 111, 103, 108, 101, 28, 2, 90,
+        0, 8, 75, 105, 110, 103, 115, 116, 111, 110, 28, 2, 0, 0, 2, 0, 4, 0, 56, 66, 73, 77, 4,
+        37, 0, 0, 0, 0, 0, 16, 67, 89, 196, 70, 206, 234, 16, 4, 50, 89, 230, 125, 147, 191, 230,
+        81,
```
197g, line 112: I assume this changed as a result of the hex decoding? At least the original test side looks like mostly hex digits from a glance.

src/codecs/png.rs
```diff
@@ -158,83 +160,165 @@ impl<R: BufRead + Seek> PngDecoder<R> {
...
+///      Metadata key    Optional info  Size  (opt.) header  payload (hex encoded)
+///
+/// This method parses such data and returns the decoded payload.
+/// If the structure doesn't match, we return and empty `Vec<u8>`.
+fn parse_raw_profile(buffer: &str, header: Option<&str>) -> Vec<u8> {
+    // Wrap logic in a closure to allow `?` for easy error propagation.
+    let parse = || {
+        let mut parts = buffer.splitn(4, '\n');
+
+        // Skip the first two parts, grab the size (3rd part), then the body.
```
197g, line 243: Nit: We're now dropping the contents of the second part, what would be contained in it? In the example it only says `Optional` but if I understand the imagemagick [source for it](https://github.com/ImageMagick/ImageMagick/blob/5285ae1de367013bf70a5a4e5feaefe3de8dda7b/coders/png.c#L1743-L1749) it is any ascii whitespace padding until a non-ws character.

src/codecs/png.rs
```diff
@@ -158,83 +160,165 @@ impl<R: BufRead + Seek> PngDecoder<R> {
...
+/// This is how ImageMagick stores metadata in zTXt chunks. By default, metadata
+/// there looks like the following:
+///
+/// "Raw profile type exif\nOptional\n264223\nExif\x00\x004578..."
+///  ^^^^^^^^^^^^^^^^^^^^   ^^^^^^^   ^^^^^^  ^^^^^^^^^^^ ^^^^
+///      Metadata key    Optional info  Size  (opt.) header  payload (hex encoded)
+///
+/// This method parses such data and returns the decoded payload.
+/// If the structure doesn't match, we return and empty `Vec<u8>`.
+fn parse_raw_profile(buffer: &str, header: Option<&str>) -> Vec<u8> {
```
fintelia, line 238: Is this something that would make sense to expose as a public method from the `png` crate?

src/codecs/png.rs
```diff
@@ -158,83 +160,165 @@ impl<R: BufRead + Seek> PngDecoder<R> {
...
+    // Wrap logic in a closure to allow `?` for easy error propagation.
+    let parse = || {
+        let mut parts = buffer.splitn(4, '\n');
+
+        // Skip the first two parts, grab the size (3rd part), then the body.
+        let size_str = parts.nth(2)?;
+        let body = parts.next()?;
+
+        // Parse size and validate a 4MB limit.
```
fintelia, line 247: Where does this 4MB come from? Is it from the underlying standards, or something you're adding to avoid running out of memory?

src/codecs/png.rs
1c3t3a: Oh yeah, that would really make sense! It should be a method on ZTXtChunk. I will open a PR shortly to add that.

PR #2655, opened 2025-11-21
Gif la8 encoding

Shnatsel:
Inspired by https://github.com/image-rs/image/pull/2654, I checked what formats image currently advertises as encodable as GIF, and La8 is one of them. It's not actually wired up so attempting to encode it would result in an error.

Depends on https://github.com/image-rs/image-gif/pull/223, marking as draft until that's merged and shipped


Shnatsel:
This is ready to go as soon as the required change to the `gif` crate is published to crates.io

Fuzzing CI fails because I didn't apply the `gif` crate override to the fuzzing Cargo.toml, the rest of CI is green.

Shnatsel:
I'm going to go ahead and merge since all of the complexity is in the `gif` crate and the changes to `image` are trivial.

Shnatsel:
Ah nevermind, I can't self-approve, the guard rails are too good :smile:

197g:
I thought you'd be able to acknowledge and merge regardless (but yea, github does not let you self-review with anything but a comment).

Issue #2659, opened 2025-11-24
hooks extensions are case sensitive


JWill392:
Image format support added by third party hooks only appears to work for lowercase extensions.  UPPERCASE or mixed do not find the registered hook and return error unsupported format.

For example, the default for images from my iphone appears to be uppercase '*.HEIC'.

Looking around the code, I think I've found the cause.

ImageReader::make_decoder accesses DECODING_HOOKS.get(&ext) directly, whereas builtin formats in ImageFormat::from_extension use ext.to_ascii_lowercase.

## Expected

/home/jackwill/Pictures/Wallpaper_Photos/GDT 25/IMG_0503.HEIC - loaded, width: 4284, height: 5712
/home/jackwill/Pictures/Wallpaper_Photos/GDT 25/IMG_0422.heic - loaded, width: 4032, height: 3024

## Actual behaviour

/home/jackwill/Pictures/Wallpaper_Photos/GDT 25/IMG_0503.HEIC - failed to load: The file extension `."HEIC"` was not recognized as an image format
/home/jackwill/Pictures/Wallpaper_Photos/GDT 25/IMG_0422.heic - loaded, width: 4032, height: 3024

## Reproduction steps

Should be reproducible with any third party integration.  I've encountered it with libheif_rs.  Note this could be worked around with with_guessed_format if other library implemented it (but there's a separate bug with libheif_rs format guessing lol; issue 35)

    libheif_rs::integration::image::register_all_decoding_hooks();
...
    let image = ImageReader::open(path)
        match image.decode() {
            Ok(img) => eprintln!("{} - loaded, width: {}, height: {}", &path.display(), img.width(), img.height()),
            Err(e) => eprintln!("{} - failed to load: {e}", &path.display()),
        }

Let me know if any other info would be helpful here.  Thanks for maintaining this crate; I'm just playing around with a hobby project, and it makes dealing with images a pleasure!

fintelia:
Want to submit a PR?

JWill392:
I'll give it a try! It's new to me, but I'd like to learn, if someone doesn't mind reviewing my attempt.

RunDevelopment:
I think this issue can be closed now that #2660 has been merged.

Issue #2616, opened 2025-10-02
Allow naming formats registered via hooks


Shnatsel:
PR  #2372 added decoding hooks, so a third-party format implementation can be registered as a plugin and work through all the usual facilities in an abstract way.

It also added a `Format` enum that lets `image` refer to these plugin formats:

https://github.com/image-rs/image/blob/1ed88e44e9acf5f08ac5236deda9b1ebeedc6054/src/io/image_reader_type.rs#L14-L18

However, that enum is private.

In `wondermagick` I keep running into the need to name formats, including plugin formats. For example, the input can be prefixed with a format modifier, e.g. `webp:some_file`, that specifies what format "some_file" should be treated as. There's [`Decoder::set_format`](https://docs.rs/image/latest/image/struct.ImageReader.html#method.set_format) that allows precisely that, but it accepts `ImageFormat` enum that can only represents built-in formats, not the ones registered via plugins.

## Draft

I think the best way is going to be to just another variant to `ImageFormat` to represent third-party plugins, in the same way as the private `Format` does today, and retire the private `Format` afterwards.

fintelia:
I'd be more inclined to add a `ImageReader::set_file_extension` method (potentially with a different name) that set the format via file extension, since that's what the `Format::Extension` variant refers to, not that the format will necessarily be decoded via a plugin.

197g:
While I also see some use for `set_file_extension` as an interface, the format type need seems more immediate. There are a bunch of methods on the builtin `ImageFormat` which could be transferred to such an (opaque) type as well. Also, `set_file_extension` is unclear when the link to `hooks` is not immediate. If I pass `jpeg`, I might expect to get matched with the builtin `Jpeg` decoder as well? Afterall, a much more intuitive way to use such a name may be closely emulating the path-based `open` with a non-`File` input. But relaxing the disjunction between hooks and `ImageFormat` is hard to realize without a type that implements a good `PartialEq` and aligns with expectations somehow, at least in a testable sense.

JWill392:
I'm encountering a related issue; it seems impossible to tell if ImageReader::with_guessed_format succeeded with a third party format.  ImageReader::format() returns Option<ImageFormat> enum rather than private Format enum, and it flattens even successful third party extension formats to None.

I was hoping to quickly read through a bunch of files and determine which are parseable images without decoding them, but I don't see a way to do that with third-party integrations.

mstoeckl:
[...]

PR #2660, opened 2025-11-24
Fix hook extension case

JWill392:
Make integration hook extensions case-insensitive, matching built-in formats.  Also added some tests to show it failing before the change; not sure if they're useful or follow standards.  New to Rust and first time contributing, so any feedback is welcome!

See issue #2659

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -13,6 +13,7 @@ use crate::{ImageDecoder, ImageResult};
 pub(crate) trait ReadSeek: Read + Seek {}
 impl<T: Read + Seek> ReadSeek for T {}
 
+/// Stores ascii lowercase extension to hook mapping
 pub(crate) static DECODING_HOOKS: RwLock<Option<HashMap<OsString, DecodingHook>>> =
     RwLock::new(None);
 
--- a/src/io/image_reader_type.rs
+++ b/src/io/image_reader_type.rs
@@ -14,7 +14,6 @@ use super::free_functions;
 #[derive(Clone)]
 enum Format {
     BuiltIn(ImageFormat),
-    /// A format based on file extension.  Contents should be normalized to ascii lowercase.
     Extension(OsString),
 }
 
@@ -163,7 +162,7 @@ impl<'a, R: 'a + BufRead + Seek> ImageReader<R> {
                 {
                     let hooks = DECODING_HOOKS.read().unwrap();
                     if let Some(hooks) = hooks.as_ref() {
-                        if let Some(hook) = hooks.get(&ext) {
+                        if let Some(hook) = hooks.get(&ext.to_ascii_lowercase()) {
                             return hook(GenericReader(BufReader::new(Box::new(reader))));
                         }
                     }
@@ -353,7 +352,7 @@ impl ImageReader<BufReader<File>> {
         let format = path
             .extension()
             .filter(|ext| !ext.is_empty())
-            .map(|ext| Format::Extension(ext.to_ascii_lowercase()));
+            .map(|ext| Format::Extension(ext.to_owned()));
 
         Ok(ImageReader {
             inner: BufReader::new(File::open(path)?),


src/hooks.rs
```diff
@@ -132,8 +134,71 @@ pub fn register_format_detection_hook(
...
+    fn decoding_hook() {
+        register_decoding_hook(
+            MOCK_HOOK_EXTENSION.into(),
+            Box::new(|_| Ok(Box::new(MockDecoder {}))),
+        );
+
+        ImageReader::open("tests/images/hook/extension.MoCkHoOk")
+            .unwrap()
+            .decode()
+            .unwrap();
```
197g, line 184: For a `should_panic` test this `unwrap` is slightly misleading considering the test expects to panic before we reach it. For the non-panicking version it's perfectly fine.

src/hooks.rs
```diff
@@ -132,8 +134,71 @@ pub fn register_format_detection_hook(
...
+    impl ImageDecoder for MockDecoder {
+        fn dimensions(&self) -> (u32, u32) {
+            (1, 1)
+        }
+        fn color_type(&self) -> ColorType {
+            ColorType::Rgb8
+        }
+        fn read_image(self, _buf: &mut [u8]) -> ImageResult<()> {
+            panic!("MOCK decoder called")
+        }
```
197g: I would avoid using panics as a test mechanism wherever possible. We want to test here that a decoder associated with a specified hook was executed. I think a nicer way to check would be using an atomic bool or `std::sync::Once`:

```
use std::sync::{Arc, Once};

struct MockDecoder {
    done: Arc<Once>,
}
```

```suggestion
        fn read_image(self, _buf: &mut [u8]) -> ImageResult<()> {
            self.done.call_once(|| {});
        }
```

```
let hook_verifier = Arc::new(Once::new());

register_decoding_hook(
    MOCK_HOOK_EXTENSION.into(),
    Box::new(move |_| {
        let decoder = MockDecoder { done: hook_verifier.clone() };
        Ok(Box::new(decoder))
    });
);
```

```
ImageReader::open("tests/images/hook/extension.MoCkHoOk")
            .unwrap()
            .decode()
            .unwrap();

assert!(hook_verifier.is_completed());
```

src/io/image_reader_type.rs
```diff
@@ -352,7 +352,7 @@ impl ImageReader<BufReader<File>> {
         let format = path
             .extension()
             .filter(|ext| !ext.is_empty())
-            .map(|ext| Format::Extension(ext.to_owned()));
+            .map(|ext| Format::Extension(ext.to_ascii_lowercase()));
```
197g: Nit: Please add a comment on the enum's variant that its contents are supposed to be normalized to ascii lowercase.

src/hooks.rs
JWill392: ahh thanks! I was trying to figure out how to do that nicely.  I'll apply these changes.

JWill392: ah couldn't figure out how to get this approach to work with multiple tests, since hook add is static.  Using a distinct extension for each test seemed too nasty.  And incrementing a static call count or something wouldn't work with concurrent tests.

I just changed it so decoder outputs known mock image data, and I check that. Hopefully that doesn't conflict with something I don't know about the image decoding process, but it seems to work?

src/io/image_reader_type.rs
RunDevelopment: Instead of lower-casing here, I think it would be better to allow arbitrary casing in `Format::Extension` and lower-case in `ImageReader::make_decoder` where casing causes problems.

I'm saying this in the context of #2662 where I added a second code path that creates a `Format::Extension`.

JWill392: hm good point, that fits the existing pattern of ImageFormat::from_extension better.  I don't want to affect that since your PR and [issue 2616](https://github.com/image-rs/image/issues/2616) might change it.

src/hooks.rs
197g: That's also a way to communicate the use explicitly, I like it.

PR #2667, opened 2025-11-26
Move benchmark temp file into `tmp/` to avoid untracked repo artifacts

### LINKED ISSUE #2666: [bench]: Running `encode` bench creates persistent `temp.bmp` file in repo root, causing constant untracked file noise
Running the `encode` benchmark suite always creates a `temp.bmp` file in the repository root. Since the file is not *gitignored*, every run leaves the working tree in a dirty state with an untracked file.

This behavior comes from here:

https://github.com/image-rs/image/blob/81b6eac19c50bd1bbe12f42eb39a5d99d0b20c60/benches/encode.rs#L75-L82

## Possible solution

We could use the `tempfile` crate or store the temp file under `std::env::temp_dir()`. But I think this would be overkill for a single use case - e.g with `std::env::temp_dir()` we *must use a secure method to create a uniquely named file*.

**Instead**, we could simply:

1. Write to a dedicated `/tmp` directory inside the repository root, e.g.: `<root>/tmp/temp.bmp`
2. Add `/tmp/` to `.gitignore`

## Additional context

I'm drafting a PR to fix the inconvenience mentioned above.

ekkolon:
`bmp` image encoding benchmark previously wrote `temp.bmp` into the workspace root, causing persistent untracked files.
Temp output is now placed under `tmp/` to keep the repo clean during bench runs.

Closes #2666

### DIFF
--- a/benches/encode.rs
+++ b/benches/encode.rs
@@ -6,6 +6,7 @@ use image::{ExtendedColorType, ImageEncoder};
 
 use std::fs::File;
 use std::io::{BufWriter, Seek, SeekFrom, Write};
+use std::path::PathBuf;
 
 trait Encoder {
     fn encode_raw(&self, into: &mut Vec<u8>, im: &[u8], dims: u32, color: ExtendedColorType);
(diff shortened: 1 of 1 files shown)


benches/encode.rs
```diff
@@ -76,7 +76,11 @@ fn encode_zeroed(group: &mut BenchGroup, with: &dyn Encoder, size: u32, color: E
         BenchmarkId::new(format!("zero-{color:?}-file"), size),
         &im,
         |b, image| {
-            let file = File::create("temp.bmp").unwrap();
+            const TEMP_DIR: &str = "tmp";
+            let tempdir = std::env::current_dir().unwrap().join(TEMP_DIR);
```
197g: Maybe we could use [`CARGO_TARGET_TMPDIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) set during the build. It seems appropriate, I hope the variable is also set for benches without harness.

```suggestion
            const TEMP_DIR: &str = env!("CARGO_TARGET_TMPDIR");
            let tempdir = std::env::current_dir().unwrap().join(TEMP_DIR);
```

ekkolon: That's indeed an elegant way to handle this. And it seems `CARGO_TARGET_TMPDIR` is set regardless of harness.

I'll also remove the redundant call to `std::env::create_dir_all`, and `/tmp/` from `.gitignore`.

197g:
Pre-emptively approved, assuming the changes applied as discussed. Seems straightforward :)

PR #2669, opened 2025-11-26
Fix thumbnail generation for floating-point images

### LINKED ISSUE #2273: Floating point image generates incorrect thumbnail
The [rounding code](https://github.com/image-rs/image/blob/v0.25.1/src/imageops/sample.rs#L682) doesn't work with color types like `Rgba32F`.

## Reproduction code

Test image: https://github.com/johannesvollmer/exrs/blob/master/tests/images/valid/custom/compression_methods/f32/b44a.exr

```rust
use image::GenericImageView;

fn main() -> anyhow::Result<()> {
    let args: Vec<String> = ::std::env::args().collect();
    let path = &args[1];

    let im = image::open(path)?;

    println!("Get pixel at (0, 0):\n");
    println!("Original = {:?}", im.get_pixel(0, 0));
    println!("Original (raw) = {:?}", im.as_rgba32f().unwrap().get_pixel(0, 0));

    let thumbnail = im.thumbnail(im.width(), im.height());
    println!("Thumbnail (same size) = {:?}", thumbnail.get_pixel(0, 0));
    println!("Thumbnail (same size, raw) = {:?}", thumbnail.as_rgba32f().unwrap().get_pixel(0, 0));
    let thumbnail = im.thumbnail(im.width() / 2, im.height() / 2);
    println!("Thumbnail (scale down) = {:?}", thumbnail.get_pixel(0, 0));
    let thumbnail = im.thumbnail(im.width() * 2, im.height() * 2);
    println!("Thumbnail (scale up) = {:?}", thumbnail.get_pixel(0, 0));

    Ok(())
}
```

The output:

```
Get pixel at (0, 0):

Original = Rgba([5, 6, 8, 255])
Original (raw) = Rgba([0.018939972, 0.022951007, 0.033104062, 1.0])
Thumbnail (same size) = Rgba([132, 133, 136, 255])
Thumbnail (same size, raw) = Rgba([0.51894, 0.522951, 0.53310406, 1.5])
Thumbnail (scale down) = Rgba([132, 133, 136, 255])
Thumbnail (scale up) = Rgba([132, 133, 136, 255])
```

RunDevelopment:
fixes #2273

Thumbnail generation used an old trick to get rounded integer division. Namely, `(sum + n/2) / n` gives a division by `n` rounded to the nearest integer (half up) for all non-negative integers. This is necessary to correctly round the average color. Unfortunately, this trick does not work for floating-point images. The effect is that all channels get 0.5 added, which produces unusable results.

This PR fixes this bug by choosing the rounding addend (previously always `n/2`) based on whether the underlying subpixel type is an integer type or a type that can represent fractions (e.g. floating point types). For integer types, `n/2` is used like before. For all other types, 0 is used. This will produce correct results for all integer types, all floating-point types, and potential rational number types.

Unfortunately, `num_traits` doesn't have a function like `is_integer` or `is_truncating_div`, so I implemented my own `is_integer`: 

```rust
let is_integer = <S::Larger as NumCast>::from(0.25).unwrap().is_zero();
```

The trick is that integer types will truncate 0.25 to 0, while types which can represent fractions will return something other than zero. LLVM can reliably optimize this to return a constant in my testing.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -103,18 +103,10 @@ pub(crate) fn vec_try_with_capacity<T>(capacity: usize) -> Result<Vec<T>, TryRes
 /// Returns whether `T` can only represent integer values.
 #[inline]
 pub(crate) fn is_integer<T: num_traits::NumCast + num_traits::Zero>() -> bool {
-    // This uses a cast of 0.25 to T, because:
-    // 1) integer types cannot represent fractional values, so they have to
-    //    round.
-    // 2) NumCast::from does not mandate a rounding mode. All common integer
-    //    types round towards zero (truncation), but there may be exotic types
-    //    that round to nearest. 0.25 is rounded to 0 for all common rounding
-    //    modes, except round towards infinity (ceil).
-    // 3) we don't want to rely on floating-point-specific properties (e.g. NaN)
-    //    to support non-float types that can represent fractional values.
-    // 4) most types that can represent fractional values can represent 0.25
-    //    exactly.
-    <T as num_traits::NumCast>::from(0.25).unwrap().is_zero()
+    // This uses a cast of 0.5 to T, because integers will truncate to zero
+    // while types that can represent fractional values will return something
+    // other than zero.
+    <T as num_traits::NumCast>::from(0.5).unwrap().is_zero()
 }
 
 #[cfg(test)]


src/imageops/sample.rs
```diff
@@ -716,7 +716,16 @@ where
     }
 
     let n = <S::Larger as NumCast>::from((right - left) * (top - bottom)).unwrap();
-    let round = <S::Larger as NumCast>::from(n / NumCast::from(2).unwrap()).unwrap();
+
+    // For integer types division truncates, so we need to add n/2 to round to
+    // the nearest integer. Floating point types do not need this.
+    let is_integer = <S::Larger as NumCast>::from(0.25).unwrap().is_zero();
```
197g: nit: I think it would help to explain `0.25` is somewhat arbitrary here. Just anything that rounds to zero for integers but not for any of the floating point types. 

Outlining it as a utils function might also provide some clarity. We don't really care about _integer_, we care that the cast truncates down. Arguably `0.5` may be a simpler constant to use that is more closely aligned to the intent? Doesn't make that much difference as long as it is explained in the code.

```
fn num_truncates_to_int<T: num_traits::NumCast + num_traits::Zero>() -> bool {
    <T as num_traits::NumCast>::from(0.5).unwrap().is_zero()
}
```


Another way to write this may be using `NaN`, avoiding the `unwrap` but it's not as clear that we're aiming to distinguish the truncation behavior anymore.

```
fn is_integer<T: num_traits::NumCast + num_traits::Zero>() -> bool {
    <T as num_traits::NumCast>::from(f64::NAN).is_none()
}
```

RunDevelopment: I selected `0.25` because I didn't know whether people might be able to use their types for subpixels. As in, custom types that implement the necessary traits. And most types that can represent fractional number, can represent 0.25 exactly.

I didn't want to use 0.5, because custom types might (for some reason) implement a cast that rounds to nearest half up. Not common, but I can see this being a good default in some cases. And `NumCast::from` only *permits* truncation, it doesn't mandate it. I don't want to use NaN, because it falsely detects things like fixed-point and rational numbers as integers.

All of these methods work for the standard numeric types, but I wanted something that works for more.

197g: But if a custom integer-like type (which we don't _support_ as `Pixel` anyways, so I'm not sure if it is applicable) implements `0.5` as rounding up, then we should __not__ apply the bias to this summation. Right? The method should return `false` here and the path be the floating point summation. All of this is to say we should be quite explicit in the property we are really testing for and relying on here—if we identify integer vs. rounding behavior to overlap but be different.

RunDevelopment: Hmmm, okay. You convinced me. Let's use 0.5.

PR #2670, opened 2025-11-26
Cleaner `Debug` output for `ImageBuffer`

### LINKED ISSUE #1613: Cleaner `Debug` output for `ImageBuffer`
Currently `ImageBuffer` uses `#[derive(Debug)]`, which prints something like this (where every channel of every pixel is printed):

```
ImageBuffer {
    width: 64,
    height: 64,
    _phantom: PhantomData,
    data: [
        0,
        0,
        0,
        0,
       ...
        0,
        0,
        0,
    ],
}
```

This is frustrating when debugging my own types which contain an `ImageBuffer`, since even a few reasonably sized images add up to tens of millions of lines of output.  It would be much nicer to see something small like `ImageBuffer<Rgb8>(64x64)`, or perhaps something like `ImageBuffer(64x64x4)` if we don't want to enforce a trait bound on the pixel format.  I'm open to ideas over what the format should be, and I'm happy to make a PR for this.  Implementation would simply be a custom `Derive` impl for `ImageBuffer` and any other image types.

RunDevelopment:
resolves #1613

I spend some time making the debug output for `ImageBuffer` easier to read, but probably went a bit too far...

```rust
assert_eq!(
    format!("{:?}", GrayImage::new(100, 100)),
    "ImageBuffer<Luma<u8>>(100x100, sRGB)"
);
```

If you want me to dial the custom debug back a little, I think I could use the standard `debug_struct` builder to make something like this:

```
ImageBuffer {
    width: 100,
    height: 100,
    pixel_type: Rgb<u8>,
    color: sRGB,
}
```

I don't mind either way, so just tell me what you think would be best.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1229,14 +1229,20 @@ where
     Container: fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let Self { width, height, .. } = self;
-
         let mut pixel = std::any::type_name::<P>();
         pixel = pixel.strip_prefix("image::color::").unwrap_or(pixel);
 
-        let color = self.color.known_name().unwrap_or("<custom color space>");
+        let mut debug_struct = f.debug_struct(&format!("ImageBuffer::<{pixel}, _>"));
+        debug_struct.field("width", &self.width);
+        debug_struct.field("height", &self.height);
 
-        write!(f, "ImageBuffer<{pixel}>({width}x{height}, {color})")
+        if let Some(color_name) = self.color.known_name() {
+            debug_struct.field("color", &color_name);
+        } else {
+            debug_struct.field("color", &self.color);
+        }
+
+        debug_struct.finish()
     }
 }
 
@@ -1785,7 +1791,9 @@ mod test {
     use super::{GrayImage, ImageBuffer, RgbImage};
     use crate::math::Rect;
     use crate::metadata::Cicp;
+    use crate::metadata::CicpMatrixCoefficients;
     use crate::metadata::CicpTransform;
+    use crate::metadata::CicpVideoFullRangeFlag;
     use crate::GenericImage as _;
     use crate::ImageFormat;
     use crate::{Luma, LumaA, Pixel, Rgb, Rgba};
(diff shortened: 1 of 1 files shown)


197g:
I think less custom structure would be better. I don't see much reason to deviate from consistency here but the type_name approach for the `debug_struct` name's parameter is pretty good for readability. I'd use that.

```
ImageBuffer::<_, Rgb<u8>> {
    width: 100,
    height: 100,
    color: "sRGB",
}
```

On the fancier side I'm pondering if a fixed-size grayscale thumbnail could acehive an overview of the contents without the dump that is `data` in the current derive.

fintelia:
Maybe also have a `data: [1, 2, 3, 4, 5, ...],` field? Enough to show the actual values of a few pixels but not overwhelm the output

RunDevelopment:
Okay, it's now this:

```
ImageBuffer::<Luma<u8>, _> {
    width: 16,
    height: 16,
    color: "sRGB",
}
```

Unfortunately, I don't think `data: [1, 2, 3, 4, 5, ...]` will be possible. The only thing known about that container is that it's `Debug` and we don't have specialization yet. So we can't even iterate the container. The only possible solution given the current constraints is to debug print the container into a separate buffer, see if it looks like an array/vec, and then truncate it after N characters (or somewhere around a `,` for nicer results). Suuuper hacky, but it would probably work for common container types.

src/images/buffer.rs
```diff
@@ -1223,6 +1223,29 @@ where
...
+impl<P, Container> fmt::Debug for ImageBuffer<P, Container>
+where
+    P: Pixel + fmt::Debug,
+    Container: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut pixel = std::any::type_name::<P>();
+        pixel = pixel.strip_prefix("image::color::").unwrap_or(pixel);
+
+        let mut debug_struct = f.debug_struct(&format!("ImageBuffer::<{pixel}, _>"));
```
RunDevelopment, line 1235: This could be done without an allocation in `format!`. Like this:

```rs
f.write_str("ImageBuffer::<")?;
f.write_str(pixel)?;
f.write_str(", _>")?;
f.debug_struct("").field(...).and_so_on
```

However, this relies on the fact that `f.debug_struct(str)` will immediately print the struct name without any additional formatting logic (like e.g. padding spaces, etc). I'm not sure if we're allowed to make this assumption.

Please tell me what you think we should do.

197g: I don't mind the allocation here (it's Debug ..) so I'd prefer readable code over a hackish solution. Would be nice if `name` could be an arbitrary `Debug` itself but really a micro-optimization to worry about.

197g (approved):
LGTM. On further thought we probably should not try to guess some arbitrary part of the image matrix to show. If anything we could provide a debug adapter (see: `OsStr::display`) that provides some configuration. Defaulting to not display pixels seems perfectly fine. (Edit: having one or more adapters could also solve the conundrum of the buffer type not having an `Deref` constraint to inspect the pixels. Different display adapters could add different trait bound to their constructor.)

PR #2671, opened 2025-11-26
Round to nearest when alpha-blending colors

### LINKED ISSUE #2533: Blend rounding of zero
In image 0.25.6

When calling overlay with a blank `DynamicImage::new_rgba8` the call to blend does some expensive computation that introduces some rounding error.

Consider this code:
```
use image::{DynamicImage, GenericImage, GenericImageView, RgbaImage, Rgba, imageops};

fn main() {
    let mut image = DynamicImage::ImageRgba8(RgbaImage::new(3, 3));
    let test_color = Rgba([45, 57, 82, 200]);
    image.put_pixel(1, 1, test_color);

    //Create padded image
    let mut padded = DynamicImage::new_rgba8(5, 5);
    imageops::overlay(&mut padded, &image, 1, 1);

    // Extract patch using view().to_image()
    let patch = padded.to_rgba8().view(1, 1, 3, 3).to_image();
    let result = patch.get_pixel(1, 1);

    //Compare values
    println!("Expected: {:?}", test_color);
    println!("Got:      {:?}", result);
}
```

For me this produces:
> Expected: Rgba([45, 57, 82, 200])
> Got:      Rgba([44, 56, 82, 200])

It appears `imageops::overlay` is calling `impl<T: Primitive> Blend for Rgba<T>` which special cases where the `other` pixel is zero but not when `self` is_zero, which means it needs to perform a bunch of calculation, and ends up rounding the values.

RunDevelopment:
fixes #2533

Alpha-blending itself was done correctly, but it used a truncating cast from f32 to the subpixel type. This caused systematic errors resulting in slightly wrong colors.

This PR fixes this by rounding to the nearest integer if the subpixel type is an integer.

Note: For `test_image_alpha_blending` I double-checked that the new blend colors are correct by calculating them all by hand. The new colors are indeed rounded correctly.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -830,8 +830,7 @@ fn from_f32_rounded<T: Primitive>(x: f32) -> T {
     // We assume that integers perform truncation when casting from float.
     // With this assumption, rounding can done simply adding 0.5 before the cast.
     // Of course, adding 0.5 must ONLY be done for integer types.
-    let round = if is_integer::<T>() { 0.5 } else { 0.0 };
-    NumCast::from(x + round).unwrap()
+    NumCast::from(if is_integer::<T>() { x + 0.5 } else { x }).unwrap()
 }
 
 impl<T: Primitive> Blend for LumaA<T> {


src/color.rs
```diff
@@ -821,6 +822,18 @@ pub(crate) trait Blend {
...
+/// Converts a f32 to a primitive type T with rounding when T is an integer.
+///
+/// E.g. 3.6 -> 4 for u8, but 3.6 -> 3.6 for f32.
+#[inline]
+fn from_f32_rounded<T: Primitive>(x: f32) -> T {
+    // We assume that integers perform truncation when casting from float.
+    // With this assumption, rounding can done simply adding 0.5 before the cast.
+    // Of course, adding 0.5 must ONLY be done for integer types.
+    let round = if is_integer::<T>() { 0.5 } else { 0.0 };
+    NumCast::from(x + round).unwrap()
```
197g: Adding `0.0` is not a no-op in float. I think we want:

```suggestion
    NumCast::from(if is_integer::<T>() { x + 0.5 } else { x }).unwrap()
```

197g (changes requested):

197g (approved):

PR #2673, opened 2025-11-27
Fixed TGA crash for images where indices have more bits than mapped colors

### LINKED ISSUE #2598: Panic in Tga decoder for color-mapped images
```
thread 'check_regressions' panicked at /home/andreas/code/projects/image/src/codecs/tga/decoder.rs:377:44:
range end index 1052672 out of range for slice of length 789504
```

## Expected

An error, the file contains no actual pixel data. But really an equivalent file with pixel data should work just fine. Alternatively, a clean error as `ImageError::Unsupported`.

## Actual behaviour

Panic. The TGA decoder relies on an assumption that the orginal color type has a smaller pixel depth than any mapped color type. This is untrue for color-mapped images in general. In an arguably degenerate case it is permissible to specify the bit-depth of the indices as `32` even when the data only contains `24` bits, i.e. RGB8.

## Reproduction steps

[tga-issue.zip](https://github.com/user-attachments/files/22248624/tga-issue.zip)

RunDevelopment:
fixes #2598

This fixes the crash by returning an error instead.

### DIFF
--- a/src/codecs/tga/decoder.rs
+++ b/src/codecs/tga/decoder.rs
@@ -383,7 +383,7 @@ impl<R: Read> ImageDecoder for TgaDecoder<R> {
         //
         // We currently assume that the indices take less space than the
         // pixels they encode, so it is safe to read the raw data into `buf`.
-        if self.raw_bytes_per_pixel > self.color_type.bytes_per_pixel() as usize {
+        if self.raw_bytes_per_pixel > self.color_type.bytes_per_pixel().into() {
             return Err(ImageError::Unsupported(
                 UnsupportedError::from_format_and_kind(
                     ImageFormat::Tga.into(),


src/codecs/tga/decoder.rs
```diff
@@ -381,9 +381,22 @@ impl<R: Read> ImageDecoder for TgaDecoder<R> {
 
         // Decode the raw data
         //
-        // We have already checked in `TgaDecoder::new` that the indices take less space than the
+        // We currently assume that the indices take less space than the
         // pixels they encode, so it is safe to read the raw data into `buf`.
+        if self.raw_bytes_per_pixel > self.color_type.bytes_per_pixel() as usize {
```
197g:
```suggestion
        if self.raw_bytes_per_pixel > usize::from(self.color_type.bytes_per_pixel()) {
```

Were possible I'd avoid `as` to clarify that this is never lossy due to the underlying type (`u8`).

197g (commented):
nit, otherwise LGTM

src/codecs/tga/decoder.rs
RunDevelopment: I used `.into()` instead since it's shorter. Hope that works for you too.

PR #2674, opened 2025-11-27
Minor optimization for TGA's BGR -> RGB conversion

RunDevelopment:
We know that `pixels.len()` is visible by `color_type.bytes_per_pixel()`, so we can use `chunks_exact_mut`. This has the effect that LLVM can optimize away the bounds checking `swap` does, since it now knows that chunks at least 3 elements. This removes a branch from the inner loop, so it should be slightly faster now. Probably not by much, since the branch is easy to predict. I haven't benchmarked it, I just looked at the assembly.

### DIFF
--- a/src/codecs/tga/decoder.rs
+++ b/src/codecs/tga/decoder.rs
@@ -307,7 +307,7 @@ impl<R: Read> TgaDecoder<R> {
         // We only need to reverse the encoding of color images
         match self.color_type {
             ColorType::Rgb8 | ColorType::Rgba8 => {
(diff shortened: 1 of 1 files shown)


197g:
Seems plausible, well spotted. As a side effect when we increase the rust-version enough there should be a replacement of `chunks_exact_mut` with `as_chunks_mut` when the argument is in fact a constant. That avoids `try_into().unwrap()` sequences, we will want to review other occurrences for optimization potential. So this makes it easier to search for.

PR #2678, opened 2025-11-27
Pick ICO entries in order

### LINKED ISSUE #1739: Incorrect decoding of .ico file
This happens in image v0.24.2 from crates.io

## Expected

GIMP renders the .ico file as it seems to be intended, with nice anti-aliasing. Here it is converted to PNG: orig

imagemagick also agrees with this rendering, with *very* minor differences.

## Actual behaviour

image-rs seems to remove the anti-aliasing and make the image a lot more jagged: image

## Reproduction steps

The code I've used for decoding the .ico and converting it to PNG is:

```rust
use std::error::Error;

    let input = std::env::args().nth(1).unwrap();
    let output = std::env::args().nth(2).unwrap();
    
    let img = ImageReader::open(input)?.decode()?;
    img.save(output)?;
    Ok(())
}
```
Invoked like this: `image-convert input.ico output.png`

GIMP version 2.10.32

imagemagick version 6.9.10.23

RunDevelopment:
ICO image files can contain any number of entries (BMP or PNG images). This is a problem for us, because we need to pick one of those entries to display. Luckily, it's pretty obvious which one to pick in most cases. Most ICO files have exactly one entry with the greatest size and greatest number of bits per pixel, ergo, the best quality.

Unfortunately, there isn't always *one* best entry. There might be multiple entries with the same size, same bpp, and even same number of bytes on disk. So. Which one would you pick?

Since there is no obvious way to pick one entry out of multiple equally good ones, programs do it differently. From my testing, they behave as follows:

1. Some pick the entry that appears first in the list of entries. Those are Chrome and MS Paint.
2. Some pick the entry that appears last in the list of entries. Those are Firefox and Windows Photo Viewer.
3. Windows explorer will pick the one that appears last in the list. Unless it needs wants to display a 16x16 icon, then it will display the one that appears first in the list.

And then there's this crate. Our ICO decoder will pick the best entry that appears first in the list, unless the very last entry of the list is one of the best entries.

So I changed the behavior of our decoder to always pick the one that appears **first** in the list of entries to be consistent with MS Paint. And Chrome, I guess. Again, there is no real right way to do this, but I would say that what we did before was definitely wrong, even if it arguably is a degenerate case.

For tests, I create a new ICO file with 2 entries of the same size and bpp. The first entry is green and the second is red. This makes it easy to see which one is picked. Here it is in GIMP:

image

---

resolves #1739

The last 2 icons attached in #1739 that were still loaded incorrectly were icons that have this exact problem. Here are the different entries in GIMP:
image

As you can see, the icon has 2 layers with the same bpp for each size. I suspect that this is an error by whoever created the icon, because the second entry for each size uses reduced colors (while still being encoded as 32 bpp).

In any case, our ICO decoder now decodes all icons from #1739 correctly, so the issue is fixed.

### DIFF
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -197,20 +197,16 @@ fn read_entry<R: Read>(r: &mut R) -> ImageResult<DirEntry> {
 /// do exist. Since we can't make an educated guess which one is best, picking
 /// the first one is a reasonable default.
 fn best_entry(entries: Vec<DirEntry>) -> ImageResult<DirEntry> {
-    let mut best = *entries.first().ok_or(DecoderError::NoEntries)?;
-    let mut best_score = (0, 0);
-
-    for entry in entries {
-        let score = (
-            entry.bits_per_pixel,
-            u32::from(entry.real_width()) * u32::from(entry.real_height()),
-        );
-        if score > best_score {
-            best = entry;
-            best_score = score;
-        }
-    }
-    Ok(best)
+    entries
+        .into_iter()
+        .rev() // ties should pick the first entry, not the last
+        .max_by_key(|entry| {
+            (
+                entry.bits_per_pixel,
+                u32::from(entry.real_width()) * u32::from(entry.real_height()),
+            )
+        })
+        .ok_or(DecoderError::NoEntries.into())
 }
 
 impl DirEntry {


197g:
Great investigation.

We can definitely do this in the `1.0` switch. Agreeing with Chrome and imagemagick is very reasonable. I wonder if it were worth it to have an option on `BmpDecoder` for the selection algorithm? Well I'm mostly thinking of the 16x16 thumbnail here—but then we could also say to use the `AnimationDecoder` (implying we fix it to just mean image sequence, with no implied order relationship but rather an explicit one for formats that are animations).

LGTM but before merging I'd like to hear you opinion on the thought.

RunDevelopment:
As I see it, icons with multiple entries being the same size and bpp are a degenerate case. So I don't think it's worth having a selection strategy option just for them. AFAIK, icons only have entries with the same size but different bpp to support older operating systems ([see wikipedia](https://en.wikipedia.org/wiki/ICO_(file_format)#History)).

I'd rather implement the decoder implements `AnimationDecoder` (or similar for sequences), so users can get all entries. This would enable more use cases and uses could use whatever strategy they want. They could even display all entries like GIMP for example.

src/codecs/ico/decoder.rs
```diff
@@ -191,13 +191,14 @@ fn read_entry<R: Read>(r: &mut R) -> ImageResult<DirEntry> {
...
-
-    let mut best_score = (
-        best.bits_per_pixel,
-        u32::from(best.real_width()) * u32::from(best.real_height()),
-    );
+///
+/// If two entries have the same color depth and size, pick the first one.
+/// While ICO files with multiple identical size and bpp entries are rare, they
+/// do exist. Since we can't make an educated guess which one is best, picking
+/// the first one is a sane default.
```
fintelia: nit: 
```suggestion
/// the first one is a reasonable default.
```

src/codecs/ico/decoder.rs
fintelia: If we're changing this function, we might as well switch it to using `max_by_key`:
```suggestion
    entries
        .into_iter()
        .rev() // ties should pick the first entry, not the last
        .max_by_key(|entry| (
            entry.bits_per_pixel,
            u32::from(entry.real_width()) * u32::from(entry.real_height()),
        ))
        .ok_or(DecoderError::NoEntries)
```

fintelia (approved):
This change seems fine to me

src/codecs/ico/decoder.rs
RunDevelopment: I also tried a version with `max_by_key`, but it didn't work, because it always gave the last one. I completely forgot about `rev` 😓

RunDevelopment:
> Would it be possible for you to start the Changelog for `1.0` and note the change in ICO down

What does "start" mean exactly? Do you only want me to add a single entry for the ICO change or do you want me to do the changelog up to new including the ICO change (so basically the whole changelog sans future PRs)?

I'm asking because you guys don't squash commits on merge, which makes the commit history on `main` a mess to decipher. Any unimportant rustfmt, clippy fix, and review suggestion commit preserved forever. How do you guys even make changelogs normally? Do you look at the list of merged PRs instead of the commit history?

197g:
I was just referring to the `Changelog.md` file, don't worry about it. When I write changelogs, I go through the auto-created merges for their issue number (`git log --first-parent` works for both rebase and merge strategies) and grab the title / contents from the Github API. Even well-done commit notes are summarized as how-and-why but that is not optimal as the user facing documentation anyways, we discuss this in the PR thread if anywhere.

PR #2672, opened 2025-11-26
Decoder metadata interface

197g:
See https://github.com/image-rs/image/issues/2245, the intended `ImageDecoder` changes.

This changes the `ImageDecoder` trait to fix some underlying issues. The main change is a clarification to the responsibilities; the trait is an interface from an implementor _towards the `image` library_. That is, the protocol established from its interface should allow _us_ to drive the decoder into our buffers and our metadata. It is _not_ optimized to be used by an external caller which should prefer the use of `ImageReader` and other inherent methods instead.

This is a work-in-progress, below motivates the changes and discusses open points.

- `ImageDecoder::peek_layout` encourages decoders to read headers after the constructor. This fixes the inherent problem we had with communicating limits. The sequences for internal use is roughly:
  ```rust
  let mut decoder = …;
  decoder.set_limits(…); // Other global configuration we have?

  { // Potentially multiple times:
    let layout_info = decoder.peek_layout()?;
    let mut buffer = allocate_for(&layout_info);
    decoder.assign_metadata()?;
    decoder.read_image(&mut buffer)?;
  }

  // … for sequences, start again from `peek_layout()`
  ```
- `ImageDecoder::read_image(&mut self)` no longer consumes `self`. We no longer need the additional `boxed` method and its trait work around, the trait is now dyn-compatible.

## Discussion

- [x] Maybe ~~`init`~~ `peek_layout` should return the full layout information in a single struct. We have a similar open issue for `png` in its own crate, and the related work for `tiff` is in the pipeline where its [`BufferLayoutPreference`](https://docs.rs/tiff/latest/tiff/decoder/struct.BufferLayoutPreference.html) already exists to be extended with said information.
  - [x] ~~Review limits and remove its size bounds insofar as they can be checked against the communicated bounds in the metadata step by the `image` side.~~ see: #2709, #2708
  - Idea: If a decoder supports builtin transforms (e.g. YCbCr -> Rgb conversion, grayscale, thumbnailing) that are more efficient than post-processing then there could be a negotiation phase here where information is polled twice / multiple times by different methods. The design should leave this negative space to be added in `1.1`, but it's not highly critical.
- [x] Fix the sequence decoder to use the new API
- [x] Tests for reading an image with `read_image` then switching to a sequence reader. But that is supposed to become mainly an adapter that implements the iterator protocol.
- [x] Remove remnants of the dyn-compatibility issue.
[...]

PR #2681, opened 2025-11-30
Rip out ImageDecoderRect

197g:
The argument is summarized in #2355 already. The details about the misfit is expanded on in <https://github.com/image-rs/image/pull/2481#discussion_r2138605372>.

	The main part of its implementation is spent on policy working around
	the underlying decoder not really cropping by itself and still yielding
	all the data. In this sense it is more like the antipole to what the
	trait really should be. That is of course part of the issue with the
	trait; none of our implementations provide the real benefit over the
	naive 'throw-away-data' strategy and that happens by them using what is,
	really, a pitfall to be quite frank. You really should not be using this
	except in the circumstance that you can't avoid it. And performant
	decoding, in general, should not work by 'scanlines' (apart from png
	where the filter/compression depends on this concept and hence the
	previous line is always necessary).

This held true for both of the removed decoder bindings. In BMP it is quite suspiciously so as that format is meant to be mmap'd—but the called utility reads every single byte regardless, wasting lots and lots of energy. We did not get complaints about such behavior but do get some for inefficient other operations. Let's face the reality that most of our users call into ImageReader and do not interact with the trait system directly—the static extension point also make it impossible to fit into as an optional path way into ImageReader since it is built around boxing and type erasing the underlying decoder early.

Closes: #2355 

----

Restricting the viewport before a readout is one of the interactions I had in mind for the reader type in #2672. I'll come up with a defaulted method in `ImageDecoder` instead. Since the trait is misbehaved regardless that is a separate matter from removing it, to me.


fintelia (approved):
Sounds good to me.

And if we do decide on something like a default-implemented method on `ImageDecoder` for adding back similar functionality, it won't be semver breaking so it is fine to land post-1.0 release.

PR #2685, opened 2025-12-04
Document removal of ImageDecoderRect

Lanzaa:
ImageDecoderRect was removed in #2681 per #2355. Update the README and CHANGES to reflect this.

Oddly enough, I was looking at the ImageDecoderRect stuff yesterday and this tripped me up a bit. Feel free to change/toss these changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@
 
 ## Changes
 
+### Version Next
+
+Breaking Changes:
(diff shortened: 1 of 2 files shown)


197g (approved):
Thanks for doing the outdated Readme mentions, too. As a heads-up for the release, I _wanted_ to accompany the breaking things with migration information (at least something, or a link to why we won't do them). This is already enough to remind us, thanks.

PR #2689, opened 2025-12-11
Adjust bounds on from_raw_rgb

197g:
These new bounds ensure that only our own pixel types can be used. That prepares changing our approach of algorithms such that we can optimize for individual cases without falling prey to needs for specialization.

More context can be found here:

<https://github.com/image-rs/image/pull/2596#issuecomment-3643133007>

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1408,9 +1408,8 @@ impl<P: Pixel> ImageBuffer<P, Vec<P::Subpixel>> {
 
 impl<S, Container> ImageBuffer<Rgb<S>, Container>
 where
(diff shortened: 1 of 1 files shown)


197g:
Tests are failing from the fix here: https://github.com/etemesi254/zune-image/pull/307

We got the same test data as there.

fintelia:
Created https://github.com/image-rs/image/pull/2692. I actually think these bounds changes may not end up being needed

197g:
We can replace `PixelWithColorType` by `Pixel` afterwards but I prefer that way to express the bound over the rather indirect additional constraints on the subpixel either way :)

PR #2692, opened 2025-12-13
Make the Primitive trait sealed

fintelia:
By sealing the `Primitive` trait, we gain flexibility for adding internally visible functionality on the trait.

For instance, https://github.com/image-rs/image/commit/123eef91b915ccc1ff661a8cf56cae2052d2bd27 demonstrates how these changes allows us to specialize the implementation of the `ImageBuffer<Rgb<u8>>` conversion methods from/to BGR order to use std::simd.

I believe this same strategy should enable performant implementations of many other methods on `ImageBuffer` without needing further changes to the API.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -33,9 +33,28 @@ impl EncodableLayout for [f32] {
     }
 }
 
+mod sealed {
+    pub trait PrimitiveSealed: Sized {}
+}
+
+impl sealed::PrimitiveSealed for usize {}
+impl sealed::PrimitiveSealed for u8 {}
+impl sealed::PrimitiveSealed for u16 {}
+impl sealed::PrimitiveSealed for u32 {}
+impl sealed::PrimitiveSealed for u64 {}
+impl sealed::PrimitiveSealed for isize {}
+impl sealed::PrimitiveSealed for i8 {}
+impl sealed::PrimitiveSealed for i16 {}
+impl sealed::PrimitiveSealed for i32 {}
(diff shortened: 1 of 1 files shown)


197g:
I'm okay with this under the assumption that we'll make `Pixel` clearly geared towards the simplified `ImageBuffer` (or `PixelImage` as proposed) buffer representation (as a superset but relative of `DynamicImage`) and `Primitive` not being used except for the `Subpixel` bound. Maybe it will show itself to be more general but I'd not want to design around such generalization preemptively. (Also, sealing this would be more contentious then). Can we rename it appropriately (e.g. `ChannelPrimitive`)? Or we punt the renaming to when we rename `Subpixel` as well cause that name is rather nondescript as well.

197g:
I've been pondering the role of `GenericImage` in all of this. We definitely have multiple different types that implement it and that should be preserved (running `imageops` over `SubImage` in particular) yet it is not universal by any means. The `Pixel` associated type really does not allow it to be.

Sealing that as well might have similar future-proofing advantages to this one. Although our long-standing issue of `copy_from` could become more performant without it, by giving having an exhaustive variation of `copy_into` methods on `GenericImageView` and making each `copy_from` impl dispatch into the appropriate one, ther uses may not be as lucky and/or it is a lot of overhead. It would be odd to add a similar variety of methods to the traits for every instance that may come up in the future; and impls be absurdly slow if they don't tick the available boxes.

PR #2694, opened 2025-12-14
Add test case for guess_format function

mina86:


src/io/free_functions.rs
```diff
@@ -162,6 +162,50 @@ pub(crate) fn guess_format_impl(buffer: &[u8]) -> Option<ImageFormat> {
...
+                None
+            }
+            Err(_) => {
+                eprintln!("{disppath}: unrecognised extension, ignoring");
+                None
+            }
+        }
+    }
+}
+
```
197g: nit: please move this all the way to the bottom of the file. I'd also appreciate a more descriptive name such as `guess_format_agrees_with_extension`.

197g (commented):
Test coverage is always good.

src/io/free_functions.rs
mina86: Done.

PR #2695, opened 2025-12-15
Simplify sub-image and rectangular selection signatures

197g:
Currently we use 4 parameters in many functions that take a rectangle selection. This is barely readable and we can not provide any convenience constructors or methods to consume this. Yet, validation is important in a few places. I've been meaning to propose this for some time.

I'm very slightly favoring the current rectangle struct over `Rect { x: Range<u32>, y: Range<u32>}`. We need to do some validation either way and this can represent the overlay of an image onto another without overflows.

### DIFF
--- a/src/math/rect.rs
+++ b/src/math/rect.rs
@@ -1,4 +1,7 @@
+use core::cmp;
 use core::ops::Range;
+
+use crate::GenericImageView;
 
 /// A Rectangle defined by its top left corner, width and height.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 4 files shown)


kornelski:
`Range` is an unusual representation. Typically I've seen `Rect { origin: Point, size: Size }`. Reusing std `Range` saves defining more types, but it's less convenient when having x,y,width,height (I have the same beef with slicing `v[start .. start + len]`).

src/math/rect.rs
```diff
@@ -6,8 +11,76 @@ pub struct Rect {
...
+    /// assert_eq!(rect.height, 20);
+    /// ```
+    pub fn from_xy_ranges(x: Range<u32>, y: Range<u32>) -> Self {
+        Self {
+            x: x.start,
+            y: y.start,
+            width: x.end.saturating_sub(x.start),
+            height: y.end.saturating_sub(y.start),
+        }
+    }
```
197g, line 37: Converting range to `(start,len)` is much more straightforward than the other way. Here I've chosen the semantics to be empty when `end <= start` in a way that preserves the origin but a (debug) assertion might also be handy for `end < start`. I'd keep the saturating semantics in any case, you'll _never_ want this to wrap.

fintelia: This seems good to me. I'd be OK with a assertion for `end  >= start`, either debug only or enabled for release builds too

PR #2697, opened 2025-12-17
Updated blur parameter

awxkee:
Here, the blur parameter was changed to radius, and some documentation was added. I believe we discussed this in #2496 and postponed it until a breaking release.


197g (approved):

197g:
So we did, thanks for ticking a box in the 1.0 progress.

PR #2701, opened 2025-12-20
Add {ImageBuffer,DynamicImage}::crop_in_place

197g:
Similar to `flipv_in_place` this never needs to grow the underlying container. Our behavior of clipping the crop selection to the images dimensions ensures that the target dimensions are always smaller.

As discussed, besides the obvious performance advantages over other forms of cropping into a new allocation this would also be relevant to #2672 .

As a complementary method introduce `ImageBuffer::shrink_to_fit`.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -996,6 +996,9 @@ where
     /// allocation takes place. The width and height of this buffer are adjusted as part of the
     /// operation. The selection is shrunk to the overlap with this image if it is out of bounds.
     ///
+    /// The pixel buffer is *not* shrunk and will continue to occupy the same amount of memory as
+    /// before. See [`ImageBuffer::shrink_to_fit`] if the container is a [`Vec`].
+    ///
     /// # Examples
     ///
     /// ```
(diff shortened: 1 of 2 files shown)


src/images/buffer.rs
```diff
@@ -989,6 +989,64 @@ where
...
+        for y in 0..selection.height {
+            let sy = selection.y + y;
+            let source = self.pixel_indices_unchecked(selection.x, sy).start;
+            let dst = y as usize * rowlen;
+            self.data.copy_within(source..source + rowlen, dst);
+        }
+
+        self.width = selection.width;
+        self.height = selection.height;
+    }
```
197g, line 1052: Note: we can't shrink the container here as this is generic over it. We could do it in `DynamicImage` on top of the original transform to avoid issues as https://github.com/rust-lang/rust/issues/120091 even though additional conditions may be necessary for the behavior to be considered particular surprising (i.e. the linked issues is because an _internal_ use of vec reserve/truncate leaves it with a lot of capacity, an analogy would thus be an operation that internally crops, but for instance `ImageReader` #2672 could truncate itself quite well).

197g: Let's at least introduce the necessary method for this in the same patch.

PR #2704, opened 2025-12-21
Replace custom HDR reference test with TIFF references and add new image

mstoeckl:
In order to make future improvements to the HDR encoder and decoder easier, I made a new reference image covering a wide range of exponents and color values*. I am licensing it to match `image` (`Apache 2.0 OR MIT`); and am also offering it under CC-0.

*This could be useful when testing improvements to Rgbe8Pixel::to_hdr -- e.g. there are a few ways to speed it up and make it more certainly deterministic that I am looking into -- and when comparing `image`'s output with other decoders.

Also, because the `raw` output of `check_hdr_references()` was somewhat awkward to work with and inspect, 
I've modified the `check_references()` test to also support TIFF reference images, rendering  `check_hdr_references()` unnecessary.

I used `tiffcp -L -c zip:3:p9` to manually recompress the new reference images.


mstoeckl:
~Thinking about this a bit more, the new image appears somewhat redundant with what `hdr::encoder::to_rgbe8_test` is checking and doesn't seem likely to catch regressions that `to_rgbe8_test` would not already catch.~

~Since the change to replace `check_hdr_references()` was motivated in large part by the new image, I am closing this MR entirely.~ (Never mind, the roundtrip testing of to_rgbe8_test is _not_ really comparable with a fixed image reference test.)

tests/reference_images.rs
```diff
@@ -27,28 +28,29 @@ where
...
+#[cfg(all(feature = "png", feature = "tiff"))]
 #[test]
 fn render_images() {
     process_images(IMAGE_DIR, None, |base, path, decoder| {
         println!("render_images {}", path.display());
         let img = match image::open(&path) {
-            Ok(DynamicImage::ImageRgb32F(_) | DynamicImage::ImageRgba32F(_)) => {
-                println!("Skipping {} - HDR codec is not enabled", path.display());
-                return;
-            }
```
197g, line 51: nit: I'd rather we not skip all the tests when `tiff` is not enabled. How about we keep the branch but skip it when `feature = tiff`?

mstoeckl: Updated.

Issue #1894, opened 2023-03-17
Investigation: animation / sub-images


fintelia:
Various image formats have ways of representing multiple images within a single file, either as frames of an animation, different sizes of the same image, and/or multiple distinct images. Currently, we only exposes the [`AnimationDecoder`](https://docs.rs/image/latest/image/trait.AnimationDecoder.html) that lets you turn an image decoder into a sequence of Rgba8 frames, but doesn't have any other functionality for working with files containing multiple sub-images. Given the prevalence of this capability, it may make sense to make it part of the core [`ImageDecoder`](https://docs.rs/image/latest/image/trait.ImageDecoder.html) trait.

| Format | Meaning of sub-images |
|--------|--------|
| PNG | animation |
| JPEG | n/a |
| GIF | animation |
| BMP | n/a |
| ICO | multiple sizes of the same image | 
| TIFF | multiple distinct images _or_ multiple sizes of the same image  |
| WebP | animation |
| AVIF | animation |
| PNM | n/a |
| DDS | multiple images layers with the same format and dimensions |
| TGA | n/a |
| OpenEXR | multiple distinct images each potentially with multiple sizes |  
| farbfeld | n/a |

<sup>(Please let me know if you notice any mistakes in the table. I'm not super familiar with most of these formats.)</sup>

darksylinc:
I was looking into this and I find the current image crate too restrictive.

I can provide more info about some mistakes or even currently-unsupported formats:

## DDS

Textures can have layers (e.g. 2D arrays, cubemaps and cubemap arrays).
Textures can have depth (e.g. 3D textures).

From what I just described so far, it seems like, layers == depth. However there's a subtle difference: mipmapping.

Mipmaps means DDS textures **can have different resolutions** (e.g. 256x256 128x128 64x64 ... 1x1). In this way it's the same as ICO in the sense that they're meant to be multiple resolutions of the same image (although not necessarily, some games intentionally exploit this functionality and place separate images into each mip level).

2D Layer do not reduce in mipmaps. e.g. 256x256x4 texture array means there are four 256x256 textures. Each can have up to 9 mipmaps, for a total of 36 different images (4 layers x 9 mipmaps). All layers have the same resolution.

3D textures however do have mipamps, e.g. 256x256x4 becomes 128x128x2, 64x64x1, 32x32x1 ... 1x1x1. That's a total of 9 images (instead of 36).

[...]

PR #2706, opened 2025-12-23
Support generic image stack decoding + tiff stack decoding

### LINKED ISSUE #2064: More convenient way to work with multipage /-layer tiffs
I would like to be able to read all images of a tiff file.

My specific use case for this functionality is writing code working with multiple instances of `DynamicImage`. Unfortunately there is no easy way to access `next_image` of  `tiff::decoder::Decoder` without heavily duplicating the wrapper code.

Following is the closest I was able to get:
```rust
fn tiff_load_layers<R: io::BufRead + io::Seek>(r: &mut R) -> Result<Vec<image::DynamicImage>, Box<dyn Error>> {
  let mut layers: Vec<image::DynamicImage> = Vec::new();
  let mut raw = tiff::decoder::Decoder::new(r)?;

  loop {
    let layer = tiff_read_image(&mut raw)?;
    layers.push(layer);

    if !raw.more_images() {
      break;
    }

    raw.next_image()?
  }

  Ok(layers)
}

fn tiff_read_image<R: io::BufRead + io::Seek>(raw: &mut tiff::decoder::Decoder<R>) -> Result<image::DynamicImage, Box<dyn Error>> {
  let (width, height) = raw.dimensions()?;
  let color_type = raw.colortype()?;
  let pixels = raw.read_image()?;

  // this code is likely `DynamicImage::from_decoder`
  let image = match (color_type, pixels) {
    (tiff::ColorType::RGB(8), tiff::decoder::DecodingResult::U8(pixels)) =>
      image::ImageBuffer::from_vec(width, height, pixels).map(image::DynamicImage::ImageRgb8),

    _ => None,
  };

  match image {
    Some(image) => Ok(image),
    None => Err(...),
  }
}
```

soraxas:
addresses #1894, closes #2064

## Features
- separate animation into a more generic image stack
- implement image stack decoding for tiff

## Open for discussion
- Some of the api had changed.
- I've rename the struct `Frames` to `Stack` as the name is more generic (i.e. frame tends to imply *width*, *height* and *data type* are the same, but it won't be true in all format, e.g., tiff)
- I've replaced the `AnimationDecoder` with the more generic `ImageStackDecoder` (need to import that trait instead), which turns into an iterator that output some generic image-like object
  - e.g. all previous implementation now returns `AnimationFrame` which is a container that contains `Frame<...>` and a `delay` float
  - The newly implemented `tiff` turns `Frame<DynamicImage>`, because tiff can contains frames of different size.

--------

Added an example for reading `tiff` image stack:
```sh
$ cargo run --example tiff_img_stack test.tiff
[examples/tiff_img_stack.rs:27:9] i = 0
Frame 0: dimensions: (1620, 1080), color: Rgb8
[examples/tiff_img_stack.rs:27:9] i = 1
Frame 1: dimensions: (2000, 4800), color: Rgb8
[examples/tiff_img_stack.rs:27:9] i = 2
Frame 2: dimensions: (1958, 849), color: Rgb8
```

-------

## Summary 
- All iterable decoder is under `ImageStackDecoder<...>`, produce -> iterable `Stack<...>`
  - **Generic image list**: `struct Stack<Frame<I>> { ... }` contains a list of frame w/ `x`, `y` offset + a generic Image buffer `I`
    - e.g. for gif, `I` is rbg image
    - for tiff, `I` is `DynamicImage`
  - **Animation**: `struct Stack<AnimationFrame> { ... }` where animation frame contains a float `delay` + a `Frame<I>` (e.g. gif)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,6 +126,10 @@ path = "benches/blur.rs"
 name = "blur"
 harness = false
 
+[[example]]
+name = "tiff_img_stack"
+required-features = ["tiff"]
+
 # because of https://github.com/image-rs/image/pull/2583
 # TODO: remove when shipping the next major release after 0.25
 [package.metadata.cargo-semver-checks.lints]


soraxas:
Can we have some indication as to whether this is a right direction to support generic image stack? Thanks

fintelia:
There's ongoing changes to `ImageDecoder` in #2672 to have it handle animation / image sequences but the PR is still in flux. Once that is complete we'll have a better sense of where things stand. My hope is that we can expose a `dyn`-compatible trait for animation decoding (which `AnimationDecoder` currently isn't)

197g:
In my view, there's a bunch of (weakly motivated) complexity here. Making the frame type generic for instance makes me wonder what other instantiations there should be, on what groups they are related, beyond some getter-style methods there is little that's there. (#2672 noted as it follows through more consequently on separating the sequence metadata from the image color data). Turning the frame into `DynamicImage` could be an orthogonal start.

Also visible on `trait ImageStackDecoder`: that interface is so generic, it's hard to distinguish from `IntoIterator<C>`. It returns `Stack<'a, C>` which is itself a lean wrapper around `Box<dyn Iterator<Item=Result<C, _>> + 'a>`. It does not add much in terms of data relationship, nor methods that would be domain specific to the image use case. The problem with such indirection is that, due to absence of specialization, it gets much harder to optimize effectively for anything all-quantized over them; hence last year's trend of toning that way down in some core concepts. I'm currently rather wary of adding new ones without data backing up a very concrete use case that they enable. (Expanding existing traits with new capabilities is much preferred where possible).

197g:
Feel free to reopen with some of the complexity concers addressed but I don't see this moving forward in the current API shape.

PR #2707, opened 2025-12-23
Optimizations and simplifications for .hdr decoder

mstoeckl:
This PR includes two changes to the .hdr decoder:
* Replacing the implementation of `Rgbe8Pixel::to_hdr()` with direct construction of floating point values using f32::from_bits. This is significantly faster.
* Inlining the function `read_image_transform()` that was only used in `read_image()`, and directly writing to `buf: &mut [u8]` instead of writing to a temporary array first.

Notes:
* The previous implementation's use of [`f32::exp2`](https://doc.rust-lang.org/std/primitive.f32.html#method.exp2) was _technically_ permitted to be implemented in a nondeterministic way. In practice, math libraries should always produce exact output for `f32::exp2` when given integer arguments in -149..=127.
* Rust documents that `f32` is IEEE 754 binary32, so constructing floats from their binary representation will work on all platforms.
* Somewhat surprisingly, the release build unrolls and vectorizes the `to_hdr()` conversion loop for me on x86_64, eliminating the branch.
* It's possible to speed up decoding of the per-component RLE images by making the calls to `decode_component()` write to four separate scanline buffers, and running a single loop that scans these  buffers in parallel and writes the decoded scanline to  `buf`. This ends up saving ~10% of decoding time, but makes the code noticeably more complicated.

### DIFF
--- a/src/codecs/hdr/decoder.rs
+++ b/src/codecs/hdr/decoder.rs
@@ -119,8 +119,6 @@ const SIGNATURE_LENGTH: usize = 10;
 #[derive(Debug)]
 pub struct HdrDecoder<R> {
     r: R,
-    width: u32,
-    height: u32,
     meta: HdrMetadata,
 }
 
@@ -251,8 +249,6 @@ impl<R: Read> HdrDecoder<R> {
         Ok(HdrDecoder {
             r: reader,
 
-            width,
-            height,
             meta: HdrMetadata {
                 width,
                 height,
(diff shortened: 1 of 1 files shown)


src/codecs/hdr/decoder.rs
```diff
@@ -301,14 +272,26 @@ impl<R: Read> ImageDecoder for HdrDecoder<R> {
...
+        // Don't read anything if image is empty
+        if self.meta.width == 0 || self.meta.height == 0 {
+            return Ok(());
+        }
 
-        for (i, Rgb(data)) in img.into_iter().enumerate() {
-            buf[(i * 12)..][..12].copy_from_slice(bytemuck::cast_slice(&data));
+        let mut scanline = vec![Default::default(); self.meta.width as usize];
+
+        let line_bytes = self.meta.width as usize * 12;
```
197g:
```suggestion
        const PIXEL_SIZE: usize = 12;
        let line_bytes = self.meta.width as usize * PIXEL_SIZE;
```

Since the two occurrences are no longer immediately following each other it may be nicer to connect the dots between the two `12` constants if we give it a name (second one in `chunks_exact_mut`).

197g (approved):
Just a readability nit, good reduction in complexity and speed otherwise.

src/codecs/hdr/decoder.rs
mstoeckl: Good point, updated. (Also using `size_of`.)

PR #2709, opened 2025-12-28
Replace `ImageDecoder::set_limits` with `ImageDecoder::set_allocation_limit`

fintelia:
The goal of this PR is to give the `Limits` struct a single purpose of controlling limits for `ImageReader` while the lower-level API directly expects the caller to check dimensions (and output buffer sizes) so the only remaining limit is memory allocation.

This also makes the concept of strict limits (if we add any) clearer. Now the `ImageReader` is responsible for figuring out which limits can be supported and directly returning an error if necessary. No more need for every decoder to check its supported limits against the requested ones.

Another inspiration for this PR was the idea from #2708 of having a shared `Arc<AtomicU64>` for the allocation limit. In particular, in a followup we could switch things so that `set_allocation_limit` took an atomic limit, while having the higher-level `Limits` object keeps its existing non-atomic `max_alloc` field. The interfacing between the two could then be fully encapsulated within the `ImageReader` type.

### DIFF
--- a/src/codecs/gif.rs
+++ b/src/codecs/gif.rs
@@ -40,16 +40,16 @@ use crate::error::{
     DecodingError, EncodingError, ImageError, ImageResult, LimitError, LimitErrorKind,
     ParameterError, ParameterErrorKind, UnsupportedError, UnsupportedErrorKind,
 };
+use crate::io::limits;
 use crate::traits::Pixel;
 use crate::{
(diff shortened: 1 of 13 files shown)


197g:
I think we still want _a_ type of sorts around the primitive type, be it just to intentionally get rid of its `Copy` impl. (Not that it should be non-exhaustive like the other structs).

PR #2712, opened 2025-12-30
Implement faster RgbaImage::(from/into)_raw_bgra variants specialized for [u8] buffers

laundmo:
This PR adds an optimized version of from_raw_bgra/into_raw_bgra for Rgba<u8> images. 

## Why

I've repeatedly run into wanting to convert a bgra buffer, like pipewire screen capture seems to prefer, to an RgbaImage. So i was very happy to see this addition:
- https://github.com/image-rs/image/pull/2596

Since I've run into this problem a few times, I've come up with a snippet I keept copying around between projects. Its based on [this rust forum thread](https://users.rust-lang.org/t/the-fastest-way-to-copy-a-buffer-bgra-to-rgba/126651/12) and it autovectorizes incredibly well.

Sadly, this simple, safe, function (shown below) needs MSRV 1.88 due to `as_chunks_mut`.

## What

This is the snippet around which i've written this PR. 
```rs
let (chunked, _) = img.as_chunks_mut::<4>();

for p in chunked {
    let bgra = u32::from_be_bytes(*p);
    let argb = bgra.swap_bytes();
    let rgba = argb.rotate_left(8);
    *p = rgba.to_be_bytes();
}
```

This PR bypasses the MSRV 1.88 requirement of `as_chunks_mut` by inlining it until only stable functions are left, which sadly needs a bunch of unsafe. I've then removed some redundant calculations, like dividing by N twice. This whole custom implementation can be removed and replaced with safe calls to `as_chunks_mut` once MSRV gets bumped to or above 1.88.

## Benchmarks

All benchmarks were done on my Ryzen 9900x using the criterion benchmark included in this PR.

target: `x86-64` (default used by rustc)
image

target: `x86-64-v3` (`avx`, `avx2`, `movbe`, ...):
image

target: `x86-64-v4` (various `avx512` sets):
image

As you can see, by default its not too much better than the built-in variant, but once i enable some more modern instruction sets it gets optimized *far* better.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1454,6 +1454,8 @@ where
     Container: DerefMut<Target = [S]>,
 {
     /// Construct an image by swapping `BgrA` channels into an `RgbA` order.
+    ///
+    /// If you have a container of u8, use the faster [Self::from_raw_bgra_bytes]
     pub fn from_raw_bgra(width: u32, height: u32, container: Container) -> Option<Self> {
         let mut img = Self::from_raw(width, height, container)?;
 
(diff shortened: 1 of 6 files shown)


src/images/buffer.rs
```diff
@@ -1563,6 +1567,75 @@ where
...
+impl<Container> ImageBuffer<Rgba<u8>, Container>
+where
+    Container: DerefMut<Target = [u8]>,
+{
+    /// SAFETY: Callers need to ensure buf.len() >= 4
+    #[inline]
+    unsafe fn _swap_bgra_rgba_buf_bytes_inplace(buf: &mut [u8]) {
+        // Replace with let (chunked, _) = img.as_chunks_mut::<4>(); when MSRV is bumped to 1.88 or above
+        // SAFETY: Ensured by caller
+        let (chunked, _) = unsafe { as_chunks_mut::<4, _>(buf) };
```
fintelia: I think it would be OK to bump the MSRV now to avoid all this unsafe code

fintelia:
I actually sketched out [here](https://github.com/image-rs/image/commit/123eef91b915ccc1ff661a8cf56cae2052d2bd27) how this functionality could be added without needing new public API methods

src/images/buffer.rs
laundmo: Are you sure? The unsafe code is basically the same which a MSRV bump would result in as well.

laundmo:
> I actually sketched out [here](https://github.com/image-rs/image/commit/123eef91b915ccc1ff661a8cf56cae2052d2bd27) how this functionality could be added without needing new public API methods

Hmm, interesting. Well, if you want to merge that, i'll migrate this PR over after you do.

src/images/buffer.rs
```diff
@@ -1563,6 +1567,75 @@ where
...
+        }
+    }
+    /// Construct an image by swapping `BgrA` channels into an `RgbA` order.
+    ///
+    /// Optimized method for u8-based containers
+    pub fn from_raw_bgra_bytes(width: u32, height: u32, container: Container) -> Option<Self> {
+        let mut img = Self::from_raw(width, height, container)?;
+        if img.len() < 4 {
+            return None;
+        }
```
197g: There's empty images, we merely don't have anything to swap. It should short-circuit by returning `Some(img)`.

src/images/buffer.rs
```diff
@@ -1563,6 +1567,75 @@ where
     }
 }
 
+/// This is a inlined version of as_chunks_mut::<4>() with some shuffling to avoid duplicate division
+/// replace usages with buf.as_chunks_mut::<4>() when MSRV is bumped to 1.88 or above
+///
+/// SAFETY: Callers need to ensure buf.len() >= N
```
197g: This precondition is odd. All the indices are in-bounds of the original slice, the `new_len` is rounded _down_, and the unsafe block inside this has an exhaustive SAFETY block that actually covers all requirements. So how is it unsafe to call? But yeah, agree with @fintelia to just bump and use the std version of this method.

fintelia:
@laundmo My patch requires nightly Rust so I'm not planning to merge it. Feel free to copy over any code from it into this PR.

src/images/buffer.rs
fintelia: Yes. We try to avoid writing unsafe wherever possible and the `as_chunks_mut` method added in Rust 1.88.0 is fully safe to call.

(Calling methods in the standard library that are implemented via unsafe is different because the standard library's code is much more heavily audited)

197g:
On a very separate note, I am rather disappointed in LLVM since even removing our abstraction layer of `PixelsMut` iterate does not lead to any better codegen. Really it seems that the `u32` byte-swapping is _better_ than `<&mut [T; 4]>::swap(0, 2)` which is absurd after monomorphization, to me at least. It should very obviously have the same effect.

laundmo:
> am rather disappointed in LLVM since even removing our abstraction layer of PixelsMut iterate does not lead to any better codegen. Really it seems that the u32 byte-swapping is better than <&mut [T; 4]>::swap(0, 2)

I have a hunch its an issue related to just how deep the compiler looks while trying to optimize. I've often found that some abstractions just end up making conceptually easy to optimize code difficult for the compiler to understand. But besides that: as_chunks_mut is basically just a transmute/cast. It doesn't really touch any data. Then its just a flat loop over that memory, which is about as efficient as it gets. The pixels iterator on the other hand creates intermediate structs etc.

> My patch requires nightly Rust so I'm not planning to merge it. Feel free to copy over any code from it into this PR.

Is it likely to have the desired effect, not needing a separate method, without the nightly parts?

My current thought of the easiest way to be able to write different implementations is to make this a trait with a default implementation, and implement it separately for each [T].

laundmo:
Well, after a conversation with my dad who knows old i468 assembly and looking at the generated asm from these conversions, we came up with this *embarrassingly* simple version
```
for [g, _, r, _] in chunked {
    (*r, *g) = (*g, *r);
}
```

And now guess what the performance of that is, on `x86-64-v3`:

Screenshot_20260101_012335

(yes, i named it `_areweidiots` because it seemed embarrassingly simple and likely to be very efficient...)

Sooo yeah, i don't think LLVM is failing us @197g, i think CPUs are incredibly strange beasts to the point that a few `mov`s are slower than whatever simd *BS* the u32 variant generates.

197g:
Extra context of previous implementation / benchmarking attempts: https://users.rust-lang.org/t/converting-a-bgra-u8-to-rgb-u8-n-for-images/67938/13

laundmo:
Allright, i've converted the code over to use the traits as @fintelia has shown, and bumped MSRV instead of reimplementing as_chunks_mut. I also mentioned these functions in the release notes, along with the MSRV bump.

Up-to-date benchmark (`x86-64-v3`)
image

With this i also changed the default implementation to use slice::swap instead of reverse. This is called `(new)` in the benchmark. `(new+optimized)` is the core of this PR. The top benchmark is just `bgr`, not `bgra` - my guess is that its slower due to worse alignment.

laundmo:
Clippy seems to be failing with lints that were added between rust 1.85 and rust 1.88

197g:
We upgraded to `1.88` independently, lints and CI should be clean. If you could rebase we can move forward here.

197g (approved):
Ack with the rebase and clean CI. `u32` logic is far simpler than the manual simd swizzle alternative and the performance difference appears miniscule. Well we can always improve further later (e.g. do the same to `u16` among other subpixel types).

197g:
Thank you :+1:

Issue #2715, opened 2026-01-04
Exposing Loop Count in `AnimationDeocder` Trait


rayguo17:
I would like to be able retrieve the loop count when using `AnimationDecoder` Trait to decode the animated image of whatever underlying format is. 

Usecase: 
In Servo, we have a unified function `decode_animated_image` function, which will take in a specific decoder that implement the `AnimationDecoder` trait, and use the API/function defined by this trait to build Servo’s representation of `ImageFrame` for each animated frame. But because `AnimationDecoder` Trait does not expose a `loop_count` interface, we are unable to know how many times the image need to be loop over.

Generally, this can be useful to any Animated Image displaying tool that use `AnimationDecoder` Trait to decode image. 

## Draft
One way that i can think of is to add a new `loop_count(&self)->LoopCount` method, and a unified `LoopCount` Enum that representing the number of loop provided by each different Animated image format. Currently there are three image format implementing `AnimationDecoder`: apng, gif, webp. And they all have way to retrieve the `loopCount` information. 
<img width="1236" height="736" alt="Image" src="https://github.com/user-attachments/assets/c85e77fb-817e-482c-a0dc-0b693f49d13b" />
Note: In the above picture, those in black are existing functions or member variables. Those in red are newly propose functions.

amyspark:
I second this; I'm implementing an image-rs based plugin for GStreamer and, for feature parity with the GIF decoder, access to this information is needed.

RunDevelopment:
Resolved by #2716 and available since v0.25.10 (Mar 2026).

PR #2716, opened 2026-01-05
Implement loop_count for AnimationDecoder Trait.

rayguo17:
As described in https://github.com/image-rs/image/issues/2715, This PR implements `loop_count` function for `AnimationDecoder` Trait, currently there are three type (Gif, Webp and Apng) implementing `AnimationDecoder` Trait, so implementation on these specific format decoders are added.

### DIFF
--- a/src/io/decoder.rs
+++ b/src/io/decoder.rs
@@ -1,9 +1,7 @@
-use std::num::NonZeroU32;
-
 use crate::animation::Frames;
 use crate::color::{ColorType, ExtendedColorType};
 use crate::error::ImageResult;
-use crate::metadata::Orientation;
+use crate::metadata::{LoopCount, Orientation};
 
 /// The trait that all decoders implement
 pub trait ImageDecoder {
@@ -182,16 +180,7 @@ pub trait AnimationDecoder<'a> {
     /// Consume the decoder producing a series of frames.
     fn into_frames(self) -> Frames<'a>;
     /// Loop count of the animated image.
-    fn loop_count(&self) -> LoopTimes;
-}
-
-/// The number of times animated image should loop over.
-#[derive(Clone, Copy)]
-pub enum LoopTimes {
-    /// Loop the image Infinitely
-    Infinite,
-    /// Loop the image within Finite times.
-    Finite(NonZeroU32),
+    fn loop_count(&self) -> LoopCount;
 }
 
 #[cfg(test)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,7 +160,7 @@ pub use crate::images::dynimage::{
 pub use crate::io::free_functions::{guess_format, load, save_buffer, save_buffer_with_format};
 
 pub use crate::io::{
-    decoder::{AnimationDecoder, ImageDecoder, LoopTimes},
+    decoder::{AnimationDecoder, ImageDecoder},
     encoder::ImageEncoder,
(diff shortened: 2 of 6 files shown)


src/io/decoder.rs
```diff
@@ -179,6 +181,17 @@ impl<T: ?Sized + ImageDecoder> ImageDecoder for Box<T> {
 pub trait AnimationDecoder<'a> {
     /// Consume the decoder producing a series of frames.
     fn into_frames(self) -> Frames<'a>;
+    /// Loop count of the animated image.
+    fn loop_count(&self) -> LoopTimes;
+}
+
+/// The number of times animated image should loop over.
+#[derive(Clone, Copy)]
+pub enum LoopTimes {
```
197g:
```suggestion

/// The number of times animated image should loop over.
#[derive(Clone, Copy)]
pub enum LoopCount {
```

nit: this should have consistent naming with the method and type. I think I prefer `LoopCount`. We could also go for `NumPlays` from the APNG spec but the contraction is not very Rusty.

src/lib.rs
```diff
@@ -160,7 +160,7 @@ pub use crate::images::dynimage::{
 pub use crate::io::free_functions::{guess_format, load, save_buffer, save_buffer_with_format};
 
 pub use crate::io::{
-    decoder::{AnimationDecoder, ImageDecoder},
+    decoder::{AnimationDecoder, ImageDecoder, LoopTimes},
```
197g: I think the whole type would be better off living under the `metadata` module.

197g (commented):
Seems agreeable. (fyi, I'd rather we merge this PR to introduce the type and then I adjust #2672 where the AnimationDecoder trait was removed. It fits well with the `is_animated` field that got added there.)

src/lib.rs
rayguo17: Done

src/io/decoder.rs
rayguo17: `LoopCount` seems good, just that it have conflict with `image_webp::decoder::LoopCount`, but it can be resolved by directly reference it with `crate::metadata::LoopCount` in webp decoder anyway.

rayguo17:
> (fyi, I'd rather we merge this PR to introduce the type and then I adjust #2672 where the AnimationDecoder trait was removed. It fits well with the `is_animated` field that got added there.)

Do you mind explain a bit on the change for AnimationDecoder in #2672 ? I look through the code, and seems to me the PR only change the implementation of the iterator, and still implementing AnimationDecoder.

197g:
The PR replaces what `AnimationDecoder` can do with simply calling `read_image` itself multiple times. That rids us of the awkward dance where you can't easily _optionally_ have an `AnimationDecoder` by moving that information into decoder or file attributes. (Deleting the actual trait would be the last commit on the PR or after it, once we're clear that all functionality replacement is good to go).

197g (approved):

rayguo17:
Hey @197g , I wonder what is the release procedure of image crate? When do you plan to do the next release?

197g:
Since the next planned to be new major bump (#2245), it may be a bit for the fully intended one (1st March?) to tie up the lose ends of animation. That said there were a few fixes in the underlying codecs and a bunch on tiff that might warrant doing a `0.25.10` and backporting this. It depends on capacity vs interested I suppose

PR #2721, opened 2026-01-07
Add streaming/incremental decoding support to BMP decoder

### LINKED ISSUE #2696: Support for resuming decoding after UnexpectedEof
Chromium image decoders have the requirement to be able to decode images even before it has been completely received. In our goal to bring BMP decoding from image crate, we need to support resuming decoding after UnexpectedEof is reached on the input stream, similar to what was done for image-png in this PR:

https://github.com/image-rs/image-png/pull/496

My initial thought is to define a new BMPRead implementation which do not requires Seek trait and which is keeping a DecodingState. I am doing preliminary work under this branch (still WIP, will create PR once I have something in a state to be reviewed):

https://github.com/image-rs/image/compare/main...telecos:image:feature/bmp-streamed-decoding

telecos:
This PR implements the proposal for resumable decoding for BMP images, allowing metadata and pixel data to be loaded incrementally when the full image data isn't immediately available (e.g., network streams, async I/O).

New API (proposal document: https://hackmd.io/@QFC5jGJdR42daEpF6AzCKg/rkmGAQmXbe):
 
- BmpDecoder::new_streaming(): Create decoder in streaming mode
- try_read_metadata(): Incrementally load image metadata
- try_read_row(): Decode pixel data row-by-row

Key features:

Fully backward compatible - existing API unchanged
State machine preserves progress across calls when data is unavailable
Returns ImageError::InsufficientData when more data needed (retryable)
Supports all BMP formats including RLE compression and ICC profiles
Async-safe I/O handling (WouldBlock, Interrupted)
Use case: This enables the Chromium experiment (tracked in https://issues.chromium.org/issues/452666425) for switching to Rust-based BMP decoding where images are received incrementally over the network.

API documentation: https://hackmd.io/@QFC5jGJdR42daEpF6AzCKg/rkmGAQmXbe

Closes #2696


src/codecs/bmp/decoder.rs
```diff
@@ -814,6 +931,1178 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+                                current_row: 0,
+                                row_byte_length,
+                            });
+                        }
+                    }
+                }
+                DecoderState::ReadingBitmasks {
+                    mut bytes_read,
+                    mut buffer,
+                } => {
```
mstoeckl, line 1264: One method to check that the streaming decoder implementation is consistent with the original is with a differential fuzzer (like https://github.com/mstoeckl/image/blob/bmp-differential-fuzz/fuzz/fuzzers/fuzzer_script_bmp_differential.rs .) I don't know if the `image` project maintainers would prefer to have the streaming implementation _replace_ or stay alongside the current decoder; but while this PR is in progress, you may find it useful to find discrepancies between the implementation that could be bugs in either the old or new implementations. (Or it may prove tedious if there are too many false positives.)

The first discrepancy that I found with it is in bitmask reading: for V4 and V5 headers, it looks like the streaming decoder always loads the bitmasks directly after the full header, while the current implementation loads the bitmasks from inside the header. See https://github.com/image-rs/image/blob/256dc9dd5501fa63cbf081a795a936c74c01abd9/src/codecs/bmp/decoder.rs#L860-L879, introduced by 140c36c21357642cc696934e5821ef5c2abc9a6d ; the BMP image documentation that it refers to appears rather confusing, so I'm not entirely certain which is right. (Or if both are wrong.)

telecos: Thanks for the differential fuzzer suggestion, it totally makes sense to guarantee parity on behavior with the reference implementation we want to match. 

Regarding having separated implementation for streamed decoding w.r.t. to current image bmp decoding, this is a conservative approach I propose because currently Chromium is using its custom BMP C++ implementation from blink, so we are expecting also some gaps/differences on current behavior for Chromium implementation w.r.t. current Image crate BMP implementation. As we will integrate the new streamed version into Chromium, having it separated may allow us to have more flexibility to converge and meet Chromium requirements without breaking current users of image crate. After the convergence with Chromium functionality is achieved, then it may make totally sense for streamed decoding implementation to replace/merge with traditional implementation, deciding which changes, may be "breaking" (if any) and thus only suitable for a major release change and which are safe to port, etc...

Also, we may want to consider in a follow up to add support for streamed/incremental decoding in a generic way from ImageDecoder (or some sister structure) if this approach may fit other codecs (for example, we also have in the backlog to bring rust ICO implementation to Chromium through image crate also, which will have also the requirement of having support for streamed decoding)

telecos: As a reference of the usage of new API, I created a draft CL in skia to add support for streamed bmp decoding through the new API:

https://skia-review.googlesource.com/c/skia/+/1138796

There will be another CL in Chromium making use of the streamed decoding support from skia but the user of image crate API for Chromium use case will be skia.

telecos: Regarding spotted issue on bitmask not being properly parsed for V4/V5, you are totally right. Current implementation validation is based mainly on the test "test_streaming_vs_traditional" which compares the decoded content for both current and new streamed decoder. This spotted a gap that we don't have a test file covering this. I will craft some sample file so we can also have coverage in this file.

src/error.rs
```diff
@@ -317,6 +338,28 @@ impl From<LimitErrorKind> for LimitError {
...
+    pub fn new() -> Self {
+        InsufficientDataError
+    }
+}
+
+impl Default for InsufficientDataError {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
197g, line 353: I don't think these are necessary. SemVer does not allow us to change the type anyways and it can be constructed publicly with `InsufficientDataError`, too. No need to go OOP here.

telecos: Thanks for taking a look on this big PR @197g ! Yes, having 2 set of method in same trait for two different usages seems prone to confusion on public API usage. This case probably would be better handled as a separated crate, however we also have mixed feelings on adding too much maintenance complexity to image-rs project by adding another new crate, for a "simple" format like BMP which usually is not having dedicated libraries but custom implementations.

I can take a look at the possibility of avoiding the new methods for streamed mode + the new InsufficientData error. Conceptually, BMP does not have interlacing like PNG does; the only reason that may justify the row by row decoding through the public API is to apply color transformation on the color profile that may be embedded, but this can also be done on a second pass on the decoded data by the caller with not much penalty. I will try to update my feasibility on Chromium/Skia side to check if there may be some limitation on this approach. In parallel, I will continue with some small refactors to make current decoder work on byte slices as this would be required in both approaches anyway.

telecos:
Discarding this PR as we changed the approach to small incremental PRs

PR #2723, opened 2026-01-09
Memory limits for HDR decoding with fixed header size limit

mstoeckl:
This is an alternative to https://github.com/image-rs/image/pull/2713 in limiting the amount of memory allocated by the .hdr file decoder. This approach sets a hard limit on the maximum length of the image header, thereby limiting the amount of memory the decoder could allocate to parse and store the header's contents. If there ever are HDR files which have very long headers, it would not be difficult to explicitly configure limits, like https://github.com/image-rs/image/pull/2713, instead.

### DIFF
--- a/src/codecs/hdr/decoder.rs
+++ b/src/codecs/hdr/decoder.rs
@@ -7,7 +7,7 @@ use crate::error::{
     DecodingError, ImageError, ImageFormatHint, ImageResult, UnsupportedError, UnsupportedErrorKind,
 };
 use crate::io::image_reader_type::SpecCompliance;
-use crate::{ColorType, ImageDecoder, ImageFormat, Rgb};
+use crate::{ColorType, ImageDecoder, ImageFormat, Limits, Rgb};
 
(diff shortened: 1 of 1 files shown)


src/codecs/hdr/decoder.rs
```diff
@@ -198,19 +222,21 @@ impl<R: Read> HdrDecoder<R> {
...
             // read header data until empty line
             loop {
-                match read_line_u8(r)? {
+                match read_line_u8(r, remaining_limit)? {
                     None => {
                         // EOF before end of header
                         return Err(DecoderError::TruncatedHeader.into());
                     }
                     Some(line) => {
+                        remaining_limit -= line.len() + 1;
```
fintelia: I think this should be `saturating_sub` so we don't underflow if `line.len() == remaining_limit`.

mstoeckl: Yes, good catch.

197g (approved):

PR #2727, opened 2026-01-14
Define BMP Compression Method constants, extract method to reuse and add missing compression method BI_ALPHABITFIELDS

telecos:
This change is including:
- Definition of constants for compression methods to remove magic numbers
- Extract compression method handling to its own method (to be reused by streamed decoder in #2721 )
- Add missing compression method BI_ALPHABITFIELDS (6) (https://en.wikipedia.org/wiki/BMP_file_format#DIB_header_(bitmap_information_header). In the case of Chromium decoder, it is handling it the same way as BI_ALPHABITFIELDS (https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/platform/image-decoders/bmp/bmp_image_reader.cc#497)

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -21,6 +21,18 @@ const BITMAPV3HEADER_SIZE: u32 = 56;
 const BITMAPV4HEADER_SIZE: u32 = 108;
 const BITMAPV5HEADER_SIZE: u32 = 124;
 
+// Compression method constants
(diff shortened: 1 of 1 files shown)


197g (approved):

197g:
It would be even cooler if the magic constants were an enum but given that the only case is use as a pattern match, and we do not have `const { … }` as a pattern-expression directly, this is pragmatically a good way to name them at least :+1:

PR #2728, opened 2026-01-15
Refactor BMP header parsing to work on byte slices

telecos:
This change is proposing a refactor of BMP header parsing to work on byte slices instead of direct read operations, to allow reusing code in streamed decoder proposed in #2696 

Introduce ParsedCoreHeader and ParsedInfoHeader structs that parse BMP header data from byte slices (&[u8]) instead of reading directly from the stream.

This refactoring:
- Separates parsing logic from I/O operations
- Makes the parsing code reusable for different reader implementations
- Enables future streaming decoder to share the same parsing code

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -87,17 +87,17 @@ struct ParsedCoreHeader {
 }
 
 impl ParsedCoreHeader {
-    /// Parse BITMAPCOREHEADER fields from buffer. Requires exactly 8 bytes.
-    fn parse(buffer: &[u8]) -> ImageResult<Self> {
-        let width = i32::from(u16::from_le_bytes([buffer[0], buffer[1]]));
-        let height = i32::from(u16::from_le_bytes([buffer[2], buffer[3]]));
+    /// Parse BITMAPCOREHEADER fields from an 8-byte buffer.
+    fn parse(buffer: &[u8; 8]) -> ImageResult<Self> {
+        let width = i32::from(u16::from_le_bytes(buffer[0..2].try_into().unwrap()));
+        let height = i32::from(u16::from_le_bytes(buffer[2..4].try_into().unwrap()));
 
-        let planes = u16::from_le_bytes([buffer[4], buffer[5]]);
+        let planes = u16::from_le_bytes(buffer[4..6].try_into().unwrap());
         if planes != 1 {
             return Err(DecoderError::MoreThanOnePlane.into());
         }
 
-        let bit_count = u16::from_le_bytes([buffer[6], buffer[7]]);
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -78,6 +78,107 @@ const MAX_WIDTH_HEIGHT: i32 = 0xFFFF;
...
+struct ParsedCoreHeader {
+    width: i32,
+    height: i32,
+    bit_count: u16,
+    image_type: ImageType,
+}
+
+impl ParsedCoreHeader {
+    /// Parse BITMAPCOREHEADER fields from buffer. Requires exactly 8 bytes.
+    fn parse(buffer: &[u8]) -> ImageResult<Self> {
```
anforowicz: Would it make sense to encode "requires exactly 8 bytes" into the type system: `buffer: &[u8; 8]`.  This will explicitly force the caller to verify that they have a slice of a sufficient length.  And I see that this PR uses `let mut buffer = [0u8; 8]` in the caller (so no extra verification is needed).

(This feedback also applies to other similar functions below.)

src/codecs/bmp/decoder.rs
```diff
@@ -78,6 +78,107 @@ const MAX_WIDTH_HEIGHT: i32 = 0xFFFF;
...
+    bit_count: u16,
+    compression: u32,
+    colors_used: u32,
+}
+
+impl ParsedInfoHeader {
+    /// Parse BITMAPINFOHEADER fields from buffer. Requires at least 36 bytes.
+    fn parse(buffer: &[u8]) -> ImageResult<Self> {
+        let width = i32::from_le_bytes([buffer[0], buffer[1], buffer[2], buffer[3]]);
+        let mut height = i32::from_le_bytes([buffer[4], buffer[5], buffer[6], buffer[7]]);
```
anforowicz: optional nit: `i32::from_le_bytes(buffer[4..8].try_into().unwrap())` would be slightly shorter.  See:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d2a75cfa157ed349a78fcb4c1f541946

anforowicz (commented):
IIUC this has been split off of https://github.com/image-rs/image/pull/2721 - i.e. this is an incremental step toward the state of the world where the old + the streaming modes are implemented on top of the same set of structs and helper functions.  If so, then this is really great!

I've left some minor comments, but I think this PR is very close to being ready to land (since this is just a refactoring with no changes in functional behavior / public API / etc).

src/codecs/bmp/decoder.rs
telecos: Thanks for the suggestion, it totally make sense to make it explicit.

anforowicz (approved):

PR #2730, opened 2026-01-16
Refactor BMP bitfields and icc parsing to work on byte slices

telecos:
This change is proposing a refactor on the parsing of bitfields and icc profile to work in byte slices instead of direct buffer reads in our way towards a support for incremental decoding.

This change is also defining a constant for ALPHA_OPAQUE which was used as magic number in several places and defined method for calculate_row_padding to improve code readibility.


telecos:
Recreating with clean commit list

PR #2731, opened 2026-01-16
Refactor BMP bitfields and icc parsing to work on byte slices

telecos:
This change is proposing a refactor on the parsing of bitfields and icc profile to work in byte slices instead of direct buffer reads in our way towards a support for incremental decoding.

This change is also defining a constant for ALPHA_OPAQUE which was used as magic number in several places and defined method for calculate_row_padding to improve code readibility.

This work is contributing to #2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -192,17 +192,10 @@ struct ParsedBitfields {
 
 impl ParsedBitfields {
     /// Parse bitfield masks from buffer.
-    /// For V2/Core headers: reads 12 bytes (RGB masks only, alpha is 0).
-    /// For V3/V4/V5 headers: reads 16 bytes (RGBA masks).
-    fn parse(buffer: &[u8], has_alpha: bool) -> ImageResult<Self> {
-        let min_size = if has_alpha { 16 } else { 12 };
-        if buffer.len() < min_size {
-            return Err(ImageError::Decoding(DecodingError::new(
-                ImageFormat::Bmp.into(),
-                "Insufficient data for bitfield masks",
-            )));
-        }
-
+    /// Caller must ensure buffer has sufficient length; this method does not validate.
+    /// Note: Caller must ensure buffer has 12 (V2/Core) or 16 (V3/V4/V5) bytes length; this method does not validate.
+    #[track_caller]
+    fn parse(buffer: &[u8], has_alpha: bool) -> Self {
         let r_mask = u32::from_le_bytes(buffer[0..4].try_into().unwrap());
         let g_mask = u32::from_le_bytes(buffer[4..8].try_into().unwrap());
         let b_mask = u32::from_le_bytes(buffer[8..12].try_into().unwrap());
@@ -212,12 +205,12 @@ impl ParsedBitfields {
             0
         };
 
-        Ok(ParsedBitfields {
+        ParsedBitfields {
             r_mask,
             g_mask,
             b_mask,
             a_mask,
-        })
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -179,6 +182,92 @@ impl ParsedInfoHeader {
...
+    fn parse(buffer: &[u8], has_alpha: bool) -> ImageResult<Self> {
+        let min_size = if has_alpha { 16 } else { 12 };
+        if buffer.len() < min_size {
+            return Err(ImageError::Decoding(DecodingError::new(
+                ImageFormat::Bmp.into(),
+                "Insufficient data for bitfield masks",
+            )));
+        }
+
+        let r_mask = u32::from_le_bytes([buffer[0], buffer[1], buffer[2], buffer[3]]);
```
anforowicz: It doesn't matter that much, but we probably should use consistent style for spelling the argument of `from_le_bytes` - https://github.com/image-rs/image/pull/2728/files#diff-e65034687b0ddb8b1517145498436e84874f048aa774f32d50e82f97f86fa018R134 uses `buffer[X..Y].try_into().unwrap()`, so maybe let's follow this style in this PR too?

src/codecs/bmp/decoder.rs
```diff
@@ -179,6 +182,92 @@ impl ParsedInfoHeader {
...
+        // bV5ProfileData is at offset 112 from header start, which is offset 108 from after size field
+        let profile_offset =
+            u32::from_le_bytes([buffer[108], buffer[109], buffer[110], buffer[111]]);
+
+        // bV5ProfileSize is at offset 116 from header start, which is offset 112 from after size field
+        let profile_size = if buffer.len() >= 116 {
+            u32::from_le_bytes([buffer[112], buffer[113], buffer[114], buffer[115]])
+        } else {
+            return Ok(None);
+        };
```
anforowicz: Not sure if I understand, but it seems to me that the `buffer.len() >= 116` condition on new line 254 is guranteed to be always true, because we already checked `buffer.len() < 116` on new line 237?

src/codecs/bmp/decoder.rs
```diff
@@ -805,71 +902,75 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+                parsed.a_mask,
+                16,
+            )?),
+            ImageType::Bitfields32 => Some(Bitfields::from_mask(
+                parsed.r_mask,
+                parsed.g_mask,
+                parsed.b_mask,
+                parsed.a_mask,
+                32,
+            )?),
```
anforowicz: nit: Can this be simplified by hoisting the shared part of `Bitfields::from_mask` call up:

```
self.bitfields = {
    let max_len = match self.image_type {
        ImageType::Bitfields16 => 16,
        ImageType::Bitfields32 => 32,
    };
    Bitfields::from_mask(
        parsed.r_mask,
        parsed.g_mask,
        parsed.b_mask,
        parsed.a_mask,
        max_len,
    )
};
```

src/codecs/bmp/decoder.rs
telecos: Totally agre, in fact I had in my mind but somehow lost the change. Fixing it

src/codecs/bmp/decoder.rs
telecos: Good catch! else clause from line 254 is not reachable due to 237 check, so we can remove it because it is redundant

src/codecs/bmp/decoder.rs
```diff
@@ -179,6 +182,92 @@ impl ParsedInfoHeader {
...
+    /// For V2/Core headers: reads 12 bytes (RGB masks only, alpha is 0).
+    /// For V3/V4/V5 headers: reads 16 bytes (RGBA masks).
+    fn parse(buffer: &[u8], has_alpha: bool) -> ImageResult<Self> {
+        let min_size = if has_alpha { 16 } else { 12 };
+        if buffer.len() < min_size {
+            return Err(ImageError::Decoding(DecodingError::new(
+                ImageFormat::Bmp.into(),
+                "Insufficient data for bitfield masks",
+            )));
+        }
```
197g: nit: In the context of (+100 lines) there is a bit of double validation happening under this change. The caller of these internal interfaces determines the buffer size and fills it with `read_exact` but then the _internal_ parse methods check the lengths again including a redundant recomputation of the expected size with a separate code path. It's also doubly confusing here since the presence of the fields does not depend on `has_alpha` but rather the header version (header length field), this error should never happen in practice.

I might put `#[track_caller]` on these and document they expect the invariant to be upheld, which should be simple given they expect the buffer in memory. Also we're reading `<<4096` bytes for the full header and have a constant maximum support length. I would consider if it were not cleaner to read it as a _whole_ based on the length field instead of multiple separate `read_exact` calls. Iirc we used the separate calls for the builtin endian conversion not any dedicated purpose of separating IO calls.

Similar to the change towards `FILE_HEADER_SIZE` below, which I find easier to understand than the multiple effects that were going on before.

PR #2732, opened 2026-01-16
Define InsufficientData error type for streaming/resumable scenario

telecos:
Define InsufficientData error type and start using it in BMP decoder's new method try_read_exact.

This change is contributing to #2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -676,12 +676,24 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
         &mut self.reader
     }
 
+    /// Try to read exactly `buf.len()` bytes into `buf`.
+    /// Returns `InsufficientData` error if not enough bytes are available.
+    fn try_read_exact(&mut self, buf: &mut [u8]) -> ImageResult<()> {
+        use std::io::ErrorKind;
+
+        match self.reader.read_exact(buf) {
+            Ok(()) => Ok(()),
+            Err(e) if e.kind() == ErrorKind::UnexpectedEof => Err(ImageError::InsufficientData),
+            Err(e) => Err(e.into()),
+        }
+    }
+
     fn read_file_header(&mut self) -> ImageResult<()> {
         if self.no_file_header {
             return Ok(());
         }
         let mut signature = [0; 2];
-        self.reader.read_exact(&mut signature)?;
+        self.try_read_exact(&mut signature)?;
 
         if signature != b"BM"[..] {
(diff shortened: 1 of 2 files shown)


anforowicz:
Code-wise LGTM, but...

IIUC the original premise behind proposing a [design using a more dedicated error](https://github.com/image-rs/image/issues/2696#issuecomment-3669707048:~:text=is%20okay%20but%20not%20great.%20A-,design%20using%20a%20more%20dedicated%20error,-variant%20would%20be%20more%20explicit.%20When) new error type was that the new type (e.g. `InsufficientData`) would be used in some scenarios and `std::io::ErrorKind::UnexpectedEof` in distinct, non-overlapping scenarios.  And this PR shows that in practice the decoder implementation's only means of detecting EOF is via `Read` trait and via checking for `std::io::ErrorKind::UnexpectedEof`.

Are there any scenarios where the new `InsufficientData` type would be returned, **without** just mapping `std::io::ErrorKind::UnexpectedEof` from one of `Read` operations?  If not, then I don't see the value of introducing a new type/variant of `image::error::ImageError` `enum`.

@197g - WDYT?  Is it ok to put the idea of a new error type on a backburner for now?

197g:
The `UnexpectedEof` is [issued by](https://doc.rust-lang.org/stable/src/std/io/error.rs.html#82-83) `Read::read_exact`'s default implementation, it's not likely to really originate from the underlying source. That is already an ugly detail but the [internals are straightforward](https://doc.rust-lang.org/stable/src/std/io/mod.rs.html#1044-1046). As written, `try_read_exact` is implemented in a misleading way. The API can in fact _not_ be simply called again since an unknown amount of data from the underlying `R: Read` must be restored–generically the reader may be in a completely unknown state in case `read_exact` was overridden by the impl. Consequently the construct as in bmp only works in conjunction with all resumption entry-points performing that seek-back in some initial setup portion of the code.

I can't totally wrap my head around the implications right away, i.e. how this would compose in practice. In `png` we always used `BufRead` where not running `consume` has very clear semantics of making `fill_buf` restartable (with no promise about the amount of data in the second buf, but at the same position).

197g:
> WDYT? Is it ok to put the idea of a new error type on a backburner for now?

Yeah. As much as I'd like an explicit acknowledgement by the decoder, communicating that via the error type as sketched does not convince me either in terms of complexity vs. gain. For a different reason apparently but still.

telecos:
I will set back this PR to draft and we can revisit if we find a situation which it will really add value

PR #2734, opened 2026-01-19
Add regression test for BMP decoder

telecos:
This change is adding a basic regression test on all the bmp test files available doing basic check of width, height and a CRC computed from pixel data. This is trying to assure that while work towards streaming decoding (#2696 ) support on BMP we are not regressing functionality. 3 additional bmp files have been added to cover some spec gaps.

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -1664,10 +1664,12 @@ mod test {
             tested += 1;
         }
 
-        // Ensure we actually tested some files
-        assert!(
-            tested >= 40,
-            "Expected to test at least 40 BMP files, but only tested {}",
+        // Ensure we actually tested all expected files
+        assert_eq!(
+            tested,
+            expected.len(),
+            "Expected to test {} BMP files, but only tested {}",
+            expected.len(),
             tested
         );
     }


telecos:
This change is adding all the files on the same test. I found that we could use rstest to have parametrized tests instead like this

```#[cfg(test)]
mod tests {
    use rstest::rstest;

    #[rstest]
    #[case("rgb24.bmp", 127, 64, 0x025bba0a)]
    #[case("rgb32.bmp", 127, 64, 0x025bba0a)]
    #[case("V4_A8_R8_G8_B8.bmp", 16, 16, 0xee3c0d72)]
    fn test_bmp_file(
        #[case] filename: &str,
        #[case] width: u32,
        #[case] height: u32,
        #[case] crc: u32,
    ) {
        // test implementation
    }
}
```

but this would require adding a dependency crate we don't have

fintelia:
We've got a generic reference testing harness here: https://github.com/image-rs/image/blob/main/tests/reference_images.rs

It works by iterating over the contents of `tests/images` and comparing the contents and CRC's from the associated file in `tests/reference`. There's probably a bunch of possible improvements but we should try to avoid duplication by adding a second reference testing harness.

telecos:
Thanks to pointing to this, as we already have coverage, I will be closing this PR.

PR #2736, opened 2026-01-20
Upgrade to jpeg-encoder 0.7

Shnatsel:
Brings safe SIMD for AVX2, faster SSE/NEON/WASM, slightly nicer API


Shnatsel:
This requires MSRV bump to 1.87 (May 2025), so I went ahead and did it.

PR #2737, opened 2026-01-20
Refactor BMP row decoding to work on byte slices

telecos:
Refactors BMP row decoding functions to separate I/O operations from pixel processing logic using pre-allocated buffers.

## Changes

**Uncompressed and bitfield formats** (`read_full_byte_pixel_data`, `read_16_bit_pixel_data`, `read_32_bit_pixel_data`):
- Allocate a single row buffer outside the `with_rows` closure  
- Read entire rows with one `read_exact()` call per row
- Process pixels from the buffer without additional I/O

**RLE-compressed formats** (`read_rle_data`):
- Introduced `RleReader` struct with 8KB internal buffer
- Reads compressed data in chunks to minimize system calls
- Provides `read_byte()` and `read_bytes()` methods for RLE instruction parsing

`read_palettized_pixel_data` already uses this pattern (row buffer allocated outside closure).

This changeset is going towards #2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -425,6 +425,21 @@ fn calculate_row_padding(bytes_per_row: usize) -> usize {
     (4 - (bytes_per_row % 4)) % 4
 }
 
+/// Allocate a row buffer with OOM protection.
+fn allocate_row_buffer(size: usize) -> ImageResult<Vec<u8>> {
+    let mut buffer = vec_try_with_capacity(size).map_err(|_| {
+        ImageError::Unsupported(UnsupportedError::from_format_and_kind(
+            ImageFormat::Bmp.into(),
+            UnsupportedErrorKind::GenericFeature(format!(
+                "Row buffer allocation ({} bytes) too large",
+                size
+            )),
+        ))
+    })?;
+    buffer.resize(size, 0);
+    Ok(buffer)
+}
+
 /// Convenience function to check if the combination of width, length and number of
 /// channels would result in a buffer that would overflow.
 fn check_for_overflow(width: i32, length: i32, channels: usize) -> ImageResult<()> {
@@ -660,22 +675,19 @@ impl Bitfields {
     }
 }
 
-/// Helper to read RLE data with buffering to minimize system calls.
-struct RleReader<R> {
-    buf_reader: io::BufReader<R>,
+/// Helper to read RLE data using the already-buffered reader.
+/// Avoids double-buffering since BmpDecoder already requires BufRead.
+struct RleReader<'a, R> {
+    reader: &'a mut R,
 }
 
-impl<R: io::Read> RleReader<R> {
-    const RLE_BUFFER_SIZE: usize = 8192;
-
-    fn new(reader: R) -> Self {
-        Self {
-            buf_reader: io::BufReader::with_capacity(Self::RLE_BUFFER_SIZE, reader),
-        }
+impl<'a, R: BufRead> RleReader<'a, R> {
+    fn new(reader: &'a mut R) -> Self {
+        Self { reader }
     }
 
     fn read_byte(&mut self) -> io::Result<u8> {
-        let buf = self.buf_reader.fill_buf()?;
+        let buf = self.reader.fill_buf()?;
         if buf.is_empty() {
             return Err(io::Error::new(
                 io::ErrorKind::UnexpectedEof,
@@ -683,35 +695,31 @@ impl<R: io::Read> RleReader<R> {
             ));
         }
         let byte = buf[0];
-        self.buf_reader.consume(1);
+        self.reader.consume(1);
         Ok(byte)
     }
 
-    fn read_bytes(&mut self, count: usize) -> io::Result<Vec<u8>> {
-        let mut output = vec_try_with_capacity(count).map_err(|_| {
-            io::Error::new(
-                io::ErrorKind::OutOfMemory,
-                "failed to allocate buffer for RLE data",
-            )
-        })?;
+    fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        let mut remaining = buf.len();
+        let mut offset = 0;
 
-        let mut remaining = count;
         while remaining > 0 {
-            let buf = self.buf_reader.fill_buf()?;
-            if buf.is_empty() {
+            let available = self.reader.fill_buf()?;
+            if available.is_empty() {
                 return Err(io::Error::new(
                     io::ErrorKind::UnexpectedEof,
                     "unexpected end of RLE data",
                 ));
             }
 
-            let to_read = remaining.min(buf.len());
-            output.extend_from_slice(&buf[..to_read]);
-            self.buf_reader.consume(to_read);
+            let to_read = remaining.min(available.len());
+            buf[offset..offset + to_read].copy_from_slice(&available[..to_read]);
+            self.reader.consume(to_read);
+            offset += to_read;
             remaining -= to_read;
         }
 
-        Ok(output)
+        Ok(())
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -663,6 +660,58 @@ impl Bitfields {
     }
 }
 
+/// This struct reads data in chunks to avoid excessive system calls when
+/// decoding RLE-compressed BMP images.
+struct RleReader<'a, R> {
```
anforowicz, line 680: Perceived problems / potential improvements:
* Reuse `std::io::BufReader`
* Avoid reading byte-by-byte in `fn read_bytes`.

----------------------

Can `reader+buffer+pos+len` fields of `RleReader` be replaced with just a single field  - created with something like `std::io::BufReader::with_capacity(RLE_BUFFER_SIZE, &mut inner_reader)`?  Then `fn read_byte` can:
* Call `fill_buf`
* Check if the returned buf has zero-length and if so, then report `UnexpectedEof`
* Otherwise get the first byte, call `buf_reader.consume(1)`, and return the first byte?

-------------------

And `fn read_bytes` can maybe:
* Loop until got required number of bytes
    * Call `fill_buf`
    * Check if the returned buf has zero-length and if so, then report `UnexpectedEof`
    * Otherwise get `std::cmp::min(remaining_bytes, fill_buf_size)` and then `resulting_vec.extend_from_slice(&fill_buf[..min_size])` and then keep looping

?

src/codecs/bmp/decoder.rs
```diff
@@ -1216,15 +1263,26 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
 
         reader.seek(SeekFrom::Start(self.data_offset))?;
 
+        // Calculate row size in bytes (2 bytes per pixel + padding)
+        let row_data_len = self.width as usize * 2;
+        let row_padding_len = calculate_row_padding(row_data_len);
```
anforowicz, line 1279: Maybe this is ok (I am not familiar with the layout/format of BMP) but I notice that `fn calculate_row_padding` does modulo-of-4 arithmetic (e.g. `(4 - (bytes_per_row % 4)) % 4` - see [here](https://github.com/image-rs/image/pull/2731/files#diff-e65034687b0ddb8b1517145498436e84874f048aa774f32d50e82f97f86fa018R427)), and the old lines of code here did modulo-of-2 artithmetic (e.g. `self.width as usize % 2 * 2`).

telecos: The BMP spec defies that rows show be padded to have a size multiple of 4 bytes (32 bit):

https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapinfoheader#:~:text=the%20following%20formula%3A-,C%2B%2B,31)%20%26%20~31)%20%3E%3E%203)%3B%0AbiSizeImage%20%3D%20abs(biHeight)%20*%20stride%3B,-For%20YUV%20formats

_For uncompressed RGB formats, the minimum stride is always the image width in bytes, rounded up to the nearest DWORD._

In the specific case of this line, this is working with 2 because it is the specific method for 16bpp, so 2 pixels already make the 32 bit pad.

src/codecs/bmp/decoder.rs
```diff
@@ -1347,23 +1424,27 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
         // two rows.
         let mut row_iter = self.rows(buf);
 
+        // Wrap reader in buffered RLE reader for efficient byte-by-byte access
+        let mut rle_reader = RleReader::new(&mut self.reader);
```
197g, line 1440: nit: The wrapper `io::BufReader` may consume more data from the underlying stream than it will require which is then dropped together with the `RleReader`. I suppose seeking would mitigate this here? Otherwise, the inner reader would not be immediately usable again as expected beyond this point.

In other codecs we avoid that semantic problem in the first place (plus potential double buffering) by having the bound `BufRead` on the inner `reader` stream. BMP has kept it as `Read` for a while but I'm okay with changing that. (Anecdotally this was also the biggest difference between our old jpeg decoder and `zune-jpeg` which unlocked a bunch of performance).

I'm okay with both outcomes (ignoring it for now, or changing the outer bound to a `BufRead`), whichever complexity is more acceptable.

src/codecs/bmp/decoder.rs
```diff
@@ -1288,15 +1357,21 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+            FormatFullBytes::RGB24 => calculate_row_padding(row_data_len),
             _ => 0,
         };
-        let row_padding = &mut [0; 4][..row_padding_len];
+        let total_row_len = row_data_len + row_padding_len;
 
         self.reader.seek(SeekFrom::Start(self.data_offset))?;
 
         let reader = &mut self.reader;
+        let mut row_buffer = vec![0u8; total_row_len];
```
197g: These (newly introduced) allocations have not be checked against the allocation limits, only the limits constraining `width` itself.

src/codecs/bmp/decoder.rs
```diff
@@ -1216,15 +1266,26 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
             |row| {
-                for pixel in row.chunks_mut(num_channels) {
-                    let data = u32::from(reader.read_u16::<LittleEndian>()?);
+                reader.read_exact(&mut row_buffer)?;
+
+                for (i, pixel) in row.chunks_mut(num_channels).enumerate() {
+                    let offset = i * 2;
+                    let data = u32::from(u16::from_le_bytes(
+                        row_buffer[offset..offset + 2].try_into().unwrap(),
+                    ));
```
```suggestion
                for (row_data, pixel) in row_buffer
                    .chunks_exact(2)
                    .zip(row.chunks_exact_mut(num_channels))
                {
                    let data = u32::from(u16::from_le_bytes(
                        row_data.try_into().unwrap(),
                    ));
```

That changes a reader's 'might this panic from invalid offsets' into a simpler correctness consideration of whether the buffer is sized correctly.

Occurs again for `* 4` below I think.

src/codecs/bmp/decoder.rs
```diff
@@ -663,6 +660,61 @@ impl Bitfields {
...
+                io::ErrorKind::UnexpectedEof,
+                "unexpected end of RLE data",
+            ));
+        }
+        let byte = buf[0];
+        self.buf_reader.consume(1);
+        Ok(byte)
+    }
+
+    fn read_bytes(&mut self, count: usize) -> io::Result<Vec<u8>> {
```
197g: Just from the signature this should be either built after `read_exact(&mut [u8])` or `read_to_end(&mut Vec<u8>)`. As written, the allocation for `indices` in `read_rle_data` occurs in a loop and is discarded each time.

If I'm not mistaken in that only use the buffer has a small, constant bound (256 at most) which suggests the use an out-buffer (`&mut [u8]`) as appropriate so this could call `BufReader<R>::read_exact`, communicate the count via the slice's length, and derive that slice from a small stack buffer preserved through loop iterations.

telecos: Good catch! Changed the signature to `read_exact(&mut [u8])` to avoid allocations in the loop. Now using a stack-allocated [u8; 256] buffer in `read_rle_data()` that's reused across all iterations. Much cleaner and no heap allocations in the hot path.

src/codecs/bmp/decoder.rs
telecos: I've refactored both the 16-bit and 32-bit pixel reading loops to use the `chunks_exact().zip()` pattern instead of indexed access. This eliminates the potential panic from slice indexing.

src/codecs/bmp/decoder.rs
telecos: Added allocation limit checks using `vec_try_with_capacity` for all row buffers (16-bit, 32-bit, and full-byte formats). Now if total_row_len exceeds limits, we return a proper error instead of potentially causing OOM.

src/codecs/bmp/decoder.rs
telecos: Since `BmpDecoder` already requires `BufRead` I updated `RleReader` to use the `BufRead` directly instead of wrapping it on another one.

197g (approved):

PR #2738, opened 2026-01-21
BMP resumable decoding with minimal state machine

telecos:
This change is introducing a new contructor `new_resumable` to be used on the resumable decoding scenario. It is also adding a minimal state machine with 2 states (one for metadata parsing and another one for ImageData reading) which are recoverable, but at this point only from the beginning of the state.

Subsequent PRs will be adding additional recoveing points by extending the state machine implementation

This change is going towards #2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -272,6 +272,8 @@ enum DecoderState {
     ReadingMetadata,
     /// Metadata has been read, ready to read image data.
     ReadingImageData,
+    /// Image data has been fully decoded.
+    ImageDecoded,
 }
 
 #[derive(PartialEq)]
@@ -759,6 +761,9 @@ pub struct BmpDecoder<R> {
 
     /// Current decoder state for resumable decoding.
     state: DecoderState,
+
+    /// Stream position to seek to on retry.
+    seek_position: Option<u64>,
 }
 
 enum RLEInsn<'a> {
@@ -792,6 +797,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
             icc_profile: None,
 
             state: DecoderState::ReadingMetadata,
+            seek_position: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -1004,103 +1074,130 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+        match self.read_metadata_impl() {
+            Ok(()) => {
+                self.state = DecoderState::ReadingImageData;
+                Ok(())
+            }
+            Err(e) if Self::is_unexpected_eof(&e) => {
+                // Seek back to the start for retry
+                // Note: For no_file_header mode, this seeks to position 0 which is where
+                // the DIB header starts (no file header to skip)
+                let _ = self.reader.seek(SeekFrom::Start(0));
```
anforowicz: I think we shouldn't ignore the result of `seek` - if seeking fails then I assume that resuming won't work.

src/codecs/bmp/decoder.rs
```diff
@@ -1594,9 +1691,25 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+    ///
+    /// Must be called after `read_metadata()` succeeds. On `UnexpectedEof`, the reader
+    /// is seeked back to allow retry on the same decoder. Buffer contents are undefined
+    /// after an error.
+    pub fn read_image_data(&mut self, buf: &mut [u8]) -> ImageResult<()> {
+        match self.read_image_data_impl(buf) {
+            Ok(()) => Ok(()),
+            Err(e) if Self::is_unexpected_eof(&e) => {
+                // Seek back to the start of image data for retry
+                self.reader.seek(SeekFrom::Start(self.data_offset))?;
```
197g: An alternative strategy in handling these (also metadata) would be to insert a seek at the start of `read_image_data_impl`, maybe optional via a marker state. That would avoid the double-error conundrum here where the seek error might override the EOF one. Also variants of decoding usually perform a seek to `data_offset` immediately anyways (see https://github.com/image-rs/image/blob/d3c7f226adfdf83550987e26e0f2de1dffc5df14/src/codecs/bmp/decoder.rs#L1372) so this might be redundant.

There's definitely more differences here. I can't put my finger on concrete cases that would further show the tradeoffs involved. Did you have any thoughts into that while writing the code?

telecos: Thanks for the suggestion, you're absolute right about the double-error codundrum, introducing not properly handled scenarios when seek fails. So now we store a new optional field `seek_position` which currently is set to the beginning of metadata (and it would be also updated in future PRs once we introduce resumable row-by-row decoding)

I've updated implementation to remove the handling of UnexpectedEof and instead assure that seek is done at beginning of the state. Also, I noticed that I was not explicitly handling the case when read was called again after fully decoded, so I also introduced the `ImageDecoded` state to make NO-OP subsequent calls.

PR #2743, opened 2026-01-25
Upgrade tiff to 0.11

197g:
Going for `0.11.2` actually here. Another notable change is the additional attribute that allows re-use of the whole decoding buffer in preparation for decoding sequences of files.

The test image is from the `image-tiff` repository.

The MSRV bump to `1.88` is quite small and we finally get `as_chunks_mut`. In a backport to the `0.25.9` we might need to re-evaluate that though. Eh.


src/codecs/tiff.rs
```diff
@@ -319,15 +366,30 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
...
 
-        match self
+        let layout = self
             .inner
+            .as_mut()
             .unwrap()
-            .read_image()
-            .map_err(ImageError::from_tiff_decode)?
-        {
+            .read_image_to_buffer(&mut self.buffer)
```
fintelia, line 376: It would be nice to use [`Decoder::read_image_bytes`](https://docs.rs/tiff/latest/tiff/decoder/struct.Decoder.html#method.read_image_bytes) when possible to avoid the extra allocation entirely (doesn't have to be this PR)

197g: Yes, thought so too. There is some overlap with the new comment in `interleave_planes` in that the solution shouldn't be over complicated. And in particular both are more easily doable when we do not run any color transforms.

PR #2744, opened 2026-01-26
Improve BMP resumable decoder to resume at row-level

telecos:
This PR extends the coarse-grained resumable decoding from #2738 to support row-level progress tracking for non-RLE BMP formats. When decoding is interrupted by UnexpectedEof, the decoder can now resume from the last successfully decoded row rather than restarting from the beginning.

This is a work towards #2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -1765,12 +1765,12 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
 
     /// Handle the result of a row-based decode operation, updating state accordingly.
     fn finish_row_decode(&mut self, result: Result<u32, (u32, io::Error)>) -> ImageResult<()> {
- 
-    let (Ok(rows) | Err((rows, _))) = result;
-    self.state = DecoderState::ReadingRowData { rows_decoded: rows };
-    match result {
-        Ok(_) => Ok(()),
-        Err((_, e)) => Err(e)?,
+        let (Ok(rows) | Err((rows, _))) = result;
+        self.state = DecoderState::ReadingRowData { rows_decoded: rows };
+        match result {
+            Ok(_) => Ok(()),
+            Err((_, e)) => Err(e)?,
+        }
     }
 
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -1697,12 +1750,38 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+
+    /// Handle the result of a row-based decode operation, updating state accordingly.
+    fn finish_row_decode(&mut self, result: Result<u32, (u32, io::Error)>) -> ImageResult<()> {
+        let rows = match &result {
+            Ok(r) => *r,
+            Err((r, _)) => *r,
+        };
+        self.state = DecoderState::ReadingRowData { rows_decoded: rows };
+        result.map(|_| ()).map_err(|(_, e)| e.into())
+    }
```
197g: I find this hard to read, suggesting:

```suggestion
 
    let (Ok(rows) | Err((rows, _))) = result;
    self.state = DecoderState::ReadingRowData { rows_decoded: rows };
    match result {
        Ok(_) => Ok(()),
        Err((_, e)) => Err(e)?,
    }
```

A single match statement with an update in both cases would also seem fine, it's just the heavy chaining of into(), map into empty tuple and so on is rather confusing.

src/codecs/bmp/decoder.rs
```diff
@@ -483,35 +489,59 @@ fn num_bytes(width: i32, length: i32, channels: usize) -> Option<usize> {
...
-        for row in buffer.chunks_mut(row_width) {
-            func(row)?;
+    }
+
+    /// Get a mutable reference to a specific row in the output buffer.
+    #[inline]
+    fn get_row_mut(buf: &mut [u8], row_index: usize, row_stride: usize) -> &mut [u8] {
+        let start = row_index * row_stride;
+        let end = start + row_stride;
+        &mut buf[start..end]
```
```suggestion
        &mut buf[start..][..row_stride]
```

It's a cute pattern, also for cases where the overflow of the addition would be a concern (it's not here, we trust the input).

PR #2745, opened 2026-01-27
Improve BMP resumable decoder to resume on metadata

telecos:
This PR extends the coarse-grained resumable decoding from #2738 and its further extension to support resumable row decoding from #2744 to also support resuming during metadata parsing. Mainly 3 different stages are defined (headers parsing, palette and icc profile), so that if UnexpectedEof occurs during their parsing, the decoder can now be resumed from the last recoverable point from metadata rather than restarting from the beginning.

This PR should complete #2696 (considering that by now resuming RLE decompression is not resumable and each retry will require to be resumed from the beginning of image data)


src/codecs/bmp/decoder.rs
```diff
@@ -825,7 +854,9 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
             bitfields: None,
             icc_profile: None,
 
-            state: DecoderState::ReadingMetadata,
+            state: DecoderState::ReadingMetadata {
+                progress: MetadataProgress::NotStarted,
+            },
```
anforowicz: nit: Would it make sense to spell this as `state: DecoderState::default(),`?

src/codecs/bmp/decoder.rs
```diff
@@ -1090,36 +1121,46 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+    /// 2. Palette: Up to 1KB for indexed color images.
+    /// 3. ICC profile: Variable size, can be several KB (V5 headers only).
     pub fn read_metadata(&mut self) -> ImageResult<()> {
-        if self.state != DecoderState::ReadingMetadata {
-            // Already read metadata, nothing to do
+        // Check if we're in a metadata reading state
+        let progress = match self.state {
+            DecoderState::ReadingMetadata { progress } => progress,
+            _ => return Ok(()), // Already past metadata phase
+        };
```
anforowicz, line 1157: nit / no strong opinion (but shorter code = probably more readable?):

```
// Check if we're in a metadata reading state
let DecoderState::ReadingMetadata { progress } = self.state else {
    return Ok(());  // Already past metadata phase
};
```

src/codecs/bmp/decoder.rs
```diff
@@ -1133,8 +1164,75 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+                }
+                self.read_palette()?;
+                // Palette complete, prepare for ICC profile phase
+                let next = if let Some((offset, size)) = icc_meta {
+                    self.seek_position = Some(u64::from(offset));
+                    MetadataProgress::ReadingIccProfile { size }
+                } else {
+                    self.seek_position = None;
+                    MetadataProgress::Complete
+                };
```
anforowicz: The code duplication (between new lines 1203-1209 and new lines 1186-1192) seems a bit unfortunate.  And it also seems a bit unfortunate that the information for resuming metadata reading has to be sometimes be looked up inside `MetadataProgress` (e.g. `icc_meta.0` is a seek position) and sometimes outside of `MetadataProgress` (e.g. `self.seek_position`).  OTOH, I am not sure what would be a good way to improve this.  One hand-wavy idea would be to:

* Introduce a separate `struct HeaderOffsets` with fields like `icc_offset_and_size: Option<(u32, u32)>` and `palette_offset: Option<usize>`.  This way the `HeaderOffsets` can be calculated **once** (in the `MetadataProgress::NotStarted` phase) and used as payload of both `MetadataProgress::ReadingPalette` and `MetadataProgress::ReadingIccProfile` phases.
* Introduce a separate `MetadataProgress::ReadingMainHeader { offset: usize }`.  This should help to remove the `BmpDecoder::seek_position` field - the transition from `MetadataProgress::NotStarted` to `MetadataProgress::ReadingMainHeader` will record the offset.  And then handling of `MetadataProgress::ReadingMainHeader` will unconditionally seek to the offset stored inside.

WDYT?

telecos: Thanks for the suggestion! I agree this is a cleaner design - it eliminates the code duplication in the transition logic between phases, and consolidates all resumability state within MetadataProgress rather than splitting it between the enum and a separate seek_position field.

I've implemented this approach in the latest commit. I added a HeaderOffsets struct to hold palette_offset and icc_profile info, and introduced a ReadingMainHeader phase that records the start offset for retry support. This allowed me to remove the seek_position field from BmpDecoder entirely.

The new flow is: NotStarted → ReadingMainHeader → ReadingPalette → ReadingIccProfile → Complete, with HeaderOffsets computed once during header parsing and carried through subsequent phases.

src/codecs/bmp/decoder.rs
```diff
@@ -1212,33 +1309,35 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
             let mut header_buffer = vec![0u8; (bmp_header_size - 4) as usize];
             let current_pos = self.reader.stream_position()?;
             self.reader.seek(SeekFrom::Start(bmp_header_offset + 4))?;
             self.reader.read_exact(&mut header_buffer)?;
-            self.read_icc_profile(&header_buffer)?;
+
+            // Extract ICC profile metadata for later reading
+            if let Some(parsed) = ParsedIccProfile::parse(&header_buffer) {
+                icc_profile = Some((u64::from(parsed.profile_offset), parsed.profile_size));
+            }
```
anforowicz: WDYT about replacing `(u64, u32)` with `ParsedIccProfile` everywhere?

Then:

* New lines 1323-1325, and new line 1313 can become an unconditional, non-mutable assignment/binding: `let icc_profile = ParsedIccProfile::parse(&header_buffer)`
* The type of `HeaderOffsets::icc_profile` field can become `Option<ParsedIccProfile>` (which seems more readable than `Option<(u64, u32)>` and maybe doesn't even need a comment)
* Signature of `read_icc_profile` can change:
    - From `fn read_icc_profile(&mut self, offset: u64, size: u32) -> ImageResult<()>`
    - To `fn read_icc_profile(&mut self, offset_and_size: &ParsedIccProfile) -> ImageResult<()>`

src/codecs/bmp/decoder.rs
```diff
@@ -1212,33 +1309,35 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
-            .seek(SeekFrom::Start(bmp_header_end + bitmask_bytes_offset))?;
+        // Calculate palette offset (position after headers)
+        let palette_offset = bmp_header_end + bitmask_bytes_offset;
 
-        match self.image_type {
-            ImageType::Palette | ImageType::RLE4 | ImageType::RLE8 => self.read_palette()?,
-            _ => {}
-        }
+        // Seek to position after headers (where palette or image data starts)
+        self.reader.seek(SeekFrom::Start(palette_offset))?;
```
anforowicz: Is this `seek` necessary?  

IIUC the same `seek` will also happen in `fn read_metadata_impl` when handling `MetadataProgress::ReadingPalette`.  And I assume that `self.data_offset` (set at the end of `fn read_metadata_impl`) will be used for reading pixel data.

src/codecs/bmp/decoder.rs
```diff
@@ -1133,20 +1170,80 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+                } else {
+                    MetadataProgress::Complete
+                };
+                self.state = DecoderState::ReadingMetadata { progress: next };
+                self.read_metadata_impl(next)
+            }
+            MetadataProgress::ReadingIccProfile { offsets } => {
+                if let Some((offset, size)) = offsets.icc_profile {
+                    self.read_icc_profile(offset, size)?;
+                }
```
anforowicz, line 1229: This is optional / depending on what you think.  This comment stems from noticing that `MetadataProgress::ReadingIccProfile` is only constructed when `offsets.icc_profile.is_some()` (and therefore this `if let` feels a bit redundant).

I think there are 3 options here:

* Do nothing (the code is fairly readable as-is, the redundancy doesn't seem that bad)
* Do `let (offset, size) = offsets.icc_profile.unwrap()`.  If passing `ParsedIccProfile` then it will become `let icc_offset_and_size = offsets.icc_profile.as_ref().unwrap()`.
    - Cons: this probably requires a comment why `unwrap` won't panic
    - Cons: this doesn't deduplicate two calls to `.is_some()`
* Stop checking `is_some` when constructing `MetadataProgress::ReadingIccProfile`.  And in general, maybe don't do any `if`s when computing `next` state and always progress in the order of `NotStarted` => `ReadingMainHeader` => `ReadingPalette` => `ReadingIccProfile` => `Complete`.
    - Cons: performance (but how bad is it really...)
    - Pros: much shorter/linear code (I hope / I think)

src/codecs/bmp/decoder.rs
telecos: Good catch! That seek is totally unnecessary, as you point it will happen in read_metadata_impl anyway

src/codecs/bmp/decoder.rs
telecos: I think that 3rd option ends on cleaner code. The extra state transitions shouldn't have a meaningful impact on performance for several reasons

- No I/O overhead: The skipped phases (e.g., ReadingPalette when no palette is needed) perform no I/O operations - they just check a condition and immediately transition to the next state.
- Tail-call-like recursion: The state transitions are implemented as tail recursion that the compiler "could" optimize. Each "empty" phase is just a match arm that immediately calls the next phase.
- Metadata is read once: This code path runs exactly once per image decode. The metadata phase is dominated by actual I/O (reading headers, palette, ICC profile), not by the state machine logic.

On the other hand I think that the linear flow is more maintainable and less error-prone than previous conditional branching.

src/codecs/bmp/decoder.rs
```diff
@@ -1133,20 +1171,77 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+                let next = MetadataProgress::Complete { offsets };
+                self.state = DecoderState::ReadingMetadata { progress: next };
+                self.read_metadata_impl(next)
+            }
+            MetadataProgress::Complete { offsets: _ } => {
+                if self.no_file_header {
+                    // Use current position as data offset (reader is positioned after
+                    // headers/palette, which is where image data starts)
+                    self.data_offset = self.reader.stream_position()?;
+                }
```
197g: The order of ICC profile vs. palette changed. So this comment is no longer true, the position we get here is not positioned after headers/palette as it may have been modified by `read_icc_profile` intermittently. Something similar was resolved in the last commit but  probably the test data does not include a profile as well. Maybe this should go into the block of `MetadatProgress::ReadingPalette`?

src/codecs/bmp/decoder.rs
```diff
@@ -2036,43 +2128,72 @@ mod test {
...
+            ), // 32-bit RGBA
+            ("tests/images/bmp/images/Info_8_Bit.bmp", false, true, false), // 8-bit palette
+            ("tests/images/bmp/images/Core_8_Bit.bmp", false, true, false), // Core header + palette
+            ("tests/images/bmp/images/pal8rle.bmp", true, true, false),     // 8-bit RLE
+            ("tests/images/bmp/images/pal4rle.bmp", true, true, false),     // 4-bit RLE
+            ("tests/images/bmp/images/rgb24prof.bmp", false, false, true), // 24-bit with ICC profile
         ];
 
-        for (path, is_rle) in test_files {
+        for (path, is_rle, has_palette, has_icc_profile) in test_files {
```
197g: nit: Since these are getting larger anyways, I'd appreciate readability of a struct with named fields to define the `test_files` array and the meaning of all the bools. The struct can be written inline in this test method and destructured by the iterator again.

src/codecs/bmp/decoder.rs
telecos: Good catch! Yeah, it seems that we are not having a specific test case for that specific scenario. I will try to find some sample file but I guess this should be a strange case having an ICO file embedding a BMP with an embedded ICC profile... most likely it will be easier to manually craft some sample file.

telecos: Another option I am thinking is to extend functionality of PartialReader helper class I am using for the test to allow skipping the file header and then mimic the ICO scenario... I will give it a try just in case it can be done with not much complexity

telecos: While feasible in PartialReader it does not work because the offset encoded in the ICC profile is relative to the start of the file, so if we fake it to remove the 14 byte header artificially from the loaded stream, then offset value is no valid for the stream we are providing.

telecos: Finally I crafted a sample file and added as regression test so we keep the coverage there. I validated that test fails with current status 

```
thread 'ico_bmp_v5_with_icc_profile' panicked at tests/regression.rs:117:34:
Failed to open ICO with BMP V5 ICC profile: IoError(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })
test ico_bmp_v5_with_icc_profile ... FAILED

failures:

failures:
    ico_bmp_v5_with_icc_profile
 ```

and passes after the proposed fix

Issue #1756, opened 2022-07-07
Color changed after decode/encode by image-rs


Brooooooklyn:
## Expected

Images are identical

## Actual behaviour

The green color of texts were changed after decode/encode by image-rs

## Reproduction steps

```rs
use std::error;
use std::fs;

use image::load_from_memory;

type Result<T> = std::result::Result<T, Box<dyn error::Error>>;

fn main() -> Result<()> {
    let raw_png = fs::read("hulu.png")?;
    let dynamic_image = load_from_memory(&raw_png)?;
    dynamic_image.save_with_format("hulu-image.png", image::ImageFormat::Png)?;
    Ok(())
}
```

- clone https://github.com/Brooooooklyn/image-rs-wrong-color
- cargo run

| Raw                       | Decode/Encode by image-rs |
| ------------------------- | ------------------------- |
| ![](https://raw.githubusercontent.com/Brooooooklyn/image-rs-wrong-color/main/hulu.png)           | ![](https://raw.githubusercontent.com/Brooooooklyn/image-rs-wrong-color/main/hulu-image.png)     |

HeroicKatora:
Can you be more precise or recheck? The images in the columns contain *literally* the same pixel matrix:

```
$ png2pnm /tmp/hulu.png > /tmp/hulu.pnm
$ png2pnm /tmp/hulu-image.png.png > /tmp/hulu-image.pnm
$ diff /tmp/hulu.pnm /tmp/hulu-image.pnm
```

HeroicKatora:
Oops, messed up paths, I see it now. Probably color profiles which we definitely are *not* doing properly through the whole chain.

Brooooooklyn:
Any workaround for this issue?

thep0y:
Any progress on fixing this issue?

fintelia:
There actually has been some progress! We've now got an [`icc_profile`](https://docs.rs/image/latest/image/trait.ImageDecoder.html#method.icc_profile) method on the ImageDecoder trait which provides the low level API to query the color profile for an image. The next step would be to expose the ability to set a color profile when saving an image

Shnatsel:
We've made a lot of progress on this, and the remaining items are tracked in #2493. I'm going to go ahead and close this in favor of that issue.

PR #2746, opened 2026-01-28
Write ICC profiles to TIFF files

Shnatsel:
Tested manually using https://github.com/Shnatsel/wondermagick/pull/84 and the image from https://github.com/image-rs/image/issues/1756

I've confirmed that the ICC profile written this way can be read back both by `image` and by imagemagick.

### DIFF
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -616,6 +616,11 @@ impl<W: Write + Seek> ImageEncoder for TiffEncoder<W> {
     ) -> ImageResult<()> {
         self.encode(buf, width, height, color_type)
     }
+
+    fn set_icc_profile(&mut self, icc_profile: Vec<u8>) -> Result<(), UnsupportedError> {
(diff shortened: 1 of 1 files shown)


Shnatsel:
Wait, this seems to be missing `set_icc_profile` implementation to actually modify the `icc` field. Then how did it pass all my manual tests? I was convinced that I verified the behavior was wrong before and correct after!

Let me make it into a draft and double-check.

Shnatsel:
Oh, I've accidentally used a TIFF file produced by imagemagick for testing instead of wondermagick. I've deleted it and re-tested with this branch, the ICC profile does correctly roundtrip through PNG->TIFF->PNG conversion and the resulting image looks the same as the starting PNG.

RReverser:
Could this be backported to 0.25 please or will it only be present in 1.0?

197g:
I'll backport it, thanks for the suggestion.

PR #2750, opened 2026-01-29
AVIF image orientation

Shnatsel:
Instead of something sensible like Exif rotation encoding, they came up with a custom scheme to encode the same thing more verbosely.

Tested on sample images with https://github.com/link-u/avif-sample-images - all pass, but the tests there are non-exhaustive.

Curiously Chromium displays them correctly while Firefox seems to outright reject the sample images with orientation metadata.

Unfortunately this requires some `unsafe` due to mp4parse only exposing pointer API for mirroring, and I don't trust them to make a timely release since the last release was over 2 years ago.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,7 @@ walkdir = "2.5.0"
 default = ["rayon", "default-formats"]
 
 # Format features
-default-formats = ["avif", "avif-native", "bmp", "dds", "exr", "ff", "gif", "hdr", "ico", "jpeg", "png", "pnm", "qoi", "tga", "tiff", "webp"]
+default-formats = ["avif", "bmp", "dds", "exr", "ff", "gif", "hdr", "ico", "jpeg", "png", "pnm", "qoi", "tga", "tiff", "webp"]
 avif = ["dep:ravif", "dep:rgb"]
 bmp = []
 dds = []


fintelia (approved):

Shnatsel:
This is now ready to go. Please take another look - I had to use some `unsafe` due to mp4parse only exposing pointer API for mirroring, and I don't trust them to make a timely release with an upstream change since the last release was over 2 years ago.

src/codecs/avif/decoder.rs
```diff
@@ -112,15 +114,55 @@ impl<R: Read> AvifDecoder<R> {
...
+        // mp4parse does not expose a safe wrapper around the pointer :(
+        let mirror_ptr = ctx.image_mirror_ptr().map_err(error_map)?;
+        let mirror: Option<ImageMirror> = if mirror_ptr.is_null() {
+            None
+        } else {
+            // SAFETY: we have verified above that the pointer is non-null.
+            // We trust mp4parse to return a pointer that is not dangling.
+            // We immediately copy/move the value, so no issues with lifetimes.
+            unsafe { Some(std::ptr::read_unaligned(mirror_ptr)) }
+        };
```
197g: We have [NonNull::read_unaligned](https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html#method.read_unaligned) since 1.80 so this could be done in a parse-dont-validate style.

Shnatsel: `NonNull:new(mirror_ptr).map(|p| p.read_unaligned())` feels a tad too magical to me. When it comes to unsafe code I prefer to keep it simple and stupid.

If you have a preferred way to rewrite it, please go ahead - edits are enabled for maintainers for this PR.

Shnatsel:
mp4parse activity is still happening in git, but releases haven't been published to crates.io in quite some time. I'm not sure what the maintenance status is.

I've opened an upstream PR for a safe wrapper today, we'll see how it goes: https://github.com/mozilla/mp4parse-rust/pull/440

src/codecs/avif/decoder.rs
197g: The clippy failure is odd.

197g: Nevermind, we'd need to cast the pointer and clippy won't let us easily avoid the map without a lint. So yeah, this is the least magical way.

fintelia:
The last published releases of `mp4parse` were actually [at our request](https://github.com/mozilla/mp4parse-rust/issues/387#issuecomment-1556294658). My understanding is that the crate [is used by Firefox](https://searchfox.org/firefox-main/source/media/mp4parse-rust/README.md) which gets it from GitHub rather than crates.io

PR #2751, opened 2026-01-29
Improve BMP resumable decoder to resume on RLE data

telecos:
This PR extends BMP resumable decoding to support row-level checkpoints for RLE4 and RLE8 compressed images. Previously, RLE decoding would restart from the beginning on `UnexpectedEof`; now it resumes from the last completed row, matching the behavior of non-RLE formats.

### Changes

**New types:**
- `RleProgress` enum tracks RLE decoding state: `NotStarted` or `Checkpoint { row, x, stream_pos }`
- `DecoderState::ReadingRleData` now carries an `RleProgress` instead of being a unit variant

**Checkpoint mechanism:**
- After each `EndOfRow` marker, the decoder saves the completed row index and stream position
- On `UnexpectedEof`, state is preserved; retry seeks to the saved position and skips already-decoded rows
- Checkpoints are saved *before* any seek operation to avoid losing progress if the seek fails

**`RleReader` refactoring:**
- Now tracks `bytes_read` internally.
- This enables accurate stream position calculation for checkpoint saves
- Methods accept the reader as a parameter, avoiding borrow conflicts with palette access

### Testing

- Existing resumable decode tests now verify RLE state transitions
- Tests assert `ReadingRleData` state with proper `RleProgress` values for RLE formats

### Performance

No regression on RLE formats. Benchmarked against `main`:

| Format | This PR | main | Change |
|--------|---------|------|--------|
| RLE8 (127×64) | ~15.26 µs | ~15.77 µs | −3% |
| RLE4 (127×64) | ~13.84 µs | ~13.81 µs | +0.2%|

### Design notes

We also considered a pre-buffering approach (read row data into a `Vec<u8>` first, then parse), but this caused a ~25–30% regression in RLE decode performance. The byte-by-byte `RleReader` approach preserves `BufRead` efficiency while still enabling precise byte counting for checkpoints.

This is the final PR for completing #2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -315,7 +315,7 @@ enum RleProgress {
 /// Decoder state for resumable decoding.
 ///
 /// This allows the decoder to recover from `UnexpectedEof` errors.
-/// Decoding can resume from the last succesfilly decoded row.
+/// Decoding can resume from the last successfully decoded row or RLE symbol.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum DecoderState {
     /// Currently reading metadata (headers, palette, ICC profile).
@@ -324,7 +324,7 @@ enum DecoderState {
     /// Stores the number of rows successfully decoded.
     ReadingRowData { rows_decoded: u32 },
     /// Currently reading RLE-compressed data.
-    /// Tracks progress at row boundaries for resumability.
+    /// Tracks progress at symbol boundaries for resumability.
     ReadingRleData { progress: RleProgress },
     /// Image data has been fully decoded.
     ImageDecoded,
@@ -1742,7 +1742,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
             // When resuming mid-row, skip to the saved x position on the first row.
             let mut x = if first_row_iteration && start_x > 0 {
                 pixel_iter.nth(start_x as usize - 1); // nth(n) consumes n+1 elements
-                start_x as usize
+                start_x
             } else {
                 0
             };
@@ -1834,7 +1834,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
                             let pixel = pixel_iter.next().ok_or(DecoderError::CorruptRleData)?;
                             pixel.fill(0);
                         }
-                        x += x_delta as usize;
+                        x += x_delta as u32;
                     }
                     RLEInsn::Absolute(length, indices) => {
                         // Absolute mode cannot span rows, so if we run
@@ -1863,7 +1863,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
                             }
                             _ => unreachable!(),
                         }
-                        x += length as usize;
+                        x += length as u32;
                     }
                     RLEInsn::PixelRun(n_pixels, palette_index) => {
                         match image_type {
@@ -1893,7 +1893,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
                             }
                             _ => unreachable!(),
                         }
-                        x += n_pixels as usize;
+                        x += n_pixels as u32;
                     }
                 }
 
@@ -1902,7 +1902,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
                 self.state = DecoderState::ReadingRleData {
                     progress: RleProgress::Checkpoint {
                         row: current_row,
-                        x: x as u32,
+                        x,
                         stream_pos: start_pos + rle_reader.bytes_read(),
                     },
                 };
@@ -1912,7 +1912,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
             self.state = DecoderState::ReadingRleData {
                 progress: RleProgress::Checkpoint {
                     row: current_row,
-                    x: x as u32,
+                    x,
                     stream_pos: start_pos + rle_reader.bytes_read(),
                 },
             };
@@ -1952,7 +1952,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
     ///
     /// - For non-RLE formats: decoding resumes from the last successfully decoded row.
     ///   Already-decoded rows are preserved in `buf`.
-    /// - For RLE formats: decoding resumes from the last checkpoint (after EndOfRow or Delta).
+    /// - For RLE formats: decoding resumes from the last checkpoint (completed instruction symbol).
     ///   Rows and pixels completed before the error are preserved in `buf`.
     pub fn read_image_data(&mut self, buf: &mut [u8]) -> ImageResult<()> {
         match self.state {


src/codecs/bmp/decoder.rs
```diff
@@ -1670,8 +1700,25 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
+                progress: RleProgress::AfterRow { row, next_row_pos },
+            } => (row + 1, next_row_pos),
+            _ => {
+                // read_rle_data called in invalid state - this is a bug
+                debug_assert!(
+                    false,
+                    "read_rle_data called in unexpected state: {:?}",
+                    self.state
+                );
+                (0, self.data_offset)
```
anforowicz: I think I would replace the final match arm with just `unreachable!()`?  This is based on Chromium guidance which recommends avoiding continuing execution once the program reaches an unexpected internal state: https://source.chromium.org/chromium/chromium/src/+/main:styleguide/c++/checks.md;l=79;drc=d13776f43240759a11dc176a770b8fc5d4408f67

src/codecs/bmp/decoder.rs
```diff
@@ -1682,16 +1729,31 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
         let mut row_iter = self.rows(buf);
 
-        // Wrap reader in buffered RLE reader for efficient byte-by-byte access
-        let mut rle_reader = RleReader::new(&mut self.reader);
+        // Skip already-decoded rows (resuming from checkpoint)
+        for _ in 0..start_row {
+            if row_iter.next().is_none() {
+                return Ok(());
+            }
+        }
```
anforowicz: I wonder if this could be rewritten as the following code (using [`Iterator::skip`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip)):

```
let mut row_iter = self.rows(buf).skip(start_row);
```

src/codecs/bmp/decoder.rs
anforowicz, line 1964: Not related to changes in this PR (so feel free to ignore, or to address in a separate PR): I think this `match` statement can be written as:

```
self.read_image_data_impl(buf).and_then(|| { self.state = DecoderState::ImageDecoded; })
```

anforowicz: I wonder what should happen when `self.state == DecoderState::ReadingMetadata`.  Maybe it's worth explicitly covering all possible variants in a `match` (rather than `if`) statement, and then calling `unreachable!()` if `fn read_image_data` is called in `ReadingMetadata` state?

PS. I do wonder if returning `ImageError::Parameter` may be a more graceful experience (e.g. maybe this is a better alternative than calling `unreachable!` here and in my other comment).  This is a bit different here, where the incorrect call/bug comes from a consumer/client of the `image` crate (unlike the `unreachable!` call I suggested in my other comment where the bug seems to be in the BMP decoder implementation).

--------------

A bit separately,  also wondered about potential redundancy between 1) `DecoderState::ReadingRowData/ReadingRleData` (selected by `fn read_metadata` based on `self.is_rle()` call which inspects `self.image_type`) and 2) checking `ImageType` again in `fn read_image_data_impl`.  But I think this is fine / I can't see how to improve it (partially because choosing the exact non-RLE reading function is still a bit complicated + I don't think exploding `DecoderState::ReadingRowData` into `ReadingPaletized` / `ReadingFullByteData` / `Reading16bitData` would be actually better).

telecos: Good point on the distinction between internal bugs vs client misuse. I've used unreachable for `read_rle_data` (internal state corruption) but added a new `DecoderError::MetadataNotRead` for the `ReadingMetadata` case since that's client misuse and deserves a graceful error. 

Regarding the `DecoderState` / `ImageType` - I agree the current design has some redundancy but it seems fine for keeping state machine simpler.

src/codecs/bmp/decoder.rs
```diff
@@ -1742,6 +1806,10 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
                         // a delta code, however IE and the windows image
                         // preview seems to replace them with black pixels,
                         // so we stick to that.
+                        //
+                        // Note: We don't save a checkpoint here. If an error occurs after
+                        // a delta, we resume from the last EndOfRow and re-zero the skipped
+                        // rows, which is idempotent and correct.
```
mstoeckl: I think this could lead to quadratic decoding time if decoding were to be attempted with every new byte of the image, and the image file contains a very long sequence of Delta(0, 0) instructions.

telecos: That's right, when I simplified when to store restore checkpoints, I was not thinking in more extreme cases which would cause a big penalty. Therefore I will also add restore checkpoints in Delta steps in which we store x and y position resulting after the processed Delta.

mstoeckl: It's not as significant, but the quadratic behavior is still present with regular RLE instructions when the image width is large, and the decoder is given one byte at a time. If the decoding is attempted with every new byte, then a line length of 65535 could lead to O(65535^2) total time by the decoder.

Also: the PixelRun instruction for RLE8 images, because it silently clamps to the row length, is a no-op like Delta(0,0) when it occurs at the end of a line.

src/codecs/bmp/decoder.rs
```diff
@@ -297,11 +297,29 @@ struct HeaderOffsets {
...
+    /// Not started yet.
+    #[default]
+    NotStarted,
+    /// Completed through row `row` (0-indexed).
+    /// `next_row_pos` is the stream position where the next row's data begins.
+    AfterRow { row: u32, next_row_pos: u64 },
+    /// Processing a Delta instruction.
+    /// `row` is the current row (0-indexed), `x` is the pixel position within the row,
+    /// `stream_pos` is the stream position immediately after the Delta instruction.
+    InDelta { row: u32, x: u32, stream_pos: u64 },
```
mstoeckl: It may be simpler to combine the `AfterRow` and `InDelta` states. As far as I can tell, the state that the RLE decoder needs to continue decoding is just a (row, x) coordinate pair, and the position in the stream (assuming that position is at an instruction boundary).

mstoeckl:
(Copying a comment on a resolved item that is easily overlooked, since it still applies to latest 7fe07b9c0fc8b16b687597a7ac76c0cf3ffaa3a4 ):

It's not as significant, but the quadratic behavior is still present with regular RLE instructions when the image width is large, and the decoder is given one byte at a time. If the decoding is attempted with every new byte, then a line length of 65535 could lead to O(65535^2) total time by the decoder.

Also: the PixelRun instruction for RLE8 images, because it silently clamps to the row length, is a no-op like Delta(0,0) when it occurs at the end of a line.

telecos:
> (Copying a comment on a resolved item that is easily overlooked, since it still applies to latest [7fe07b9](https://github.com/image-rs/image/commit/7fe07b9c0fc8b16b687597a7ac76c0cf3ffaa3a4) ):
> 
> It's not as significant, but the quadratic behavior is still present with regular RLE instructions when the image width is large, and the decoder is given one byte at a time. If the decoding is attempted with every new byte, then a line length of 65535 could lead to O(65535^2) total time by the decoder.
> 
> Also: the PixelRun instruction for RLE8 images, because it silently clamps to the row length, is a no-op like Delta(0,0) when it occurs at the end of a line.

Implemented the checkpoint storage for each operation, so that we store after each iteration in the `loop` for pixels in row and again after breaking the `loop` due to an EndOfRow. I measured performance impact of this change, and it is having a 3% penalty on the existing RLE tests (`pal4rle.bmp` and `pal8rle.bmp`).

mstoeckl (approved):

src/codecs/bmp/decoder.rs
```diff
@@ -297,11 +297,25 @@ struct HeaderOffsets {
...
+    /// On resume, decoding continues from this exact pixel position.
+    Checkpoint { row: u32, x: u32, stream_pos: u64 },
+}
+
 /// Decoder state for resumable decoding.
 ///
 /// This allows the decoder to recover from `UnexpectedEof` errors.
-/// For row-based formats, decoding can resume from the last successfully
-/// decoded row. For RLE formats, decoding restarts from the beginning.
+/// Decoding can resume from the last succesfilly decoded row.
```
```suggestion
/// Decoding can resume from the last successfully decoded row or RLE symbol.
```

src/codecs/bmp/decoder.rs
```diff
@@ -310,9 +324,8 @@ enum DecoderState {
     /// Stores the number of rows successfully decoded.
     ReadingRowData { rows_decoded: u32 },
     /// Currently reading RLE-compressed data.
-    /// RLE cannot be resumed at row boundaries due to its sequential nature,
-    /// so we restart from the beginning on UnexpectedEof.
-    ReadingRleData,
+    /// Tracks progress at row boundaries for resumability.
```
```suggestion
    /// Tracks progress at symbol boundaries for resumability.
```

src/codecs/bmp/decoder.rs
```diff
@@ -1680,18 +1721,33 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
             let mut pixel_iter = row.chunks_exact_mut(num_channels);
 
-            let mut x = 0;
+            // When resuming mid-row, skip to the saved x position on the first row.
+            let mut x = if first_row_iteration && start_x > 0 {
+                pixel_iter.nth(start_x as usize - 1); // nth(n) consumes n+1 elements
+                start_x as usize
+            } else {
+                0
+            };
```
197g, line 1748: nit: It may be more readable to store `x` as a `u32` here instead of `usize`. It is used for an iteration bound (type does not matter), to add RLE lengths (all of which are `u8`) and to construct the next state (requires it to be `u32`). That would also match `current_row`.

src/codecs/bmp/decoder.rs
```diff
@@ -1867,23 +1948,19 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
     /// Must be called after `read_metadata()` succeeds. On `UnexpectedEof`, the decoder
-    /// can be retried - each format reader seeks to the correct position at the start.
+    /// can be retried:
     ///
-    /// For row-based formats (non-RLE), already-decoded rows are preserved in the buffer
-    /// and decoding resumes from the next row. For RLE formats, decoding restarts from
-    /// the beginning and buffer contents are undefined after an error.
+    /// - For non-RLE formats: decoding resumes from the last successfully decoded row.
+    ///   Already-decoded rows are preserved in `buf`.
+    /// - For RLE formats: decoding resumes from the last checkpoint (after EndOfRow or Delta).
```
```suggestion
    /// - For RLE formats: decoding resumes from the last checkpoint (completed instruction symbol).
```

Issue #2753, opened 2026-02-01
Support for resuming decoding after UnexpectedEof in ICO format


telecos:
Following work done #2696 , next format we plan to integrate in Chromium from image crate is ICO. We should update it to also support streamed/resumable decoding so after an UnexpectedEof it is able to recover from the latest decoded state when additional data is available in the input stream. We will try to follow a similar approach than for BMP, proposing smaller self-contained PRs.

PR #2754, opened 2026-02-01
Refactor ICO decoder to work on byte slices

telecos:
This refactor is modifying the read entries and mask reading to work in byte slices instead of direct read operations to the stream.

This work is towards #2753

### DIFF
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -1,3 +1,4 @@
+use byteorder_lite::ReadBytesExt;
 use std::io::{BufRead, Read, Seek, SeekFrom};
 use std::{error, fmt};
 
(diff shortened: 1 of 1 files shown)


src/codecs/ico/decoder.rs
```diff
@@ -349,11 +349,12 @@ impl<R: BufRead + Seek> ImageDecoder for IcoDecoder<R> {
                 // for that claim, so we can't be sure which is correct.
                 if data_end >= image_end + mask_length {
                     // If there's an AND mask following the image, read and apply it.
+                    let mut mask_row = vec![0u8; mask_row_bytes as usize];
```
197g: That is bounded by roughly `u32::MAX/8` which is a bit too much to (potentially) allocate without checking the allocation limits. Can you explain the motivation better? Similar to the previous discussion on run-length encoding, the newly introduced requirement that everything be available at once while doing `O(n)` work up-front could be a performance tarpit for malformed inputs on top of the allocation.

I'm not sure if this is the right direction to go or if need to complete the rework to avoid an awkward intermediate state between two different strategies that fails to deliver requirements from either.

telecos: Good catch! Yes, while mask "should" be small, AFAIU, there is nothing in the spec preventing from getting an ICO of an arbitrarily high size. We could limit the size of the buffer used for read, but then we would need special handling for the last set of bytes read that do not fill the buffer completely... Though giving it a second though, I think that if we keep it as it was (i.e. direct read_u8), we can capture UnexpectedEof and store a restore point in row y and pixel x-1 as the last successfully read position. So I will revert this change to the original state as it seems better to keep it as it is.

PR #2755, opened 2026-02-02
Add a knob for progressive mode in JPEG encoding

Shnatsel:
Really straightforward since the underlying encoder already supports it and we just need to expose the toggle.

Needed for `wondermagick`


telecos (approved):
LGTM

I see that JPEGDecoder is exposing is_progressive member, so maybe we could add a roundtrip test checking that. Not a high value test anyway.

PR #2757, opened 2026-02-03
Fix icc profile returned properly in BMPDecoder

### LINKED ISSUE #2756: Wrong bytes returned for ICC profile in BMPDecoder
During the final integration with Chromium, we detected that the returned bytes for icc profiles are shifted because it is not taking into consideration the bmp header bytes.

## Expected

icc_profile() from BMPDecoder returns exact ICC profile content

## Actual behaviour

icc_profile() is returned content shifted by 14 bytes

telecos:
During integration in Chromium we found that returned bytes for icc profile were shifted. The issue was not detected before because unit tests only verified that returned size of ICC profile was the expected, but it was not validating that it could properly be parsed. With this change we also improve the test to parse the profile with moxcms and do a minimal validation that it can be parsed without error.

Fixes #2756

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -215,7 +215,7 @@ impl ParsedBitfields {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct ParsedIccProfile {
     /// Absolute file offset where the ICC profile data starts.
-    profile_offset: u32,
+    profile_offset: u64,
     profile_size: u32,
 }
 
@@ -244,7 +244,7 @@ impl ParsedIccProfile {
         }
 
         // Compute the absolute file offset by adding the header's position to the relative offset
-        let profile_offset = bmp_header_offset as u32 + profile_offset_from_header;
+        let profile_offset = bmp_header_offset + u64::from(profile_offset_from_header);
 
         Some(ParsedIccProfile {
             profile_offset,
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -233,15 +234,18 @@ impl ParsedIccProfile {
...
         // bV5ProfileSize is at offset 116 from header start, which is offset 112 from after size field
         let profile_size = u32::from_le_bytes(buffer[112..116].try_into().unwrap());
 
-        if profile_size == 0 || profile_offset == 0 {
+        if profile_size == 0 || profile_offset_from_header == 0 {
             return None;
         }
 
+        // Compute the absolute file offset by adding the header's position to the relative offset
+        let profile_offset = bmp_header_offset as u32 + profile_offset_from_header;
```
197g: nit: the cast being very lossy here is smell of an inconsisteny: either the smaller offset should part of the function signature (i.e. turn `bmp_header_offset` into a `u32` and accept that we only support bmps at the start of files) or the offset should be tracked as proper `u64` stream position. Since the use is a `seek(FromStart(…))` with conversion to `u64` anyways I would heavily gravitate towards the latter.

197g (commented):
What a catch, sorry for having missed this in review apparently. I agree with the underlying issue being the test infrastructure, we can learn from this for other format's metadata support.

src/codecs/bmp/decoder.rs
telecos: Updated to work with `u64` to have clean usage of it in Seek and just converting the field from header which is `u32` to `u64`

PR #2758, opened 2026-02-03
Expose rows_decoded method and improve it to handle all states

telecos:
For the fully support of incremental decoding in Chromium, it is required that once `UnexpectedEof` is obtained, the succesfully decoded rows are provided so that rendering of them can start in the browser. In order to be able to do that, we are missing to expose the number of successfully decoded rows. This change is exposing the method, which already existed, and also updating it as it was only used for non-RLE images.

This is a gap detected on #2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -1929,12 +1929,17 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
         matches!(self.image_type, ImageType::RLE4 | ImageType::RLE8)
     }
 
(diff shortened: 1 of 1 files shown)


197g:
Alright. I'm usually a cautious exposing underlying details of the formats in `image` but we're building a more general restart/continuation interface here and that needs to start somewhere concrete. For bmp it's at least very clear how the format works and highly unlikely to change to something entirely different (unlike different tilings and interpolation, loop filtering, etc. in other formats that make 'progress' harder to model).

PR #2759, opened 2026-02-06
Expose is_top_down method in BMPDecoder

telecos:
During the codereview integrating streamed/resumabled BMP decoding in Skia that will be used from Chromium, we found a gap when copying partially decoded rows:

https://skia-review.googlesource.com/c/skia/+/1093916

The issue is that, despite BMP decoder is always doing conversion to always provide the same row ordering independently of the type of file, when we have partially decoded rows and only first n rows are available, it is required to know if the image is top down or not to know if the rows to copy are the first n rows or the last n rows.

This is a gap identified for completing #2696


telecos:
PR created from a poluted branch, recreating...

PR #2760, opened 2026-02-06
Expose is_top_down function in BMP Decoder

telecos:
During the codereview integrating streamed/resumabled BMP decoding in Skia that will be used from Chromium, we found a gap when copying partially decoded rows:

https://skia-review.googlesource.com/c/skia/+/1138796

The issue is that, despite BMP decoder is always doing conversion to always provide the same row ordering independently of the type of file, when we have partially decoded rows and only first n rows are available, it is required to know if the image is top down or not to know if the rows to copy are the first n rows or the last n rows.

This is a gap identified for completing https://github.com/image-rs/image/issues/2696

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -85,13 +85,13 @@ const PROFILE_EMBEDDED: u32 = 0x4D424544;
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum RowsDecoded {
     /// Rows were decoded sequentially from the top of the image.
-    Sequential {
-        /// Number of rows decoded so far.
+    TopDown {
+        /// Number of top rows decoded so far.
         rows: u32,
     },
-    /// Rows were decoded with a vertical flip (bottom-up to top-down).
-    FlippedVertically {
-        /// Number of rows decoded so far.
+    /// Rows were decoded from the bottom of the image (vertical flip).
+    BottomUp {
+        /// Number of bottom rows decoded so far.
         rows: u32,
     },
 }
@@ -101,7 +101,7 @@ impl RowsDecoded {
     #[inline]
     pub fn rows(&self) -> u32 {
         match *self {
-            RowsDecoded::Sequential { rows } | RowsDecoded::FlippedVertically { rows } => rows,
+            RowsDecoded::TopDown { rows } | RowsDecoded::BottomUp { rows } => rows,
         }
     }
 }
@@ -1511,7 +1511,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
 
         let rows_decoded = self.rows_decoded();
         let start_row = rows_decoded.rows();
-        let top_down = matches!(rows_decoded, RowsDecoded::Sequential { .. });
+        let top_down = matches!(rows_decoded, RowsDecoded::TopDown { .. });
 
         let file_offset = self.data_offset + (start_row as u64 * row_byte_length as u64);
         self.reader.seek(SeekFrom::Start(file_offset))?;
@@ -1575,7 +1575,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
 
         let rows_decoded = self.rows_decoded();
         let start_row = rows_decoded.rows();
-        let top_down = matches!(rows_decoded, RowsDecoded::Sequential { .. });
+        let top_down = matches!(rows_decoded, RowsDecoded::TopDown { .. });
         let width = self.width;
         let height = self.height;
 
@@ -1626,7 +1626,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
 
         let rows_decoded = self.rows_decoded();
         let start_row = rows_decoded.rows();
-        let top_down = matches!(rows_decoded, RowsDecoded::Sequential { .. });
+        let top_down = matches!(rows_decoded, RowsDecoded::TopDown { .. });
         let width = self.width;
         let height = self.height;
 
@@ -1688,7 +1688,7 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
 
         let rows_decoded = self.rows_decoded();
         let start_row = rows_decoded.rows();
-        let top_down = matches!(rows_decoded, RowsDecoded::Sequential { .. });
+        let top_down = matches!(rows_decoded, RowsDecoded::TopDown { .. });
         let width = self.width;
         let height = self.height;
 
@@ -1977,9 +1977,9 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
             DecoderState::ReadingMetadata { .. } => 0,
         };
         if self.top_down {
-            RowsDecoded::Sequential { rows }
+            RowsDecoded::TopDown { rows }
         } else {
-            RowsDecoded::FlippedVertically { rows }
+            RowsDecoded::BottomUp { rows }
         }
     }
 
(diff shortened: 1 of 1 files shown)


telecos:
Thinking about an alternative, I think that probably it is a better option to return as a second parameter of rows_read function instead of a new dedicated method as knowing orientation from the consumer side only makes sense while having partially decoded rows; once fully decoded, orientation does not matter for the consumer of BMPDecoder as it manages internally the row inversion (the only case where it would be required is that one in which you want to encode exact same file content you decoded, but I don't think it is a valued use case and would require exposing all other header attributes).

telecos:
I updated to return the tuple in rows_decoded and removed the public function which could cause confusion when used on a fully decoded image.

src/codecs/bmp/decoder.rs
```diff
@@ -1929,9 +1926,13 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
 
-    /// Returns the number of fully decoded rows currently available in the buffer.
-    pub fn rows_decoded(&self) -> u32 {
-        match self.state {
+    /// Returns the number of fully decoded rows and the image orientation.
+    ///
+    /// The returned tuple contains:
+    /// - `rows`: The number of rows that have been fully decoded and are available in the buffer.
+    /// - `top_down`: Whether the BMP is stored top-down (`true`) or bottom-up (`false`).
+    pub fn rows_decoded(&self) -> (u32, bool) {
```
197g: Can we make this an enumeration? I think the problem you found generalizes: The decoder must return a precise description of the pixels that it has touched; and that description must be for the layout of the buffer including all transforms done to it (top-down vs. bottom up is a mirror) and not a description in terms of the file layout.

We don't want to overgeneralize as there's too many ways to go about it across formats but we're doing this to get more formats into such support eventually. I'm looking at solving this somewhat pragmatically: some formats and some files we can decode progressively and we would then expand that support as practical by adding possible variants describing their partial state. These two possibilities are just the most clear-cut and widely applicable.

Let's introduce a type for `bmp` specifically first, i.e. an exhaustive enumeration. Then we can add conversion to a general variant supported by other image formats later.

You'll want to use `matches!` in the tests.

telecos: Thanks for the suggestion, I really appreciate it. Having an enum seems more scalable for other transformations that may occur. For BMP we only have this vertical flipping of rows, but I can see other codecs needing additional variants, such as interlacing (e.g. PNG's Adam7 or GIF's 4-pass mode), where rows arrive in a non-sequential order during progressive decoding; or horizontal mirroring (e.g. TGA's origin bits can specify any corner).

src/codecs/bmp/decoder.rs
```diff
@@ -78,6 +78,34 @@ const MAX_WIDTH_HEIGHT: i32 = 0xFFFF;
...
 
+/// During progressive decoding, the decoder applies transforms (e.g. a vertical
+/// flip for bottom-up BMP files) as it writes rows into the output buffer.
+/// This enum describes which rows contain valid pixel data by indicating the
+/// transform that was applied.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RowsDecoded {
+    /// Rows were decoded sequentially from the top of the image.
+    Sequential {
+        /// Number of rows decoded so far.
```
anforowicz: nit: `/// Number of top rows decoded so far.`?

src/codecs/bmp/decoder.rs
```diff
@@ -78,6 +78,34 @@ const MAX_WIDTH_HEIGHT: i32 = 0xFFFF;
...
+        rows: u32,
+    },
+    /// Rows were decoded with a vertical flip (bottom-up to top-down).
+    FlippedVertically {
+        /// Number of rows decoded so far.
```
anforowicz: nit: `/// Number of bottom rows decoded so far.`?

src/codecs/bmp/decoder.rs
```diff
@@ -78,6 +78,34 @@ const MAX_WIDTH_HEIGHT: i32 = 0xFFFF;
...
 const PROFILE_EMBEDDED: u32 = 0x4D424544;
 
+/// During progressive decoding, the decoder applies transforms (e.g. a vertical
+/// flip for bottom-up BMP files) as it writes rows into the output buffer.
+/// This enum describes which rows contain valid pixel data by indicating the
+/// transform that was applied.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RowsDecoded {
+    /// Rows were decoded sequentially from the top of the image.
+    Sequential {
```
anforowicz: optional nit / bike-shedding: Let's discuss variant names with @197g and others?

* Current proposal: `Sequential` and `FlippedVertically`
* Alternative: `TopDown` and `BottomUp`

WDYT?

src/codecs/bmp/decoder.rs
```diff
@@ -78,6 +78,34 @@ const MAX_WIDTH_HEIGHT: i32 = 0xFFFF;
 /// The value of the V5 header field indicating an embedded ICC profile ("MBED").
 const PROFILE_EMBEDDED: u32 = 0x4D424544;
 
+/// During progressive decoding, the decoder applies transforms (e.g. a vertical
+/// flip for bottom-up BMP files) as it writes rows into the output buffer.
+/// This enum describes which rows contain valid pixel data by indicating the
+/// transform that was applied.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RowsDecoded {
```
anforowicz, line 86: just a thought / no actionable feedback here: I notice that `RowsDecoded::TopDown(image_height)` has in practice the same meaning as `RowsDecoded::BottomUp(image_height)`.  I think that's okay and we don't want to do something about this if it would complicate things (e.g. introducing a 3 variants: `RowsDecoded::AllRows`, `RowsDecoded::PartialTopDown(rows)`, `RowsDecoded::PartialBottomUp(rows)`).  But I just wanted to share this in case you or others have any other thoughts on this.

src/codecs/bmp/decoder.rs
anforowicz: > Can we make this an enumeration?

FWIW I am trying to augment the upstream guidelines to explicitly ask for this: https://github.com/rust-lang/api-guidelines/pull/290

src/codecs/bmp/decoder.rs
197g: Good note, I have no concern about it. The type is only complete with the layout and we don't need to be comparable for correctness here. In practice it'll be odd for any decoder to switch the variant during decoding or to go backwards, but also just _odd_. In the sense of the struct not being `Ord` we might assume monotonicity while decoding but similar to the standard library's collections we could panic in the worst case. And we can be significantly more lenient. The one interface idea I'd really want for ergonomically _doing_ something with the type is:

```
impl ImageBuffer<_, _> {
    fn copy_decoded(&mut self, other: &Self, up_to: RowsDecoded);
    fn copy_decoded_incremental(&mut self, other: &Self, from_to: Range<&RowsDecoded>);
}
```

Those have no problem with misbehaved arguments though. They can make a best effort (i.e. second can fallback to a full copy on mismatched variants), do nothing (logic bug stays logic bug) or be written to return a result. I'd say we'll see. Either way there's no problem here with multiple variants of the same meaning either.

src/codecs/bmp/decoder.rs
197g: I'd favor the second variant and `TopDown`, `BottomUp` is pretty good. If we are to add others then the description should elude to the algorithm that defines the direction itself. Terms use for the direction are consistent with existing documentation and variables where bottom = biggest Y in several locations (and always assuming X-major layout). One concern motivating clarity here is that this must refer to the matrix order of pixels and not some visual notion, `ImageReader` will apply the orientation _after_ decoding the pixel matrix.

Maybe one alternative: `HeightIncreasing`, `HeightDecreasing`. Not more concise but `Height` is even more coupled to Y in the public part of the API than `bottom` is. My mind isn't made up between them.

I had considered `imagemagick`'s gravity option which uses `North`, `South`, …; but maybe that's a good distinction in that we can then keep those available for possible other interpretations in 'intended' image orientation and explicitly avoid using them for the pixel matrix's sample order.

telecos: Updated to TopDown / BottomUp to keep consistency with current internal naming and comments.

Issue #2747, opened 2026-01-28
BMP decoder feature parity with Chromium


telecos:
In our integration with Chromium we have executed the test suites for BMP and compared current results from Chromium BMP decoder with the BMP decoder from image crate. We are creating this issue for tracking the gaps and reference it in following PRs that will try to close the gap whenever it makes sense.

After analyzing differences, we have spot these discrepancies:

1. ~OS/2 v2 palette format not supported - The decoder doesn't recognize OS/2 BITMAPINFOHEADER2 format (64-byte header with variable palette entry sizes of 3 or 4 bytes). Chromium supports this legacy format. Affected tests: `pal8os2v2.bmp, pal8os2v2-16.bmp, pal8os2v2-sz.bmp`~ PR #2770
2. ~RLE transparency handling - When RLE-encoded images have "delta" escapes or truncated data leaving unfilled pixels, Chromium treats these as transparent (alpha=0), while the Rust decoder leaves them as opaque black. Affected tests: `pal4rlecut.bmp, pal4rletrns.bmp, pal8rletrns.bmp`~ PR #2773 
3. ~Oversize palette validation - Chromium accepts BMPs where biClrUsed exceeds the maximum for the bit depth (e.g., 300 colors for 8-bit) by clamping to the valid range. The Rust decoder rejects these files entirely. Affected test: `pal8oversizepal.bmp`~ PR #2764 
4. ~Stricter Header validation - Chromium tolerates invalid header values like biPlanes != 1 or biClrUsed > 256 and proceeds with decoding. The Rust decoder returns an error. Affected tests: `badplanes.bmp, badpalettesize.bmp`~ PR #2764
5. ~Truncated RGB16 data handling - For truncated 16-bit RGB files, Chromium decodes available rows and fills remaining pixels. The Rust decoder fails at metadata parsing. Affected test: `rgb16-880.bmp`~ PR #2764
6. ~RLE top-down validation — Chromium accepts top-down RLE images (negative height) as valid, while the Rust decoder rejects them during header validation. Affected test: `rletopdown.bmp`~ PR #2764 
7. ~BI_ALPHABITFIELDS alpha mask not read - For BITMAPINFOHEADER with compression type BI_ALPHABITFIELDS (6), the alpha bitmask wasn't being read, causing transparent pixels to render as opaque. Affected test: `rgba32abf.bmp`~ PR #2762 
[...]

PR #2764, opened 2026-02-10
Allow lenient validation in BMP decoder

telecos:
Chromium decoder accept certain files which are not valid per strict BMP spec. This change is introduce a lenient mode to allow these files only when lenient mode is enabled, while Strict mode is still default mode.

This change should implement discrepancies 3, 4, 5 and 6 from #2747


src/codecs/bmp/decoder.rs
```diff
@@ -11,6 +11,18 @@ use crate::error::{
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum BmpSpec {
+    /// Strictly follow the BMP specification.
+    /// Rejects files that violate spec constraints (e.g., RLE with top-down).
+    #[default]
+    Strict,
+    /// Allow some non-conformant files that violate some spec constraints
+    /// but still can be decoded at best effort.
+    Lenient,
+}
```
anforowicz, line 24: not necessarily immediately actionable / I just want to explicitly discuss: I note that we have a choice here, whether or not we want to expose strict-vs-lenient modes through the public API.  FWIW in case of the `png` crate, we [made a choice](https://github.com/image-rs/image-png/pull/569#issuecomment-2642062285) to *not* expose this through the public API, but retain an option to do this later by still detecting strict errors internally (and then ignoring them).  OTOH, this "ignore errors in unimportant metadata" behavior may be easier to implement in the `png` crate, where `fn parse_chunk` is a central place that parses auxiliary PNG chunks (I am not sure if such a central place exists in BMP decoder here).

At any rate, I think we should explicitly decide which of the following 3 options we want to take:

1. Just switch to lenient behavior
2. Introduce `BmpSpec::Strict` vs `BmpSpec::Lenient` but don't expose it through the public API + always internally use `BmpSpec::Lenient` when constructing the decoder through public API.
3. Introduce `BmpSpec::Strict` vs `BmpSpec::Lenient` and expose it through the public API

IIUC in `png` case we took option2.  But I think I would vote for whichever option results in the simplest/most-readable BMP decoder code (which may be option1?).  And initial version of this PR proposes option3.

WDYT?  /cc @197g and @fintelia

telecos: Thanks for sharing the context on what was done for png, really helps me. My initial thought was to respect current behavior as default and then add the new lenient mode as opt-in. However, as you point, exposing the new lenient mode parameter in public API seems not optimal and better if we can avoid it. If we go with option 1 and make lenient as default, I think that for current consumers of BMPDecoder it seems less impactful to accept some edge cases which weren't accepted before than the opposite.

Assuming we made option lenient be the default, we should also consider then that there are some `DecoderError` cases that no longer will happen: `MoreThanOnePlane`, `ImageTypeInvalidForTopDown`, `BitfieldMaskMissing`

Also, there is one caveat with exposing the Strict/Lenient parameter on API and setting either of the values as default: Controlling it from public API only works for resumable decoder because most of these validation happen during header parsing and this happen on ctor for the regular ctor.

@197g / @fintelia what are your thoughts on this?

197g: For context, we do have the same open point with jpeg (#2684) right now; And in terms of exposed API it would be very good to have a uniform interface for them. It's somewhat unfortunate that JPEG defaults to strict whereas png defaults to permissive but the context explains that: for png (and bmp) we had looked very specifically at browsers as references for acceptable fallback behavior and have general test coverage for expectations. Although arguably we could always do better with more extensive summarized documentation of that.

Getting back to the issue at hand, as stated with png, (2.) is agreeable and (1.) would also work. If the complexity of an option stays similarly small as here, I concur with the assessment.

telecos: Thanks for the additional context @197g ! I have updated the PR to go with option 2. : remove public function and make Lenient be the default.

Seeing that we have similar issues across codecs, it may make sense to define the strictness setter in ImageDecoder trait so that all codecs may have same exposed method (maybe it can be introduced after the change in #2684 ? If that looks fine I can go ahead after completing this PR and the JPEG PR with a follow up PR to introduce it and implement it initially for JPEG and BMP)

telecos: After changing default to `BmpSpec::Lenient` bad_bmps regression test started failing because it is expecting all these files to be rejected, while some of them are now accepted with lenient validation. I have disabled it so we can re-enable once we can externally configure `BmpSpec::Strict` (or whatever enum we define for strictness in ImageDecoder)

telecos: After looking more carefully at these regression tests, I found that not all of them apply to the lenient validation, there are still bad cases that should be also rejected in Lenient mode. In fact some of these files I already am using for the lenient vs strict test, so I removed them from bad_bmps test

197g (approved):
Good approach of having `lenient` separately to really test the mode while preserving existing `bad_bmp` test. LGTM.

PR #2765, opened 2026-02-11
Refactor: Convert `chunks_exact[_mut]` iterators to `as_chunks[_mut]` iterators

okaneco:
- Use the `as_chunks[_mut]` iterators when the chunk size is constant.

There's a clippy lint for this in the works but it's not merged yet.

---

This is a refactoring PR with no functional changes intended.

The newer chunk iterators give the compiler more information to potentially eliminate bounds checks or autovectorize/unroll beyond what the old style iterators could do. There are still cases today, with the latest LLVM upgrade, where simple-looking `chunks_exact` code doesn't autovectorize until upgraded to use `as_chunks` iterators. The new iterators are also more resilient to autovectorizer regressions such as the ones we saw starting in 1.87 (LLVM 20) which were fixed recently with the bump to LLVM 22.

This allowed for some other local cleanups, such as:
- Cleaning up several `try_into().unwrap()` array casts
- Removing the need for some byteorder read/write calls
- Changing some small function signatures to take arrays as input arguments instead of slices

### DIFF
--- a/src/imageops/fast_blur.rs
+++ b/src/imageops/fast_blur.rs
@@ -332,11 +332,10 @@ fn box_blur_horizontal_pass_impl<T, const CN: usize>(
             let dst_chunks = dst.as_chunks_mut::<CN>().0.iter_mut();
             let data_section_chunks = data_section.as_chunks::<CN>().0.iter();
             let advanced_kernel_part_chunks = advanced_kernel_part.as_chunks::<CN>().0.iter();
-            for ((dst, src_previous), src_next) in dst_chunks
+            for ((dst_chunk, src_previous), src_next) in dst_chunks
                 .zip(data_section_chunks)
                 .zip(advanced_kernel_part_chunks)
             {
-                let dst_chunk = &mut dst[..CN];
                 dst_chunk[0] = rounding_saturating_mul(weight0, weight);
                 if CN > 1 {
                     dst_chunk[1] = rounding_saturating_mul(weight1, weight);
--- a/src/imageops/filter_1d.rs
+++ b/src/imageops/filter_1d.rs
@@ -100,7 +100,7 @@ where
     for (x, dst) in (image_size.width..(image_size.width + pad_w))
         .zip(row_buffer.as_chunks_mut::<N>().0.iter_mut().rev())
     {
-        let old_x = x.max(0).min(image_size.width - 1);
+        let old_x = x.min(image_size.width - 1);
         let old_px = old_x * N;
         let src_iter = &source_row[old_px..(old_px + N)];
         for (dst, src) in dst.iter_mut().zip(src_iter.iter()) {
@@ -269,13 +269,11 @@ fn filter_symmetric_column<T, F, const N: usize>(
                 }
             }
 
-            let shaped_dst0 = &mut chunk[..16];
+            let (shaped_dst0, shaped_dst1) = chunk.split_at_mut(16);
 
             for (src, dst) in store0.iter().zip(shaped_dst0.iter_mut()) {
                 *dst = src.to_();
             }
-
-            let shaped_dst1 = &mut chunk[16..32];
 
             for (src, dst) in store1.iter().zip(shaped_dst1.iter_mut()) {
                 *dst = src.to_();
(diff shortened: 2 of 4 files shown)


src/imageops/fast_blur.rs
```diff
@@ -329,10 +329,12 @@ fn box_blur_horizontal_pass_impl<T, const CN: usize>(
...
-                .zip(data_section.chunks_exact(CN))
-                .zip(advanced_kernel_part.chunks_exact(CN))
+            let dst_chunks = dst.as_chunks_mut::<CN>().0.iter_mut();
+            let data_section_chunks = data_section.as_chunks::<CN>().0.iter();
+            let advanced_kernel_part_chunks = advanced_kernel_part.as_chunks::<CN>().0.iter();
+            for ((dst, src_previous), src_next) in dst_chunks
+                .zip(data_section_chunks)
+                .zip(advanced_kernel_part_chunks)
             {
                 let dst_chunk = &mut dst[..CN];
```
197g: Note: should no longer be necessary. Afaik we used this to force an early bounds check (not sure if we really looked at the produced byte code back then) as the order of accessing the destination produces several consecutive but the array now has constant length and this should all be a simple constant propagation.

src/imageops/sample.rs
```diff
@@ -1509,38 +1510,9 @@ where
...
-                    .unwrap()
-                    .map(|v| NumCast::from(FloatNearest(v)).unwrap());
-
-                *dst = *Pixel::from_slice(&channels);
-            }
-            _ => unreachable!(),
-        }
+    let transient_dst_chunks = transient_dst.as_chunks_mut::<CN>().0.iter_mut();
+    for (dst, src) in out.pixels_mut().zip(transient_dst_chunks) {
+        *dst = *Pixel::from_slice(&src.map(|v| NumCast::from(FloatNearest(v)).unwrap()))
```
197g: This feels a little terse, I'd appreciate a two line version with the `array::map` on its own line. Also makes it a little more obvious that we're not calling `Iterator::map` here by producing a sized result.

src/imageops/filter_1d.rs
```diff
@@ -97,8 +97,8 @@ where
         }
     }
 
-    for (x, dst) in
-        (image_size.width..(image_size.width + pad_w)).zip(row_buffer.chunks_exact_mut(N).rev())
+    for (x, dst) in (image_size.width..(image_size.width + pad_w))
+        .zip(row_buffer.as_chunks_mut::<N>().0.iter_mut().rev())
     {
         let old_x = x.max(0).min(image_size.width - 1);
```
okaneco: Wondering why `max(0)` on `usize` isn't triggering `clippy::unnecessary_min_or_max`.

src/imageops/fast_blur.rs
okaneco: Good eye, I spotted another place in `imageops/filter_1d.rs` where that occurred. Otherwise, I don't think it was a common pattern.

src/imageops/sample.rs
okaneco: I wasn't sure if it was too much one line.

I agree generally and assume every `map` to be an iterator at first glance unless I see the explicit type annotation in an assignment. There were a few cases of map on array inside closures here that I had to read carefully during this PR, like in the cicp file.

src/imageops/filter_1d.rs
awxkee: I believe `.max(0)` is a clippy bug and can be safely removed. Interestingly, this presents in my original code for years.

src/codecs/avif/ycgco.rs
```diff
@@ -119,12 +119,14 @@ fn process_halved_chroma_row_cgco<
 
     let bias_y = range.bias_y as i32;
     let bias_uv = range.bias_uv as i32;
-    let y_iter = y_plane.chunks_exact(2);
-    let rgb_chunks = rgba.chunks_exact_mut(CHANNELS * 2);
```
awxkee, line 123: Interesting fact: Rust with LLVM was always bad in autovectorizing YUV chroma subsampling.
Now, with `as_chunks[_mut]` it can at least autovectorize YUV [encoding with chroma subsampling](https://godbolt.org/z/djeMchTKT) but not [decoding](https://godbolt.org/z/3hef6bncn).

src/imageops/filter_1d.rs
okaneco: Just pushed a commit to remove it.

197g:
Thank you!

Issue #2417, opened 2025-02-05
Faster `copy_from` implementation


Jasper-Bekkers:
After https://github.com/image-rs/image/pull/909 w/ the move to remove unsafe code, I think (didn't test extensively) the implementation of `copy_from` got significantly slower. At the very least, it became a recent bottleneck for me that I spent some time looking at.

The implementation of `copy_from` is a pretty straight forward `get_pixel`/`put_pixel` loop over the image data where the function calls aren't inlined and are performing bounds checks on every access, even tho the preconditions of the function already do the bounds check beforehand.

In my specific case I managed to ~3x increase the performance, without re-introducing unsafe code, by copying over scanlines at a time. The code below lives in my project because I can live without it having to be a generalized implementation (as you see - it would need to be adapted to `GenericImage`/`GenericImageView` to be able to live in this project properly).

I'm submitting it here (rather then as a pull request) because I found a way to entirely not go down this code path at all anymore, however, I'm sharing it because I think it's useful to this project still and can potentially be adapted.

```rust
trait FastCopyFrom {
    type Pixel: image::Pixel;

    fn fast_copy_from(&mut self, other: &image::RgbaImage, x: u32, y: u32);
}

impl FastCopyFrom for image::RgbaImage {
    type Pixel = <image::RgbaImage as GenericImageView>::Pixel;
    fn fast_copy_from(&mut self, other: &image::RgbaImage, x: u32, y: u32) {
        // Do bounds checking here so we can use the non-bounds-checking
        // functions to copy pixels.
        if self.width() < other.width() + x || self.height() < other.height() + y {
            return;
        }

        let other_data = other.as_raw();
        let self_w = self.width();

        let mut self_data = self.as_flat_samples_mut();
        let mut self_slice = self_data.as_mut_slice();
        let other_w = other.width();

        for k in 0..other.height() {
            let other_start = ((k * other_w) * 4) as usize;
            let other_end = (((k + 1) * other_w) * 4) as usize;

            let self_start = (((x) + ((k + y) * self_w)) * 4) as usize;
            let self_end = self_start + (other_w * 4) as usize;

            self_slice[self_start..self_end].copy_from_slice(
                &other_data[other_start..other_end]
            );
        }
    }
}
```

## Timings before

![Image](https://github.com/user-attachments/assets/984c0f08-c486-4e29-9a04-d657cd1802d6)

## Timings after 

[...]

Issue #2553, opened 2025-08-24
Proposal: Have `GenericImage` only abstract over `ImageBuffer` and `SubImage`


fintelia:
The `GenericImage` trait has been in a weird place for a while. The way it is currently defined, it isn't really possible to write efficient algorithms using it (#2300), but at the same time if we impose too many layout constraints then it just ends up being redundant with the `ImageBuffer` struct (#1118, probably others). 

But if we made the trait abstract over both `ImageBuffer` and sub-images of `ImageBuffer`s, then I think the trait could be useful. It would mean all our image processing algorithms would work on both without requiring inefficient pixel-by-pixel accesses to implement them. The traits would look something like:

```rust
pub trait GenericImage {
    type Pixel: Pixel;
    fn buffer(&self) -> &[<Self::Pixel as Pixel>::Subpixel];
    fn dimensions(&self) -> (u32, u32);
    fn row_stride(&self) -> usize;

    // Provided methods
    fn get_pixel(&self, ...) -> &Self::Pixel { ... }
    fn sub_image(&self, ...) -> SubImage<Self::Pixel, ...> { ... }
    ...
}

pub trait GenericImageMut: GenericImage {
    fn buffer_mut(&mut self) -> &mut [<Self::Pixel as Pixel>::Subpixel];

    // Provided methods
    ...
}
```

(This also switches `GenericImageView` -> `GenericImage` and `GenericImage` -> `GenericImageMut`, which I think is more aligned with naming conventions, but I don't feel strongly about.)

The `SubImage` struct could then be something like this (or be broken into separate `SubImage` and `SubImageMut` types):
```rust
pub struct SubImage<P: Pixel, Buffer: Deref<Target = [P::Subpixel]>> {
    inner: Buffer,
    dimensions: (u32, u32),
    row_stride: usize,
    _phantom: std::marker::PhantomData<P>,
}

impl<P: Pixel, Inner: Deref<Target = [P::Subpixel]>> GenericImage for SubImage<P, Inner> { ... }
impl<P: Pixel, Inner: Deref<Target = [P::Subpixel]> + DerefMut> GenericImageMut for SubImage<P, Inner> { ... }
```

197g:
[...]

PR #2766, opened 2026-02-12
Speed up copy for ImageBuffer

197g:
Benchmarks TBD but expecting ~60% performance or more. Motivation, context and approach from the first commit and follows pretty directly what was presented in the thoughts on #2553.

This has been a long standing issue. The lack of generalization in Rust
makes it generally hard to provide specialized implementations of
`copy_from` that can make full use of the layout information to improve
code generation. Since the argument is generic we stay with the default
implementation even for the very strict targets (and source) of
`ImageBuffer`.

We sidestep a large part of the problem by speeding up only the copy
from a well-described layout: `View` (a strided matrix). Then an
optional method on `GenericImageView` lets us query if the source is
equivalent to such a copy. We wont reap the benefits for user defined
trait impls but importantly the crate's own `ImageBuffer` and `SubImage`
can provide it The latter also has a generic impl for

	SubImage<D> were <D as Deref>::Target: GenericImageView

Due to this we do not use a double-dispatch approach (adding a
`copy_into_buffer`) but this which composes purely at runtime.

Also note that DynamicImage can not provide it but the goal is to
replace this, too.

Closes: #2417, possibly #2553 (@fintelia addresses the immediate concern but we expanded the scope for reasons of extensibility concerns; but sealing the trait is already represented in another issue; your call)

### DIFF
--- a/src/images/generic_image.rs
+++ b/src/images/generic_image.rs
@@ -1,5 +1,5 @@
 use crate::error::{ImageError, ImageResult};
-use crate::flat::{View, ViewOfPixel};
+use crate::flat::ViewOfPixel;
 use crate::math::Rect;
 use crate::traits::Pixel;
 use crate::{ImageBuffer, SubImage};
@@ -126,8 +126,17 @@ pub trait GenericImageView {
         ImageBuffer::new(width, height)
     }
 
-    /// If the buffer has a fitting layout, return a view of the samples descriptor.
-    fn as_samples(&self) -> Option<ViewOfPixel<'_, Self::Pixel>> {
+    /// If the buffer has a fitting layout, return a canonical view of the samples.
+    ///
+    /// This is the basis of optimization and by default return `None`. It lets consumers of
+    /// generic images access the sample data through a canonical descriptor of its layout directly
+    /// instead of pixel-by-pixel. This provides more efficient forms of access that the
+    /// [`GenericImageView`] trait itself does not demand from all its implementations.
+    ///
+    /// Implementation of this method should be cheap to call.
+    ///
+    /// If implemented, a [`SubImage`] proxy of this image will provide a sample view as well.
+    fn to_pixel_view(&self) -> Option<ViewOfPixel<'_, Self::Pixel>> {
         None
     }
 }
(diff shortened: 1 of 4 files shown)


src/images/generic_image.rs
```diff
@@ -253,6 +263,32 @@ pub trait GenericImage: GenericImageView {
...
+    /// Copy pixels from a regular strided matrix of pixels.
+    fn copy_from_samples(
+        &mut self,
+        samples: View<&[<Self::Pixel as Pixel>::Subpixel], Self::Pixel>,
+        x: u32,
+        y: u32,
+    ) -> ImageResult<()> {
+        // Even though the implementation is the same, do not just call `Self::copy_from` here to
+        // avoid circular dependencies in careless implementations.
+        if self.width() < samples.width() + x || self.height() < samples.height() + y {
```
telecos: Use test_in_bounds from Rect to consider integer overflows?

197g: Good spot, that can be moved to the new code. Which I also changed to return a Result since the error is the same in all uses.

197g:
Rebased so we can run complete benchmarks. Although, it does not seem necessary to justify this any further.

<details>

```
copy_from               time:   [2.9246 ms 2.9460 ms 2.9688 ms]
                        change: [-25.357% -24.766% -24.149%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

copy_at                 time:   [345.24 µs 346.54 µs 348.18 µs]
                        change: [-64.403% -63.765% -63.034%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 19 outliers among 100 measurements (19.00%)
  1 (1.00%) low mild
  9 (9.00%) high mild
  9 (9.00%) high severe

copy_view               time:   [552.29 µs 553.94 µs 556.02 µs]
                        change: [-78.457% -77.900% -77.345%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild
  11 (11.00%) high severe

copy_fill               time:   [704.57 µs 705.18 µs 705.79 µs]
                        change: [-78.583% -78.402% -78.128%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

copy_strides            time:   [875.41 µs 876.36 µs 877.22 µs]
                        change: [-74.813% -74.428% -73.418%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
```

</details>

telecos (approved):

PR #2767, opened 2026-02-12
Define SpecCompliance enum for controlling spec strictness

telecos:
This PR is originated after #2764 and #2684 where it was spotted the need to allow controlling strictness on the decoder for accepting files which may not be fully compliant to the spec. A solution was implemented to #2764 defaulting to lenient validation for BMP (which is also the case for JPEG and PNG for example). In #2684 it was spotted the need to allow forcing strictness of validation to allow control from the caller and a solution based on a boolean parameter was proposed. This PR is proposed a common solution for codecs allowing such control (at this point BMP and JPEG).

### DIFF
--- a/src/codecs/hdr/decoder.rs
+++ b/src/codecs/hdr/decoder.rs
@@ -6,6 +6,7 @@ use std::{error, fmt};
 use crate::error::{
     DecodingError, ImageError, ImageFormatHint, ImageResult, UnsupportedError, UnsupportedErrorKind,
 };
+use crate::io::image_reader_type::SpecCompliance;
 use crate::{ColorType, ImageDecoder, ImageFormat, Rgb};
 
 /// Errors that can occur during decoding and parsing of a HDR image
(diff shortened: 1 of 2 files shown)


mstoeckl:
> This PR is proposed a common solution for codecs allowing such control (at this point BMP and JPEG).

There is also HDR; it currently defaults to strict instead of lenient, but that may be worth changing:
https://github.com/image-rs/image/blob/48ec3b30876434ba876f3791b5a5cfa93fe9af32/src/codecs/hdr/decoder.rs#L172-L174

telecos:
> > This PR is proposed a common solution for codecs allowing such control (at this point BMP and JPEG).
> 
> There is also HDR; it currently defaults to strict instead of lenient, but that may be worth changing:
> 
> https://github.com/image-rs/image/blob/48ec3b30876434ba876f3791b5a5cfa93fe9af32/src/codecs/hdr/decoder.rs#L172-L174

Thanks for pointing to this. I added a similar new_with_strictness ctor which receives the SpecCompliance enum as parameter. Also found that there was a "new_nostrict" and tagged it as deprecated suggesting usage of the new one.

RunDevelopment:
I think this is a good idea.

I want to mention that this feature would also be useful for plugin decoders. E.g. the (hopefully soon to be) DDS decoder in `image-extras` currently uses [a permissive mode](https://github.com/RunDevelopment/image-extras/blob/678ec1c97517c1c4a6332d0aa54f6adeb5434ca8/src/dds.rs#L42) to parse DDS headers. So if this PR gets merged, I think #2672 should be updated to add a method like: `ImageDecoder::set_spec_compliance(&mut self, compliance: SpecCompliance) -> ImageResult<()>` (mirroring the setter for limits). This would allow plugins to take advantage of this setting too.

PR #2770, opened 2026-02-13
Adds support for decoding BMP files using the OS/2 2v2 and OS/2-specific 24-bit RLE compression.

telecos:
This change is bringing support for OS/2 BMP files, as part of the work to bring Chromium feature parity #2747 :

- OS/2 v2 headers have variable size from 16 to 64 bytes. The first 40 bytes are layout-compatible with BITMAPINFOHEADER, so the existing `ParsedInfoHeader::parse` is reused with zero-padding for fields beyond the actual header size. OS/2-specific fields at offsets 44-64 (halftoning, color encoding, etc.) are read and discarded. Sizes that match known Windows headers (40, 52, 56) are still decoded via their existing paths since the layouts are compatible.

- RLE24 (compression value 4) encodes inline BGR triples instead of palette indices. On Windows, value 4 means BI_JPEG (legacy pass-through printing, already rejected as unsupported), so the decoder disambiguates based on header type. Rather than creating a separate function, RLE24 support is integrated directly into the existing `read_rle_data` method alongside RLE4 and RLE8, with only the encoded-run and absolute-mode branches dispatching on `image_type`. This eliminates the need for the RLEInsn intermediate enum and ensures all three RLE modes share the same escape handling, delta logic, checkpoint bookkeeping, and resumable decoding infrastructure.

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -2474,6 +2474,13 @@ mod test {
                 has_icc_profile: true,
                 top_down: false,
             },
+            TestCase {
+                path: "tests/images/bmp/images/rgb24rle24.bmp",
+                is_rle: true,
+                has_palette: false,
+                has_icc_profile: false,
+                top_down: false,
+            },
         ];
 
         for TestCase {


197g:
Can you expand on the halftoning fields that are discarded? If I'm not mistaken confusing them for a Windows header would get them interpreted as color masks. Also, is the data in those halftoning fields necessary for color fidelity and is discarding them convention?

telecos:
> Can you expand on the halftoning fields that are discarded? If I'm not mistaken confusing them for a Windows header would get them interpreted as color masks. Also, is the data in those halftoning fields necessary for color fidelity and is discarding them convention?

This proposed implementation is trying to replicate behavior from Chromium's described in the comment here:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/bmp/bmp_image_reader.cc;l=258

As there was not much detail on the fields not read from the OS/2 header, I tried to dig a bit on the specifics of these headers, and found this description document:

https://www.fileformat.info/format/os2bmp/egff.htm

The OS/2 v2 header (up to 64 bytes) extends the Windows (40 bytes) with these additional parameters:

```
	WORD   Units;            /* Type of units used to measure resolution */
	WORD   Reserved;         /* Pad structure to 4-byte boundary */
	WORD   Recording;        /* Recording algorithm */
	WORD   Rendering;        /* Halftoning algorithm used */
	DWORD  Size1;            /* Reserved for halftoning algorithm use */
	DWORD  Size2;            /* Reserved for halftoning algorithm use */
	DWORD  ColorEncoding;    /* Color model used in bitmap */
	DWORD  Identifier;       /* Reserved for application use */
```

Specifically for halftoning parameters, my understanding is that they are mainly informative about halftoning algorithms that may be used for generating the pixel data stored in that BMP.

> Rendering specifies the halftoning algorithm used on the bitmap data. A value of 0 indicates that no halftoning algorithm was used; 1 indicates error diffusion halftoning; 2 indicates Processing Algorithm for Noncoded Document Acquisition (PANDA); and 3 indicates super-circle halftoning.

Also, regarding the ColorEncoding parameter, there is a comment stating that RGB is the only value supported:

> ColorEncoding specifies the color model used to describe the bitmap data. The only valid value is 0, indicating the RGB encoding scheme.

197g (approved):
Thank you, that is sufficient elaboration for me.

PR #2771, opened 2026-02-15
Support BMP V4/V5 color space fields

telecos:
This PR is extending the BMP decoder to parse color space information from [`BITMAPV4HEADER`][v4] and [`BITMAPV5HEADER`][v5] structures. Previously, only the `PROFILE_EMBEDDED` ICC path in V5 headers was handled; this PR adds support for the `bV4CSType` / `bV5CSType` field and its associated data across all V4/V5 color space types.

The parsed parameters are used to create an equivalent ICC profile which can be exposed through existing icc_profile() method, removing thus the need to expose the bmp-specific V4/V5 parameters.

This work is covering discrepancy number 8 from #2747 

[v4]: https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapv4header
[v5]: https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapv5header


197g (commented):
LGTM, looks to me like the ICC approach is tenable.

@fintelia Still I don't want to decide this alone since it'll be a template for other formats. The advantage here is that we don't invent our own representation; the disadvantage that if ICC proves too rigid for some representations we may have to diverge in some decoder and that certainly creates complexity. Also in case we need access to concrete data after all (i.e the ICC file fails to represent important information that later turns out is important) then we will end up with more code than having _one_ solution with all the maintenance overhead.

So, preliminary approval from my side pending other directional input.

Shnatsel (approved):
Looks good to me for BMP. We can cross the bridge of applying this to other formats once we get to it.

Really the question for other format is "what happens if there's several colorspace indicators at once" like CICP and ICC in PNG at the same time, but the answer is probably going to be "pick one to expose" and if people need lower-level access they can just use the format crate directly.

PR #2777, opened 2026-02-21
Remove unnecessary type constraint for `ImageBuffer::as_flat_samples`

RunDevelopment:
I noticed that the constraint `Container: AsRef<[P::Subpixel]>` was completely unnecessary for the implementation of `as_flat_samples`. Since the impl already has the constraint `Container: Deref<Target = [P::SubPixel]>`, we can just use that.

However, while unnecessary for the implementation of `as_flat_samples`, callers of `as_flat_samples` still had to satisfy the constraint. With the unnecessary restriction gone, most generic instances of `ImageBuffer` can use `as_flat_samples` now.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1320,13 +1320,7 @@ where
     }
 
     fn to_pixel_view(&self) -> Option<ViewOfPixel<'_, Self::Pixel>> {
-        let samples = FlatSamples {
-            samples: &*self.data,
-            layout: self.sample_layout(),
-            color_hint: None,
-        };
-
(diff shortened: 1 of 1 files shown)


197g:
There is code in the `<ImageBuffer as GenericImageView>::to_pixel_view` block that could previously not make use of `as_flat_samples` due to the mentioned bound but can. Would you get that as internal motivation for the relaxation?

RunDevelopment:
Sure. I updated `to_pixel_view` to use `as_flat_samples` as you suggested.

197g (approved):
Thanks!

PR #2780, opened 2026-02-24
Use leaked reference for decoding hooks

RReverser:
We don't actually need the overhead of `Arc` reference counting - since we're doing append-only registration, it's simpler to leak the `Box` and then deal with the static reference we can easily pass around with no overhead.

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -5,7 +5,7 @@ use std::{
     collections::HashMap,
     ffi::{OsStr, OsString},
     io::{BufRead, BufReader, Read, Seek},
-    sync::{Arc, RwLock},
+    sync::RwLock,
 };
 
 use crate::{ImageDecoder, ImageResult};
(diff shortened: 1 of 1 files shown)


197g:
I don't really see this as much easier? It'll certainly be harder to reverse the change if the `'static` infects systems that can (for some reason) not work with `Arc` to the hook. And that'd be necessary to support deletion of hooks.

RReverser:
> It'll certainly be harder to reverse the change if the `'static` infects systems that can (for some reason) not work with `Arc` to the hook.

It's purely an internal change, not touching the public API. Or do you mean if more parts of codebase somehow start relying on it?

> And that'd be necessary to support deletion of hooks.

Is that something you plan to do? I thought (was hoping) registration is intended to be one-way, sort of merely as a workaround for Rust's lack of static registration.

RReverser:
For me it was mostly not so much about ease as about avoiding tiny but unnecessary overhead.

When you do `Box::new(|| { ...some closure that doesn't capture anything from outside environment... })`, Rust doesn't actually allocate anything as it can determine that closure size is 0 and so allocation would be no-op anyway. Subsequently converting to reference is also zero-cost, as well as passing it around when retrieving the hook.

In constract, `Arc::new(closure)` must always actually allocate on heap at least for the reference counter, and then increment/decrement it on each decoding hook retrieval.

Considering that registration is append-only, that seemed like unnecessary complexity and overhead, but if you plan to change it to allow hook deletion, then yeah, this wouldn't make sense and feel free to close this PR.

197g:
> Is that something you plan to do? I thought (was hoping) registration is intended to be one-way, sort of merely as a workaround for Rust's lack of static registration.

Not sure. Following Debian's adoption it seems at least plausible that one may load `image` via a shared library. The only way to properly destroy the resources on unloading is to de-register all the hooks again. I'd like to see that option stay open. A few cycles of overhead don't seem like a good tradeoff for unavoidable leaks.

RReverser:
Fair enough.

PR #2785, opened 2026-02-25
Add `TiffEncoder::set_compression()` to allow controlling compression mode

Shnatsel:
Allow controlling TIFF compression algorithm and predictor in `TiffEncoder`.

This PR also changes the default from uncompressed to balanced DEFLATE with a horizontal predictor. This is as a more sensible default, and matches imagemagick.

`TiffEncoder::set_compression()` diverges from the crate's apparent convention of `new_with_quality`, because I think it's rather silly to call this quality when it's a lossless format. I can change this if you prefer.

I've manually tested the default compression method with `wondermagick` (which is what I need this for) but I haven't verified the more obscure options like PackBits produce correct results.

### DIFF
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -699,7 +699,7 @@ pub enum Predictor {
     /// and the floating-point predictor (predictor 3) to floating-point samples.
     ///
     /// Predictor 3 improves compression of floating-point data, but is not
-    /// supported by all software. In particular, Safari, macOS Preview,
+    /// supported by all software. For example, Safari, macOS Preview,
     /// and ImageJ cannot read TIFF files that use predictor 3.
     IntegerAndFloat,
 }


RunDevelopment:
Related: #2251

> `TiffEncoder::set_compression()` diverges from the crate's apparent convention of `new_with_quality`

Given the way decoders are currently moving towards setters instead of constructor args, I would expect that encoders will move towards setters too. So you're likely just ahead of the trend.

src/codecs/tiff.rs
```diff
@@ -600,3 +640,22 @@ impl<W: Write + Seek> ImageEncoder for TiffEncoder<W> {
...
+#[non_exhaustive]
+pub enum CompressionType {
+    /// No compression whatsoever
+    Uncompressed,
+    /// [PackBits](https://en.wikipedia.org/wiki/PackBits) compression, a variant of RLE compression scheme. Fast, primitive compression with a poor compression ratio.
+    PackBits,
+    /// Legacy [LZW](https://en.wikipedia.org/wiki/Lempel%E2%80%93Ziv%E2%80%93Welch) algorithm. Not recommended.
+    Lzw,
+    /// Fast compression using DEFLATE algorithm
+    Fast,
```
fintelia: We should probably call this `DeflateFast` or similar, since we may want to add `zstd` and/or `webp` compression in the future and both of them can support multiple compression levels

Shnatsel: Done

Shnatsel:
Float compression was implemented in the `tiff` crate in https://github.com/image-rs/image-tiff/pull/342 but I don't think that should block this PR.

fintelia:
I'm now wondering if it would it make sense to land support for deflate levels 1-9 in the `tiff` crate first, and then just have a single `Deflate(u8)` variant?

Shnatsel:
I think it would, but it's an API break for the `tiff` crate.

197g:
I'm very much okay with that. There's a bunch of API to clean up there anyways (all the direct tag reads on `Decoder`). Maybe we could fit some more writer cleanup in; if you have found some more after the immediate pain points resolved recently?

Shnatsel:
Turns out we *can* expose precise DEFLATE compression level from `tiff` crate without an API break: https://github.com/image-rs/image-tiff/pull/348

I'll update this PR once the `tiff` crate PR is merged.

Shnatsel:
Ah, sorry, that wasn't sufficient after all. I've opened https://github.com/image-rs/image-tiff/pull/351 for a semver-breaking change instead.

Shnatsel:
Marking as draft pending upstream work in `tiff` crate, notably https://github.com/image-rs/image-tiff/issues/357

Shnatsel:
This is now done, but I'm keeping it as a draft until the `tiff` crate with the necessary changes is released to crates.io

PR #2788, opened 2026-02-26
Implement seek_relative for hooks::GenericReader

mstoeckl:
Directly forwarding Seek::seek_relative to the underlying BufReader can make decoders using the format hook mechanism much more efficient in edge cases with many short seeks. See also https://github.com/image-rs/image-extras/pull/38.


fintelia (approved):

RunDevelopment:
@197g I think this change should also be backported to 0.25.*. We missed it in 0.25.10, which causes perf issues for `image-extras`' [ICNS decoder](https://github.com/image-rs/image-extras/pull/38#issuecomment-3963380354).

PR #2797, opened 2026-02-26
pnm: Avoid allocations while parsing binary data

mstoeckl:
This PR pushes the reading of binary PNM data into the `pnm::decoder::Sample` trait. (The current approach first reads it into a separately allocated buffer and applies a trait function to the buffer.) All postprocessing needed to map from the sample structure to `image`'s types can then be done in place.

Note: the inline PbmBit::from_bytes expansion, while hard for a compiler to unroll, is still a bit faster than the more generic utils::expand_bits. It is possible to reduce its evaluation time to ~40% of the current value by unrolling it, like https://github.com/mstoeckl/image/commit/4b4b7bb113b68528655bff67a98a7bb994fbae1f. (I wouldn't be surprised if, using SIMD, bit unpacking is ultimately memory bandwidth limited; but getting to that point would be a lot more work.) However, I have not included this because PBM decoding is already reasonably fast, loop unrolling produces a lot of edge cases to test, and PBM decoding time is currently dominated by something else (sample rescaling).

To help avoid regressions if the PBM unpacking is changed in the future, I've added a new 1x9 PBM sample image. I created it and license it to match the project (Apache-2.0 OR MIT), and also under CC-0.

### DIFF
--- a/src/codecs/pnm/decoder.rs
+++ b/src/codecs/pnm/decoder.rs
@@ -11,7 +11,6 @@ use crate::color::{ColorType, ExtendedColorType};
 use crate::error::{
     DecodingError, ImageError, ImageResult, UnsupportedError, UnsupportedErrorKind,
 };
-use crate::io::ReadExt;
 use crate::{utils, ImageDecoder, ImageFormat};
 
 /// All errors that can occur when attempting to parse a PNM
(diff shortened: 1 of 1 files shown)


197g:
> Note: the inline PbmBit::from_bytes expansion … is still a bit faster than the more generic utils::expand_bits.

I believe we should at one point improve the utils method. Since it was written we learned a bit about designing such things for performance and there are just three variants that should be generated separately at compiletime (1, 2, 4 bits). One difference being here of course that it also operations within the same slice so this does not necessarily relate to this PR at all. Good information bit though.

197g (approved):
I like this approach. LGTM.

PR #2804, opened 2026-02-28
Less cast pixel codegen 2

197g:
Simplifies `cast_pixels_from_subpixels` with _almost_ on runtime losses and confusingly some small gains.. I'm not too worried about them since LLVM is very bad at vectorizing these still. (As indicated by the comment maybe that must be pushed into the `FromPrimitive` trait where we have a customization point). So an optimization will come at some point and that will as well be easier with the structure after the PR.

In any case, by having a slice of initialized contents we sidestep LLVM tripping over the capacity check associated with `push` or `extend_from_slice`. Improving codegen necessitates removing extend via `flat_map` here but the standard library specialization for these is the only _safe_ API with reasonable performance here in the general case. I tried an approach where each case was manually chunked up so to reduce the number of checks but ultimately that did not work transparently with monomorphization—trusting LLVM with simple code is just better than complexity.

Removes ~10% of LLVM lines from our `--no-default-features` crate.

### DIFF
--- a/src/metadata/cicp.rs
+++ b/src/metadata/cicp.rs
@@ -1277,6 +1277,8 @@ impl CicpRgb {
         }
     }
 
+    // Note: ~50% faster than the output-based fallback in Luma8->LumaA8 and Luma8->LumaA16 codegen
+    // so this one stays with `flat_map` for now.
     fn subpixel_cast_luma_to_luma_alpha<FromSubpixel, IntoSubpixel>(
         output: &mut Vec<IntoSubpixel>,
         buffer: &[FromSubpixel],
@@ -1284,17 +1286,16 @@ impl CicpRgb {
         FromSubpixel: ColorComponentForCicp,
         IntoSubpixel: ColorComponentForCicp + FromPrimitive<FromSubpixel> + Primitive,
     {
-        let pixels = buffer.len() / LayoutWithColor::Luma.channels();
-        Self::create_output::<IntoSubpixel>(output, pixels, LayoutWithColor::LumaAlpha);
-
         let map_channel = <IntoSubpixel as FromPrimitive<FromSubpixel>>::from_primitive;
-        let default_alpha = <IntoSubpixel as Primitive>::DEFAULT_MAX_VALUE;
 
-        let output_chunks = output.as_chunks_mut::<2>().0;
-
-        for (&l, out) in buffer.iter().zip(output_chunks) {
-            *out = [map_channel(l), default_alpha];
-        }
+        output.extend(buffer.iter().copied().flat_map(|l| {
+            [
+                map_channel(l),
+                // Crucially inlined here. When I tried to move this out then it no longer
+                // optimizes any better than the output method. (#2804).
+                <IntoSubpixel as Primitive>::DEFAULT_MAX_VALUE,
+            ]
+        }));
     }
 
     fn subpixel_cast_luma_alpha_to_luma<FromSubpixel, IntoSubpixel>(
@@ -1304,15 +1305,11 @@ impl CicpRgb {
         FromSubpixel: ColorComponentForCicp,
         IntoSubpixel: ColorComponentForCicp + FromPrimitive<FromSubpixel> + Primitive,
     {
-        let pixels = buffer.len() / LayoutWithColor::LumaAlpha.channels();
-        Self::create_output::<IntoSubpixel>(output, pixels, LayoutWithColor::Luma);
-
         let map_channel = <IntoSubpixel as FromPrimitive<FromSubpixel>>::from_primitive;
+
         let buffer_chunks = buffer.as_chunks::<2>().0;
 
-        for (&[l, _], out) in buffer_chunks.iter().zip(output) {
-            *out = map_channel(l);
-        }
+        output.extend(buffer_chunks.iter().map(|&[l, _]| map_channel(l)));
     }
 }
 


fintelia (approved):

197g:
Some insane residual result:

## Fast

```
        let map_channel = <IntoSubpixel as FromPrimitive<FromSubpixel>>::from_primitive;

        output.extend(buffer.iter().copied().flat_map(|l|[map_channel(l), <IntoSubpixel as Primitive>::DEFAULT_MAX_VALUE]));

```

## Slow



```
        let map_channel = <IntoSubpixel as FromPrimitive<FromSubpixel>>::from_primitive;
        let default_alpha = <IntoSubpixel as Primitive>::DEFAULT_MAX_VALUE;

        output.extend(buffer.iter().copied().flat_map(|l|[map_channel(l), default_alpha]));

```

You see that correctly, the only difference is manually hosting the constant _into_ the function. I mean, this is far too brittle to be reliable but I'll just have a go at doing that manually for the non-flat_map variants and see if that gains something.

197g:
@fintelia For Luma<>LumaA it seems that the regression would be more than a bit. Only one of them uses the `flat_map` construct. We obviously pay a little bit in terms of compile time but the gains are unfortunately a bit too large for my comfort. Overall this PR is quite a win for the RGBA->RGB cases.

```
cast_dynamic_rgba8_rgb8 time:   [66.344 µs 66.365 µs 66.392 µs]
                        change: [-23.295% -21.370% -19.196%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
```

I'd say we take the tradeoff route and prioritize performance. Alright with you?

<details>

```
  Lines                 Copies              Function name
  -----                 ------              -------------
  295862                7624                (TOTAL)
   15804 (5.3%,  5.3%)    12 (0.2%,  0.2%)  image::imageops::filter_1d::filter_symmetric_column
   12678 (4.3%,  9.6%)    12 (0.2%,  0.3%)  image::imageops::fast_blur::box_blur_horizontal_pass_impl
    7332 (2.5%, 12.1%)    12 (0.2%,  0.5%)  image::imageops::filter_1d::filter_2d_separable
    7047 (2.4%, 14.5%)   126 (1.7%,  2.1%)  <core::iter::adapters::zip::Zip<A,B> as core::iter::adapters::zip::ZipImpl<A,B>>::next
    6336 (2.1%, 16.6%)    12 (0.2%,  2.3%)  image::imageops::filter_1d::filter_symmetric_row
    6324 (2.1%, 18.8%)    12 (0.2%,  2.4%)  image::imageops::filter_1d::filter_2d_separable_ring_queue
    5841 (2.0%, 20.7%)   149 (2.0%,  4.4%)  core::iter::traits::iterator::Iterator::zip
    5580 (1.9%, 22.6%)    12 (0.2%,  4.6%)  image::imageops::filter_1d::make_arena_row
    5532 (1.9%, 24.5%)    63 (0.8%,  5.4%)  core::array::try_from_fn_erased
    5392 (1.8%, 26.3%)   141 (1.8%,  7.2%)  <core::iter::adapters::zip::Zip<A,B> as core::iter::adapters::zip::ZipImpl<A,B>>::new
```

</details>

src/metadata/cicp.rs
```diff
@@ -1195,42 +1209,108 @@ impl CicpRgb {
...
+        let map_channel = <IntoSubpixel as FromPrimitive<FromSubpixel>>::from_primitive;
+
+        output.extend(buffer.iter().copied().flat_map(|l| {
+            [
+                map_channel(l),
+                // Crucially inlined here. When I tried to move this out then it no longer
+                // optimizes any better than the output method. (#2804).
+                <IntoSubpixel as Primitive>::DEFAULT_MAX_VALUE,
+            ]
+        }));
```
RunDevelopment, line 1298: From my experience, `flat_map` is almost always horrible for perf in hot paths. Have you tried using `map` and then flattening the `Vec<[T; N]>` with `Vec::into_flattened`?

197g: I did not, thank you for the recommendation. However it does not appear to be doing that well here (-20%? need to look at the assembly). Also, the inverse does not yet appear to be stable so there's no way to utilize an existing allocation here like in the other alternatives. We'd rely on codegen fusing a dealloc-alloc pair by taking the existing `Vec`, dropping it, and `collect`-ing into a new vector instead. Not doing that and starting from a vector without capacity results in even worse perf for some reason ….

If you have ideas of how to get this to work in codegen feel free to run a follow up PR. Maybe I'm measuring wrong.

197g:
```diff
diff --git a/src/metadata/cicp.rs b/src/metadata/cicp.rs
index 82cb762b..aadb062f 100644
--- a/src/metadata/cicp.rs
+++ b/src/metadata/cicp.rs
@@ -1183,10 +1183,12 @@ impl CicpRgb {
         assert!(buffer.len().is_multiple_of(from_layout.channels()));
         let pixels = buffer.len() / from_layout.channels();

-        let mut output: Vec<IntoSubpixel> = vec_try_with_capacity(pixels * into_layout.channels())
+        let mut output: Vec<IntoSubpixel> = Vec::new();
+
+            // vec_try_with_capacity(pixels * into_layout.channels())
             // Not entirely failsafe, if you expand luma to rgba you can get a factor of 4 but at
             // least this will not overflow. And that's why I'm a fan of in-place operations.
-            .expect("input layout already allocated with appropriate layout");
+            //.expect("input layout already allocated with appropriate layout");

         // In most cases we perform a seemingly wasteful initialization, by initializing the output
         // before writing. However, we gain it back in codegen. If we did not have the vector exist
@@ -1287,15 +1289,22 @@ impl CicpRgb {
         IntoSubpixel: ColorComponentForCicp + FromPrimitive<FromSubpixel> + Primitive,
     {
         let map_channel = <IntoSubpixel as FromPrimitive<FromSubpixel>>::from_primitive;
-
-        output.extend(buffer.iter().copied().flat_map(|l| {
-            [
-                map_channel(l),
-                // Crucially inlined here. When I tried to move this out then it no longer
-                // optimizes any better than the output method. (#2804).
-                <IntoSubpixel as Primitive>::DEFAULT_MAX_VALUE,
-            ]
-        }));
+        core::mem::take(output);
+
+        let chunk_output = buffer
+            .iter()
+            .copied()
+            .map(|l| {
+                [
+                    map_channel(l),
+                    // Crucially inlined here. When I tried to move this out then it no longer
+                    // optimizes any better than the output method. (#2804).
+                    <IntoSubpixel as Primitive>::DEFAULT_MAX_VALUE,
+                ]
+            })
+            .collect::<Vec<_>>();
+
+        *output = Vec::into_flattened(chunk_output);
     }

     fn subpixel_cast_luma_alpha_to_luma<FromSubpixel, IntoSubpixel>(
```

This is what I tried

PR #2807, opened 2026-02-28
Remove const N in fast blur vertical pass

awxkee:
Removed redundant const in fast blur vertical pass. There are no functional changes.


197g:
@fintelia As a data point for PR strategy, there's a difference for the backporting workflow. With squash merges the only commit that is easily backported is the squash itself since the other objects are never pulled from upstream. On the other hand it's only a single commit that needs to be cherry-picked for it.

PR #2814, opened 2026-03-01
Version 0.25 clippy

197g:


197g:
Ahhh, I missed the SemVer compatibility for the new `loop_count` method.

PR #2817, opened 2026-03-02
Expose available YCbCr support from image-tiff crate in image crate (jpeg-compressed tiff and (1,1) ycbcr sampling for now)

mpospelova:
This PR enables support for available YCbCr Tiff images decoding from the `image-tiff` crate in the image crate. Previously, any Tiff with YCbCr was rejected immediately. This change exposes the YCbCr color type from the image-tiff decoder and implements the necessary color space conversion to Rgb8. Coefficients to transform between YCbCr and Rgb are extracted from tag 529.

Following Tiff images as per the `image-tiff` crate will be decoded successfully:
- JPEG-compressed Tiffs (handled by the JPEG library)
- Non-JPEG Tiffs (supported only for (1,1) sampling)

Test images are from https://gitlab.com/libtiff/libtiff-pics, and transformed with ImageMagick (specifically oxford.tif and ycbcr-cat.tif) 

Since `image-tiff` does not yet support other types of YCbCr Tiff images, such as other compression methods or subsamplings, the decoding will not work for those specific cases for now.  However the plan is to add upsampling support to be more libtiff conform.


src/codecs/tiff.rs
```diff
@@ -460,6 +469,22 @@ pub struct TiffEncoder<W> {
     icc: Option<Vec<u8>>,
 }
 
+fn ycbcr_to_rgb8(ycbcr: &[u8]) -> [u8; 3] {
```
awxkee: There is broad support for YUV formats in the AVIF module which I believe could be reused.

src/color.rs
```diff
@@ -164,6 +160,9 @@ pub enum ExtendedColorType {
     /// Pixel is 16-bit CMYK
     Cmyk16,
 
+    /// Pixel is 8-bit YCbCr
+    YCbCr8,
```
RunDevelopment, line 168: I don't think adding YCbCr here is a good idea. 

The main problem I see is that YCbCr isn't strictly *one* color type. There are multiple ways to [convert RGB <-> YCbCr](https://en.wikipedia.org/wiki/YCbCr#Conversions), each producing different results. I would say that all of them are technically different color types.

src/codecs/tiff.rs
```diff
@@ -460,6 +469,22 @@ pub struct TiffEncoder<W> {
...
+
+    let r = y + 1.402 * cr;
+    let g = y - 0.34414 * cb - 0.71414 * cr;
+    let b = y + 1.772 * cb;
+
+    [
+        r.clamp(0.0, 255.0).round() as u8,
+        g.clamp(0.0, 255.0).round() as u8,
+        b.clamp(0.0, 255.0).round() as u8,
+    ]
```
RunDevelopment: 2 tips for better performance:

1. `as u8` already clamps to 0-255. Rust guarantees this, so we don't need to do it ourselves.
2. `f32::round` is ***super slow*** on x86. Luckily, there's a trick we can use to very closely approximate `f32::round`: just add 0.5. This is around 10-20x faster. (The trick doesn't work well for negative inputs, but that's not an issue here.)
```suggestion
    [ (r + 0.5) as u8, (g + 0.5) as u8, (b + 0.5) as u8]
```

src/color.rs
197g: Neither is any of the others (hence the Cicp need in ImageBuffer). It should really be read as color model instead. That might be a rename worth for `1.0` but I always found the term more confusing to beginners than ours.

RunDevelopment: Right. Color type only describes channel number+types+bits, not the way those colors are interpreted. My bad.

In that case, YCC8 should be fine.

mpospelova:
@RunDevelopment Thank you so much for your comment! I have added code to read out tag 529, and use its coefficients when transforming between ycbcr and rgb. The default coefficients 601 are used if the tag is not set. I also added a check on tag 530, and now an error is thrown if the tiff image is non-jpeg compressed and subsampling is anything else than (1,1) (the image-tiff crate also throws an error in this case).

src/codecs/tiff.rs
```diff
@@ -190,7 +281,8 @@ fn check_sample_format(sample_format: u16, color_type: tiff::ColorType) -> Resul
         ColorType::RGB(k) => k,
         ColorType::RGBA(k) => k,
         ColorType::GrayA(k) => k,
-        ColorType::Palette(k) | ColorType::YCbCr(k) => {
+        ColorType::YCbCr(k) if k == 8 => k,
+        ColorType::Palette(k) => {
```
197g:
```suggestion
        ColorType::YCbCr(8) => 8,
        ColorType::Palette(k) | ColorType::YCbCr(k) => {
```

src/codecs/tiff.rs
```diff
@@ -413,6 +505,13 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
                     out_row.copy_from_slice(&utils::expand_bits(1, width, in_row));
                 }
             }
+            DecodingResult::U8(v) if self.original_color_type == ExtendedColorType::YCbCr8 => {
+                let [lr, lg, lb] = self.ycbcr_coefficients;
+                let mut out_cur = Cursor::new(buf);
+                for ycbcr in v.chunks_exact(3) {
+                    out_cur.write_all(&ycbcr_to_rgb8(ycbcr, lr, lg, lb))?;
+                }
```
197g: See the surrounding code on style here. Use of `chunks_exact(3)` is no longer up to date and `Cursor` is odd over slice iteration. Was there a point of error handling length expectations? That is better handled as an explicit branch into the appropriate `ImageError` rather than a fake IO

src/codecs/tiff.rs
```diff
@@ -460,6 +559,18 @@ pub struct TiffEncoder<W> {
...
+    let y = f32::from(ycbcr[0]);
+    let cb = f32::from(ycbcr[1]) - 128.0;
+    let cr = f32::from(ycbcr[2]) - 128.0;
+
+    let r = y + cr * 2.0 * (1.0 - lr);
+    let b = y + cb * 2.0 * (1.0 - lb);
+    let g = (y - lr * r - lb * b) / lg;
+
+    [(r + 0.5) as u8, (g + 0.5) as u8, (b + 0.5) as u8]
+}
```
197g, line 575: Since `(1 - lr)` etc. are constants across calls, this is going to be more efficient if it processed a whole chunk of `ycbcr` items at a time. The interface would then also be more explicit if it were:

```rust
fn ycbcr_to_rgb8(ycbcr: &[[u8; 3]], lr: f32, lg: f32, lb: f32, out: &mut [[u8; 3]] {

    let coeff_r = 2.0 * (1.0 - lr);
    let coeff_g = 2.0 * (1.0 - lb);
    let coeff_g = 1.0 / lg;
```


[`slice::as_chunks`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.as_chunks) can be used for conversion in the caller as appropriate.

src/codecs/tiff.rs
RunDevelopment: The first two match arms also use Cursor in this style. They likely just copied that. And I agree that it is super weird to use `Cursor` in this way.

src/codecs/tiff.rs
```diff
@@ -182,6 +195,84 @@ where
...
+
+    let subsampling = value.into_u16_vec().map_err(ImageError::from_tiff_decode)?;
+    if subsampling != [1, 1] {
+        return Err(ImageError::Unsupported(UnsupportedError::from_format_and_kind(
+            ImageFormat::Tiff.into(),
+            UnsupportedErrorKind::GenericFeature(
+                format!("Subsampling {:?} is not supported. Only (1,1) is supported for non-JPEG YCbCr.", subsampling),
+            ),
+        )));
+    }
```
RunDevelopment: I think it would be better to specify the default value for YCbCr Subsampling tag in code. When looking over this, I was quite confused why it errors if no value is defined until I looked it up in the spec.

(Sorry if my suggestion code doesn't compile. I wrote it without an IDE. I hope what it does is clear nonetheless.)

```suggestion
    let subsampling = decoder
        .find_tag(TAG_YCBCR_SUBSAMPLING)
        .map_err(ImageError::from_tiff_decode)?
        .map(|value| value.into_u16_vec())
        .transpose()
        .map_err(ImageError::from_tiff_decode)?;
        
    // defaults to 2,2 (half width, half height)
    let subsampling = subsampling.as_ref().map_or(&[2, 2], |v| v.as_slice());
    
    if subsampling != &[1, 1] {
        return Err(ImageError::Unsupported(UnsupportedError::from_format_and_kind(
            ImageFormat::Tiff.into(),
            UnsupportedErrorKind::GenericFeature(
                format!("Subsampling {:?} is not supported. Only (1,1) is supported for non-JPEG YCbCr.", subsampling),
            ),
        )));
    }
```

src/codecs/tiff.rs
197g: I had just looked at the closer arms, now the choice makes sense. #2824 demonstrates the better style well so may be used as a reference. Thank you.

RunDevelopment:
Code looks good IMO. 

Just one last nit about your test files: they are quite large. You added 381 KB in test files to the repo. Small images are generally preferred. Do you have smaller test images that you could use?

tests/images/tiff/testsuite/ycbcr_lzw_22.tif
197g, line 1: Where's this one from and is is encumbered copyright that we can't have in the repository? It's used as a test for subsampling _not_ being supported right now which does not require the image data itself. So if in doubt I would remove that one.

mpospelova: All of the images that I have added are from https://gitlab.com/libtiff/libtiff-pics. for `tests/images/tiff/testsuite/ycbcr_lzw_22.tif)`, I used the `ycbcr-cat.tif` image, and all other images are modified versions of `oxford.tif`.  But it does make sense to remove this test as you said, as it only tests if the code fails, which might become irrelevant after support is added. Thank you for your comment! :)

ronjakoi:
I would greatly benefit from this functionality. When could I expect a release that includes this contribution?

PR #2826, opened 2026-03-05
Update png dependency to 0.18.1

telecos:
Update png dependency to 0.18.1, which includes a fix required by Chromium. Chromium's dependency on image crate is currently based on main branch for BMP experiment (as we require all the changes for supporting resumable decoding). Now, we are starting work on rustification of ICO codec (preliminary work in https://skia-review.googlesource.com/c/skia/+/1165477) which requires enabling png feature, so we would need it including the latest fix from 0.18.1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ exr = { version = "1.74.0", default-features = false, optional = true }
 gif = { version = "0.14.1", optional = true }
 image-webp = { version = "0.2.0", optional = true }
 mp4parse = { version = "0.17.0", optional = true }
-png = { version = "0.18.0", optional = true }
+png = { version = "0.18.1", optional = true }
 qoi = { version = "0.4", optional = true }
 ravif = { version = "0.13", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


Shnatsel:
Cargo already permits you to use 0.18.1 with `image` crate. This change only makes it impossible to use `image` with earlier versions of `png`, which I don't think is desired in your case.

To use 0.18.1 in Chromium you should change your own `Cargo.lock` file (or its Blaze equivalent) to point to 0.18.1.

telecos:
> Cargo already permits you to use 0.18.1 with `image` crate. This change only makes it impossible to use `image` with earlier versions of `png`, which I don't think is desired in your case.
> 
> To use 0.18.1 in Chromium you should change your own `Cargo.lock` file (or its Blaze equivalent) to point to 0.18.1.

Thanks for pointing this. It seems I was having a local build issue due to the old patch we had to bring the png fix that did not cleanup well. I tried again cleaning output and I am avoiding the build error, sorry for the confussion.

PR #2827, opened 2026-03-05
Update to moxcms 0.8.0 version

telecos:
In Chromium we are also switch ICC parsing to use moxcms (WIP CL: https://chromium-review.googlesource.com/c/chromium/src/+/7588078). This requires some fix and struct exposure which are included in moxcms 0.8.0. However, this makes a dependency conflict as image crate depends on 0.7.4. This change is allowing both 0.7.4 and 0.8.x as at the moment there are no API changes. @awxkee , do you think it is ok to keep both as supported dependencies or better switch to 0.8.0?

Update: There is in fact some API change causing this clippy error when building. To support both we should tag with allow `clippy::useless_conversion` the callilng method 
```
error: useless conversion to the same type: `std::sync::Arc<dyn moxcms::TransformExecutor<f32> + std::marker::Send + std::marker::Sync>`
   --> src/metadata/cicp.rs:369:79
    |
369 |                    from.create_transform_f32(from_layout, into, into_layout, opt)
    |   _______________________________________________________________________________^
    |  |_______________________________________________________________________________|
370 | ||                     .map(Arc::<dyn moxcms::TransformExecutor<f32> + Send + Sync>::from)
    | ||_______________________________________________________________________________________^
371 | |                      .ok()
    | |_____________________- help: consider removing
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
    = note: `-D clippy::useless-conversion` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::useless_conversion)]`
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 bytemuck = { version = "1.8.0", features = ["extern_crate_alloc"] } # includes cast_vec
 byteorder-lite = "0.1.0"
-moxcms = { version = ">=0.7.4, <0.9" }
+moxcms = "0.8.0"
 num-traits = { version = "0.2.0" }
 
 # Optional dependencies
--- a/src/metadata/cicp.rs
+++ b/src/metadata/cicp.rs
@@ -342,9 +342,6 @@ impl CicpTransform {
     /// This is used with [`ConvertColorOptions`][`crate::ConvertColorOptions`] in
     /// [`ImageBuffer::copy_from_color_space`][`crate::ImageBuffer::copy_from_color_space`],
     /// [`DynamicImage::copy_from_color_space`][`DynamicImage::copy_from_color_space`].
-    // `Arc::from` is needed for moxcms 0.7.x (returns Box) but is a no-op in 0.8.x (returns Arc
-    // directly); allow the lint so both versions compile cleanly under `-D warnings`.
-    #[allow(clippy::useless_conversion)]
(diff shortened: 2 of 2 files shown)


awxkee:
The policy on my crates is like: once a new version published the previous one is effectively abandoned unless there is a security or functionality disaster. So yes, I do recommend to update the crate.

Additionaly, the builds from 0.7.* track are the worst in compiles times and binary size. This is not a problem for me I don't really care if it compiles 2s or 4s, but it's a known problem for some of image users. Because of that, I think allowing usage of two version or 0.7.* track isn't really wanted.

The migration shouldn't be actually complex. As far as I remember mostly redundant `Arc` removing here is needed.

However, the feature set in 0.8.* has changed. There are some features that may be dropped to compile only what image actually uses so @197g consulting might be needed.

telecos:
> The policy on my crates is like: once a new version published the previous one is effectively abandoned unless there is a security or functionality disaster. So yes, I do recommend to update the crate.
> 
> Additionaly, the builds from 0.7.* track are the worst in compiles times and binary size. This is not a problem for me I don't really care if it compiles 2s or 4s, but it's a known problem for some of image users. Because of that, I think allowing usage of two version or 0.7.* track isn't really wanted.
> 
> The migration shouldn't be actually complex. As far as I remember mostly redundant `Arc` removing here is needed.
> 
> However, the feature set in 0.8.* has changed. There are some features that may be dropped to compile only what image actually uses so @197g consulting might be needed.

Updated to 0.8.0 and fixed the clippy issue instead of disabling it.

PR #2829, opened 2026-03-07
Flip processing order of reference test

RunDevelopment:
As [requested](https://github.com/image-rs/image/pull/2802#issuecomment-4010011104).

This PR flips the processing order of the reference test, ensuring that every image in `tests/images` gets tested. This also adds missing reference images, totaling about 2-3 MB (guess).


mstoeckl:
The reference images for IPTC data (`tests/reference/jpg/iptc.jpg.png` and `tests/reference/jpg/iptc.jpg.png`) are unusually large; since the original tests for the IPTC images do not really need the precise image content, I will try to make an PR to resize the images later today. (Once I find a way to resize a PNG while exactly preserving metadata.) Edit: done!

mstoeckl:
Reminder, since it's been a while since this was looked at, and we may otherwise forget: please rebase/squash this PR, so that if it is merged without squashing commits, the old IPTC reference file PNG images will not be included in the repository history. (A brief review of the code finds no other new issues.)

RunDevelopment:
Done and thanks the review. I had to change the implementation a little though, so you might want to re-review.

Also, I found a new bug in GIF. Disposal methods don't work anymore. Seems like all animations with transparency are decoded incorrectly. Fun.

@197g Could you please also take a look at this? This has been open for 3 months. Almost every time I revisit this, I find new bugs in the code base simply by running `cargo test`.

197g (approved):
Alright, this sat for longer than necessary and I haven't come up with any real downsides. We don't have a good process for blessing images and this does not change it; if anything this surfaces bugs earlier and the eventual blessing process is quite a bit cleaner. With prior logic it would have to guess which references are meant to exist, with this logic the function points in the right direction.

PR #2835, opened 2026-03-10
Avif EXIF and bump ravif dav1d versions

197g:
Bump `dav1d` and `ravif` dependencies, use new `with_tiff` capability.


197g:
Note: targets the `v0.25` branch as I intend to cut a release after this.

PR #2838, opened 2026-03-10
Don't use `f32::round` for filter 1D

RunDevelopment:
Same as #2837. I removed `f32::round` and replaced it with `+ 0.5` for better perf on x86..

The `self.min(u16::MAX as f32)` might be unnecessary too. I left it in for now, because it maps NaN to `u16::MAX` instead of 0 (by the `as` cast). If NaN isn't a concern for filter 1D, then it can be removed too. Just let me know.

### DIFF
--- a/src/imageops/filter_1d.rs
+++ b/src/imageops/filter_1d.rs
@@ -194,7 +194,7 @@ impl ToStorage<u16> for u32 {
 impl ToStorage<u16> for f32 {
     #[inline(always)]
     fn to_(self) -> u16 {
-        self.round().min(u16::MAX as f32) as u16
+        (self.min(u16::MAX as f32) + 0.5) as u16
     }
 }
 


197g:
I'm actually not sure myself..

awxkee:
NaNs typically saturated to the lower bound if we assume ieee standard. However, NaNs appearing here is unlikely so I don’t think it’s important.

197g:
We might put it into compatibility notes but we don't provide a method that would filter float images but convert the result into a uint image. I don't think you could provoke a NaN-conversion from the image input except by the kernel; but then it applies to the whole image which does not seem so useful.

awxkee:
I think we've overlooked this a bit in those PRs, since I already put [filter](https://github.com/image-rs/image/blob/3a92b1efc2e3e8fe44b730157946cd291273e054/src/imageops/sample.rs#L1160) on not normal values. However, it's still possible to submit [here](https://github.com/image-rs/image/blob/3a92b1efc2e3e8fe44b730157946cd291273e054/src/imageops/sample.rs#L1033) NaN values. I don't see much point to mess with NaN inputs, if you see that different we could just return for any NaN input a default image filled with pre-defined value. `exp` is defined for all inputs except NaN so in practice except a system has malformed libm here should not be NaNs. As well we could scan kernel and if there is a NaN somewhere it's also possible to return a default image.
I could do PR with NaNs filter or return default image, or something else?

PR #2839, opened 2026-03-10
Return exit code to collect test coverage

RunDevelopment:
`cargo-llvm-cov` cannot write coverage data on Windows when using `.exit()`. See https://github.com/LukasKalbertodt/libtest-mimic/issues/39.

In this PR, I changed it to `.exit_if_failed()`. This allows LLVM cov to report coverage, fixing the issue.

### DIFF
--- a/tests/reference_images.rs
+++ b/tests/reference_images.rs
@@ -15,7 +15,7 @@ use image::{DynamicImage, ImageFormat};
 use libtest_mimic::{Arguments, Failed, Trial};
 use walkdir::WalkDir;
 
-fn main() {
+fn main() -> std::process::ExitCode {
     let mut trials = Vec::new();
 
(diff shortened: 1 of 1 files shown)


tests/reference_images.rs
```diff
@@ -187,7 +187,7 @@ fn main() {
     }
 
     let args = Arguments::from_args();
-    libtest_mimic::run(&args, trials).exit();
+    libtest_mimic::run(&args, trials).exit_if_failed();
 }
```
mstoeckl: I occasionally check test coverage while some tests are failing, so using [Conclusion::exit_code](https://docs.rs/libtest-mimic/latest/libtest_mimic/struct.Conclusion.html#method.exit_code) and returning the std::process::ExitCode from main() may work better.

PR #2841, opened 2026-03-10
Split PNM ImageFormat into subformats for PBM, PGM, etc.

### LINKED ISSUE #2810: DynamicImage::save writes PAM into .pgm paths
This happens in `image` v0.25.9 and the current git version (558bb4b4d7096afde40b82e9ea1624c6b93ce6a9, as I write).

## Expected

If `im` is a grayscale DynamicImage, then `im.save("out.pgm")` should create a [PGM](https://netpbm.sourceforge.net/doc/pgm.html) file (header P5).

## Actual behaviour

`im.save("out.pgm")` always produces a [PAM](https://netpbm.sourceforge.net/doc/pam.html) file (header P7, tupltype GRAYSCALE). 

(Support for PAM (P7) in other image format libraries is generally weaker than the simpler and older PNM formats (P1-P6), so some image viewers or programs cannot open this.)

## Reproduction steps

Build `image`'s `convert` example, run `convert gray.png gray.pgm` on a grayscale image (ex. [gray.png](https://github.com/user-attachments/assets/2999d6a6-50a6-4fa7-aa48-9326489e4531)), and inspect the resulting image content (ex. [gray_image.pgm.txt](https://github.com/user-attachments/files/25640789/gray_image.pgm.txt)). A reference PGM format output can be produced with `pngtopnm  < gray.png  > gray_netpbm.pgm` (ex. [gray_netpbm.pgm.txt](https://github.com/user-attachments/files/25640811/gray_netpbm.pgm.txt)).

## Details

`DynamicImage::save` calls `ImageFormat::from_path`, which maps the ".pgm" extension to ImageFormat::Pnm, and then calls `DynamicImage::save_with_format` with that format. The information about the extension is not passed on `PnmEncoder::new()`, which defaults to writing in PAM format.

A similar issue occurs for the ".pbm" and ".ppm" extensions. (The way ".pnm" is handled may also be incorrect, since https://netpbm.sourceforge.net/doc/index.html#formats describes PNM as referring to any  of PBM, PGM, or PPM.)

I'm not certain about the best way to resolve this at the moment; one solution could be to split ImageFormat::Pnm into four or five distinct enum values (ImageFormat::Pbm, ImageFormat::Pgm, ImageFormat::Ppm, which write in their specific format; ImageFormat::Pnm which writes any of Pbm/Pgm/Ppm, and ImageFormat::Pam which writes P7.)

mstoeckl:
This PR tries to fix https://github.com/image-rs/image/issues/2810.

It splits ImageFormat::Pnm into five different formats, so that each of the NetPBM formats has its own enum entry. This makes it possible for methods like DynamicImage::save to set the exact format based on the file extension, so that saving to a ".pgm" file will only try to save a PGM image.

This behavioral change may surprise anyone who previously was writing Rgba8 images to a ".pnm" file (actually producing PAM); this will now fail because PNM does not support alpha.

Other than this change, I have tried to avoid changing the behavior of the PnmDecoder, so that anyone using it directly will continue to see the same output. The only addition is a `with_dynamic_pnm_header()` method to allow autoselecting a PNM header.

I've also added some test images that cover all the main netpbm format variations; some of them are also used as reference images for updated PNM _encoding_ tests in `tests/save_pnm.rs`.

The test images were all made by me, and licensed Apache-2.0 OR MIT OR CC-0. I've checked their references against NetPBM's conversion (using `pamdepth 65535 <$in | pamtopng -verbose >$out.png`; the test image maxvals are chosen to be exactly convertible to 16 bit without rounding.).


fintelia:
I'm torn on this because I don't love having so many NetPBM formats, but I also don't see a better approach to fixing the issue

fintelia:
I think there might be an alternative approach to this. 

As I understand it, the problem is that some of the encoding methods in this crate take a `Path` but are implemented in terms of other methods that work with an `ImageFormat`. But there's nothing stopping us from changing them to internally pass both the desired format and the original path (if known) when constructing an encoder. If we did that, PNM output could properly dispatch to the right variant based on the actual extension and we wouldn't need to change the public API other than adding a note to the relevant docs.

mstoeckl:
> But there's nothing stopping us from changing them to internally pass both the desired format and the original path (if known) when constructing an encoder. If we did that, PNM output could properly dispatch to the right variant based on the actual extension and we wouldn't need to change the public API other than adding a note to the relevant docs.

While this could work, I would actually consider this a larger break in behavior than splitting the PNM format, because users of the API now need to consider the note and think about whether they need to pass file extension information, even if they do not use PNM specifically. It would also introduce a mismatch between methods like DynamicImage::write_to and DynamicImage::save, where the latter can be used to write PAM files due to extension autodetection and the former can't.

fintelia:
We can make the change in the 1.0 release, so I'm less worried about behavior changes.

There is already the fully general `DynamicImage::write_with_encoder` so the other methods mostly serve as more convenient short-hands. I'm imagining that we'd keep the public method signatures completely intact and just make sure than any method which takes a path uses the file extension when constructing an encoder.

My main goal is avoiding clearly wrong behavior while avoiding an increase in the overall API surface to handle an edge case.

Issue #2842, opened 2026-03-11
Gaussian blur API is inconsistent with image editing programs


RunDevelopment:
In response to #2508, `DynamicImage::blur` was changed from taking a sigma to a radius. I believe that this was an error considering the reference to GIMP in [the discussion here](https://github.com/image-rs/image/pull/2496#discussion_r2155748765).

> I was actually confused by the sigma vs kernel size distinction.
>
> I just want a high-level API that I can call that roughly matches what I'd get from "blur radius" in GIMP

GIMP *uses* sigma but *calls* it size ([source code](https://gitlab.gnome.org/GNOME/gegl/-/blob/master/operations/common/gaussian-blur.c#L36-50)). So the API was changed from exactly what GIMP was doing to something else.

While I can't say what proprietary programs like Paint.net and Photoshop are doing under the hood, I can say that all these programs have one common property: if the size/radius/sigma is 0, no blurring takes place. This is not the case for `image`'s radius=0. radius=0 produces a sigma=0.5, which is still a significant blur. 

Considering the problems of radius (and the fact that it's basically arbitrary), I would propose changing `blur` back to using sigma. This makes the simple API consistent with other programs and all other blur-related methods on `DynamicImage`.

197g:
It's not a binary choice. We had done the rework into a `*Parameter` struct so that different constructors can be provided. Instead of having just one `blur` function there can be one exposed via that struct that let's you do _either_ parameter choice.

Then we can also discuss what is a reasonable parameter choice separately:
- For radius I tend towards clamping below at `3` pixels; that is not allowing `radius=0` by either method (panic, `Option` return, `.max(3)` on the diameter).
- For _sigma_ it seems reasonable to allow values small enough to not blur _at all_. Those should get fixed to return the image unchanged. And then returning it unchanged for subnormals and `0` as well is just a steady continuation into a limit result. I don't see any reason to panic here if the nearby open parameter space apart from the border is a no-op. Just no-op too. Advising that an unknown value is to be passed with `v.max(0.0)` to ensure there are no panics is easy; advising to do `v.max(std::f32::EPSILON)` is complicated for what reason?

RunDevelopment:
I'm specifically talking about `DynamicImage::blur`, because I want it to use a sensible default. 

> Instead of having just one blur function there can be one exposed via that struct that let's you do either parameter choice.

Sorry, but I'm not really sure what you are proposing here. Could you maybe give pseudo-code of the function signatures you envision?

[...]

PR #2843, opened 2026-03-11
Add sigma threshold for `new_from_sigma`

RunDevelopment:
As [I pointed out here](https://github.com/image-rs/image/issues/2842#issuecomment-4038951260), small sigmas quickly approach a blur kernel of `[0, 1, 0]`. This kernel performs no blurring and outputs the image unchanged. As such, we can define a sigma threshold below which we assume that the image is returned unchanged (i.e. no blurring).

To repeat the table of sigma values and resulting kernels:

| sigma | kernel |
| --- | --- |
| 0.5 | `0.10651 0.78699 0.10651` |
| 0.4 | `0.04039 0.91922 0.04039` |
| 0.3 | `0.00384 0.99233 0.00384` |
| 0.2 | `0.000004 0.999993 0.000004` |
| 0.15 | `2.2e-10 0.99999999956 2.2e-10` |
| 0.1 | `1.9e-22 1.0 1.9e-22` |
| 0.05 | `1.4e-87 1.0 1.4e-87` |

I selected sigma=0.2 as the threshold, because it's close enough to `0 1 0` that no blurring is performed on u16 images. If this isn't good enough for f32 images, we can lower it to 0.15.

Ideally, we could pick the threshold based on the precision of the image being blurred, but the API currently doesn't allow this. If we want to go this route, I believe this should be done in future PRs.

---

This PR is currently blocked by a bug that causes 1x1 kernels to be applied incorrectly. This bug should be fixed before this PR is merged. I'm currently working on a fix.

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -1207,9 +1207,9 @@ impl GaussianBlurParameters {
         /// precision.
         const IDENTITY_THRESHOLD: f32 = 0.16986436;
         if sigma < IDENTITY_THRESHOLD {
-            // Any kernel of size 1 is the identity, so sigma doesn't matter.
-            // However, we pick sigma=1 to avoid  potential issues with NaN,
-            // infinities, and subnormals.
+            // Any (normalized) kernel of size 1 is the identity, so sigma
+            // doesn't matter. However, we pick sigma=1 to avoid  potential
+            // issues with NaN, infinities, and subnormals.
             return GaussianBlurParameters {
                 x_axis_kernel_size: 1,
                 x_axis_sigma: 1.0,


src/imageops/sample.rs
```diff
@@ -1183,12 +1183,35 @@ impl GaussianBlurParameters {
...
-        assert!(sigma > 0.0, "Sigma must be positive");
+
+        /// As sigma gets smaller, the blur kernel quickly approaches all zeros
+        /// with the center being 1, which is an identity for 1D convolution.
+        /// I.e. no blur. So we define a threshold for sigma below which we
+        /// consider the blur to be an identity.
+        ///
+        /// The threshold is 0.2, because at sigma = 0.2, the center weight of
+        /// the kernel will 99.9993% of the sum of all weights.
+        const IDENTITY_THRESHOLD: f32 = 0.2;
```
197g: We could argue the threshold be based on this 'identity':

`1.0f32 + E.powf(-1.0 / 2. / vf32.powi(2)) == 1.0`

This seems to hold for `v=0.1725`. At that point the summation with a central value that is almost the same is a no-op. If we instead choose `.15` it would even hold with the central value being `1.0f32/255.` and at `0.125` it also holds for `1.0f32/65535`.

In the end I don't mind too much which of these is used. We could alternatively check this _later_ when converting to a kernel. With type information about the sample type this could be _much_ larger.

RunDevelopment: I really like that. I would just change it to `1.0f32 + 2.0 * E.powf(...) == 1.0`, because it accounts for all weights in a 3x1 kernel before normalization.

And, yes, we could pick a much larger threshold if we knew the sample type, but I don't think that really matters. So what if non-blurring blurs aren't perfectly optimized? They're still correct. Plus, my main goal was to remove the somewhat arbitrary restrictions around 0 and subnormals, and any decent threshold achieves that goal.

PR #2844, opened 2026-03-11
Improved Gauss kernel generation

RunDevelopment:
Changes:
- Added debug assertions to check assumptions in debug.
- Define a minimum sigma of 0.01. This prevents division by zero.
- Removed the scaling by $1/\sqrt{2\pi\sigma^2}$. Since kernel weights are normalized, this scaling is unnecessary.
- Remove use of `f32::powf` and cache more divisions. The performance impact will likely be negligible, but I'd rather remove functions of unknown complexity. 

Taken together, `get_gaussian_kernel_1d` now generates correct kernels for all $\sigma\ge 0$.

Since I only changed the behavior for edge cases (very small sigma), the behavior of our Gaussian blur remains unchanged.

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -8,7 +8,6 @@
 
 use num_traits::{NumCast, ToPrimitive, Zero};
 use std::f32;
-use std::ops::Mul;
 
 use crate::imageops::filter_1d::{
     filter_2d_sep_la, filter_2d_sep_la_f32, filter_2d_sep_la_u16, filter_2d_sep_plane,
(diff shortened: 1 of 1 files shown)


mstoeckl (approved):
This looks like a strict improvement over the preceding code; I do not see any new floating point issues.

(It doesn't fix all of them -- the calculation of `sum_norm` is of course increasingly affected by floating point error when `sigma` and `width` are huge. But fixing this would make the function much slower or much more complicated, and has no benefit when the image filtering code doesn't do the same. And huge `sigma` and `width` are incredibly slow to evaluate anyway.)

RunDevelopment:
Thanks for the review!

>  the calculation of sum_norm is of course increasingly affected by floating point error when sigma and width are huge.

Yes, but I don't think we need to worry about the precision of `sum_norm`. For a sigma of 10k, [I see a relative error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3f97c7794f090d326baf54d83744bdef) of around 0.0003%. (That said, we can lower this error by exploiting that the kernel is symmetric. This (approximately) halves the relative error.)

PR #2845, opened 2026-03-11
Simplified and explained `kernel_size_from_sigma`

RunDevelopment:
Changes:
- I simplified the arithmetic expression to make it obvious that it's just a simple linear function of `sigma`.
- I documented the reason for chosing this specific formula and explained why it works.
- I renamed `sigma_size` to `sigma_from_kernel_size` in order to make the relationship with `kernel_size_from_sigma` more clear.
- I replaced the branch with a `| 1` to make the number odd. This behaves the same.

---

@awxkee (git blame pointed to you, so I hope I got the right person.)
Could you please explain why you picked this formula? I think it's the inverse of the formula for kernel size->sigma you used above (and maybe [got from OpenCV](https://docs.opencv.org/4.x/d4/d86/group__imgproc__filter.html#gac05a120c1ae92a6060dd0db190a61afa)). However, [OpenCV itself uses a different formula](https://github.com/opencv/opencv/blob/4391bd4cb40a3702619f965a4d74eeae148a7ed7/modules/imgproc/src/smooth.dispatch.cpp#L289) for sigma->kernel size based on the 3 (or 4) sigma accounting for 99.73% (or 99.9937%) of the area of the curve. This one:

```c
ksize= round(sigma * (depth == CV_8U ? 3 : 4) * 2 + 1) | 1;
```

Why did you use this one? Edit: [See below](https://github.com/image-rs/image/pull/2845#issuecomment-4042886789).

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -1135,7 +1135,7 @@ impl GaussianBlurParameters {
         );
         let i_kernel_size = GaussianBlurParameters::round_to_nearest_odd(kernel_size);
         assert_ne!(i_kernel_size % 2, 0, "Kernel size must be odd");
-        let v_sigma = GaussianBlurParameters::sigma_size(kernel_size);
+        let v_sigma = GaussianBlurParameters::sigma_from_kernel_size(kernel_size);
         GaussianBlurParameters {
             x_axis_kernel_size: i_kernel_size,
             x_axis_sigma: v_sigma,
@@ -1172,8 +1172,8 @@ impl GaussianBlurParameters {
         assert_ne!(x_kernel_size % 2, 0, "Kernel size must be odd");
         let y_kernel_size = GaussianBlurParameters::round_to_nearest_odd(y_axis_kernel_size);
         assert_ne!(y_kernel_size % 2, 0, "Kernel size must be odd");
-        let x_sigma = GaussianBlurParameters::sigma_size(x_axis_kernel_size);
-        let y_sigma = GaussianBlurParameters::sigma_size(y_axis_kernel_size);
+        let x_sigma = GaussianBlurParameters::sigma_from_kernel_size(x_axis_kernel_size);
+        let y_sigma = GaussianBlurParameters::sigma_from_kernel_size(y_axis_kernel_size);
         GaussianBlurParameters {
             x_axis_kernel_size: x_kernel_size,
             x_axis_sigma: x_sigma,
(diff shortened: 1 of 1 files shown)


awxkee:
This is a sort of inversion of “optimal perceptual sigma”. ‘sigma_from_kernel’ ( or how does it named, I don’t remember ) should be optimal sigma itself. Kernel generation is not validated against OpenCV, and I tend to think it should not since OpenCV mostly assumes only noise removal whereas I’m not sure if noise removal is a goal here.

awxkee:
I thought about it for some time: I don’t see any obvious reason why exactly size->sigma should use one assumption, and backwards it’ll use something different.

awxkee:
```
ksize= round(sigma * (depth == CV_8U ? 3 : 4) * 2 + 1) | 1;
```
This kernel flattens too quick and you’ll see an ugly aliasing. You can find an example [here](https://github.com/image-rs/image/pull/2388) or check out how does it looks. As I mentioned this is fine for small kernels and aggressive noise removal, but it’s not something good for general imaging.

RunDevelopment:
> I don’t see any obvious reason why exactly size->sigma should use one assumption, and backwards it’ll use something different.

That's fair.

> This kernel flattens too quick and you’ll see an ugly aliasing.

Have you actually tried this formula? Because it doesn't do that. 

For 16- and 32-bit images, the OpenCV formula always selects greater kernel sizes. Ergo, better quality. Even for 8-bit images, the OpenCV formula selects kernel sizes at least as large as the inversion formula until sigma=6.25. And for sufficiently large sigma, the difference is between mean±3σ and mean±3.33σ, or 99.73% and 99.91% of the area of the curve. A difference so small that it will vanish with the quantization to u8.

As I understand it, this formula cannot produce artifacts of the magnitude you're claiming.

awxkee:
>  Have you actually tried this formula? Because it doesn't do that.

No. Prevously here was this formula, I skimmed the implementation and it seems that it was it, and it was producing aliasing.

However, I tried now and aliasing is obvious.

alias
ref

> For 16- and 32-bit images, the OpenCV formula always selects greater kernel sizes. Ergo, better quality. Even for 8-bit images, the OpenCV formula selects kernel sizes at least as large as the inversion formula until sigma=6.25. And for sufficiently large sigma, the difference is between mean±3σ and mean±3.33σ, or 99.73% and 99.91% of the area of the curve. A difference so small that it will vanish with the quantization to u8.

OpenCV isn't an important reference to me if that's something you're trying to say. I don't verify the most implementation against theirs because they don't do general imaging. Especially in terms of blur I rather don't want to know what they do.

RunDevelopment:
> However, I tried now and aliasing is obvious.

When I plug in their formula (=I replace the impl of `kernel_size_from_sigma`), I get the following when blurring with 50 sigma (=`.blur_advanced(new_from_sigma(50.))`):

Current formula:
old 50o

OpenCV formula:
ocv 50o

Near identical results. By which I mean that the maximum pixel error is 1 (u8, L1). In other words, the difference is within the quantization error.

<details><summary>Code for kernel_size_from_sigma</summary>

```rs
// ours
    fn kernel_size_from_sigma(sigma: f32) -> u32 { 
        (6.66666 * sigma - 2.3333333).max(3.) as u32 | 1
    }

// OpenCV
    fn kernel_size_from_sigma(sigma: f32) -> u32 { 
        (6.0 * sigma + 1.).round() as u32 | 1
    }
```

</details>

So what did you try?

awxkee:
My bad it used to be 2*sigma + 1 in previous implementation. I looked at the formula above and thought it was the same, but it isn't.

Yes, the "optimal" gaussian kernel size is roughly 6*σ, so when the kernel is reasonably large, the difference is negligible.

RunDevelopment:
Ahh, no worries. 

So in conclusion, you picked the inverse formula of kernel size->sigma for consistency and so both are "optimal" (by some definition of optimal). OpenCV's formula also works but has no significantly different effect.

Alright, I'll write that down. Thank you for your help!

Issue #2809, opened 2026-02-28
1.67× fast_blur speedup + gaussian blur improvements


art049:
Hello,

I've been profiling the blur operations and have a PR ready with three targeted optimizations: https://github.com/art049/image-rs/pull/2

### Wall-time results (`cargo bench`)

| Benchmark | Before | After | Speedup |
|-----------|--------|-------|---------|
| fast blur σ=3.0 | 40.1 ms | 24.0 ms | **1.67×** |
| fast blur σ=7.0 | 40.1 ms | 24.1 ms | **1.66×** |
| fast blur σ=50.0 | 40.8 ms | 24.6 ms | **1.65×** |
| gaussian blur σ=3.0 | 21.6 ms | 18.5 ms | **+17%** |
| gaussian blur σ=7.0 | 36.3 ms | 33.1 ms | **+10%** |
| gaussian blur σ=50.0 | 361.4 ms | 348.7 ms | **+4%** |

I also have a separate PR setting up [CodSpeed](https://codspeed.io/): https://github.com/art049/image-rs/pull/1. That's what I used to find these hot spots and iterate.
Full disclaimer: I created CodSpeed, and we'd be happy to provide it for free to `image-rs`, but I prefer to be upfront about it. 
Setting it up would require installing the GitHub App on the org (one-click, org admin needed), and you'd get PR comments like this: https://github.com/art049/image-rs/pull/2#issuecomment-3977609693

Both PRs are independent, happy to open the perf PR against upstream on its own. Let me know how you'd like to proceed!

awxkee:
I can't speak on behalf on maintainers but to me it seems those PRs a bit controversial compared to the approach I tried to implement in those blur algorithms and to optimization itself:

- It introduces unsafe code, which we specifically wanted to avoid.
- It moves toward modular arithmetic but doesn't use actual integer reciprocals. Instead, it uses integer division which kills throughput since division requires ~20 cycles and is a blocking operation.
- It changes `to_u8` and `to_u16` under assumption they are faster, but it holds only in two cases: target platform is not `aarch64` (or any other having direct rounding instruction ) and code is not compiled against `sse4.1` on x86 which have rounding instructions. Also a lot of branching isn't optimization as well. Real fast rounding should move your float to a range where float can be **the only integer** and move it back as so:
 ```rust
  const SHIFTER: f32 = ((1u32 << 23) + (1u32 << 22)) as f32;
[...]

PR #2846, opened 2026-03-12
perf: 5.5x faster `fast_blur` with u32 accumulators

art049:
Following #2809 
This was already reviewed a bit [here](https://github.com/art049/image-rs/pull/2)

## Changes
- **Use u32 integer accumulators for u8 fast blur** — The box blur hot path used f32 accumulators for all pixel types. For u8 images (the dominant case), every pixel went through to_f32/from_f32 conversions and software roundf. Replaced with u32 integer arithmetic.
- ~Replace roundf with fast truncation in FloatNearest::to_u8/to_u16 for non optimized architectures~

## Results

Walltime on x86 on a build without sse4.1
Simulation results from CodSpeed

### Walltime

| Benchmark | Baseline | Optimized | Speedup |
|-----------|----------|-----------|---------|
| fast blur: sigma 3.0 | 41.8 ms | 7.7 ms | **×5.5** |
| fast blur: sigma 7.0 | 42.1 ms | 7.7 ms | **×5.4** |
| fast blur: sigma 50.0 | 42.5 ms | 8.4 ms | **×5.0** |
| fast_blur | 41.9 ms | 8.5 ms | **×4.9** |

### Simulation

| Benchmark | Baseline | Optimized | Speedup |
|-----------|----------|-----------|---------|
| fast blur: sigma 3.0 | 254.3 ms | 45.3 ms | **×5.6** |
| fast blur: sigma 7.0 | 254.8 ms | 45.6 ms | **×5.6** |
| fast blur: sigma 50.0 | 260.6 ms | 50.3 ms | **×5.2** |
| fast_blur | 260.6 ms | 50.3 ms | **×5.2** |

### DIFF
--- a/src/imageops/fast_blur.rs
+++ b/src/imageops/fast_blur.rs
@@ -146,19 +146,53 @@ fn rounding_saturating_mul<T: Primitive>(v: f32, w: f32) -> T {
     T::clamp_nearest_from(v * w)
 }
 
+/// Precomputed reciprocal for fast integer division of u32 accumulators.
+///
+/// Replaces `(acc + ks/2) / ks` with a multiply-shift using the identity:
+///   `floor(n / d) == floor(n * ceil(2^32 / d) / 2^32)`
+/// which is exact for all `n` where `(d-1)*n < d * 2^32` (Granlund-Montgomery '94).
+/// For blur accumulators (`n <= kernel_size * 255`), this holds for all practical sizes.
+#[derive(Clone, Copy)]
+pub(crate) struct U8Weight {
+    reciprocal: u32, // ceil(2^32 / kernel_size)
+    rounding_bias: u32, // kernel_size / 2
+}
+
+impl U8Weight {
+    #[inline]
+    fn new(kernel_size: u32) -> Self {
+        debug_assert!(kernel_size >= 2);
+        Self {
+            // ceil(2^32 / ks) = floor((2^32 - 1) / ks) + 1 = (u32::MAX / ks) + 1
+            reciprocal: (u32::MAX / kernel_size) + 1,
+            rounding_bias: kernel_size / 2,
+        }
+    }
+
+    /// Compute `(acc + bias) / kernel_size` using reciprocal multiplication.
+    #[inline(always)]
+    fn apply(self, acc: u32) -> u8 {
+        let n = acc + self.rounding_bias;
+        ((n as u64 * self.reciprocal as u64) >> 32) as u8
+    }
+}
+
 /// Accumulator abstraction for box blur.
 /// `u8` uses `u32` integer accumulators; other types use `f32`.
 pub(crate) trait BlurAccum: Primitive + Sized {
     type Acc: Copy + Add<Output = Self::Acc> + AddAssign + Sub<Output = Self::Acc> + SubAssign;
+    type Weight: Copy;
 
     fn to_acc(self) -> Self::Acc;
     fn acc_zero() -> Self::Acc;
     fn acc_scale(acc: Self::Acc, count: usize) -> Self::Acc;
-    fn acc_to_store(acc: Self::Acc, kernel_size: usize) -> Self;
+    fn make_weight(kernel_size: usize) -> Self::Weight;
+    fn acc_to_store(acc: Self::Acc, weight: Self::Weight) -> Self;
 }
 
 impl BlurAccum for u8 {
     type Acc = u32;
+    type Weight = U8Weight;
     #[inline(always)]
     fn to_acc(self) -> u32 {
         self as u32
@@ -172,9 +206,12 @@ impl BlurAccum for u8 {
         acc * count as u32
     }
     #[inline(always)]
-    fn acc_to_store(acc: u32, kernel_size: usize) -> u8 {
-        let ks = kernel_size as u32;
-        ((acc + ks / 2) / ks) as u8
+    fn make_weight(kernel_size: usize) -> U8Weight {
+        U8Weight::new(kernel_size as u32)
+    }
+    #[inline(always)]
+    fn acc_to_store(acc: u32, weight: U8Weight) -> u8 {
+        weight.apply(acc)
     }
 }
 
@@ -182,6 +219,7 @@ macro_rules! impl_blur_accum_f32 {
     ($($t:ty),+) => { $(
         impl BlurAccum for $t {
             type Acc = f32;
+            type Weight = f32;
             #[inline(always)]
             fn to_acc(self) -> f32 {
                 self.to_f32().unwrap()
@@ -195,8 +233,12 @@ macro_rules! impl_blur_accum_f32 {
                 acc * count as f32
             }
             #[inline(always)]
-            fn acc_to_store(acc: f32, kernel_size: usize) -> Self {
-                rounding_saturating_mul(acc, 1.0 / kernel_size as f32)
+            fn make_weight(kernel_size: usize) -> f32 {
+                1.0 / kernel_size as f32
+            }
+            #[inline(always)]
+            fn acc_to_store(acc: f32, weight: f32) -> Self {
+                rounding_saturating_mul(acc, weight)
             }
         }
     )+ };
@@ -262,6 +304,7 @@ fn box_blur_horizontal_pass<P: BlurAccum, const CN: usize>(
     test_radius_size(width as usize, radius);
 
     let kernel_size = radius * 2 + 1;
+    let weight = P::make_weight(kernel_size);
     let edge_count = (kernel_size / 2) + 1;
     let half_kernel = kernel_size / 2;
     let width_bound = width as usize - 1;
@@ -299,7 +342,7 @@ fn box_blur_horizontal_pass<P: BlurAccum, const CN: usize>(
 
             let dst_chunk = &mut dst[x * CN..x * CN + CN];
             for c in 0..CN {
-                dst_chunk[c] = P::acc_to_store(sums[c], kernel_size);
+                dst_chunk[c] = P::acc_to_store(sums[c], weight);
             }
 
             let next_chunk = &src[next..next + CN];
@@ -330,7 +373,7 @@ fn box_blur_horizontal_pass<P: BlurAccum, const CN: usize>(
                 .zip(advanced_kernel_part_chunks)
             {
                 for c in 0..CN {
-                    dst_chunk[c] = P::acc_to_store(sums[c], kernel_size);
+                    dst_chunk[c] = P::acc_to_store(sums[c], weight);
                 }
                 for c in 0..CN {
                     sums[c] += src_next[c].to_acc();
@@ -347,7 +390,7 @@ fn box_blur_horizontal_pass<P: BlurAccum, const CN: usize>(
 
             let dst_chunk = &mut dst[x * CN..x * CN + CN];
             for c in 0..CN {
-                dst_chunk[c] = P::acc_to_store(sums[c], kernel_size);
+                dst_chunk[c] = P::acc_to_store(sums[c], weight);
             }
 
             let next_chunk = &src[next..next + CN];
@@ -376,6 +419,7 @@ fn box_blur_vertical_pass<P: BlurAccum>(
     test_radius_size(width as usize, radius);
 
     let kernel_size = radius * 2 + 1;
+    let weight = P::make_weight(kernel_size);
     let edge_count = (kernel_size / 2) + 1;
     let half_kernel = kernel_size / 2;
     let height_bound = height as usize - 1;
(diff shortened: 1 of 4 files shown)


src/math/utils.rs
```diff
@@ -46,6 +46,37 @@ pub(crate) fn multiply_accumulate<
...
+    target_arch = "aarch64",
+    all(
+        any(target_arch = "x86", target_arch = "x86_64"),
+        target_feature = "sse4.1"
+    )
+)))]
+#[inline(always)]
+pub(crate) fn fast_round_positive_f32(x: f32) -> f32 {
+    const MAGIC: f32 = (1u32 << 23) as f32; // 8_388_608.0
+    (x + MAGIC) - MAGIC
```
awxkee: We typically want rounding to nearest everywhere, this won't round '0.5' to the nearest, this is relatively important if we're unlucky and if alpha channel is blurred to `254.5`, then this will make opaque image transparent; this isn't good at all.

This behaviour can be broken by adding EPS to x.
In perfect world we could do `(f32::from_bits(x.to_bits() + 1)` but world is not perfect, so float number and integral uses different ports and this addition will cost almost twice more, accounting data transfer between units.

This doesn't round negatives properly because this MAGIC is wrong, it should be:
```
const MAGIC: f32 = ((1u32 << 23) + (1u32 << 22)) as f32;
```

src/traits.rs
```diff
@@ -34,26 +34,181 @@ impl EncodableLayout for [f32] {
...
+    impl PrimitiveSealed for i64 {}
+    impl PrimitiveSealed for f32 {}
+    impl PrimitiveSealed for f64 {}
+
+    /// Precomputed reciprocal for fast integer division of u32 accumulators.
+    ///
+    /// Replaces `(acc + ks/2) / ks` with a multiply-shift using the identity:
+    ///   `floor(n / d) == floor(n * ceil(2^32 / d) / 2^32)`
+    /// which is exact for all `n` where `(d-1)*n < d * 2^32` (Granlund-Montgomery '94).
+    /// For blur accumulators (`n <= kernel_size * 255`), this holds for all practical sizes.
```
197g: I'm wondering, we obviously want to make it _fast_ for all practical sizes but we still want it to be correct for the others, I guess? We could preserve the float accumulation variant but call the integer weight variant for all `kernel_width,height < 1 << 24`. I wouldn't really mind the complexity budget; this seems to require nothing more than an associated const on `BlurAccum` to check the kernel size.

src/imageops/fast_blur.rs
```diff
@@ -137,60 +137,37 @@ fn ceil_to_odd(x: usize) -> usize {
...
-                .min(T::DEFAULT_MAX_VALUE.to_f32().unwrap())
-                .max(T::DEFAULT_MIN_VALUE.to_f32().unwrap()),
-        )
-        .unwrap()
-    }
-}
-
-fn box_blur_horizontal_pass_strategy<T, P: Primitive>(
-    src: &[P],
+fn box_blur_horizontal_pass_strategy<T: Pixel>(
```
RunDevelopment: These attributes are left over from `fn rounding_saturating_mul`. They are unnecessary. Please remove them.

src/imageops/fast_blur.rs
```diff
@@ -233,36 +206,20 @@ fn box_blur_horizontal_pass_impl<T, const CN: usize>(
         .chunks_exact_mut(dst_stride)
         .zip(src.chunks_exact(src_stride))
     {
-        let mut weight1: f32 = 0.;
-        let mut weight2: f32 = 0.;
-        let mut weight3: f32 = 0.;
+        let mut sums = [P::ZERO; CN];
```
RunDevelopment: `P::ZERO` is *highly* confusing. Took me about 5 minutes to realize that `u8::ZERO == 0_u32`. Not obvious that this is `BlurAccum::ZERO`.

Maybe we could rename it to `EMPTY_ACCUMULATOR`? Or maybe even restructure the trait like I suggested below?

src/traits.rs
```diff
@@ -34,26 +34,188 @@ impl EncodableLayout for [f32] {
...
+        type Acc: Copy + Add<Output = Self::Acc> + AddAssign + Sub<Output = Self::Acc> + SubAssign;
+        type Weight: Copy;
+
+        const ZERO: Self::Acc;
+
+        fn to_acc(self) -> Self::Acc;
+        fn scale(acc: Self::Acc, count: usize) -> Self::Acc;
+        fn make_weight(kernel_size: usize) -> Self::Weight;
+        fn to_store(acc: Self::Acc, weight: Self::Weight) -> Self;
+    }
```
RunDevelopment: This might be a stupid suggestion, but I feel like this trait is the wrong way around.

This trait is implemented for primitives, but the primitives are not the blur accumulators. `Self::Acc` is. So I think it would be more natural to make the accumulator one trait (separate from `Primitive`) and give each primitive an associated type like this:

```rs
pub trait WithBlurAcc { // sealed supertrait of `Primitive`
    type BlurAcc: BlurAccumulator<Self>;
}

pub(crate) trait BlurAccumulator<T>: Copy + Sized + Add<Output = Self> + AddAssign + Sub<Output = Self> + SubAssign {
    const ZERO: Self;
    fn from_primitive(value: T) -> Self;
    fn scale(self, count: usize) -> Self;

    type Weight: Copy;
    fn create_weight(self, kernel_size: usize) -> Self::Weight;
    fn to_store(self, weight: Self::Weight) -> T;
}

// general implementation
impl<T: Primtive> BlurAccumulator<T> for f32 { ... }
// optimized implementation for `u8`
impl BlurAccumulator<u8> for u32 { ... }

// pick the right implementation
impl WithBlurAcc for u8 { 
    type BlurAcc = u32; 
}
impl WithBlurAcc for u16, u32, u64, usize, i8, i16, i32, i64, isize, f32, f64 { 
    type BlurAcc = f32; 
}
```

This might be cleaner, since `trait BlurAccumulator<T>` and its implementations can live inside `fast_blur.rs`. This will keep all blurring logic local to one file.

(@197g ping for thoughts)

src/imageops/fast_blur.rs
```diff
@@ -137,60 +137,37 @@ fn ceil_to_odd(x: usize) -> usize {
 
 #[inline]
 #[allow(clippy::manual_clamp)]
```
RunDevelopment: These attributes are left over from `fn rounding_saturating_mul` and unnecessary. Please remove them.

RunDevelopment (commented):
Amazing stuff @art049!

Before I go into details about the code itself, I'd like to talk about codspeed. Codspeed integration should be a separate PR, so the maintainers of `image` can discuss it without that blocking your improvements to `fast_blur`. So please revert the changes to `Cargo.toml`, `README.md`, and `.github/workflows/codspeed.yml`.

I only have a few nits about the code. Please see my comments.

I also want to say that I love the way you documented all the non-trivial fixed-point math tricks. It's easy to understand and verify.

src/math/utils.rs
RunDevelopment: Floating-point numbers round the results of addition (and other operations) using round ties even. So all integers of the form X.5 (X is a positive/negative integer or zero) will be rounded to the nearest even number. There's no way around that. Even [the "correct" magic number has that issue](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b9a9a5172a3013a61dc845afcf06b050).

So I would say that this function is implemented as correctly as can be.

awxkee: I meant something like [this one](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=60d456ab64975d9287e03591f37527b4)

awxkee: As well [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5bd37ef49a2c5c46250699f10f48a032) should correctly break ties almost everywhere (where this method can work, it can't work on the whole range ) because that's should be enough by float definition to break the ties. However, I never performed an exhaustive check.

awxkee: Part about changed magic touches only "negatives" because mantissa is wrong, see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3cf5f671540697aeb5d0a97cfadd19da)

src/imageops/fast_blur.rs
art049: Definitely this is not obvious, changed to what you're suggesting.

src/math/utils.rs
art049: Changed the magic, if it can handle negative as well, better do it.

Also determined the domain limit of the ULP nudge (-2^22, 2^22]

Currently benching this to see the impact on  the throughput since as you mentioned, this is using another port and from the simulation I ran this might be twice as slow in term of latency.

For context, here is the assembly:
```asm

fast_round:
        addss   xmm0, dword ptr [rip + .LCPI1_0]
        addss   xmm0, dword ptr [rip + .LCPI1_1]
        ret

fast_round_awxkee:
        movd    eax, xmm0
        inc     eax
        movd    xmm0, eax
        addss   xmm0, dword ptr [rip + .LCPI0_0]
        addss   xmm0, dword ptr [rip + .LCPI0_1]
        ret

 ```

RunDevelopment:
@art094 You might want to look at #2868. It just got merged and is the main reason for the merge conflicts here.

Since `NearestFrom` replaced `FloatNearest`, I think this means that `fast_round_f32` is no longer used and can be removed.

src/traits.rs
art049: I agree it would be better to scope that bluring logic. This is close to what I did in the beginning but [after a first review](https://github.com/art049/image-rs/pull/2#discussion_r2869798067) I changed it ([here is the commit](https://github.com/art049/image-rs/commit/a40865de5b3e5739799d589ceeb126d33ecd016c#diff-b0593df772dce4980fcd17f3280796ee3af23cd5719ca19a95d1ae99637b6523))..

Honestly, I'm opened to both. So it would really help to have your input @197g

art049:
> Since `NearestFrom` replaced `FloatNearest`, I think this means that `fast_round_f32` is no longer used and can be removed.

Thanks for the notice. I think there might still be the same fix to apply here
https://github.com/RunDevelopment/image/blob/274d02bf90a9d7405de17f26472816c5ccadc43a/src/primitive_sealed.rs#L106-L108
But I'll have a look.

RunDevelopment:
> I think there might still be the same fix to apply here
> https://github.com/RunDevelopment/image/blob/274d02bf90a9d7405de17f26472816c5ccadc43a/src/primitive_sealed.rs#L106-L108

You're right, but this feels like it should be a separate PR to me. I feel like this has little to do with `fast_blur` itself.

src/math/utils.rs
art049: Measured it and there is no measurable impact in throughput, neat!

src/traits.rs
197g: Those two review, I think, refer to independent properties. In mine I said that any additional trait should be a super trait of `Primitive` (or `PrimitiveSealed`), which just refers to the way of dispatching into the summation algorithm, while @RunDevelopment is referring to the mechanism itself, i.e. that each primitive might refer to an accumulator type and not implement the accumulation directly.

As to that idea, I suppose having only two accumulator types here is neat? Having a generic on `BlurAccumulator<T>` allows us to express the idea that `u8` can be summed and averaged in multiple different ways and `WithBlurAcc::BlurAcc` only picks one of them, so conceivably it will be easier to code a dispatch into an particular accumulator for particular kernel sizes. It does not seem to require a lot of changes  to the code as written otherwise. This ties in to the other comment.

art049: Implemented it in 061bbffa
Let me know what you think @RunDevelopment @197g! Agree that in the end this is much cleaner since the actual logic lives in the right place and in the meantime this is much smoother thant the very first implementation I had.

src/traits.rs
art049: Implemented it in https://github.com/image-rs/image/pull/2846/changes/d4fdbd6268d8f894eb56e234770e190ca43c0c04. Let me know what you think.

197g: I like it, thank you :+1: Very nice that `f32` ends up being a concrete type and using the generic impl, so `WithBlurAcc` itself can be simpler than I had thought.

197g (approved):

197g:
If you want to point to this PR as performance work supported / enabled by CodSpeed, feel free to. And yes, we can discuss an integration PR. To start that, from past experience (with a fuzzer product) as well as Github's rather incomplete story of security I'm wary of *any* external service integrated in the Actions pipeline, particularly for pull requests. On the other hand, rustc's perf integration is obviously incredibly helpful to its PR process so if it is implemented similarly, that'd be great.

Issue #2621, opened 2025-10-20
enable re_rav1d by default for avif decoding


valarauca:
## I would like to be able ...

Decode avif images without having to fight `pkg-config`, `ninja`, `meson`, and `vckpg` to build this crate on Microsoft-Windows.

## My specific use case for this functionality is ...

Decoding avif images on Microsoft Windows.

## This is more generally applicable to ...

Anyone who wants to use this crate & decode avif images on Windows.

## Draft

`re_rav1d::dav1d` is full compatible with the `dav1d` crate, as its API is a fork.

I've managed to get this to build with the following operation:

```bash
$ git clone https://github.com/image-rs/image
$ cd image

# Update our Cargo.toml
$ sed -i '59i re_rav1d = { version = "0.1.3", optional = true }' Cargo.toml 
$ sed -i 's/nasm = \["ravif?\/asm"\]/nasm = ["ravif?\/asm", "re_rav1d?\/asm"]/' Cargo.toml
$ sed -i 's/avif-native = \["dep:mp4parse", "dep:dav1d"\]/avif-native = ["dep:dav1d"]/' Cargo.toml
$ sed -i 's/avif = \["dep:ravif", "dep:rgb"\]/avif = ["dep:ravif", "dep:rgb", "dep:re_rav1d", "dep:mp4parse"]/' Cargo.toml
```

Then `src/codecs/av1f/decoder.rs` is updated to the import statement becomes

```
#[cfg(all(feature = "avif", not(feature = "avif-native")))]
use re_rav1d::dav1d::{PixelLayout, PlanarImageComponent};
#[cfg(all(feature = "avif", not(feature = "avif-native")))]
use re_rav1d::dav1d;
#[cfg(all(feature = "avif", feature = "avif-native"))]
use dav1d::{PixelLayout, PlanarImageComponent};
#[cfg(all(feature = "avif", feature = "avif-native"))]
use dav1d::{self};
```

To ensure there is always a `dav1d` prefix.

It seems to just work provided you update the `src/codecs/av1f/mod.rs` `#feature` gates.

Shnatsel:
We've been waiting on https://github.com/memorysafety/rav1d/pull/1439 or for dav1d-sys crate to support rav1d before we can switch to rav1d for decoding.

Depending on a private fork of rav1d that is not getting updated as frequently and may not receive timely security patches does not seem like a great idea for a major crate like `image`.

PR #2849, opened 2026-03-14
Switch to rav1d for AVIF decoding

Shnatsel:
Depends on https://github.com/memorysafety/rav1d/pull/1439

We don't have any in-tree AVIF tests, so I tested manually with https://github.com/Shnatsel/wondermagick/pull/92 and it seems to work great!


fintelia:
We'll probably want to get a sense of the crate with fuzzing/Miri before switching over.

From what I understand zlib-rs was created by the same organization. And my experience with zlib-rs was that the marketing got a bit ahead of the implementation: When I saw it declared as ready to try for production use, I ran it through some fuzzing and discovered that it instantly crashed because a few decode code-paths were still marked `todo!`.

Shnatsel:
FWIW I don't see the two projects sharing any key developers. But I don't see a fuzzing harness in-tree, which is a potential red flag.

Shnatsel:
Oh, https://github.com/imazen/rav1d-safe also exists where people ported rav1d's handwritten assembly to Rust's safe SIMD intrinsics plus `safe_unaligned_simd`. And it already comes with a safe Rust API. I believe that could be a better fit for image-rs, provided it works as advertised; we're not operating under video codec level of performance constraints.

However, the license is AGPL-3.0, which is a very aggressive copyleft variant. But the README includes the following:

> ##Upstream Contribution
> 
> This fork exists because maintaining a separate safe SIMD implementation is the fastest path to getting safe Rust AV1 decoding into production. If the rav1d maintainers are interested in upstreaming any of this work under the original BSD-2-Clause license, we'd be happy to contribute. Open an issue or reach out.

Should I talk to the memorysafety.org folks (with you cc'd) about getting the safe Rust codepaths upstreamed?

Or perhaps we could reach out to rav1d-safe ourselves and discuss inclusion into `image` directly?

197g:
In any case it brought [`archmage`](https://docs.rs/archmage/latest/archmage/) to my attention which proxies `safe-unaligned-simd` for the memory access :tada:

awxkee:
Whole rav1d-safe seems to me have been made by AI. Don't get me wrong I'm fine with AI, but there is like ~100K lines fairly complex AI-generated code that likely no one at least partially understands. If there are logical or safety errors, which is almost inevitable, they will be quite difficult to track and fix, especially since no one really knows what the code is actually supposed to do.

Shnatsel:
`#[forbid(unsafe_code)]` insulates us from safety errors. Since the entire thing is safe Rust, the consequences of AI mistakes are far less drastic. I'd take 100k of safe Rust over 160k unsafe handwritten assembly.

> especially since no one really knows what the code is actually supposed to do.

That's not true: there are scalar Rust functions implementing the same behavior. So we always have known-good reference code and can automatically generate tests that compare the output of any optimized implementation to the scalar reference. This is precisely the property that could make an AI-driven conversion manageable; without this ground truth it would be a lost cause.

Still, the fact that such a faithful conversion is potentially feasible doesn't mean that rav1d-safe did it with an appropriate level of rigor. Their claims still need to be verified.

awxkee:
> That's not true: there are scalar Rust functions implementing the same behavior. So we always have known-good reference code and can automatically generate tests that compare the output of any optimized implementation to the scalar reference. This is precisely the property that could make an AI-driven conversion manageable; without this ground truth it would be a lost cause.

I meant that if something goes wrong in the decoding path without an obvious panic, and the chatbot can't guess where the mistake is, one would have to debug the entire decoding path from scratch, because no one would be able to point out even where to start looking.

197g:
I have to do that in other repositories, too, both that I wrote and did not write myself. The argument is not exactly a quantitative methodology for getting rid of faults in comitted code. Using an LLM for translation of code is as close to a qualified use case for it as it gets; and intrinsic instructions rarely have panic side effects so really besides weird behavior I'm not too concerned.

Shnatsel:
The upstream `rav1d` (not -safe) seems to be decently complete. I've [written a fuzzing harness](https://github.com/Shnatsel/rav1d/blob/fuzzing/fuzz/fuzz_targets/fuzz_target_1.rs) and the fuzzer didn't find any `todo!()`s. It did find a panic on invalid indexing due to arithmetic overflow, but that's precisely why I want a Rust implementation. No AddressSanitizer errors so far.

leo030303:
https://github.com/memorysafety/rav1d/pull/1439 is now merged so this should be good to progress @Shnatsel

Shnatsel:
Nice, I've contributed a fuzzing harness using it: https://github.com/memorysafety/rav1d/pull/1483

I'll see if I can revive this PR too

Shnatsel:
Okay, starting with fuzzing was a good call. I've opened https://github.com/memorysafety/rav1d/pull/1484 to remove some panics that the newly added Rust API forces.

Shnatsel:
This is currently blocked on upstream PRs https://github.com/memorysafety/rav1d/pull/1484 and https://github.com/memorysafety/rav1d/pull/1485

Shnatsel:
The required upstream PRs are merged, this is now ready for review.

Shnatsel:
I've resolved the merge conflicts caused by #2977

This is now ready for review. Once it's approved, I'll ask upstream to cut a new release and we can merge.

RunDevelopment:
Nice stuff. Code looks good to me aside from one thing: the `avif-native` feature.

AFAIK, this feature exists because dav1d requires a C compiler to build. Since that's not the case anymore, maybe we could remove the feature and gate `mp4parse` + `rav1d` behind the `avif` feature? This would also finally enable full AVIF support by default (#2621).

Shnatsel:
**TL;DR:** I prefer to split `avif` format into two features: `avif-encode` and `avif-decode`. We can also make an `avif` feature that implies both for convenience.

Unlike most other formats, AVIF encoding and decoding is handled by entirely different crates. They are the most complex implementations we have, with plenty of `unsafe`, and it is nice to be able to remove one of them from the attack surface.

PR #2855, opened 2026-03-15
Fast path for `crop_in_place`

RunDevelopment:
`crop_in_place` did unnecessary copying for certain rectangles. For example, if the selection only shrinks the height of the image. Now it will check whether copying is necessary before calling `slice::copy_within`.

This won't make the average case faster, but the performance of certain edge cases is now close to optimal.

Note: I am making the assumption that `slice::copy_within` doesn't already perform this optimization. It doesn't seem to be, but I couldn't verify this conclusively.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1026,6 +1026,7 @@ where
     /// ```
     pub fn crop_in_place(&mut self, selection: Rect) {
         let selection = selection.crop_dimms(self);
+        assert!(selection.test_in_bounds(self).is_ok());
 
         fn copy_within<T: Copy>(data: &mut [T], src: usize, len: usize, dst: usize) {
             if src == dst || len == 0 {


src/images/buffer.rs
```diff
@@ -1026,6 +1026,19 @@ where
     /// ```
     pub fn crop_in_place(&mut self, selection: Rect) {
         let selection = selection.crop_dimms(self);
+        debug_assert!(
+            selection.x + selection.width <= self.width
+                && selection.y + selection.height <= self.height,
+            "Selection must be within the image"
+        );
```
197g: There's a utility method on `Rect` for the check that should be used instead of a manual calculation. It returns `Result`, I would prefer asserting over `is_ok()` over unwrapping but not by much.

RunDevelopment: My first implementation of a similar optimization relied on `selection.width == self.width` $\implies$ `selection.x == 0` $\land$ both row pitches are the same. So I wanted to double-check my assumptions. 

But then I thought of this simpler solution: the one in this PR. The check probably isn't necessary anymore, so I'll just remove it.

197g: It makes sense to have it (as a proper assert), we do not return an error and panic in similar methods.

RunDevelopment:
I did a little more digging and `slice::copy_within` seems to compile to `memmove`. Looking at older versions of glibc, I don't think `memmove` optimizes for `src_prt == dst_ptr`.

src/images/buffer.rs
RunDevelopment: Not really. `selection` is already in bounds (because it's the result of `crop_dimms` (horrible name btw)), so the assert will always succeed (unless `crop_dimms` or `test_in_bounds` is implemented incorrectly).

But I don't have strong feelings on this. I added a proper assert like you suggested.

PR #2856, opened 2026-03-15
Clean up internal `Rect` methods

RunDevelopment:
Changes:

- Renamed `crop_dimms` to `shrink_to_bounds_of`.
- Consistent naming for the 3 methods. They are now called `test_in_bounds_of`, `assert_in_bounds_of`, and `shrink_to_bounds_of`.
- Move condition for testing in-bounds-ness into its own function. (I also didn't like that one version of it relied on `None < Some(T)` for correctness. This is non-obvious to anyone not deep into Rust.)


197g (approved):
Sounds good, the only reason for that naming was legacy. Feels nice to give it up for much more descriptive names.

PR #2862, opened 2026-03-15
Remove deprecated `TiffReader`

RunDevelopment:
`TiffReader` was deprecated and otherwise unused. I removed it for 1.0.


RunDevelopment:
Jep. I'm currently going through everything that's deprecated one by one.

PR #2865, opened 2026-03-15
Remove deprecated constructor for `HdrDecoder`

RunDevelopment:
I removed the deprecated and internally unused `new_nonstrict` constructor for `HdrDecoder`.

This is the last deprecated item in `image`.

### DIFF
--- a/src/codecs/hdr/decoder.rs
+++ b/src/codecs/hdr/decoder.rs
@@ -184,9 +184,11 @@ impl Rgbe8Pixel {
 }
 
 impl<R: Read> HdrDecoder<R> {
-    /// Reads Radiance HDR image header from stream ```r```
-    /// if the header is valid, creates `HdrDecoder`
-    /// strict mode is enabled
+    /// Reads Radiance HDR image header from `reader` with strict mode enabled.
+    ///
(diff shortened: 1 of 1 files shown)


fintelia:
This method was only deprecated 3 weeks ago, and the deprecation note points at a method that isn't public. I think this might need a bit more attention first...

RunDevelopment:
Good catch! I didn't check when it was deprecated.

I think just explaining things a bit more in the docs of `new` and `with_strictness` should clear things up.

Also, should `with_strictness` be renamed to `new_with_strictness`? Sounds more like a constructor.

197g:
Since Rust has no special  'constructor', `new_` as part of names is superfluous. If you look at the standard library then [`from_`](https://doc.rust-lang.org/stable/std/?search=from_) and [`with_`](https://doc.rust-lang.org/stable/std/?search=with_) are both heavily used (with the former preferred where the latter could mistaken for a builder pattern).

RunDevelopment:
Special or not, what's a constructor if not a static factory method? I just like to call them that here, since that's what their role is.

But after looking at `BufReader::with_capacity`, I think `with_strictness` should be okay.

PR #2867, opened 2026-03-16
Add support for reading extended XMP metadata and implement it for JPEG.

1c3t3a:


src/io/decoder.rs
```diff
@@ -39,6 +39,15 @@ pub trait ImageDecoder {
         Ok(None)
     }
 
+    /// Returns the raw [Extended XMP](https://en.wikipedia.org/wiki/Extensible_Metadata_Platform) metadata, if present.
```
Shnatsel, line 42: https://en.wikipedia.org/wiki/Extensible_Metadata_Platform contains no mentions of the word "extended", could you clarify where this terminology comes from and where this behavior is specified?

Issue #2858, opened 2026-03-15
`fast_blur` does not approximate a Gaussian blur well for small sigma


RunDevelopment:
`fast_blur` produces significantly different results for small sigma (around 1.0). The smaller sigma is, the greater the differences become.

## Expected

`fast_blur` should approximate a true Gaussian blur well for all sigma.

## Actual behaviour

`fast_blur` only approximates well for larger sigma.

## Reproduction steps

Blur any image with `fast_blur` and `blur` with a sigma of e.g. 1.0 and observe very different results.

197g:
If `fast_blur` and Gaussian blur were the same, we would just use the former, no? If there's an algorithmic detail that was missed, sure enough, but it should be regarded as two different forms of blur. That said, I think our implementation will never produce box blurs of size `1` (i.e. no-ops) despite that being necessary to approximate cases where the 'ideal' sizes is `1 < ideal < 3`. Do we have test vectors for `boxes_for_gauss` somewhere?

awxkee:
It's fairly not feasible, box blur is discrete, minimum what we can do is radius 1. Currently it's set here for 3 passes it's commonly assumed that 3 averages is enough to approximate normal distribution. For small sigmas, we could effectively use 2 passes ( with radius 1 ), but it still won't match exactly, and it will produce a tent blur, because using averaging kernel twice results in a triangular shape.

RunDevelopment:
The first sentence of the documentation for `fast_blur` is "Approximation of Gaussian blur."

The issue isn't that `fast_blur` isn't the same as Gaussian blur. The issue is that `fast_blur` is supposed to be an approximation for Gaussian blur but isn't a good one for certain inputs.

We're not limited to boxes to approximate a Gaussian blur. If boxes for gaussian isn't a good approximation for small sigma, then we can just use something else. E.g. a small convolution (3x3 or 5x5 separated) should be reasonably fast and provide sufficient quality for small sigma. That's also what programs like GIMP and Photoshop do too IIRC.

> box blur is discrete, minimum what we can do is radius 1

Depends. You probably already know that, but box blur can be extended to real radii. I.e. with kernel size = 2*r+1, the convolution kernel for r=0 is `[1]` and r=1 is `[1 1 1]`. Then the kernel for r=0.8 is `[0.8 1 0.8]`. (Separated x and y, of course.)

What box blur algorithm do we use? It's probably one using a running sum/average to be O(1) per pixel, right? That one would be difficult to extend to real radii...

[...]

PR #2868, opened 2026-03-16
Add `NearestFrom` for unified f32 rounding/casting

RunDevelopment:
After talking about it [here](https://github.com/image-rs/image/issues/2858#issuecomment-4063925529), I removed the `f32::round` in the hot path of `fast_blur`.

I did this by adding a new private trait: `NearestFrom`. This has similar semantics to `NumCast::from`, but it will round to nearest for float->int conversions and saturate on numeric bounds instead of returning `Option`. This makes it a natural fit for performance-sensitive code that needs to convert f32 to subpixels.

For now, I only used the trait in `fast_blur`, but other operations can use it as well for both correctness and performance. I specifically designed the trait to have uses beyond `fast_blur`.

This PR does make `fast_blur` significantly faster, even for the `u8` case. Here are the benchmark results from my machine (Intel i7-8700K):

| bench | before | after | change |
| --- | --- | --- | --- |
| fast blur: sigma 3.0 | 125.60 ms | 17.616 ms | -85.975% |
| fast blur: sigma 7.0 | 112.48 ms | 17.462 ms | -84.475% |
| fast blur: sigma 50.0 | 108.44 ms | 18.044 ms | -83.360% |

That's between 6-7 times faster. 

---

Note that this is no competition for #2846 and its fixed-point implementation for u8. On my machine, #2846 reaches 8ms on the same benchmark.

---

I also want to mention that some implementations of `NearestFrom` still leave performance on the table. See [this comment](https://github.com/image-rs/image/pull/2846#discussion_r2928069963) for example. I just optimized the important primitives for now. Everything else can follow later.


fintelia:
We may want to spin everything related to `PrimitiveSealed` into a separate source file. I suspect we're going to end up with a bunch

RunDevelopment:
Depends. For example, I also want to add a trait to make RGB->Luma conversions faster. This would also need to be a super trait of `PrimitiveSealed`, but I'd like to keep its definition and implementation right next to the (internal) code that uses it for local reasoning.

So I'm not too sure that the code in `traits.rs` is going to grow a lot, but I'm also not against moving things into a separate source file if it does.

src/traits.rs
```diff
@@ -33,8 +33,57 @@ impl EncodableLayout for [f32] {
...
+/// - For a float -> float conversion:
+///     - The float is clamped to the range `[0.0, 1.0]`.
+///     - NaN is mapped to 0.0.
+pub(crate) trait NearestFrom<T> {
+    fn nearest_from(value: T) -> Self;
+}
+
+impl NearestFrom<f32> for u8 {
+    fn nearest_from(value: f32) -> Self {
+        (value + 0.5) as u8
```
mstoeckl: This rounds `0.5f32.nextdown()` (`0.49999997f32`) to 0 instead of to 1, which I think is fine for the simple and color-space unaware mathematical operations of imageops. (There is little point in making the final step of interpolate_bilinear or fast_blur exact when the others introduce floating point calculation error.) Still, if you agree and plan to keep this behavior, could you document that NearestFrom may deviate from the ideal rounding in edge cases?

RunDevelopment: Sure. Documenting that rounding may use a fast approximation should clear that up.

RunDevelopment:
> We may want to spin everything related to `PrimitiveSealed` into a separate source file. I suspect we're going to end up with a bunch

After thinking about it a bit more, I implemented your suggestion. I don't see it growing a lot now, but there's also no harm in doing so.

RunDevelopment:
I took some more time to think about this and slightly changed the interface of `NearestFrom`. It now has two methods:

- `::nearest_from`: This returns the nearest value in the target type. No clamping is performed. E.g. 3.6_f32 to f64 will return 3.6_f64.
- `::clamp_nearest_from`: This is roughly equivalent to `T::nearest_from(input).clamp(T::DEFAULT_MIN_VALUE, T::DEFAULT_MAX_VALUE)`.

These two methods seem to cover all existing patterns of casting f32 to primitives. This makes it possible to use `NearestFrom` in more places throughout the codebase. I initially only intended this as a way to unify and simplify f32 casting/rounding (the same logic was implemented like 4 times and all differently), but it sped up some operations as a side effect.

Here's Gaussian blur:

| bench                       |    `main` |   this PR | change % |   change Δ |
| --------------------------- | --------: | --------: | -------: | ---------: |
| `gaussian blur: sigma 3.0`  | 27.827 ms | 16.616 ms | -40.287% | -11.211 ms |
| `gaussian blur: sigma 7.0`  | 47.775 ms | 33.254 ms | -30.394% | -14.521 ms |
| `gaussian blur: sigma 50.0` | 251.09 ms | 237.90 ms |  -5.255% | -13.190 ms |

Quite the speed-up, given that I only changed the very last step where the temporary f32 image is cast back to the original pixel type. This is why the speedup is a roughly constant 13 ms and doesn't depend on sigma.

For a similar reason, resizing also got faster:

| bench                         |    `main` |   this PR | change % |   change Δ |
| ----------------------------- | --------: | --------: | -------: | ---------: |
| `resize 400x300 Nearest`      | 7.1668 ms | 3.6532 ms | -49.027% |   3.514 ms |
| `resize 400x300 Triangle`     | 15.170 ms | 10.652 ms | -29.784% |   4.518 ms |
| `resize 400x300 CatmullRom`   | 23.738 ms | 18.171 ms | -23.452% |   5.567 ms |
| `resize 400x300 Gaussian`     | 31.634 ms | 26.268 ms | -16.963% |   5.366 ms |
| `resize 400x300 Lanczos3`     | 30.669 ms | 26.416 ms | -13.866% |   4.253 ms |
| `resize 2000x2000 Nearest`    | 148.07 ms | 38.930 ms | -74.067% | 109.140 ms |
| `resize 2000x2000 Triangle`   | 190.85 ms | 55.490 ms | -70.445% | 135.360 ms |
| `resize 2000x2000 CatmullRom` | 226.97 ms | 73.044 ms | -67.103% | 153.926 ms |
| `resize 2000x2000 Gaussian`   | 243.09 ms | 90.549 ms | -61.879% | 152.541 ms |
| `resize 2000x2000 Lanczos3`   | 226.83 ms | 90.264 ms | -59.044% | 136.566 ms |

(It's honestly ridiculous how much having `f32::round` in a hot path hurts performance on x86. Maybe we should consider banning this function.)

197g (approved):

PR #2870, opened 2026-03-17
Make `huerotate` support any number of channels

### LINKED ISSUE #2859: `huerotate` produces incorrect results for grayscale images.
Obviously, `huerotate` should be a noop for grayscale images, since they have no hue. However, that's not what happens. It will interpret a grayscale pixel `g` as RGB=`[g, 0, 0]` (a red tone), rotate the hue of this color, and then output the R channel of the rotated color.

For grayscale+alpha images, it will interpret `[g, a]` as RGB=`[g, a, 0]`.

## Expected

`huerotate` should output grayscale image unchanged.

## Actual behaviour

See above.

RunDevelopment:
Fixes #2859

I changed the implementations of `huerotate` and `huerotate_in_place` to be a noop for pixels with less than 3 channels. Both will now assume that the first 3 channels are R,G,B (like before) and leave all others untouched (not quite like before). This is done by directly modifying the the first 3 channels of the original pixel, instead of trying to create a new pixel, which previously only supported up to 4 channels.

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -213,11 +213,24 @@ where
 }
 
 /// Hue rotate the supplied image.
-/// `value` is the degrees to rotate each pixel by.
-/// 0 and 360 do nothing, the rest rotates by the given degree value.
-/// just like the css webkit filter hue-rotate(180)
 ///
-/// *[See also `huerotate_in_place`.][huerotate_in_place]*
+/// `value` is the angle in degrees to rotate the hue of each pixel by. 0 and 360
+/// do nothing. -15 is the same as 360-15 = 345. This behaves the same as the CSS
+/// filter [`hue-rotate()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/filter-function/hue-rotate).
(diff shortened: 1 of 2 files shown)


src/imageops/colorops.rs
```diff
@@ -224,6 +224,13 @@ where
     P: Pixel<Subpixel = S>,
     S: Primitive,
 {
+    if <I::Pixel as Pixel>::CHANNEL_COUNT < 3 {
+        // ignore grayscale and gray+alpha images
+        // TODO: Find a faster way to convert GenericImageView to ImageBuffer
+        let (width, height) = image.dimensions();
+        return ImageBuffer::from_fn(width, height, |x, y| image.get_pixel(x, y));
```
197g: Re: Find a faster way to convert GenericImageView to ImageBuffer

```suggestion
        let mut buffer = image.buffer_like();
        buffer.copy_from(&image, 0, 0).unwrap();
        buffer
```

Not  only faster but importantly preserves color space attributes.

RunDevelopment:
Done.

PR #2874, opened 2026-03-18
Clamp abyss policy for `filter3x3`

### LINKED ISSUE #3026: Panic in `filter3x3()` on zero-dimension image due to unsigned integer underflow
### Describe the bug

`imageops::filter3x3()` panics with "attempt to subtract with overflow" when called on an image with zero width or zero height. 

Stack trace: 

For zero-width:

```
thread 'main' panicked at src/imageops/sample.rs:928:21:
attempt to subtract with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ec7c02612527d185c379900b613311bc1dcbf7dc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/ec7c02612527d185c379900b613311bc1dcbf7dc/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_sub_overflow
             at /rustc/ec7c02612527d185c379900b613311bc1dcbf7dc/library/core/src/panicking.rs:175:17
   3: image::imageops::sample::filter3x3
             at ./src/imageops/sample.rs:928:21
   4: poc_03_filter3x3_zero_dim::main
             at ./examples/poc_03_filter3x3_zero_dim.rs:30:5
```

For zero-height:
```
thread 'main' panicked at src/imageops/sample.rs:927:17:
attempt to subtract with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ec7c02612527d185c379900b613311bc1dcbf7dc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/ec7c02612527d185c379900b613311bc1dcbf7dc/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_sub_overflow
             at /rustc/ec7c02612527d185c379900b613311bc1dcbf7dc/library/core/src/panicking.rs:175:17
   3: image::imageops::sample::filter3x3
             at ./src/imageops/sample.rs:927:17
   4: poc_03_filter3x3_zero_dim::main
             at ./examples/poc_03_filter3x3_zero_dim.rs:31:5
```

### Root cause

In `src/imageops/sample.rs:927-928`, the iteration bounds use unchecked unsigned subtraction:

```rust
pub fn filter3x3<I, P, S>(image: &I, kernel: &[f32; 9]) -> ImageBuffer<P, Vec<S>>
// ...
{
    let (width, height) = image.dimensions();
    // ...
    for y in 1..height - 1 {        // ← line 927: panics when height == 0 (0u32 - 1 underflows)
        for x in 1..width - 1 {      // ← line 928: panics when width == 0 (0u32 - 1 underflows)
```

`height` and `width` can be zero, both leading to unsigned integer underflow when `-1`. 

### Expected behavior

May return an empty output image without panicking when either dimension is 0.

### To reproduce

```rust
// examples/poc_03_filter3x3_zero_dim.rs
fn main() {
    let kernel: [f32; 9] = [0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0];

    // Case 1: zero width
    let img = image::open("poc_input_0x5.ppm").unwrap();
    let rgba = img.to_rgba8();
    let _ = image::imageops::filter3x3(&rgba, &kernel);  // panic

    // Case 2: zero height
    let img = image::open("poc_input_5x0.ppm").unwrap();
    let rgba = img.to_rgba8();
    let _ = image::imageops::filter3x3(&rgba, &kernel);  // panic
}
```

Input files can be created by the following  commands (upload failed): 
`echo "P6\n0 5\n255" > poc_input_0x5.ppm`
`echo "P6\n5 0\n255" > poc_input_5x0.ppm`

### Test environment

- Version: image crate `main` branch 
- OS: Ubuntu 24.04, 64-bit
- Rustc version: rustc 1.91.0-nightly (ec7c02612 2025-08-05)

### Suggested fix

Add an early return at the top of `filter3x3`:

```rust
pub fn filter3x3<I, P, S>(image: &I, kernel: &[f32; 9]) -> ImageBuffer<P, Vec<S>>
where
    I: GenericImageView<Pixel = P>,
    P: Pixel<Subpixel = S>,
    S: Primitive,
{
    let (width, height) = image.dimensions();
    let mut out = image.buffer_like();

    if width <= 2 || height <= 2 {
        return out;  // Not enough pixels for a 3×3 kernel
    }
    // ... rest unchanged
```

RunDevelopment:
This adds the clamp/replicate abyss policy to `filter3x3`. This is consistent with `blur` and `fast_blur`.

**Why is this necessary?**
Previously, `filter3x3` would leave a 1-pixel border around the filtered image. This made the function a lot less useful. This PR fixes this defect by applying the filter to all pixels of the image.

---

As the TODO says, this function is currently slower than it has to be. This can be addressed in future PRs. I focused on making it correct for now.

---

Closes: #3026


src/imageops/sample.rs
```diff
@@ -924,16 +931,16 @@ where
...
+    for y in 0..height {
+        for x in 0..width {
             let mut t = [0.0; MAX_CHANNEL];
 
             // TODO: There is no need to recalculate the kernel for each pixel.
             // Only a subtract and addition is needed for pixels after the first
             // in each row.
             for (&k, &(a, b)) in kernel.iter().zip(taps.iter()) {
-                let x0 = x as isize + a;
-                let y0 = y as isize + b;
```
197g, line 936: Regarding slowness, do you mean it would be better to handle the near-border in a separate loop? If so, do you want to do that in a follow-up PR or did you intend that comment to be a blocker? I'm fine with either.

In an case, the cast to `isize` is suspicious here after the fix. The only use is with type `u32` in `get_pixel` and we have ful control over the type of `taps` which is a local. It seems rather wasteful to suppose `isize` for what is a tri-state bool, no?

197g (commented):
In direct comparison to #3055 , I like clamp-abyss much better as an explainable policy for _this_ operation. That also fits with internalizing complexity instead of a weirder interface. There's imageproc for all the customization you may want. Dynamically sized kernel are instead reweighted with outside pixels not counting to the total but I think that'd be least intuitive behavior for the most common use of a 3x3 kernel in particular.

src/imageops/sample.rs
RunDevelopment: > Regarding slowness, 

Definitely follow-up PR. This PR changes behavior, so I can't just add a benchmark and say "it's 2x faster now," since the before and after don't do the same thing.

> In an case, the cast to `isize` is suspicious here after the fix. The only use is with type `u32` in `get_pixel` and we have ful control over the type of `taps` which is a local. It seems rather wasteful to suppose `isize` for what is a tri-state bool, no?

It's not any more sus than before IMO. But we could change tabs to go 0,1,2 instead of -1,0,1 and then subtract 1. I.e. `(x + a).saturating_sub(1).min(width - 1)`. This makes the clamping less explicit, but it's probably fine.

197g: Actually, this exposes an overflow case when `x = u32::MAX` (previous `>= i32::MAX` on 32-bit targets). It's good to handle / fail that the same regardless of platform.

RunDevelopment: Well, `x = u32::MAX` isn't possible because `x < width`, but `x = u32::MAX - 1` overflows so your point stands. 

I think I'll just add an assert `width < u32::MAX`. Same for height. Not supporting images that large is probably fine.

Alternatively, we could also handle `x=0` separately. This would also turn the `saturating_sub(1)` into a simple `- 1` and allow us to write `x - 1 + a` (underflow and overflow impossible). But I don't want to write that code, because unrolling the nested loops in that way is going to be annoying.

RunDevelopment:
CI is super flaky for some reason. Third time a job failed due to a network error today.

RunDevelopment:
@197g Can we get this in? I think this is the third time now that someone tried to fix #3026 the obvious way.

197g (approved):
Okay 👍

PR #2876, opened 2026-03-19
fix into frames for static webp

### LINKED ISSUE #2761: `WebPDecoder::into_frames` returns an empty iterator for non-animated images
image 0.25.9

## Expected

I would expect the iterator to yield precisely one frame, since the image contains one frame.

Alternatively, yielding an error ("image is not animated") might also make sense.

## Actual behaviour

The operation silently fails and produces zero frames.

## Reproduction steps

Load a non-animated WebP with the `WebPDecoder`. Then call `WebPDecoder::into_frames().collect_frames()?`.

adhami3310:
fixes #2761


RunDevelopment:
No longer a bug because of #2672.

Thank you, though!

PR #2877, opened 2026-03-19
Add reference tests for image operations

RunDevelopment:
I added reference tests for image operations. For now, I tested `blur`, `fast_blur`, `invert`, `filter3x3`, and `huerotate`.

Reference images currently capture the bugs of these functions as is:
- `blur` produces incorrect outputs if kernel size=1.
- All outputs of `fast_blur` with a sigma <=1.5 are the same. (#2858)
- `filter3x3` adds a black border. (#2874)

### DIFF
--- a/tests/imageops_reference.rs
+++ b/tests/imageops_reference.rs
@@ -7,22 +7,25 @@ use std::{
 use libtest_mimic::{Arguments, Trial};
 
 fn main() -> std::process::ExitCode {
-    let mut trials = Vec::new();
+    let tests_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("tests");
+    let assets_dir = tests_dir.join("assets");
+    let reference_dir = tests_dir.join("imageops");
 
     static IMAGES: OnceLock<[Image; 3]> = OnceLock::new();
     let [bw_edge, noise, cat] = IMAGES.get_or_init(|| {
         [
-            Image::open(&tests_dir().join("assets/bw-edge.png")),
-            Image::open(&tests_dir().join("assets/noise.png")),
-            Image::open(&tests_dir().join("assets/cat.png")),
+            Image::open(&assets_dir.join("bw-edge.png")),
+            Image::open(&assets_dir.join("noise.png")),
+            Image::open(&assets_dir.join("cat.png")),
         ]
     });
+
+    let mut trials = ImageTrials::new(reference_dir);
 
     // blur & fast_blur with various sigmas
     for image in [bw_edge, noise] {
         for sigma in [0.1, 0.5, 1.0, 1.5, 2.0, 5.0, 10.0, 50.0] {
-            add_trial(
-                &mut trials,
+            trials.add(
                 format!("blur/{} blur sigma={sigma:.1}", image.name),
                 move || {
                     image::imageops::blur_advanced(
@@ -31,8 +34,7 @@ fn main() -> std::process::ExitCode {
                     )
                 },
             );
-            add_trial(
-                &mut trials,
+            trials.add(
                 format!("blur/{} fast_blur sigma={sigma:.1}", image.name),
                 move || image::imageops::fast_blur(&image.rgb, sigma),
             );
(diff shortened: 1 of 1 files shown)


RunDevelopment:
I forgot to mention: The cat image is CC0. I got it from [here](https://free-images.com/display/cat_hangover_red_cute_9.html). The other two images I created myself.

tests/imageops_reference.rs
```diff
@@ -0,0 +1,151 @@
...
+    path::{Path, PathBuf},
+    sync::LazyLock,
+};
+
+use libtest_mimic::{Arguments, Trial};
+
+static TESTS_DIR: LazyLock<PathBuf> =
+    LazyLock::new(|| PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("tests"));
+static ASSETS: LazyLock<PathBuf> = LazyLock::new(|| TESTS_DIR.join("assets"));
+static REFERENCE: LazyLock<PathBuf> = LazyLock::new(|| TESTS_DIR.join("imageops"));
```
197g: I think the complexity of types etc. outweighs whatever kind of efficiency we gain from a shared static. The tests are built with a lot of move-closures as well so I wouldn't shy away from more arguments (or a struct for grouping them).

tests/imageops_reference.rs
```diff
@@ -0,0 +1,151 @@
...
+        } else {
+            DynamicImage::new_rgb8(8, 8)
+        };
+
+        Box::leak(Box::new(Self {
+            name,
+            rgba: image.to_rgba8(),
+            rgb: image.to_rgb8(),
+            gray: image.to_luma8(),
+        }))
```
197g: Is this because of closures and the discussion around making it more ergonomic to clone? I'd prefer if this returned `Arc<Self>` with an explicit clone in the body before it's being moved. For tests that only have one case it'll probably be ergonomic to add a block around them.

```
    {
        let cat = cat.clone();
        add_trial(&mut trials, "filter3x3 laplace", move || {
            image::imageops::filter3x3(&cat.rgb, &[1.0, 1.0, 1.0, 1.0, -8.0, 1.0, 1.0, 1.0, 1.0])
        });
    }
```

I would even prefer that since it makes the code more uniform overall.

tests/imageops_reference.rs
RunDevelopment: I didn't do this for efficiency. These are essentially constants/variables of the test runner, and I don't want these constants to be scattered all over the code.

If `Path::join` was const, these would just be `static NAME: PathBuf = ...;`.

tests/imageops_reference.rs
RunDevelopment: I did this because I didn't want to write `let cat = cat.clone()`. I wanted test definitions to be as little code and as simple as possible. By returning `Arc`, every test has to write `image.clone()` boilerplate (and add a scope around it for name shadowing reasons).

What do we gain from using `Arc`?

197g: Not leaking. I don't want to treat tests and benchmarks different from other programs, we're just as likely to need to modify them. So general principles of data organization still apply including cleaning up.

I'd be fine with `Box::leak` if these were inside `static` constant, then leaking them is expected enough.

RunDevelopment:
> Should we bit-for-bit test these methods yet?

I would like to answer it like this: The right time to add such tests for these functions was years ago when they were added. The second best time is right now.

Suppose we change `fast_blur`. Having existing tests doesn't get in the way of that. If anything, they would show that the function did change and how. So I don't see what future changes should get in the way of testing.

tests/imageops_reference.rs
RunDevelopment: > I'd be fine with `Box::leak` if these were inside `static` constant, then leaking them is expected enough.

Okay, but then we're back to `LazyLock` since reading an image isn't `const`. Would that be okay?

197g: I'd prefer a `OnceLock` still constructed in main by `get_or_init` here, not a `LazyLock` and still _not_ have the closures access the static themselves.

To expand: the rather bothering thing about the other point is that it also has a large amount of control flow between them. They are not just used for setup work that produces a `&'static` that works easily with move closures (issue with Rust that adds visual noise), they instead obfuscate very real control and data flow through multiple statics instead of a parameter struct. `static` data is just data leaked by the loader, doing it manually has all the same properties except an explicit control and data flow where we can pass in data.

For instance, some packaging wants benchmarks to be executable but we do not ship the data in a `.crate` file. We don't need to do anything here but (I belive) it's nicer if their patch can adjust the input by a normal function instead of the dependency graph data flow of static construction order.

RunDevelopment: Sure.

Issue #2850, opened 2026-03-14
`brighten` and `threshold` are broken for images with subpixel types other than u8 and u16


RunDevelopment:
`brighten`, `unsharpen`, and related methods have the following signatures (simplified):

```rs
fn brighten<I, P, S>(image: &I, value: i32) -> ImageBuffer<P, Vec<S>>;
fn unsharpen<I, P, S>(image: &I, sigma: f32, threshold: i32) -> ImageBuffer<P, Vec<S>>;
```

Notice the `i32` parameters. These parameters are used without adjustments to the subpixel type of the image. E.g. `brighten(img, 16)` will make u8 images 6% brighter, u16 images 0.02% brighter, and f32 images 1600% brighter (all relative to maximum brightness). While rare, images with a subpixel type of `u32` can have their brightness raised/reduced by at most 50% and `u64` by at most 0.0000000012%.

As a result, high-level functions like `DynamicImage::brighten` fail to provide an API that works consistently for all underlying image types.

## Expected

`brighten` and similar should produce roughly the same images for different subpixel types.

## Actual behaviour

`brighten` and similar produce completely different outputs unless the parameter is set to 0.

## Possible solutions

Change the parameter to take an `f32` in the range `0.0..=1.0` (or `-1.0..=1.0`) and scale it internally by the maximum value of the subpixel type.

RunDevelopment:
Correction: I just re-read the code for `threshold`. It panics for (random) images of subpixel type `u32`, `u64`, `usize`, `i32`, `i64`, and `isize`. `f32` and `f64` can panic too if the user sets values outside the range of `i32` (they aren't supposed to do that, but we shouldn't panic if they do).

Relevant code:

```rs
let p = a.map2(b, |c, d| { // c and d are pixel components
    let ic: i32 = NumCast::from(c).unwrap(); // might panic for all listed 
    let id: i32 = NumCast::from(d).unwrap(); // subpixel type except i32

    let diff = ic - id; // might panic for i32 as well
```

PR #2886, opened 2026-03-22
Fix `brighten` API and f32 behavior

RunDevelopment:
Partial fix for #2850.

Changes:
- `brighten` and `brighten_in_place` now take a `f32` parameter between -1 and 1 to indicate the change in brightness.
- The implementation of these functions is now `f32` based.

In this PR, I fixed the issues with `brighten` and added some new ones. Firstly, the API now behaves consistently across subpixel types. I.e. `brighten(img, 0.5)` will always raise brightness by 50% (of total brightness). Secondly, the implementation is now based on `f32`, which finally produces correct outputs for `f32` images. Of course, `u8` and `u16` images still produce correct results (* see 3 below). Finally, signed integer types are properly supported now.

As for the new issues this introduces:

1. The `f32` implementation is slower than the previous implementation for `u8` and `u16`. (We only have to consider the slowdown for these subpixel types, because the previous implementation was incorrect for all others.) This should be easy to fix by implementing the brighten logic for each primitive instead of using a generic implementation.
2. `f64` and all integer types >24 bits lose precision due to the conversion to `f32`.
3. Amount is not rounded correctly for integer types. It currently truncates because of `NumCast::from`. This mostly affects signed integer types. This will be easy to fix once #2868 is in.

That said, I don't think any of these issues are blocking, since the main subpixel types are handled correctly.


mstoeckl:
The new interface looks reasonable; searching through Github, I find a few scripts have built a similar floating point interface for brightening. And the API migration after the break would just require passing (value as f32 / 255.0).

> As for the new issues this introduces:

I think it should now be possible to resolve all three issues by using NearestFrom to map `value.abs()` to P::Subpixel and doing arithmetic directly on the P::Subpixel. Although this would probably require adding a SaturatingAdd and SaturatingSub trait to PrimitiveSealed.

RunDevelopment:
I implemented your suggestion and it works quite well. Performance is back to normal for u8 and better for f32.

PR #2889, opened 2026-03-26
Add alpha premultiplication to tiff based on ExtraSamples tag

mpospelova:
libtiff premultiplies the alpha channel based on the ExtraSamples tag (Tag 338). This pull request mimics the behavior of libtiff.

For value 0 (unspecified), libtiff does not perform any premultiplication and this PR sets the premultiplication flag to false 

For value 1 (associated alpha), the rgb channels are already premultiplied and libtiff doesn't do anything. This PR sets the premultiplication flag to false in this case

For value 2 (unassociated alpha), libtiff multiplies the color channels by the alpha value. This PR mimics this by performing premultiplication with the pic-scale-safe library

### DIFF
--- a/src/codecs/tiff.rs
+++ b/src/codecs/tiff.rs
@@ -125,10 +125,8 @@ where
                 .find_tag_unsigned_vec::<u16>(Tag::ExtraSamples)
                 .map_err(ImageError::from_tiff_decode)?;
 
-            should_premultiply_alpha = match extra_samples.as_deref().and_then(|v| v.first()) {
-                Some(&2) => true,
-                _ => false,
-            };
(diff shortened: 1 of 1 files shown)


197g:
Our `ImageBuffer` does not have premultiplied samples so the output should undo that on associated samples (semantics we already use for color conversion and for rescaling where we premultiply in an extra step). From the description I had expected the change to be on the encoder. We should only emulate libtiff insofar as our own types align with it.

197g:
@mpospelova Of course it means our decoder is wrong in all the other cases right now. If you want to contribute a PR, I'd gladly see one fixing color fidelity in that direction by checking the tag.

PR #2895, opened 2026-04-01
Support ICO decoding mid stream

RunDevelopment:
Based on [this comment](https://github.com/image-rs/image-extras/issues/41#issuecomment-4158901403), I investigated which decoders assume that they start reading the reader at stream position 0. I found that ICO, BMP, and TIFF make that assumption and fail to decode (or decode incorrectly) when the assumption is broken.

In this PR, I fixed ICO. I did this by adjusting the code to support arbitrary starting positions for the reader. 

However, I am not sure if this is the best approach. It adds non-trivial complexity. Looking at TIFF, the code making the assumption may not even be part of `image`. So I was also considering another approach: Make the assumption true. We could make a thin wrapper around the reader which offsets the stream position such that decoders start at *their* zero. This has the advantage that (almost) no code changes to decoders are necessary, which makes it trivial to support decoders implemented elsewhere.

### DIFF
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -1,12 +1,12 @@
 use byteorder_lite::ReadBytesExt;
-use std::io::{BufRead, Read, Seek, SeekFrom};
+use std::io::{BufRead, Read, Seek};
 use std::{error, fmt};
 
 use crate::color::ColorType;
 use crate::error::{
     DecodingError, ImageError, ImageResult, UnsupportedError, UnsupportedErrorKind,
 };
-use crate::utils::OffsetReader;
+use crate::utils::seek_start_with_offset;
 use crate::{ImageDecoder, ImageFormat};
 
 use self::InnerDecoder::*;
@@ -108,12 +108,13 @@ impl From<IcoEntryImageFormat> for ImageFormat {
 /// An ico decoder
 pub struct IcoDecoder<R: BufRead + Seek> {
     selected_entry: DirEntry,
+    reader_offset: u64,
     inner_decoder: InnerDecoder<R>,
 }
 
 enum InnerDecoder<R: BufRead + Seek> {
-    Bmp(BmpDecoder<OffsetReader<R>>),
-    Png(Box<PngDecoder<OffsetReader<R>>>),
+    Bmp(BmpDecoder<R>),
+    Png(Box<PngDecoder<R>>),
 }
 
 #[derive(Clone, Copy, Default)]
@@ -143,18 +144,15 @@ struct DirEntry {
 
 impl<R: BufRead + Seek> IcoDecoder<R> {
     /// Create a new decoder that decodes from the stream ```r```
-    pub fn new(r: R) -> ImageResult<IcoDecoder<R>> {
-        // This decoder assumes that we start reading at stream_position() == 0,
-        // because the ICO format is offset based. The make that assumption true,
-        // we wrap the reader.
-        let mut r = OffsetReader::start_at_zero(r)?;
-
+    pub fn new(mut r: R) -> ImageResult<IcoDecoder<R>> {
+        let reader_offset = r.stream_position()?;
         let entries = read_entries(&mut r)?;
         let entry = best_entry(entries)?;
-        let decoder = entry.decoder(r)?;
+        let decoder = entry.decoder(r, reader_offset)?;
 
         Ok(IcoDecoder {
             selected_entry: entry,
+            reader_offset,
             inner_decoder: decoder,
         })
     }
@@ -236,13 +234,13 @@ impl DirEntry {
             && u32::from(self.real_height()) == height.min(256)
     }
 
-    fn seek_to_start<R: Read + Seek>(&self, r: &mut R) -> ImageResult<()> {
-        r.seek(SeekFrom::Start(u64::from(self.image_offset)))?;
+    fn seek_to_start<R: Read + Seek>(&self, r: &mut R, reader_offset: u64) -> ImageResult<()> {
+        seek_start_with_offset(r, reader_offset, u64::from(self.image_offset))?;
         Ok(())
     }
 
-    fn is_png<R: Read + Seek>(&self, r: &mut R) -> ImageResult<bool> {
-        self.seek_to_start(r)?;
+    fn is_png<R: Read + Seek>(&self, r: &mut R, reader_offset: u64) -> ImageResult<bool> {
+        self.seek_to_start(r, reader_offset)?;
 
         // Read the first 8 bytes to sniff the image.
         let mut signature = [0u8; 8];
@@ -251,9 +249,13 @@ impl DirEntry {
         Ok(signature == PNG_SIGNATURE)
     }
 
-    fn decoder<R: BufRead + Seek>(&self, mut r: OffsetReader<R>) -> ImageResult<InnerDecoder<R>> {
-        let is_png = self.is_png(&mut r)?;
-        self.seek_to_start(&mut r)?;
+    fn decoder<R: BufRead + Seek>(
+        &self,
+        mut r: R,
+        reader_offset: u64,
+    ) -> ImageResult<InnerDecoder<R>> {
+        let is_png = self.is_png(&mut r, reader_offset)?;
+        self.seek_to_start(&mut r, reader_offset)?;
 
         if is_png {
             let limits = crate::Limits {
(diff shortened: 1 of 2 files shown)


src/codecs/ico/decoder.rs
```diff
@@ -39,6 +39,10 @@ enum DecoderError {
         /// The dimensions of the image itself
         image: (u32, u32),
     },
+
+    /// The starting stream position of the reader is so large that the
+    /// positions of some bytes in the image data cannot be represented in a u64.
+    StreamPositionTooLarge,
```
197g: Should this just be an IO error? We could write an offset-stream wrapper that modifies `Seek` except that this is probably not very efficient. This would however be its best and consistent way of handling these errors. I believe `FileTooLarge` may be an appropriate kind?

197g:
Yes, I'm aware of the TIFF case. Currently working on that but due to the ability to have a wrapper outside and in particular for a `read_at`-based file this being trivial I don't think it's a priority bug for anyone. I could be convinced otherwise by seeing that in the wild. We should definitely fix this but not at considerable cost.

RunDevelopment:
> We should definitely fix this but not at considerable cost.

Agreed.

I think I'll implement the wrapper-around-the-reader approach. The wrapper type can be reused across implementations, so this should be less complexity over all.

fintelia:
Another option is to just return an error at the start if the stream position isn't zero

src/utils/mod.rs
```diff
@@ -156,6 +157,96 @@ pub(crate) fn is_integer<T: num_traits::NumCast + num_traits::Zero>() -> bool {
...
+            _ => pos,
+        };
+
+        let stream_pos = self.inner.seek(inner_pos)?;
+
+        if stream_pos < self.offset {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!("Seek to {pos:?} would seek before the start of the image data (which starts at {}).", self.offset),
+            ));
```
197g: By here the seek has already happened. I suppose returning an error isn't forbidden but it is odd.

RunDevelopment:
It's an option if you document this behavior for all relevant [functions](https://docs.rs/image/latest/image/fn.load.html) and [types](https://docs.rs/image/latest/image/struct.ImageReader.html) (prob missed a few). This should of course include an up-to-date list of all formats with this restriction, an explanation of it, and how to work around it. 

Seems to me that it's easier to fix it and add a test to ensure everything's correct.

src/utils/mod.rs
```diff
@@ -156,6 +157,96 @@ pub(crate) fn is_integer<T: num_traits::NumCast + num_traits::Zero>() -> bool {
...
+    #[inline]
+    #[track_caller]
+    fn seek(&mut self, pos: io::SeekFrom) -> io::Result<u64> {
+        let inner_pos = match pos {
+            io::SeekFrom::Start(pos) => {
+                let Some(inner_pos) = pos.checked_add(self.offset) else {
+                    return Err(io::Error::new(
+                        io::ErrorKind::FileTooLarge,
+                        format!("SeekFrom::Start({pos}) is invalid, because it would cause an overflow when adding the offset (which is {}).", self.offset),
+                    ));
```
197g: I've thought about the appropriate error kind here a bit more, and reconsidered. It may be more convenient to treat this as a very simple case of seeking beyond the end, or overflow as a result of adding the offset to the current position. Compare [`std::io::Cursor`](https://doc.rust-lang.org/stable/src/std/io/cursor.rs.html#293-312).

src/utils/mod.rs
```diff
@@ -156,6 +157,96 @@ pub(crate) fn is_integer<T: num_traits::NumCast + num_traits::Zero>() -> bool {
...
+        Ok(stream_pos - self.offset)
+    }
+    fn seek_relative(&mut self, offset: i64) -> io::Result<()> {
+        self.inner.seek_relative(offset)?;
+
+        debug_assert!(
+            self.inner.stream_position()? >= self.offset,
+            "Seek relative by {offset} would seek before the start of the image data (which starts at {}).",
+            self.offset
+        );
```
197g: I don't think this holds as an assert so that should be a proper error return and utilize an actual `Seek` underneath. Note that this is always just an optimization—if we can't do it that is fine. The important case will be that the `BufReader` _around_ us can avoid discarding data.

However, we can runtime dispatch on `self.offset == 0` here. On other methods it is just a control flow, here it is an opportunity to call a different underlying method.

src/utils/mod.rs
```diff
@@ -156,6 +157,96 @@ pub(crate) fn is_integer<T: num_traits::NumCast + num_traits::Zero>() -> bool {
...
+
+        Ok(())
+    }
+    fn stream_position(&mut self) -> io::Result<u64> {
+        self.inner.stream_position().map(|pos| pos - self.offset)
```
197g: > This is equivalent to self.seek(SeekFrom::Current(0)).

And I believe it should also behave that way with regards to an invalid current offset—an error instead of silent wrapping.

197g (commented):
I do believe that the wrapper is the right direction. Particularly reviewing the implementation it seems simple enough to provide it. Specifically because there are a few comments that I'd like to have changed; but they were easy enough to spot. Can you add comments indicating some of the reasoning for the implementation while doing that?

The only weirdness is that formats that are positioned will always go through the type even if their offset is `0`. This is more of a maintenance/performance perspective, Rust will only add transparently defaulted methods I assume but still explicit support for new methods requires modifying the wrapper.

src/utils/mod.rs
RunDevelopment: You're right. `self.offset == 0` is our common case, so only that must to be fast. Merely correct is good enough for everything else.

src/utils/mod.rs
RunDevelopment: I did this to reduce overhead. If we want to do this before seeking, we need to call `inner.stream_position()`. Not too expensive, but still nice to avoid.

Plus, `OffsetReader` is only used (and will ever be used) internally. Once an IO operation fails, a decoder is arguably in an invalid state. As I see it, the only reasonable ways to deal with IO failures are to stop decoding (= never use the reader again) or to (try to) reset the reader to a known state (e.g. with a seek to `Start(0)`). In either case, it's okay for the underlying reader to be in an invalid position (from the perspective of `OffsetReader`).

197g: Alright. I concur, an invalid position after an error is not forbidden.

src/codecs/ico/decoder.rs
```diff
@@ -111,8 +112,8 @@ pub struct IcoDecoder<R: BufRead + Seek> {
 }
 
 enum InnerDecoder<R: BufRead + Seek> {
-    Bmp(BmpDecoder<R>),
-    Png(Box<PngDecoder<R>>),
+    Bmp(BmpDecoder<OffsetReader<R>>),
+    Png(Box<PngDecoder<OffsetReader<R>>>),
```
fintelia: This is likely going to monomorphize 2x the number of copies of our PNG and BMP decoders because we'll have one with `R` as the reader type and one with `OffsetReader<R>`.

RunDevelopment: Yes. Is this an issue? If so, what possible solutions have you considered?

fintelia: It isn't great because it adds to compile time and binary size (2x as many copies get compiled and included in the output binary).

One alternative is to not use `OffsetReader` and instead directly track the base offset in the `IcoDecoder`. Since there's only a single method where the ICO decoder does a seek, I don't think it would be too bad? There's a certain elegance to the wrapper approach but I'm not sure it is worth it here.

RunDevelopment: It's not just here. BMP and TIFF have the same problem. TIFF seems to be in the works (see above comment by @197g). We have control over BMP, so we can fix it directly.

The approach you're proposing will work for ICO and BMP. Assuming that TIFF is fixed before v1.0, it shouldn't be an issue either. So let's go with that.

RunDevelopment:
Alright, I implement the offset-tracking approach.

---

Note for merger: All 3 approaches I implemented so far are their own commit, so please squash & merge when merging. (I did this so we can easily switch between different approaches.)

197g:
Sure, I'm okay with either approach. I'm going to add it to the list of reasons _against_ adding the reader to the decoder structs, especially in the dependent crates. The more years of Rust I spend writing the less appealing it is that the ecosystem seems to have converged on this approach (or did it ever do anything else? Especially for `deflate` and so on, providing new forms of `Read` was on-vogue very early on with lots of grief when finally `async` runtimes diverged into different traits).

PR #1173, opened 2020-03-27
farbfeld followup

nabijaczleweli:
Marked as draft since I can't get `load_rect()` to stop panicking in code I don't understand; see commit marked [WIP], particularly the diff with `read_scanline()`.

`cargo test --features farbfeld farbfeld`:
```
running 1 test
test farbfeld::tests::read_rect ... FAILED

failures:

---- farbfeld::tests::read_rect stdout ----
inner
magic = [102, 97, 114, 98, 102, 101, 108, 100]
magic ok
[src\farbfeld.rs:66] read_dimm(&mut inner) = Ok(
    2,
)
[src\farbfeld.rs:67] read_dimm(&mut inner) = Ok(
    3,
)
rrwp(1, 1, 1, 2)
start = 16
we in this
Progress { current: 0, total: 16 }
|s, scanline=12|
[src\image.rs:244] bytes_read = 0
[src\image.rs:244] &mut buf[(dbg!(bytes_read) as usize)..] = [
    0,
    64,
]
[src\image.rs:244] &mut dbg!(& mut buf
          [(dbg ! (bytes_read) as usize) ..])[..(scanline_bytes as usize)] = [
    0,
    64,
]
|s, buf=[00, 40]|
Progress { current: 2, total: 16 }
[src\image.rs:244] bytes_read = 2
[src\image.rs:244] &mut buf[(dbg!(bytes_read) as usize)..] = []
thread 'farbfeld::tests::read_rect' panicked at 'index 2 out of range for slice of length 0', src\libcore\slice\mod.rs:2674:5
stack backtrace:
   0:           0xd369ab - backtrace::backtrace::dbghelp::trace::h2c26bcbdf690198b
                               at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.40\src\backtrace/dbghelp.rs:88
   1:           0xd369ab - backtrace::backtrace::trace_unsynchronized::h5cce0bcde2e3f784
                               at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.40\src\backtrace/mod.rs:66
   2:           0xd369ab - std::sys_common::backtrace::_print_fmt::ha46fdf8ef77a98a0
                               at src\libstd\sys_common/backtrace.rs:84
   3:           0xd369ab - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h75ab5013ba686b22
[...]

PR #2902, opened 2026-04-05
Farbfeld: Simpler decoder

RunDevelopment:
Changes:
- Remove `FarbfeldReader`.
- Read the entire image at once and swap endianness on LE.

---

Frankly, I have no idea why `FarbfeldReader` existed in the first place. I read through the discussion on the PRs touching it (#1171, #1173) and nobody seems to have raised any objections. Reading pixel data channel by channel and swapping the endianness while reading is about as inefficient as it gets. We make 4 calls to the underlying reader *per pixel*. There was also a bug in `FarbfeldReader::read` that could cause a panic.

So I just ripped it out. Now the decoder reads the entire image with one call to the underlying reader total (ignoring header parsing) and then swaps endianness on LE. The compiler is going to love vectorizing the swapping function, so this is FAST.

I've also written `FarbfeldDecoder` such that generic instances of it are basically free in terms of code size and compile times. The actual parsing/decoding logic is in non-generic functions now.

### DIFF
--- a/src/codecs/farbfeld.rs
+++ b/src/codecs/farbfeld.rs
@@ -16,7 +16,7 @@
 //! # Related Links
 //! * <https://tools.suckless.org/farbfeld/> - the farbfeld specification
 
-use std::io::{self, Read, Seek, SeekFrom, Write};
+use std::io::{self, Read, Write};
 
 use crate::color::ExtendedColorType;
 use crate::error::{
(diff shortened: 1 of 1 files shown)


src/codecs/farbfeld.rs
```diff
@@ -16,186 +16,78 @@
...
-                    } else {
-                        Some(off)
-                    }
-                }
-            }
+fn u16_be_to_ne(data: &mut [u8]) {
+    #[cfg(target_endian = "little")]
+    {
+        for [low, high] in data.as_chunks_mut::<2>().0 {
+            std::mem::swap(low, high);
```
nabijaczleweli, line 63: I'll be honest with you, I don't know why I did it like that.
```rust

        for bytes in data.as_chunks_mut::<2>().0 {
            bytes.swap(0, 1);
```
?

RunDevelopment: I used `std::mem::swap`, because it optimizes more consistently from my experience. In other projects, I found that `slice::swap` sometimes produces worse assembly, so I try to avoid it for fixed indexes.

197g (approved):
Sure.

PR #2903, opened 2026-04-05
Farbfeld: Buffer encoder

RunDevelopment:
Instead of calling the writer 4 times per pixel, I buffered a few pixels and give them to the writer at once. My main intention was to reduce overhead (calling a (virtual) function isn't free), but this also reduces the need for buffered writers as a side effect.


mstoeckl (approved):
While I haven't tested this on a big-endian machine, only little-endian, the change to the logic is simple enough that I'm confident this will cause no issues.

197g (approved):

PR #2905, opened 2026-04-07
Convert most CICP whitepoints and transfer

197g:
Relax the constraints of accepting only two whitepoints and transfer functions. Moxcms implements them all and I do expect that any CMS should be able to at least provide basic support by converting through linear, unclamped floats (i.e. coded up in a day). This does not enable support for PQ due to (potential) patent coverage. The one I found details neither the method of defining the curve nor the method for computing the EOTF. So this is either intentionally incompetent or covered separately by something I did _not_ find, i.e. some rogue defensive patent. Either way—no.

We do not accept more matrices though. Yet.

### DIFF
--- a/src/metadata/cicp.rs
+++ b/src/metadata/cicp.rs
@@ -91,6 +91,25 @@ pub enum CicpColorPrimaries {
 }
 
 impl CicpColorPrimaries {
+    /// Do support these primaries in color [`CicpTransform`]?
+    pub const fn is_supported_for_transform(self) -> bool {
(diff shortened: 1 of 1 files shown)


awxkee (approved):
> This does not enable support for PQ due to (potential) patent coverage. 

Doing PQ and HLG is anyway pointless without built-in luminance mapping and gamut mapping where needed. moxcms can do it, but "do it" should be assumed to mean extremely straightforward processing, and the output is going to be wrong in almost all PQ cases.

fintelia (approved):

PR #2906, opened 2026-04-07
Add IPTC metadata support for tiff

### LINKED ISSUE #2610: IPTC metadata support

mpospelova:
This PR adds IPTC support for TIFF images.

An IRB reader was added to parse Tag 34377 and look for the IPTC block (ID 0x0404). It acts as a fallback when RichTIFFIPTC isn't there. It iterates through the blocks, checking the 8BIM signature and handling padding, without allocating extra memory

Closes https://github.com/image-rs/image/issues/2610


src/codecs/tiff.rs
```diff
@@ -431,6 +433,43 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
...
+            }
+            // Fallback: Adobe software sometimes incorrectly writes this as LONG (u32).
+            // We convert the u32 integers back to raw little-endian bytes to recover the payload.
+            if let Some(vec) = value
+                .into_u32_vec()
+                .ok()
+                .map(|vec| {
+                    vec.into_iter()
+                        .flat_map(|v| v.to_le_bytes())
+                        .collect::<Vec<u8>>()
```
RunDevelopment, line 462: What I'm suggesting is overoptimized for this use case, but hear me out:
```suggestion
                    Vec::into_flattened(
                        vec.into_iter().map(u32::to_le_bytes).collect()
                    )
```

The map should now reuse the allocation and `into_flattened` is basically just a pointer cast.

src/codecs/tiff.rs
```diff
@@ -431,6 +433,43 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
         Ok(())
     }
 
+    fn iptc_metadata(&mut self) -> ImageResult<Option<Vec<u8>>> {
+        let Some(decoder) = &mut self.inner else {
+            return Ok(None);
+        };
+
+        // Try Photoshop tag
+        if let Ok(data) = decoder.get_tag_u8_vec(TAG_PHOTOSHOP) {
```
RunDevelopment, line 442: Any reason why the PS tag takes precedence over the RichTIFFIPTC tag?

src/codecs/tiff.rs
```diff
@@ -692,3 +731,69 @@ impl<W: Write + Seek> ImageEncoder for TiffEncoder<W> {
         Ok(())
     }
 }
+
+struct IrbReader<'a> {
+    data: &'a [u8],
+}
+
+impl<'a> IrbReader<'a> {
```
RunDevelopment, line 739: I don't like that the IRB reader isn't reading a lot of IRB. It has no real logic specific to IRB. It's basically a generic reader.

I would suggest adding a method `fn read_block(&mut self) -> Option<(u16, &[u8], &[u8])>` which parses a whole block and returns the ID, name, and data if valid.

This would also simplify the extraction logic to just

```rs
    while let Some((id, _, data)) = reader.read_block() {
        if id == IPTC_ID {
            return Some(block_data);
        }
    }
```

---

Also, please link the spec: https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577409_pgfId-1037504

src/codecs/tiff.rs
```diff
@@ -692,3 +731,69 @@ impl<W: Write + Seek> ImageEncoder for TiffEncoder<W> {
...
+    fn skip_padding(&mut self, size: usize) {
+        if !size.is_multiple_of(2) && !self.data.is_empty() {
+            self.data = &self.data[1..];
+        }
+    }
+}
+
+fn extract_iptc_from_photoshop_irb(data: &[u8]) -> Option<&[u8]> {
+    const SIGNATURE: &[u8] = b"8BIM";
+    const IPTC_ID: u16 = 0x0404;
```
RunDevelopment, line 772: A link to the spec for this constant would be nice here too: https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577409_38034

RunDevelopment (changes requested):
Sorry for the delay and thank you for making this!

Regarding the Photoshop tag, how common is for Adobe to not write the slightly more standard RichTIFFIPTC tag and use their own instead?

And on that note, not related to this PR, but the Photoshop tag and its IRBs can store other metadata too. E.g. ICC, XMP, EXIF. Does Photoshop use the standard TIFF tags for that for those or does it use its IRBs?

---

Closes #2610

PR #2907, opened 2026-04-09
Assorted optimizations for GStreamer's image-rs plugin

amyspark:
Hi,

This is a MR to add a few things for feature parity in my GStreamer reader plugin:

- PNG: `iCCP` chunk's profile name
- AVIF: speed up the reader by delaying actual decoding until `read_image`

Both require upstream work, so marking as draft until those are landed: https://github.com/image-rs/image-png/pull/679, https://github.com/mozilla/mp4parse-rust/pull/446

### DIFF
--- a/src/io/decoder.rs
+++ b/src/io/decoder.rs
@@ -23,6 +23,13 @@ pub trait ImageDecoder {
         Ok(None)
     }
 
+    /// Return the ICC profile's name as embedded in the image.
+    ///
+    /// For formats that don't store embedded profiles with separate names this function should always return `Ok(None)`.
+    fn icc_profile_name(&mut self) -> ImageResult<Option<&String>> {
+        Ok(None)
+    }
+
     /// Returns the raw [Exif](https://en.wikipedia.org/wiki/Exif) chunk, if it is present.
     /// A third-party crate such as [`kamadak-exif`](https://docs.rs/kamadak-exif/) is required to actually parse it.
     ///
@@ -142,6 +149,9 @@ impl<T: ?Sized + ImageDecoder> ImageDecoder for Box<T> {
     }
     fn original_color_type(&self) -> ExtendedColorType {
         (**self).original_color_type()
+    }
+    fn icc_profile_name(&mut self) -> ImageResult<Option<&String>> {
+        (**self).icc_profile_name()
     }
     fn icc_profile(&mut self) -> ImageResult<Option<Vec<u8>>> {
         (**self).icc_profile()
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -1,7 +1,6 @@
 //! Decoding of AVIF images.
 use crate::error::{
-    DecodingError, ImageFormatHint, LimitError, LimitErrorKind, UnsupportedError,
-    UnsupportedErrorKind,
+    DecodingError, LimitError, LimitErrorKind, UnsupportedError, UnsupportedErrorKind,
 };
 use crate::{ColorType, ImageDecoder, ImageError, ImageFormat, ImageResult};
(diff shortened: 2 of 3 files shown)


src/io/decoder.rs
```diff
@@ -23,6 +23,13 @@ pub trait ImageDecoder {
         Ok(None)
     }
 
+    /// Return the ICC profile's name as embedded in the image.
+    ///
+    /// For formats that don't store embedded profiles with separate names this function should always return `Ok(None)`.
```
fintelia, line 28: Are there formats other than PNG which _do_ store names for the embedded profiles?

amyspark: That's the only one. I introduced this API, however, because I didn't know how to downcast from `impl ImageDecoder` to `PngDecoder` (and so access a decoder specific API).

197g: We may need a novel mechanism for this. Downcast ties it to the `png` version which conflicts with the requirement of being able to bump it (through breaking changes). On the other hand a new API on the trait is also rather awkward. Maybe something similar to [`Error::provide`](https://doc.rust-lang.org/stable/core/error/trait.Error.html#method.provide) would fit the bill in letting us translate types provided by version-encumbered crates to stable types defined in `image` directly (that we then only let include the 'stable' options / data from the formats). The converse holds for encoder, another unsatisfied user ask.

fintelia: You're trying to cast between a `Box<dyn ImageDecoder>` and a `image::codecs::png::PngDecoder`, right?

For that, I think we might just be able switch to a `Box<dyn ImageDecoder + Any>` or make `Any` a super-trait of `ImageDecoder`?

197g: `Any` does not really solve the problem. It only makes the dependency problem a runtime error but I'd consider that _worse_ than a breaking change everytime we bump dependencies. There's no reasonable way to use `Any::downcast` right here for downstream—they'd have to add fallback cases for potentially _unpublished_ versions of our dependencies preemptively to avoid functionality breaking and that just pollutes the dependency graph greatly.

fintelia: No, if I'm understanding correctly, the goal isn't to downcast to the decoder struct from the `png` crate but rather to cast to the [`PngDecoder`](https://docs.rs/image/latest/image/codecs/png/struct.PngDecoder.html) struct in this crate

amyspark: > No, if I'm understanding correctly, the goal isn't to downcast to the decoder struct from the png crate but rather to cast to the [PngDecoder](https://docs.rs/image/latest/image/codecs/png/struct.PngDecoder.html) struct in this crate

Correct, I already have to do something similar when dealing with image-extras formats that do not support guessing. It is also required when accessing the animation API IIRC, because only explicitly typed instances of `AnimationDecoder` allow calling `into_frames`.

RunDevelopment: > make Any a super-trait of ImageDecoder?

`Any` requires `Self: 'static`. Decoders are generic over readers and some useful readers (e.g. `&[u8]`) aren't `'static`. So `ImageDecoder: Any` would be a major limitation.

It's also unclear to me how the downcast would even work. Say we did require `ImageDecoder: Any`, you can't downcast unless you know the exact reader type. E.g. `image_decoder.downcast_ref::<images::codecs::PngDecoder<Reader>>()`. Would that even be useful?

> only explicitly typed instances of AnimationDecoder allow calling into_frames.

In 0.25.x, yes, but this will change soon with the new decoding API. `ImageDecoder` will handle single images, sequences, and animations. #2672

fintelia: Hm, yeah the `'static` requirement does make that less workable

PR #2917, opened 2026-04-23
BMP: Make RLE decoder lenient on row overflow

mpospelova:
This PR modifies the BMP decoder to handle malformed RLE data more leniently. When an RLE run specifies a count that exceeds the remaining space in the current row, the decoder now silently truncates the run and breaks, rather than returning a fatal CorruptRleData error.

Reason: 
Many legacy BMP encoders have alignment bugs that cause RLE runs to overshoot the scanline width. Those files are still visually valid and are being decoded correctly in other software like Chrome or Firefox. Even if the images are actually corrupt, Chrome for example is still being able to decode them without failing. This change allows the rust BMP decoder to decode such images now as well. I took corrupted RLE images for testing from https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html


197g (approved):
Ah, nice. You already fixed what I was about to remark, gating these under the spec complicance flag.

LGTM then.

PR #2919, opened 2026-04-24
BMP: clamp invalid data_offset pointing into headers

mpospelova:
Malformed BMP files may declare a `data_offset` that falls within the DIB header range [14, bmp_header_end). This caused the decoder to seek backwards into header bytes and interpret them as pixel data.  This PR clamps such offsets to `bmp_header_end` during metadata parsing.


telecos:
Thanks for the contribution! Given that this special handling malformed BMP, shouldn't this new handling be done only if Lenient mode is enabled?

PR #2920, opened 2026-04-24
Fixed `GrayImage::expand_palette` at pixel 0,0

### LINKED ISSUE #2918: `GrayImage::expand_palette` does not expand pixel at `(0,0)` based on the palette
This happens in `image` `v0.25.10` using rust `v1.95.0`. When using `GrayImage::expand_palette`, the pixel at `(0,0)` does not get correctly expanded to the value in the palette.

## Expected

The pixel at `(0,0)` should get expanded based on the provided palette.

## Actual behaviour

The pixel at `(0,0)` gets expanded into a `Rgba` with the actual palette index for each component.

## Reproduction steps

Rust Playground link:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=103094beb0697215768b93ed1acde483

Code:
```rust
fn main() {
    use image::{GrayImage, Luma, Rgba};

    let imggr1 = GrayImage::from_pixel(2, 2, Luma([0]));
    let palette = [(255, 255, 255)];

    let expandes = imggr1.expand_palette(&palette, None);

    assert_eq!(expandes.get_pixel(1, 0), &Rgba([255, 255, 255, 255]));
    assert_eq!(expandes.get_pixel(0, 1), &Rgba([255, 255, 255, 255]));
    assert_eq!(expandes.get_pixel(1, 1), &Rgba([255, 255, 255, 255]));
    assert_eq!(expandes.get_pixel(0, 0), &Rgba([255, 255, 255, 255])); // fails here
}
```
Prints
```
thread 'main' (13) panicked at src/main.rs:12:5:
assertion `left == right` failed
  left: Rgba([0, 0, 0, 0])
 right: Rgba([255, 255, 255, 255])
```
If you change the fill index to fill the image with `1` instead, you see how the pixel at `(0,0)` is now expanded to four `1`s:

```rust
fn main() {
        use image::{GrayImage, Luma, Rgba};

        let imggr1 = GrayImage::from_pixel(2, 2, Luma([1]));
        let palette = [(255, 255, 255), (0, 255, 0)];

        let expanded = imggr1.expand_palette(&palette, None);

        assert_eq!(expanded.get_pixel(1, 0), &Rgba([0, 255, 0, 255]));
        assert_eq!(expanded.get_pixel(0, 1), &Rgba([0, 255, 0, 255]));
        assert_eq!(expanded.get_pixel(1, 1), &Rgba([0, 255, 0, 255]));
        assert_eq!(expanded.get_pixel(0, 0), &Rgba([0, 255, 0, 255])); // still fails here
}
```
Output:
```
thread 'main' (10) panicked at src/main.rs:12:9:
assertion `left == right` failed
  left: Rgba([1, 1, 1, 1])
 right: Rgba([0, 255, 0, 255])
```

Changing the `Luma` value to `2` gives `expanded.get_pixel(0,0) == Rgba([2, 2, 2, 2])` etc.

RunDevelopment:
Fixes #2918.

The bug was in `expand_packed` in this line:

```rs
    let buf_len = buf.len();
    let j_inv = (channels..buf_len).step_by(channels); // <---
    for ((shift, i), j_inv) in i.zip(j_inv) {
        let j = buf_len - j_inv;
        let pixel = (buf[i] & (mask << shift)) >> shift;
        func(pixel, &mut buf[j..(j + channels)]);
```

The range in `j_inv` goes to `buf_len` *exclusive*, so `j` is never zero. That `..` should have been `..=`.

However, that's not how I fixed it. I removed `expand_packed`. The function is much more general than necessary for `GrayImage::expand_palette` (the only caller of that function), and all it does is a loop. So I just moved the simple loop inside `GrayImage::expand_palette` and removed `expand_packed`. Should be faster too.

I added a test too.

---

Unrelated: `GrayImage::expand_palette` seems like a very limited API to me. I also don't get why `expand_palette` performs everything inplace. The commit that did that said it was for performance, but `data.resize` most likely re-allocs anyway, so it seems more efficient to me to alloc a new vec and initializing it the indexed colors directly (e.g. via `Vec::from_iter`). This would also enable a more expressive API.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1629,7 +1629,7 @@ pub trait ConvertBuffer<T> {
 impl GrayImage {
     /// Expands a color palette by re-using the existing buffer.
     /// Assumes 8 bit per pixel. Uses an optionally transparent index to
-    /// adjust it's alpha value accordingly.
+    /// adjust its alpha value accordingly.
     #[must_use]
     pub fn expand_palette(
         self,
@@ -1647,22 +1647,15 @@ impl GrayImage {
             let write_index = read_index * 4;
 
             let palette_index = bytes[read_index];
-            let pixel = &mut bytes[write_index..write_index + 4];
 
             let (r, g, b) = palette[palette_index as usize];
-            let a = if let Some(t_idx) = transparent_idx {
-                if t_idx == palette_index {
-                    0
-                } else {
-                    255
-                }
+            let a = if Some(palette_index) == transparent_idx {
+                0
             } else {
                 255
             };
-            pixel[0] = r;
-            pixel[1] = g;
-            pixel[2] = b;
-            pixel[3] = a;
+
+            bytes[write_index..][..4].copy_from_slice(&[r, g, b, a]);
         }
 
         buffer
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1643,8 +1643,7 @@ impl GrayImage {
         let mut buffer = ImageBuffer::from_vec(width, height, data).unwrap();
 
         let bytes = &mut *buffer;
-        for i in 0..entries {
-            let read_index = entries - 1 - i;
+        for read_index in (0..entries).rev() {
             let write_index = read_index * 4;
 
             let palette_index = bytes[read_index];
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -14,7 +14,6 @@ use crate::flat::{FlatSamples, SampleLayout, ViewOfPixel};
 use crate::math::Rect;
 use crate::metadata::cicp::{CicpApplicable, CicpPixelCast, CicpRgb, ColorComponentForCicp};
 use crate::traits::{EncodableLayout, Pixel, PixelWithColorType};
-use crate::utils::expand_packed;
 use crate::{
     metadata::{Cicp, CicpColorPrimaries, CicpTransferCharacteristics, CicpTransform},
     save_buffer, save_buffer_with_format, write_buffer_with_format, ImageError,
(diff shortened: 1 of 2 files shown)


src/images/buffer.rs
```diff
@@ -1642,10 +1641,18 @@ impl GrayImage {
         let entries = data.len();
         data.resize(entries.checked_mul(4).unwrap(), 0);
         let mut buffer = ImageBuffer::from_vec(width, height, data).unwrap();
-        expand_packed(&mut buffer, 4, 8, |idx, pixel| {
-            let (r, g, b) = palette[idx as usize];
+
+        let bytes = &mut *buffer;
+        for i in 0..entries {
+            let read_index = entries - 1 - i;
```
okaneco:
```suggestion
        for read_index in (0..entries).rev() {
```
`i` is only used on this line, I find it easier to understand the intention by using an iterator like the old method.

src/images/buffer.rs
okaneco: I think you should be able to write something like this to collapse the ifs?
```rust
let a = if Some(palette_index) == transparent_idx {
    0
} else {
    255
};

File: src/images/buffer.rs [hunk 2]
```diff
@@ -1642,10 +1641,18 @@ impl GrayImage {
...
+            let write_index = read_index * 4;
+
+            let palette_index = bytes[read_index];
+            let pixel = &mut bytes[write_index..write_index + 4];
```
okaneco: This line would make more sense if it were placed after getting the alpha value.

Instead of creating the mutable reference, we could also build up the array and copy to slice.
```rust
bytes[write_index..][..4].copy_from_slice(&[r, g, b, a]);
```

197g:
@okaneco Thanks for the review, seems pretty thorough to me already.

Re: `Unrelated …`. It was previously used a bit more. `png` in particular did not always expand and we may want to expand and `ImageDecoderRect`, passing scanline-by-scanline, would invoke it on piecewise to avoid the allocation of another buffer with full size. This was another motivation for the in-place code as there may have been another short-lived allocation. Regardless, that mess is fixed now so it is more or less legacy here.

197g:
To put the conclusion explicitly if you want to got for a rewrite that explicitly uses a separate temporary buffer to then process items in-order, you could also try that route with one or two benchmarks thrown in. I'll guess that those will agree with a separate allocation at most sizes due to improvements in codegen.

RunDevelopment:
> if you want to go for a rewrite that explicitly uses a separate temporary buffer to then process items in-order, you could also try that route with one or two benchmarks thrown in.

Sure, but in a follow-up PR.

RunDevelopment:
Thanks for the review @okaneco! I address your comments. Could you please re-review?

okaneco (approved):

197g (approved):

PR #2914, opened 2026-04-20
GIF: Zip `pixels_mut()` for faster animation decoding

RunDevelopment:
I replaced the `enumerate_pixels_mut()` + `get_pixel_mut()` combo with two zipped `pixels_mut()`. This removes the overhead of bounds checking and coordinate tracking.

I also added a benchmark for animation decoding to measure the perf gain. Before: 7.6806 ms, after: 6.2113 ms (-19.130%).

<details><summary>Full benchmark results</summary>

Before:

```
load-Gif/large-gif-anim-full-frame-replace.gif
                        time:   [1.8674 ms 1.8902 ms 1.9169 ms]
Found 19 outliers among 100 measurements (19.00%)
  6 (6.00%) high mild
  13 (13.00%) high severe
load-Gif/Animation: large-gif-anim-full-frame-replace.gif
                        time:   [7.5737 ms 7.6806 ms 7.8073 ms]
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild
  10 (10.00%) high severe
load-Gif/large-gif-anim-combine.gif
                        time:   [2.7986 ms 2.8495 ms 2.9086 ms]
Found 14 outliers among 100 measurements (14.00%)
  2 (2.00%) high mild
  12 (12.00%) high severe
load-Gif/Animation: large-gif-anim-combine.gif
                        time:   [9.8276 ms 9.9649 ms 10.126 ms]
Found 13 outliers among 100 measurements (13.00%)
  3 (3.00%) high mild
  10 (10.00%) high severe
```

After:

```
load-Gif/large-gif-anim-full-frame-replace.gif
                        time:   [1.8580 ms 1.8831 ms 1.9115 ms]
                        change: [-3.7041% -0.7083% +2.3049%] (p = 0.66 > 0.05)
                        No change in performance detected.
Found 19 outliers among 100 measurements (19.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  16 (16.00%) high severe
load-Gif/Animation: large-gif-anim-full-frame-replace.gif
                        time:   [6.0712 ms 6.2113 ms 6.3758 ms]
                        change: [-21.222% -19.130% -16.598%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) high mild
  8 (8.00%) high severe
load-Gif/large-gif-anim-combine.gif
                        time:   [2.7384 ms 2.7886 ms 2.8459 ms]
[...]

PR #2923, opened 2026-04-27
Fix panic on every benchmark introduced by #2914

Shnatsel:
Without this the code always panics when running nearly any benchmark.


RunDevelopment (approved):
Woops, my bad. I only benchmarked GIF images, so I didn't notice...

Issue #2295, opened 2024-07-28
`DynamicImage::crop()` is slow because `to_image()` is slow


Shnatsel:
In v0.25.2, the crop operation is implemented as:

https://github.com/image-rs/image/blob/e176cd414ac6cc73909ff162e5d0b677f9d4fb08/src/dynimage.rs#L422

which obtains a view into an image, and then calls `to_image()` on it to turn it into a new image buffer. And `to_image()` shuffles pixels one by one in a purely scalar fashion and with bounds checks on every pixel:

https://github.com/image-rs/image/blob/e176cd414ac6cc73909ff162e5d0b677f9d4fb08/src/image.rs#L1131-L1136

A much better way would be iterating over the rows, and copying over entire rows with `slice::copy_from_slice()`. That would both leverage vector instructions and reduce the amount of bounds checks from n^2 to n.

PR #2928, opened 2026-04-27
Improve perf of `DynamicImage::crop`

RunDevelopment:
Partial fix for #2295

I couldn't make `SubImage::to_image` faster, so I side-stepped the issue by implementing a new `ImageBuffer::crop` and deferring `DynamicImage::crop` to that. This makes the operation 3x faster on my machine.

I also added a benchmark for the operation.

<details><summary>Details</summary>

Before:

```
dyn image crop          time:   [2.1008 ms 2.1250 ms 2.1536 ms]
Found 16 outliers among 100 measurements (16.00%)
  5 (5.00%) high mild
  11 (11.00%) high severe
```

After:

```
dyn image crop          time:   [664.53 µs 708.05 µs 759.58 µs]
                        change: [-68.370% -66.769% -64.819%] (p = 0.00 < 0.05)
                        Performance has improved.
```

</p>
</details>

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -14,6 +14,7 @@ use crate::flat::{FlatSamples, SampleLayout, ViewOfPixel};
 use crate::math::Rect;
 use crate::metadata::cicp::{CicpApplicable, CicpPixelCast, CicpRgb, ColorComponentForCicp};
 use crate::traits::{EncodableLayout, Pixel, PixelWithColorType};
+use crate::utils::vec_try_with_capacity;
 use crate::{
     metadata::{Cicp, CicpColorPrimaries, CicpTransferCharacteristics, CicpTransform},
     save_buffer, save_buffer_with_format, write_buffer_with_format, ImageError,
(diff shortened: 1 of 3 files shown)


src/images/buffer.rs
```diff
@@ -986,6 +987,27 @@ where
...
+        let row_pitch = data_len / selection.height as usize;
+
+        for y in 0..selection.height {
+            let sy = selection.y + y;
+            let source = self.pixel_indices_unchecked(selection.x, sy).start;
+            data.extend_from_slice(&self.data[source..source + row_pitch]);
+        }
+
+        let mut out = ImageBuffer::from_raw(selection.width, selection.height, data).unwrap();
+        out.color = self.color;
```
197g, line 1007: Preferably this would use `copy_color_space_from` instead of a direct copy.

197g (approved):
Makes sense that this would be more optimized on a concrete type. Hm, I'm a little disappointed that `to_image` is not optimized though. Maybe it could have a fast path for [`GenericImageView::to_pixel_view`], which might use precisely the added code but from a generic, borrowed, matrix instead of the owned buffer.

But the PR is also self contained as-is.

RunDevelopment:
> a fast path [...] which might use precisely the added code but from a generic, borrowed, matrix [...].

Urgh, this sounds annoying.... but doable.

Let's keep this PR on hold for now. I'll make a separate PR for this approach. If it works out, we just close this PR, and if it doesn't, we still have this PR as a backup.

RunDevelopment:
Closing since #2929 got merged

PR #2929, opened 2026-04-28
Improve perf of `SubImage::to_image`

### LINKED ISSUE #2295: `DynamicImage::crop()` is slow because `to_image()` is slow

RunDevelopment:
Fixes https://github.com/image-rs/image/issues/2295

This PR implements @197g's idea for [a fast path](https://github.com/image-rs/image/pull/2928#pullrequestreview-4184934836) in `SubImage::to_image` using `to_pixel_view`. At the heart of this lies the new `View::iter_rows` method. This returns an iterator over all pixel rows in the flat buffer, if those rows are contiguous in memory and pixels within them are packed.

Just like in #2928, this makes `DynamicImage::crop` 3x faster. I also added the same benchmark as in #2928.

<details>

```
dyn image crop          time:   [676.30 µs 718.70 µs 769.05 µs]
                        change: [-67.332% -65.684% -63.447%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 19 outliers among 100 measurements (19.00%)
  3 (3.00%) high mild
  16 (16.00%) high severe
```

</details>

---

Aside: `iter_rows` could potentially be used to simplify the implementation of `ImageBuffer::copy_from_samples`. It also seems related to #2300.

### DIFF
--- a/src/images/flat.rs
+++ b/src/images/flat.rs
@@ -1260,6 +1260,10 @@ where
         let row_len = layout.width as usize * channels as usize;
         let row_stride = layout.height_stride;
 
+        // Note: We do **NOT** assume row_len <= row_stride. Rows may overlap
+        // (or even all be the same slice when row_stride == 0). So the usual
+        // iter().chunks() strategy does not work here.
+
         Some((0..layout.height as usize).map(move |y| {
             let start = y * row_stride;
             &data[start..start + row_len]
--- a/src/images/flat.rs
+++ b/src/images/flat.rs
@@ -1243,6 +1243,28 @@ where
             phantom: PhantomData,
         })
     }
+
+    /// If the underlying samples are in row-major form and packed pixels, return
+    /// an iterator over the rows of the image.
+    pub(crate) fn iter_rows(&self) -> Option<impl Iterator<Item = &[P::Subpixel]> + '_> {
+        let layout = self.flat().layout;
+        let channels = P::CHANNEL_COUNT;
+        if layout.channel_stride != 1
+            || layout.channels != channels
+            || layout.width_stride != channels as usize
+        {
+            return None;
+        }
+
+        let data = self.samples().as_ref();
+        let row_len = layout.width as usize * channels as usize;
+        let row_stride = layout.height_stride;
+
+        Some((0..layout.height as usize).map(move |y| {
+            let start = y * row_stride;
+            &data[start..start + row_len]
+        }))
+    }
 }
 
 impl<Buffer, P: Pixel> ViewMut<Buffer, P>
--- a/src/images/sub_image.rs
+++ b/src/images/sub_image.rs
@@ -88,12 +88,9 @@ impl<I> SubImage<I> {
         // fast path for row-major packed views
         if let Some(view) = self.to_pixel_view() {
             if let Some(row_iter) = view.iter_rows() {
-                let out_data = out.deref_mut();
-                let row_len = out_data.len() / h as usize;
-                for (y, row) in row_iter.enumerate() {
-                    out_data[y * row_len..(y + 1) * row_len].copy_from_slice(row);
+                for (row, out_row) in row_iter.zip(out.rows_mut().into_slices()) {
+                    out_row.copy_from_slice(row);
                 }
-
                 return out;
             }
         }
--- a/src/images/sub_image.rs
+++ b/src/images/sub_image.rs
@@ -80,11 +80,26 @@ impl<I> SubImage<I> {
         I: Deref,
         I::Target: GenericImageView + 'static,
     {
+        let w = self.inner.xstride;
+        let h = self.inner.ystride;
         let borrowed = &*self.inner.image;
-        let mut out = borrowed.buffer_with_dimensions(self.inner.xstride, self.inner.ystride);
+        let mut out = borrowed.buffer_with_dimensions(w, h);
 
-        for y in 0..self.inner.ystride {
-            for x in 0..self.inner.xstride {
+        // fast path for row-major packed views
+        if let Some(view) = self.to_pixel_view() {
+            if let Some(row_iter) = view.iter_rows() {
+                let out_data = out.deref_mut();
+                let row_len = out_data.len() / h as usize;
+                for (y, row) in row_iter.enumerate() {
+                    out_data[y * row_len..(y + 1) * row_len].copy_from_slice(row);
+                }
+
+                return out;
+            }
+        }
+
+        for y in 0..h {
+            for x in 0..w {
                 let p = borrowed.get_pixel(x + self.inner.xoffset, y + self.inner.yoffset);
                 out.put_pixel(x, y, p);
             }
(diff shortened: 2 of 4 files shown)


src/images/flat.rs
```diff
@@ -1243,6 +1243,28 @@ where
...
+        }
+
+        let data = self.samples().as_ref();
+        let row_len = layout.width as usize * channels as usize;
+        let row_stride = layout.height_stride;
+
+        Some((0..layout.height as usize).map(move |y| {
+            let start = y * row_stride;
+            &data[start..start + row_len]
+        }))
```
197g, line 1270: Interesting: I thought it might be clearer to rewrite in terms of an iterator over the `data`-slice but that requires `slice::windows`; this loop is more general than `slice::chunks_exact` since it does not require that `row_len <= row_stride` (overlap is permitted by `View`).

So best variant I could come up with would be:


```suggestion
        Some(data
            .windows(row_len)
            .step_by(row_stride)
            .take(layout.height as usize))
```

I'm not certain I find that particularly readable though.

Both seem fine, generally?

It's going to be significant for a `iter_rows_mut` though.

src/images/sub_image.rs
```diff
@@ -80,11 +80,26 @@ impl<I> SubImage<I> {
...
+        let h = self.inner.ystride;
         let borrowed = &*self.inner.image;
-        let mut out = borrowed.buffer_with_dimensions(self.inner.xstride, self.inner.ystride);
+        let mut out = borrowed.buffer_with_dimensions(w, h);
+
+        // fast path for row-major packed views
+        if let Some(view) = self.to_pixel_view() {
+            if let Some(row_iter) = view.iter_rows() {
+                let out_data = out.deref_mut();
+                let row_len = out_data.len() / h as usize;
```
197g: I'd rather the output row_len be calculated from `w` here or the view since there's no guarantee that the container underlying `out` is sized _exactly_ for the image data it contains. Extra elements are necessary to preserve initialized parts of the allocation.

197g (commented):
I'd like to hear thoughts on the implementation comment but there's no inherent problem with it. `View`'s invariants should have taken care of any potential index overflows and out-of-bounds that might surface differences in either.

src/images/flat.rs
RunDevelopment: I didn't even think about overlapping slices, so good point. As you said, for a mutable iterator this will be a problem. Maybe we should have the same requirements for both? So including `row_len <= row_stride`. This would be an artificial limitation for the iterator of immutable slices though.

Also, `chunks_exact` can't be used to implement that even if `row_len <= row_stride`. E.g. `data.chunks_exact(row_stride).map(|s| &s[..row_len]).take(h)` is incorrect. `data` might only be `row_stride * (h-1) + row_len` elements long. `chunks` would work though.

As for `slice::window`, I don't know. I prefer my explicit slicing. Makes it obvious that certain iterator features (e.g. `nth`, exact size iter) are easy to implement.

src/images/sub_image.rs
RunDevelopment: I did this to avoid potential overflow.

> Extra elements are necessary to preserve initialized parts of the allocation.

Not sure what you mean by that.

src/images/flat.rs
197g: It was a pretty conscious choice on `View` _not_ to have the same requirements for both since that allows it to express constant color images and repeated things. There was no real upside of having the artificial limitation.

Regarding the `Map` iterator: `nth` is _not_ specialized (much) for it since it'll [call the closure for all elements](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=536860a5a4723437c93925b0534ab835); compare to `windows` which does specialize it by advancing the start so `step_by` should be comparatively efficient.

I think the slicing and explicit indices makes it more obvious which elements are yielded by each iteration, so let's stay with that.

Could you add a note about not assuming `rowlen <= row_stride` and `chunks` being ineligible for that?

src/images/sub_image.rs
197g: How do you mean to avoid overflow? Having an `ImageBuffer` should be enough to know the indices work. It's up to `buffer_with_dimensions` to actually return the right buffer with `(w, h)` but that is a contract, it's fine to do something weird if it is violated.

> Extra elements are necessary to preserve initialized parts of the allocation.

You're allowed to construct an `ImageBuffer` with _more_ space in the vector than required by the pixel count. That's useful if you want to preserve a vector's length, maybe reuse it later, with having to re-zero the tail. `out.len()` is the underlying size, not the number of channels involved so `out_data.len() / h as usize;` can be wrong.

It also does not account for `w = 0`.

Maybe a nice way of solving this is adding `into_remainder` for `RowsMut` so that you can use `ImageBuffer::rows_mut` as an iterator over each row's raw channels instead of only being able to use it to iterate individual pixels per row. That could prove quite useful anyways. (Note its behavior: when the image is empty it won't return any rows itself, bypassing the chunk length `0` problem).

src/images/flat.rs
RunDevelopment: Alrighty, no unnecessary requirements then. Sounds good.

> Regarding the `Map` iterator: `nth` is _not_ specialized (much) for it

Oh, that's not what I meant. I meant that *we* can implement a (custom) iterator that does these things very efficiently.

> Could you add a note

Sure 👍

src/images/sub_image.rs
RunDevelopment: I implement *something* for RowsMut. Since the inner ChunksExactMut is exactly what I needed, I just returned that using RowsMut::into_slices(self). I don't like the name, but it works.

---

Also, is there any reason why Rows{,Mut} yield Pixels{,Mut} instead of plain &{mut}[Pixel] slices? Slices seem a lot more useful to me. 

It also seems to me that the methods Pixel::from_slice{,_mut} implicitly require types implementing the trait to have an alignment <= align_of(SubPixel) and a size == size_of(SubPixel) * CHANNEL_COUNT. This would mean that not only is casting a slice to a pixel reference safe, but casting a slice of subpixels to a slice of pixels should also be safe (if the length of the subpixel slice is divisible by channel count).

<sub>No code formatting because my windows randomly decided that the backtick key should stop working. It'll work again after a reboot.</sub>

197g (approved):

PR #2930, opened 2026-04-28
BMP: Implement lenient decoding for missing final byte on last row

mpospelova:
This PR implements lenient handling for BMP files that are truncated by exactly one byte at the end of the file. This is an issue with some legacy BMP encoders that omit the trailing padding byte on the final scanline. The image is fine otherwise.

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -564,8 +564,8 @@ fn allocate_row_buffer(size: usize) -> ImageResult<Vec<u8>> {
 /// last one, it performs a normal read. Otherwise, the special case applies:
 /// Apparently many BMPs are missing the final byte at the end of the file.
 /// This function checks if the stream is exactly one byte short of the
-/// required final scanline length. If so, it reads the available bytes, leaving
-/// the missing trailing byte as zero. Otherwise, it performs a normal `read_exact`.
+/// required final scanline length. If so, it reads the available bytes and
+/// explicitly zeroes the missing trailing byte. Otherwise, it performs a normal `read_exact`.
 fn read_scanline(
     reader: &mut (impl io::Read + Seek),
     buf: &mut [u8],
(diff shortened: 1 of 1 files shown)


src/codecs/bmp/decoder.rs
```diff
@@ -1642,7 +1666,13 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
             start_row,
             |row| {
-                reader.read_exact(&mut indices)?;
+                let is_last_row = current_file_row == last_row;
+                current_file_row += 1;
+                if is_last_row && spec_strictness == SpecCompliance::Lenient {
+                    read_last_scanline_lenient(reader, &mut indices)?;
+                } else {
+                    reader.read_exact(&mut indices)?;
+                }
```
telecos: I see that this code snippet is defined in several place. Maybe we can extract a helper method to avoid duplication if possible

mpospelova: Thank you! I have extracted this functionality to the `read_scanline` method.

telecos (approved):
LGTM. Thanks!

src/codecs/bmp/decoder.rs
```diff
@@ -560,6 +560,37 @@ fn allocate_row_buffer(size: usize) -> ImageResult<Vec<u8>> {
...
+    spec_strictness: SpecCompliance,
+) -> io::Result<()> {
+    let is_last_row = *current_file_row == last_row;
+    *current_file_row += 1;
+
+    if is_last_row && spec_strictness == SpecCompliance::Lenient {
+        let current_pos = reader.stream_position()?;
+        let end_pos = reader.seek(SeekFrom::End(0))?;
+        reader.seek(SeekFrom::Start(current_pos))?;
+        let remaining = (end_pos - current_pos) as usize;
```
197g: On 32-bit systems this `as` conversion could drop some relevant data and then incidentally match the length. Also the documentation says the last byte is *left* as zero, which is a confusingly worded: it can be left untouched; or explicitly zeroed. In this case it seems to be the former so isn't the value dependent on the prior row?

Writing the block more robustly:

```
let Some((last, head)) = buf.last_mut() else {
    // Empty row, nothing to read.
    return Ok(());
};

if Ok(head.len()) == usize::try_from(end_pos - current_pos) {
    reader.read_exact(head)?;
    *last = b'\0';
    Ok(());
}
```

197g (commented):
The stylistic question isn't my top-concern, both seem fine so happy to accommodate your preference here. Moving a loop-constant (and control flow based on it) into an opaque function call is a trade-off to me, despite the clear duplication. Eh, taste.

PR #2931, opened 2026-04-29
Faster and non-destructive `GrayImage::expand_palette`

RunDevelopment:
This implements a non-destructive version of `GrayImage::expand_palette` I suggested in #2920.

The implementation pulls a few tricks for speed:
- It uses an exact-size iterator, so collecting into a vec allocates only once and then initializes the memory with the yielded data.
- It creates a full 256-long palette to avoid bounds checking and the "branch" for `transparent_idx`.

In the first commit of this PR, you can see the benchmark code comparing the old function and the new one. I tested the old function in two scenarios: (1) the vec of the gray image is too small and requires a realloc, and (2) the vec of the gray image is exactly the right size for the RGBA image and no new memory needs to be allocated. Scenario 2 is the ideal scenario for the old implementation.

Results:

| Image Size | Old (1)   | Old (2)   | New       |
| ---------: | --------- | --------- | --------- |
|       64^2 | 6.3112 µs | 6.1838 µs | 1.7512 µs |
|      256^2 | 99.176 µs | 97.005 µs | 24.220 µs |
|     1024^2 | 2.1806 ms | 1.9472 ms | 1.1150 ms |
|     2048^2 | 8.6819 ms | 7.7218 ms | 4.4289 ms |

As we can see, the new implementation outperforms the old one by a significant margin.

---

I also fixed the bug that the old implementation assumed `data.len() == w * h`.

### DIFF
--- a/benches/imageops.rs
+++ b/benches/imageops.rs
@@ -1,11 +1,11 @@
-use std::time::Duration;
+use std::{hint::black_box, time::Duration};
 
 use criterion::{criterion_group, criterion_main, Criterion};
 use image::{
     buffer::ConvertBuffer,
     imageops::{self, FilterType},
     math::Rect,
-    DynamicImage, GrayImage, ImageBuffer, Rgb,
+    DynamicImage, GrayImage, ImageBuffer, Luma, Rgb,
 };
 
 pub fn bench_imageops(c: &mut Criterion) {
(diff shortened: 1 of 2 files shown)


benches/imageops.rs
```diff
@@ -85,6 +86,11 @@ pub fn bench_imageops(c: &mut Criterion) {
             })
         });
     });
+
+    c.bench_function("gray image expend_palette", |b| {
```
197g:
```suggestion
    c.bench_function("gray image expand_palette", |b| {
```

nit: simple typo

src/images/buffer.rs
```diff
@@ -1631,38 +1631,34 @@ pub trait ConvertBuffer<T> {
...
+            full_palette[palette_index as usize][3] = 0;
         }
 
-        buffer
+        let rgba_data: Vec<[u8; 4]> = self.as_raw()[..len]
+            .iter()
+            .map(|&palette_index| full_palette[palette_index as usize])
+            .collect();
+
+        ImageBuffer::from_vec(width, height, rgba_data.into_flattened()).unwrap()
```
197g, line 1662: `into_flattened` is a neat trick 👍 . This was stabilized all the way back in `1.80`? That went totally over my head.

197g (commented):
The new tight loop for constructing the data Vec is great for large images no question. Plus I see the point that the destructive form does not really contribute much anyways.

However, it is a rather large amount of stack usage; plus I would suppose the numbers are less convincing on small images where the expansion of the LUT is significant in runtime. So maybe the expanded palette could be moved into a `Box`-ed allocation and only be used for large-ish images (cut-off something arbitrary) instead? Just a thought to explore, if it does not get too complex—constructing data with an intermediate `Vec<[u8; 4]>` to-be-flattened which is collected from an iterator with reliable size_hint can be done in both paths.

Shnatsel:
FWIW we've already explored similar tricks for `png` crate and concluded that going the lookup table route is beneficial almost always, and keeping the fallback path for small images isn't worth it.

The benchmarks for this PR show a 64x64 image improving 6x so I'm not too concerned about small image performance.

RunDevelopment:
> it is a rather large amount of stack usage

It's 1024 bytes. That's really not a lot. Some structs in some projects are that big.

Anyway. I heap-allocated the full palette. It's probably smaller for small images now since it adds another allocation, but I couldn't measure any perf change in the benchmark.

197g (approved):
Thanks, yeah it's not too bad overall (std's io::copy allocates a full DEFAULT_BUF_SIZE) but stack overflows deep down hurt. Anyways, maybe I tend to be overcautious but if benchmarks do not detect it, eh. Open to adjust it afterwards with more evidence of usage / cost.

PR #2933, opened 2026-04-29
BmpDecoder - Add support for alpha mask in RGB file which was ignored

### LINKED ISSUE #2932: BmpDecoder - Missing support for Alpha mask in BI_BMP
During the Chromium field trial there has been reported a regression in which BMP files encoded using BI_BMP and setting an alpha_mask are being rendered as RGB

## Expected

These files should render as RGBA with transparency

## Actual behaviour

Transparent pixels are rendered black

## Reproduction steps

Issue details: https://issues.chromium.org/issues/507939858

telecos:
When a V4/V5 BMP is encoding an alpha mask, it should treat file as RGBA to allow rendering transparency.

Example https://jsfiddle.net/wx51o4q7/

Expected: 

image

Current behaviour:

image

Fix #2932


src/codecs/bmp/decoder.rs
```diff
@@ -1486,6 +1486,19 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
...
                 };
             }
+        } else if self.image_type == ImageType::RGB32 && bmp_header_size >= BITMAPV4HEADER_SIZE {
+            // V4/V5 headers may declare an alpha channel via alpha_mask even under BI_RGB.
+            let mut masks_buf = [0u8; 16];
+            self.reader.read_exact(&mut masks_buf)?;
+            let alpha_mask = u32::from_le_bytes(masks_buf[12..16].try_into().unwrap());
+            if alpha_mask != 0 {
+                // BI_RGB implies fixed BGRA layout; only 0xFF000000 is valid.
+                if self.spec_strictness == SpecCompliance::Strict && alpha_mask != 0xFF000000 {
```
197g, line 1500: Presuming the answer motivated this extra code: Do other alpha masks fit the bill of 'lenient' behavior, i.e. are there samples of such images and is interpreting them with `0xff000000` despite their contradictory header value expected / useful behavior? It would be nice to have that spelled out or referenced somewhere.

197g (approved):
LGTM, question mostly out of curiosity but I also wouldn't mind an extra code comment for it.

src/codecs/bmp/decoder.rs
telecos: Updated comment explaining why lenient mode accepts other mask values and added a specific file with bad alpha mask to the lenient_vs_strict test to validate that file is accepted in lenient mode but rejected in strict mode.

Shnatsel (approved):
The clarification and the extra test look good. Thanks!

PR #2934, opened 2026-05-01
Add L32F and La32F to ColorType and DynamicImage

mstoeckl:
This adds new L32F and La32F ColorType entries.

My motivation here is mainly to make it easier to decode image formats that can contain L32F pixels (TIFF, EXR, and [perhaps](https://github.com/mstoeckl/image/tree/pfm) [PFM](https://pauldebevec.com/Research/HDR/PFM/)), but this change also the nice benefit of making DynamicImage's subformats form a grid of pixel dimensions and subpixel type, so that operations like DynamicImage::grayscale reliably produce a Luma or LumaA image, and we could later add methods like `to_32f` that affect the subpixel type but not the pixel component count.

See also a previous attempt at this in https://github.com/image-rs/image/pull/1846, and the recent change adding L32F and La32F ExtendedColorTypes in https://github.com/image-rs/image/pull/2793.

There is a lot of casework involved in this MR; I _think_ I've updated or added every function that needs adding, but it's not impossible that I missed something.

### DIFF
--- a/src/images/dynimage.rs
+++ b/src/images/dynimage.rs
@@ -7,8 +7,8 @@ use crate::error::{ImageError, ImageResult, ParameterError, ParameterErrorKind};
 use crate::flat::FlatSamples;
 use crate::imageops::{gaussian_blur_dyn_image, GaussianBlurParameters};
 use crate::images::buffer::{
-    Gray16Image, GrayAlpha16Image, GrayAlphaImage, GrayImage, ImageBuffer, Rgb16Image, Rgb32FImage,
-    RgbImage, Rgba16Image, Rgba32FImage, RgbaImage,
+    Gray16Image, Gray32FImage, GrayAlpha16Image, GrayAlpha32FImage, GrayAlphaImage, GrayImage,
+    ImageBuffer, Rgb16Image, Rgb32FImage, RgbImage, Rgba16Image, Rgba32FImage, RgbaImage,
 };
 use crate::io::encoder::ImageEncoderBoxed;
 use crate::io::free_functions::{self, encoder_for_format};
@@ -94,6 +94,12 @@ pub enum DynamicImage {
     /// Each pixel in this image is 16-bit Rgb with alpha
     ImageRgba16(Rgba16Image),
 
+    /// Each pixel in this image is 32-bit float Luma
+    ImageLuma32F(Gray32FImage),
+
+    /// Each pixel in this image is 32-bit float Luma with alpha
+    ImageLumaA32F(GrayAlpha32FImage),
+
     /// Each pixel in this image is 32-bit float Rgb
     ImageRgb32F(Rgb32FImage),
 
@@ -113,6 +119,8 @@ macro_rules! dynamic_map(
                 ImageLumaA16($image) => ImageLumaA16($action),
                 ImageRgb16($image) => ImageRgb16($action),
                 ImageRgba16($image) => ImageRgba16($action),
+                ImageLuma32F($image) => ImageLuma32F($action),
+                ImageLumaA32F($image) => ImageLumaA32F($action),
                 ImageRgb32F($image) => ImageRgb32F($action),
                 ImageRgba32F($image) => ImageRgba32F($action),
             }
@@ -128,6 +136,8 @@ macro_rules! dynamic_map(
                 DynamicImage::ImageLumaA16($image) => $action,
                 DynamicImage::ImageRgb16($image) => $action,
                 DynamicImage::ImageRgba16($image) => $action,
+                DynamicImage::ImageLuma32F($image) => $action,
+                DynamicImage::ImageLumaA32F($image) => $action,
                 DynamicImage::ImageRgb32F($image) => $action,
                 DynamicImage::ImageRgba32F($image) => $action,
             }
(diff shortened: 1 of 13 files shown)


src/images/dynimage.rs
```diff
@@ -1484,15 +1577,15 @@ impl From<Rgba32FImage> for DynamicImage {
...
 }
 
-impl From<ImageBuffer<LumaA<f32>, Vec<f32>>> for DynamicImage {
-    fn from(image: ImageBuffer<LumaA<f32>, Vec<f32>>) -> Self {
-        DynamicImage::ImageRgba32F(image.convert())
+impl From<GrayAlpha32FImage> for DynamicImage {
+    fn from(image: GrayAlpha32FImage) -> Self {
+        DynamicImage::ImageLumaA32F(image)
     }
 }
```
RunDevelopment, line 1629: Yikes. I didn't know that these conversions created a whole new image before. Definitely good that this is changed.

src/images/dynimage.rs
```diff
@@ -346,7 +376,7 @@ impl DynamicImage {
 
     /// Returns a copy of this image as a Luma image.
     #[must_use]
-    pub fn to_luma32f(&self) -> ImageBuffer<Luma<f32>, Vec<f32>> {
+    pub fn to_luma32f(&self) -> Gray32FImage {
         dynamic_map!(self, ref p, p.cast_in_color_space())
```
RunDevelopment: This should be a `match` like above.

src/images/dynimage.rs
```diff
@@ -370,7 +400,7 @@ impl DynamicImage {
 
     /// Returns a copy of this image as a `LumaA` image.
     #[must_use]
-    pub fn to_luma_alpha32f(&self) -> ImageBuffer<LumaA<f32>, Vec<f32>> {
+    pub fn to_luma_alpha32f(&self) -> GrayAlpha32FImage {
         dynamic_map!(self, ref p, p.cast_in_color_space())
```
RunDevelopment: Same here.

RunDevelopment (changes requested):
Great stuff!

I really like the symmetry between color types this creates. I'm also surprised by how little code you needed to change to make this happen. I expected it to be much more.

src/images/dynimage.rs
mstoeckl: Good point, updated.

src/images/dynimage.rs
mstoeckl: Also updated.

fintelia (changes requested):
If you're going to relitigate intentional design decisions that we previously made (even citing the previous thread where I stated that these variants intentionally left off!) then please start with a issue thread explaining what has changed/why we should reconsider, rather than jumping straight to a PR.

RunDevelopment:
@fintelia Could you please state the intentions/reasons behind this design decision? The [previous comment](https://github.com/image-rs/image/pull/1846#issuecomment-1404334570) you referred to only states that this decision exists, not why.

197g:
What definitely _did_ surface since then is that, while implementing color conversion, it was noted that [`DynamicImage::to`](https://docs.rs/image/latest/image/enum.DynamicImage.html#method.to) already had to deal with all the complexity of having `LF32` and `LaF32`, without having any of the tools of `ColorType` which made the whole interface somewhat absurdly odd. (And as a consequence, it was impossible to make it color space aware). If `ColorType` was designed as a product-type between channel type and color model instead, which this PR would result in, that would allow other designs to much better rely on dispatch as a solution to this instead so they deal with each of those dimensions individually rather than coupled.

197g:
Expanding on the above, concretely it was impossible to add [`to_alpha_mask`](https://github.com/image-rs/image/pull/2544) and a setter for `DynamicImage` due to the lack of pure luma variants for all color types.

The design decision I remember is to use a small subset of variants we can get away with as informed by combinations occurring real world image files. We also have those here now or in the forseeable future: unassociated alpha, uncolored image data, 3d-depth data, and more scientific files (tiff but also JPEGXL) support 24-bit or 32-bit colors in floating point. We additional made another design choice of performing color conversion when loading into `DynamicImage` (i.e. simplify it so not everyone has to deal with YCbCr and so on); but such conversion results in banding for non-linear rescaling; will always require clipping which makes it impossible to represent HDR data where non-SDR has larger-than-1 samples; in particular ICC support would benefit from non-clipped 32-bit data with single-channel images.

mstoeckl:
Apologies @fintelia , I see now that you've replied on this topic many times in the past.  I've linked to some of the issues/PRs in https://github.com/image-rs/image/issues/2935.

src/images/dynimage.rs
```diff
@@ -264,8 +292,10 @@ impl DynamicImage {
             + FromColor<color::Rgb<u16>>
             + FromColor<Luma<u8>>
             + FromColor<Luma<u16>>
+            + FromColor<Luma<f32>>
             + FromColor<LumaA<u16>>
-            + FromColor<LumaA<u8>>,
+            + FromColor<LumaA<u8>>
+            + FromColor<LumaA<f32>>,
```
197g, line 310: This breaking change being a good reason to do it _now_ rather than delay it.

197g (approved):
I'm okay with this for aforementioned reason of making `ColorType` and `DynamicImage` a proper product of channel type (`u8`, `u16`, `f32`) and color model (`L`, `La`, `Rgb`, `Rgba`).

I'm also confident that Tiff, and generic ICCv4 tables (note it defines `f64` arrays but does not allow its use in any baseline tags), as well as any necessarily _transformed_ inputs of depth ~16 (e.g. allowed in Jpeg which does Yuv->Rgb) that would otherwise suffer banding, warrant at least having the option of a floating point representation—and these are not guaranteed to have at 3 channels or a sensible Rgb interpretation.

Issue #2936, opened 2026-05-04
Extend subpixel slice casting for pixels


RunDevelopment:
Right now, `Pixel` has 2 methods to cast subpixel slices to a pixel reference:

- `fn Pixel::from_slice(slice: &[Self::Subpixel]) -> &Self`
- `fn Pixel::from_slice_mut(slice: &mut [Self::Subpixel]) -> &mut Self`

Additionally, `Pixel::channels()` and `Pixel::channels_mut()` can be used to "cast" a pixel reference to a slice of subpixels. While these methods aren't semantically defined that way, this is what they are doing in effect.

I would like to propose extending this interface with 2 new methods to cast a slice of subpixels to a slice of pixels:

- `fn Pixel::slice_from_slice(slice: &[Self::Subpixel]) -> &[Self]`
- `fn Pixel::slice_from_slice_mut(slice: &mut [Self::Subpixel]) -> &mut [Self]`

The direction from pixel slice to subpixel slice might also be useful, but I don't see them as vital.

## Why

To make `Pixels` and `PixelsMut` unnecessary. These 2 types currently exist seemingly solely for the purpose of casting slices of subpixels to a pixel reference. Returning a slice directly in e.g. `ImageBuffer::pixels_mut()` is strictly more powerful than returning a `PixelsMut` iterator.

This can also unlock better performance and enable new use cases. E.g. parallel pixel processing is trivial with a slice and rayon.

This arguably also removes the need for `ImageBuffer::par_pixels{,_mut}`, so users can just use rayon on the slice directly. I.e. `image.pixels().par_iter()`.

## Requirements regarding alignment and size

Making these functions part of the interface implicitly imposes the following requirements on all implementing types:

- `size_of::<Self>() == Self::CHANNEL_COUNT * size_of::<Self::Subpixel>()`
- `align_of::<Self>() == align_of::<Self::Subpixel>()`

However, these requirements were arguably already imposed on implementing types. Any implementation that implements `Pixel::from_slice_mut` and `Pixel::channels_mut` with the property `Self::channels_mut().len() == Self::CHANNEL_COUNT`, is already forced into being a wrapper around `[Self::Subpixel; N]`. Almost anything else would lead to UB.

## Draft

I made [a draft](https://github.com/RunDevelopment/image/tree/pixel-slice-cast) implementing the basics of my proposal. It just removes `Pixels{,Mut}` and otherwise doesn't use the returned slices yet.

RunDevelopment:
Closed by #2938

PR #2938, opened 2026-05-04
Add slice casting to `Pixel`

RunDevelopment:
This PR implements my proposal #2936. For motivation, see the proposal. (The main point is that slices are strictly more powerful than iterators.)

API changes:
- Add `fn Pixel::slice_from_slice(slice: &[Self::Subpixel]) -> &[Self]` and `fn Pixel::slice_from_slice_mut(slice: &mut [Self::Subpixel]) -> &mut [Self]` to cast a slice of subpixels to a slice of pixels.
- Change `ImageBuffer::pixels()` and `ImageBuffer::pixels_mut()` to return pixel slices inside of iterators.
- Change `Rows` and `RowsMut` iterators to yield pixel slices.
- Remove `Pixels`, `PixelsMut`, `PixelsPar`, and `PixelsMutPar` iterators.

Notes:
- I removed `Pixels` and `PixelsPar` because they are now unnecessary. Use `image.pixels().iter()` and `image.pixels().par_iter()` instead. Same for the mut variants.
- I ran all benchmarks and found no perf regressions.
- This unfortunately adds 2 new `unsafe` to implement slice casting.

### Future work

- `Rows` and `RowsMut` are now wrappers around `ChunksExact{,Mut}` and don't implement any real logic. They could be removed as well.
- We could expose `ImageBuffer::inner_pixels() -> &[P::Subpixel]` under the name `ImageBuffer::subpixels()`. This would create a nice symmetry with `ImageBuffer::pixels() -> &[P]`. Same for mut, of course.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -395,6 +395,16 @@ impl<T: $($bound+)*> Pixel for $ident<T> {
         unsafe { std::slice::from_raw_parts_mut(slice.as_mut_ptr() as *mut $ident<T>, len) }
     }
 
+    fn pixels_as_channels(slice: &[$ident<T>]) -> &[T] {
+        let len = slice.len() * $channels;
+        unsafe { std::slice::from_raw_parts(slice.as_ptr() as *const T, len) }
+    }
+
+    fn pixels_as_channels_mut(slice: &mut [$ident<T>]) -> &mut [T] {
+        let len = slice.len() * $channels;
+        unsafe { std::slice::from_raw_parts_mut(slice.as_mut_ptr() as *mut T, len) }
+    }
+
     fn broadcast(val: T) -> $ident<T> {
         $ident([val; $channels])
     }
--- a/src/color.rs
+++ b/src/color.rs
@@ -401,16 +401,6 @@ impl<T: $($bound+)*> Pixel for $ident<T> {
         unsafe { std::slice::from_raw_parts_mut(slice.as_mut_ptr() as *mut $ident<T>, len) }
     }
 
-    fn to_subpixel_slice(slice: &[$ident<T>]) -> &[T] {
-        let len = slice.len() * $channels;
-        unsafe { std::slice::from_raw_parts(slice.as_ptr() as *const T, len) }
-    }
-
-    fn to_subpixel_slice_mut(slice: &mut [$ident<T>]) -> &mut [T] {
-        let len = slice.len() * $channels;
-        unsafe { std::slice::from_raw_parts_mut(slice.as_mut_ptr() as *mut T, len) }
-    }
-
     fn broadcast(val: T) -> $ident<T> {
         $ident([val; $channels])
     }
--- a/src/color.rs
+++ b/src/color.rs
@@ -385,6 +385,28 @@ impl<T: $($bound+)*> Pixel for $ident<T> {
         unsafe { &mut *(slice.as_mut_ptr() as *mut $ident<T>) }
     }
 
+    #[track_caller]
+    #[allow(clippy::modulo_one)]
+    fn slice_from_slice(slice: &[T]) -> &[ $ident<T>] {
+        assert_eq!(slice.len() % $channels, 0);
+        unsafe { std::slice::from_raw_parts(slice.as_ptr() as *const $ident<T>, slice.len() / $channels) }
+    }
+
+    #[track_caller]
+    #[allow(clippy::modulo_one)]
+    fn slice_from_slice_mut(slice: &mut [T]) -> &mut [ $ident<T>] {
+        assert_eq!(slice.len() % $channels, 0);
+        unsafe { std::slice::from_raw_parts_mut(slice.as_mut_ptr() as *mut $ident<T>, slice.len() / $channels) }
+    }
+
+    fn to_subpixel_slice(slice: &[$ident<T>]) -> &[T] {
+        unsafe { std::slice::from_raw_parts(slice.as_ptr() as *const T, slice.len() * $channels) }
+    }
+
+    fn to_subpixel_slice_mut(slice: &mut [$ident<T>]) -> &mut [T] {
+        unsafe { std::slice::from_raw_parts_mut(slice.as_mut_ptr() as *mut T, slice.len() * $channels) }
+    }
+
     fn broadcast(val: T) -> $ident<T> {
         $ident([val; $channels])
     }
--- a/src/color.rs
+++ b/src/color.rs
@@ -385,12 +385,12 @@ impl<T: $($bound+)*> Pixel for $ident<T> {
         unsafe { &mut *(slice.as_mut_ptr() as *mut $ident<T>) }
     }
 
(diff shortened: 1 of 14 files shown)


src/traits.rs
```diff
@@ -474,6 +474,20 @@ pub trait Pixel: Copy + Clone {
     /// that the slice is long enough to prevent panics if the pixel is used later on.
     fn from_slice_mut(slice: &mut [Self::Subpixel]) -> &mut Self;
 
+    /// Returns a slice of pixels from a slice of subpixels.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if `slice.len()` is not a multiple of `CHANNEL_COUNT`.
+    fn slice_from_slice(slice: &[Self::Subpixel]) -> &[Self];
```
197g: The equivalent note on from_slice is not worded well either but I feel like we should not promise this as a part of the API contract of the trait. It could be encouraged, sure. See impl note.

197g: Very unfortunate that we can't write this as:

```
    fn _(slice: &[[u8; Self::CHANNEL_COUNT as usize]]);
```

But `slice_from_arrays` will be available for when that will become possible and that should never panic.

src/images/sub_image.rs
```diff
@@ -88,7 +88,9 @@ impl<I> SubImage<I> {
         // fast path for row-major packed views
         if let Some(view) = self.to_pixel_view() {
             if let Some(row_iter) = view.iter_rows() {
-                for (row, out_row) in row_iter.zip(out.rows_mut().into_slices()) {
+                let row_len = (w as usize) * <DerefPixel<I> as Pixel>::CHANNEL_COUNT as usize;
+                for (row, out_row) in row_iter.zip(out.inner_pixels_mut().chunks_exact_mut(row_len))
+                {
```
197g: This is a clear indication that we should introduce the inverse operation in the same PR.

src/color.rs
```diff
@@ -385,6 +385,22 @@ impl<T: $($bound+)*> Pixel for $ident<T> {
         unsafe { &mut *(slice.as_mut_ptr() as *mut $ident<T>) }
     }
 
+    #[track_caller]
+    #[allow(clippy::modulo_one)]
+    fn slice_from_slice(slice: &[T]) -> &[ $ident<T>] {
+        assert_eq!(slice.len() % $channels, 0);
```
197g: I'm surprised clippy did not complain about [manual_is_multiple_of](https://rust-lang.github.io/rust-clippy/master/index.html#manual_is_multiple_of) for the assert. In any case this only ensures that the returned does not hold less data than the input—which may be suboptimal but not unsound.

Maybe this method should return an `Option`?

src/traits.rs
197g: Bikeshedding, so to be done last: The name seems rather poor. If we have the inverse then of course it should be consistent with it. That would point to .

1. `pixels_from_slice`, but `slice_from_pixels`.
2. `slice_as_pixels` / `pixels_as_slice`.
3. `channels_as_pixels` / `pixels_as_channels`.

Option 1. seems a little odd. To me it does not communicate we're talking about a slice of channels. Option 3. fits with the existing method. I'm not fixed on either yet, what seems hard is fitting the eventual array variant in with `std` suggesting `chunks_as` and `as_chunks` though.

197g (commented):
Yeah, that seems quite a lot better. Very unfortunate that we can't write this as:

    fn _(slice: &[[Self::Subpixel; Self::CHANNEL_COUNT as usize]]) -> &[Self];

To me that is the largest motivation for posing the question of returning `Option<&[Self]>` in the slice conversion instead of panicking. Most of the existing code, that goes through an image buffer, would then have that panic path lifted into `ImageBuffer::rows{,_mut}` with little impact. Of course, that's not entirely consistent with the existing method.. I'm not entirely sure. `slice::as_chunks` of course offers a third alternative of returning `(&[Self], &[Self::Subpixel])`..

src/color.rs
RunDevelopment: Me too, but I'm not complaining. I don't like `is_multiple_of` very much. To me, `num % f == 0` is more intuitive.

> Maybe this method should return an Option?

That, or we change the semantics to return the longest prefix of pixels. I.e. all `slice.len() / channels` of them. This would make the impl panic-free and branch-free at the cost of not catching weird inputs.

src/images/sub_image.rs
RunDevelopment: Kinda? I don't disagree that it feels like the inverse op is missing, but this code isn't clear proof of that. I initially wrote this as copying slices of pixels (instead of subpixels) but then changed to this because I feared that RGB with its 3 components might mess up performance.

My goal was to keep the changes in this PR as small as I reasonably could, but I can add the inverse ops if you think they should be added.

src/traits.rs
RunDevelopment: Yeah, I couldn't think of any good names.

I think `pixels_from_slice` creates a nice symmetry with `fn Pixel::from_slice(...) -> &Self`. But that's only the case because `from_slice` arguably isn't the best name either...

Maybe we could call them `ref_from_channels`, `pixels_from_channels`, and `pixels_as_channels`? Or `ref_from_channels`, `slice_from_channels`, and `slice_as_channels`?

I think that if we give the new methods better names, the names of the existing methods should fit in and make sense.

RunDevelopment:
> posing the question of returning `Option<&[Self]>`

I don't mind returning an option, but I feel like it might not fit here from an API perspective. We would return `None` when the cast went wrong, but if the caller supplies a slice of the wrong length, the input is arguably invalid. But I also don't understand your concern about panic paths, so maybe I'm missing something.

I also want to repeat my suggestion for returning a valid prefix. This would remove the panic by making all inputs allowed, even if some make more sense than others.

> `slice::as_chunks` of course offers a third alternative of returning `(&[Self], &[Self::Subpixel])`

Not a huge fan. It makes sense to return the tail for `as_chunks`, but is that tail useful to use? I feel like we would only ever do `Pixel::chunk_pixels(slice).0` like we do with `slice::as_chunks` in this codebase. I don't see the merit in making the API more complex that it has to be.

src/images/sub_image.rs
RunDevelopment: I added an inverse op for now. If we deem that it is not necessary or adds too much complexity, we can remove it at anytime.

src/color.rs
RunDevelopment: I changed it to prefix semantics.

src/traits.rs
RunDevelopment: The method now guarantees that it returns a prefix of whole pixels. I think this is reasonable as the API contract, but you might feel different, so please let me know if you would like to change it.

197g: I do like `{from,as}_channels`. That makes sense all around, I don't think the `ref_` prefix is necessary even for any of the cases.

197g (approved):
Good, naming and semantics seems good to me.

PR #2939, opened 2026-05-06
Omit LUT features from moxcms

charliermarsh:
## Summary

Given:

```rust
use image::metadata::Cicp;
use image::{ImageBuffer, Rgb};

fn main() {
    let mut image = ImageBuffer::from_fn(1, 1, |_, _| Rgb([255u8, 0, 0]));
    image.set_rgb_primaries(Cicp::SRGB.primaries);
    image.set_transfer_function(Cicp::SRGB.transfer);
    image
        .apply_color_space(Cicp::DISPLAY_P3, Default::default())
        .unwrap();
    std::hint::black_box(image);
}
```

This PR reduces the binary size from 1,052,208 B to 831,152 B, by removing the unused LUT features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,14 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 bytemuck = { version = "1.8.0", features = ["extern_crate_alloc"] } # includes cast_vec
 byteorder-lite = "0.1.0"
-moxcms = "0.8.0"
+moxcms = { version = "0.8.0", default-features = false, features = [
+    "avx_shaper_fixed_point_paths",
+    "avx_shaper_paths",
+    "neon_shaper_fixed_point_paths",
+    "neon_shaper_paths",
+    "sse_shaper_fixed_point_paths",
+    "sse_shaper_paths",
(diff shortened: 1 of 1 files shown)


RunDevelopment:
This looks okay to me and tests pass, but I also don't know much about color. I just wonder why the compiler couldn't remove supposedly unused code by itself.

@awxkee, you're probably most qualified to review this.

awxkee:
This unsurprisingly looks good to me since I did this. However, last time I discussed using features from cms there was an organizational point that using `default` is simpler https://github.com/image-rs/image/pull/2827#pullrequestreview-3899353366 and this likely needs resolution from someone who knows how it wanted to be organized. Currently image uses only fixed point paths, so actually everything else can be dropped.

197g (approved):
Well, I had only reviewed compile times so the size is a new data point to me. It seems pretty a convincing to separate features for this. The organization question is ensuring that the feature set is sufficient for what we require but given that we just need to be correct with the minimum set most of this will only pop up when we try to process ICC ourselves. I'll make a note to mention the performance outcomes in the release notes.

PR #2940, opened 2026-05-07
Expose `ImageBuffer::subpixels*`

RunDevelopment:
Follow-up to #2938

### Changes

- Publicly expse `ImageBuffer::subpixels` and `ImageBuffer::subpixels_mut`. (Formerls `inner_pixels` and `inner_pixels_mut` respectively)
- Add documentation.

### Motivation

Before this PR, the most obvious and simple ways to get a slice of subpixels were `image.deref()` or `image.as_mut()` or `image.as_slice()` or `&*image` (depending on context). The problem is that these are all wrong. `ImageBuffer` derefs to the underlying container which is allowed to contain more subpixels than necessary for image. I.e. `image.len()` is required to be >= `w*h*c`, not equal. This makes it easy to create bugs since a lot of code assumes that ``image.as_slice()`` and co. have exactly `w*h*c` elements. To make things worse, not only was it easy to do the wrong, there was also no easy way to do the right thing.

This changes with `ImageBuffer::subpixels`. Now there at least is a function that gives a slice of subpixels that is guaranteed to be `w*h*c` subpixels long.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -572,6 +572,7 @@ where
     /// Returns a slice of the subpixels of this image.
     ///
     /// This is guaranteed to contain exactly `width * height * channels` subpixels.
+    #[doc(alias = "channels")]
     pub fn subpixels(&self) -> &[P::Subpixel] {
         let len = Self::image_buffer_len(self.width, self.height).unwrap();
         &self.data[..len]
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -569,17 +569,21 @@ where
         self.height
     }
 
-    // TODO: choose name under which to expose.
-    pub(crate) fn inner_pixels(&self) -> &[P::Subpixel] {
+    /// Returns a slice of the subpixels of this image.
+    ///
+    /// This is guaranteed to contain exactly `width * height * channels` subpixels.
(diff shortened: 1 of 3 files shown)


src/images/buffer.rs
```diff
@@ -569,17 +569,21 @@ where
         self.height
     }
 
-    // TODO: choose name under which to expose.
-    pub(crate) fn inner_pixels(&self) -> &[P::Subpixel] {
+    /// Returns a slice of the subpixels of this image.
+    ///
+    /// This is guaranteed to contain exactly `width * height * channels` subpixels.
+    pub fn subpixels(&self) -> &[P::Subpixel] {
```
197g, line 576: nit: despite our naming convention, I think people are likely enough to search for 'channels' as a term. (If you want you could add that alias to the documentation of the trait, too).

```suggestion
    #[doc(alias = "channels")]
    pub fn subpixels(&self) -> &[P::Subpixel] {
```

197g (approved):
LGTM, the name is consistent with everything we have. Definitely makes more sense than `inner` and `pixels` 🙂

src/images/buffer.rs
RunDevelopment: If you think that would be beneficial, then we can add this alias.

PR #2944, opened 2026-05-07
Use `.subpixels()` to access subpixels consistently

RunDevelopment:
I looked at how the codebase accessed subpixels prior to #2940 and found multiple issues. They all follow the same pattern: the backing container is assumed to be sized exactly `w*h*c` when that is not guaranteed.

In this PR, I unified multiple different access patterns to always use `.subpixels()`. This should ensure consistency and correctness throughout the codebase.

---

One remaining bug is `DynamicImage::into_bytes`.

### DIFF
--- a/src/imageops/colorops.rs
+++ b/src/imageops/colorops.rs
@@ -585,7 +585,7 @@ mod test {
         let mut image = ImageBuffer::from_raw(2, 2, vec![127, 127, 127, 127]).unwrap();
         let cmap = BiLevel;
         dither(&mut image, &cmap);
-        assert_eq!(&*image, &[0, 0xFF, 0xFF, 0]);
+        assert_eq!(image.subpixels(), &[0, 0xFF, 0xFF, 0]);
         assert_eq!(index_colors(&image, &cmap).into_raw(), vec![0, 1, 1, 0]);
     }
 
--- a/src/codecs/openexr.rs
+++ b/src/codecs/openexr.rs
@@ -353,7 +353,7 @@ mod test {
     fn write_rgb_image(write: impl Write + Seek, image: &Rgb32FImage) -> ImageResult<()> {
         write_buffer(
             write,
-            bytemuck::cast_slice(image.as_raw().as_slice()),
+            bytemuck::cast_slice(image.subpixels()),
             image.width(),
             image.height(),
             ExtendedColorType::Rgb32F,
@@ -366,7 +366,7 @@ mod test {
     fn write_rgba_image(write: impl Write + Seek, image: &Rgba32FImage) -> ImageResult<()> {
         write_buffer(
             write,
-            bytemuck::cast_slice(image.as_raw().as_slice()),
+            bytemuck::cast_slice(image.subpixels()),
             image.width(),
             image.height(),
             ExtendedColorType::Rgba32F,
--- a/src/imageops/fast_blur.rs
+++ b/src/imageops/fast_blur.rs
@@ -35,7 +35,7 @@ pub fn fast_blur<P: Pixel>(
         return input_buffer.clone();
     }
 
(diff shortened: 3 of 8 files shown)


src/imageops/colorops.rs
```diff
@@ -585,7 +585,7 @@ mod test {
         let mut image = ImageBuffer::from_raw(2, 2, vec![127, 127, 127, 127]).unwrap();
         let cmap = BiLevel;
         dither(&mut image, &cmap);
-        assert_eq!(&*image, &[0, 0xFF, 0xFF, 0]);
+        assert_eq!(image.subpixels(), &[0, 0xFF, 0xFF, 0]);
```
197g, line 588: This one is fine (it's constructed just above with an explicit buffer provided) but for the sake of clarity, sure.

197g (commented):
re: `as_bytes`/`into_bytes`. Hm. The inconsistency of `as_mut_bytes` always having been implemented in terms of subpixels certainly suggests that we should truncate here; I'll take it but maybe the names could be clarified (`as_pixel_bytes`)? Not entirely sure, I don't want to jump the gun on making the name more verbose than necessary.

Also `DynamicImage::into_bytes` is only missing a call to truncate on the owning path and a slice indexing on the other, right? That is the length consistent with the above is given by `size_of_val(self.subpixels())` that is easy to construct before. (the panic induced of slice indexing instead of a pure calculation may be a feature, so some care should be taken that the owning path also panics even though it need not strictly verify the length).

RunDevelopment:
> re: `as_bytes`/`into_bytes`.

I mostly went by the document for intended semantics. It says that it returns "this image's pixels." So a slice of the buffer including data beyond the image's pixels seems wrong to me. Similar for `into_bytes`. 

I would also expect that `image.as_bytes().to_vec() == image.as_bytes_mut().to_vec() == image.into_bytes()`, so they should definitely behave consistently.

> I'll take it but maybe the names could be clarified (`as_pixel_bytes`)? Not entirely sure, I don't want to jump the gun on making the name more verbose than necessary.

IMO, buffers being allowed to be longer is an implementation detail that users shouldn't have to worry about (too much). If they use the buffer directly, they obviously have to know and handle this, but I think that all image/pixel-related APIs should just deal with it transparently.

197g:
>  I think that all image/pixel-related APIs should just deal with it transparently.

Totally agree on this, this is mainly me not being sure if `as_bytes` qualifies for this or if it rather just acknowledges the other fact that all channel coefficients are PODs. That said, since the method does not appear on `ImageBuffer` directly but only on `DynamicImage` no one is going to find it intuitive to use that way.. Alright, the type may be enough of a hint and we do not need to `pixels` qualifier for semantics.

src/imageops/colorops.rs
RunDevelopment: Yes, this is mostly an artifact from how I discovered direct uses of the underlying buffer. I removed the `Deref` impl for `ImageBuffer` and let the compiler guide me.

Also, with `.pixels()` and `.subpixels()`, the `Deref` impl really isn't all that needed anymore. After this PR, there are only around 5 more places that use it, all of which can be replaced with `.subpixels()`. This might be better for users also, since things like `image.to_vec()` and `image.into_vec()` can be confusing.

Still gathering my thoughts on that though.

197g: > I removed the Deref impl for ImageBuffer and let the compiler guide me.

Ooooh, nice idea. I had already wondered how. 

And `Deref` is definitely rather odd in many regards as it is not a 'smart pointer'. So if you have a good argument for removing it after this (there are alternatives afterall) then sure, I'm ready to hear that discussed in a PR or issue.

197g (approved):

PR #2946, opened 2026-05-08
Fix links in documentations

RunDevelopment:
All the renaming in #2672 caused some doc links to become stale. This PR fixes that.

I also removed one sentence from `finish` that *I think* no longer applies.

---

I thought we already had something like `RUSTDOCFLAGS="-D warnings" cargo doc` in CI, but apparently not. Should I add that?


197g:
> I thought we already had something like RUSTDOCFLAGS="-D warnings" cargo doc in CI, but apparently not. Should I add that?

I was under the same impression. Sure, wouldn't hurt.

197g (approved):

PR #2947, opened 2026-05-08
Make `WithBlurAcc` super of `PrimitiveSealed`

RunDevelopment:
After it was merged, I noticed that #2846 had a small issue: it used `PrimitiveSealed` incorrectly. 

`PrimitiveSealed` is supposed to be *the* trait all internal traits attach themselves to. So `WithBlurAcc` (being internal) should be super of `PrimitiveSealed`, not `Primitive`. Boring bookkeeping that doesn't affect functionality, but I like to keep things tidy.


197g (approved):
They probably had referred to discussion where we hadn't nailed it down yet (both leave us in the same technical state with regards to compatibility) but sure, let's make all bounds we have added so far consistent with each other in this regard.

PR #2948, opened 2026-05-08
CI: Error on doc warnings

RunDevelopment:
I changed the existing toolchains test to error when `cargo doc` warnings occur.

CI for this PR will fail until #2946 is merged and this PR is updated.


RunDevelopment:
Hmmm, it works, but maybe this should be its own job?

Issue #1646, opened 2022-01-04
Support monochrome BMP ?


lufengd3:
I would like to be able create monochrome bmp,  1 bit per pixel.

My specific use case for this functionality is e-ink screen displays. The basic device only support monochrome bitmap.

fintelia:
Does your use case involve encoding, decoding, or both?

lufengd3:
> Does your use case involve encoding, decoding, or both?

Currently only use encoding.

Sheepyhead:
I'll second the need for 1bpp bmp, in this case to write the bmp to an eink screen. I'd like to be able to create the bmp in code, then write it to a file, and there doesn't seem to be a crate that supports this for 1bpp bmps

fintelia:
At this point, the groundwork is in place if anyone wants to work on a PR: The `write_image` method can be called with `ColorType::L1` (currently it returns an error) and [this comment](https://github.com/image-rs/image/issues/2425#issuecomment-2692454046) describes the expected data layout.

ajacquet-evolis:
Supporting the 1bpp BMP encoding/decoding seems to "only" require to edit `BmpDecoder` and `BmpEncoder` in order to accept the `ExtendedColorType::L1`.

But you would still need to provide the image data (`&[u8]`) correctly based on your previous link, which means it's up to the user to make the conversion from `L8` to 1bpp for example, right ?
Would it make sense to integrate 1bpp image in the `ColorType` and `DynamicImage` enum to ease the conversion and export of such image type ? This would for example allow to use the `DynamicImage::save` function.

fintelia:
> Would it make sense to integrate 1bpp image in the `ColorType and `DynamicImage` enum to ease the conversion and export of such image type ? This would for example allow to use the `DynamicImage::save` function.

No. The problem is that a bunch of places (both in this library and in user code) assume that pixels take an integer number of bytes to store. The most obvious is [`ColorType::bytes_per_pixel`](https://docs.rs/image/latest/image/enum.ColorType.html#method.bytes_per_pixel) but there's a bunch of others that show up because all Rust structs have to be an integer number of bytes

ajacquet-evolis:
Ok, thanks for the insight.

What about introducing a new utilitary function somewhere (on `DynamicImage` maybe ?) to export the image as a 1bpp `Vec<u8>` that would be a valid input for `BmpEncode::WriteImage` with `ExtendedColorType::L1` ?
[...]

Issue #2425, opened 2025-02-23
Panic documentation for `ImageEncoder::write_image` is outdated


RunDevelopment:
The condition for when `ImageEncoder::write_image` is expected to panic is currently documented as follows:

```rs
    /// # Panics
    ///
    /// Panics if `width * height * color_type.bytes_per_pixel() != buf.len()`.
    fn write_image(
        self,
        buf: &[u8],
        width: u32,
        height: u32,
        color_type: ExtendedColorType,
    ) -> ImageResult<()>;
```

This doesn't make sense, because `ExtendedColorType` doesn't have a `bytes_per_pixel()` method, only [`bits_per_pixel`](https://docs.rs/image/latest/image/enum.ExtendedColorType.html#method.bits_per_pixel). This is only natural as `ExtendedColorType` supports formats like `L1`.

Currently, it is unclear how `ImageEncoder::write_image` behaves when `color_type.bits_per_pixel()` is not divisible by 8.

(Also, this doc comment has been copied a lot. Search for "Panics if `width * height" and you'll find 7 copies in 6 files. There might be more, I didn't search for long.)

As for how this came to be: 9a8591ac35e89fb6043f31574e3955c3fb798e5f added the documentation and #2142 changed the signature of `write_image` without updating it.

---

Somewhat related, it would also be nice if the data layout of `ExtendedColorType::L1` was explained. I assume that it stores 8 pixels in one byte, but it's not clear to me (1) how it handles images with `width % 8 != 0` and (2) which bit corresponds to which pixel (i.e. is the pixel at x=0,y=0 the LSB or MSB in the first byte of `buf`?).  Same for a lot of the other variants.

fintelia:
Thanks for raising this issue! The real requirement on `write_image` is that the provided byte slice length is precisely what's expected given the color type. 

We probably need to write out a description somewhere explaining the expected memory layout for pixel data. The general overview is:

* 8-bit pixel data is arranged sequentially in row-major format without any padding.
* 16-bit and 32-bit data is expected to be in native endian
* sub 8-bit pixel data is packed with multiple samples in a single byte. Rows of pixels are padded to byte boundaries. IIRC both PNG and TIFF follow the same convention on the ordering of bits within a byte and I think we do the same. But I'd have to double check.
* Places that accept byte slices do not require any alignment.

PR #2953, opened 2026-05-13
adding 1bit bmp encode support

libesz:
Implements: https://github.com/image-rs/image/issues/1646

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,6 +109,10 @@ avif-native = ["dep:mp4parse", "dep:dav1d"] # Enable native dependency libdav1d
 benchmarks = [] # Build some inline benchmarks. Useful only during development (requires nightly Rust)
 serde = ["dep:serde"]
 
+[[example]]
+name = "bmp_1bit_encoding"
+required-features = ["bmp"]
+
 [[bench]]
 path = "benches/decode.rs"
 name = "decode"
--- a/examples/bmp_1bit_encoding.rs
+++ b/examples/bmp_1bit_encoding.rs
@@ -1,0 +1,66 @@
+// Example: How to encode 1-bit BMP files using the image-rs library
+//
+// This demonstrates 1-bit palette BMP encoding, useful for:
+// - Monochrome displays (e-ink, embedded systems)
+// - Binary image data (QR codes, barcodes)
+// - Minimal file size (8x smaller than 8-bit)
+
+use image::codecs::bmp::BmpEncoder;
+use image::ExtendedColorType;
+
+fn main() {
+    // Create an 8x8 checkerboard pattern
+    // Each pixel is 0 (black) or 1 (white)
+    let width = 8;
+    let height = 8;
+    let mut image_data = Vec::new();
+
+    for y in 0..height {
+        for x in 0..width {
+            let pixel = if (x + y) % 2 == 0 { 0 } else { 1 };
+            image_data.push(pixel);
+        }
+    }
+
+    // Example 1: Encode with default black/white palette
+    let mut output = Vec::new();
+    BmpEncoder::new(&mut output)
+        .encode(&image_data, width, height, ExtendedColorType::L1)
+        .expect("Failed to encode 1-bit BMP");
+
+    std::fs::write("checkerboard_1bit.bmp", &output).expect("Failed to write file");
+    println!("Saved checkerboard_1bit.bmp ({} bytes)", output.len());
+
+    // Example 2: Encode with custom red/blue palette
+    let custom_palette = vec![
+        [255, 0, 0], // Red for 0
+        [0, 0, 255], // Blue for 1
+    ];
+
+    let mut output_custom = Vec::new();
+    BmpEncoder::new(&mut output_custom)
+        .encode_with_palette(
+            &image_data,
+            width,
+            height,
+            ExtendedColorType::L1,
+            Some(&custom_palette),
+        )
+        .expect("Failed to encode with custom palette");
+
+    std::fs::write("checkerboard_custom.bmp", &output_custom).expect("Failed to write file");
+    println!(
+        "Saved checkerboard_custom.bmp with red/blue palette ({} bytes)",
+        output_custom.len()
+    );
+
+    // Example 3: Convert existing image to 1-bit
+    println!("\nTo convert an existing image to 1-bit BMP:");
+    println!("  let img = image::open(\"input.png\")?;");
+    println!("  let gray = img.into_luma8();");
+    println!("  let l1_data: Vec<u8> = gray.as_raw().iter()");
+    println!("      .map(|&v| if v > 127 {{ 1 }} else {{ 0 }})");
+    println!("      .collect();");
+    println!("  BmpEncoder::new(&mut output)");
+    println!("      .encode(&l1_data, gray.width(), gray.height(), ExtendedColorType::L1)?;");
+}
--- a/examples/bmp_1bit_encoding.rs
+++ b/examples/bmp_1bit_encoding.rs
@@ -9,17 +9,21 @@ use image::codecs::bmp::BmpEncoder;
 use image::ExtendedColorType;
 
 fn main() {
-    // Create an 8x8 checkerboard pattern
-    // Each pixel is 0 (black) or 1 (white)
+    // Create an 8x8 checkerboard pattern in packed 1-bit format
+    // L1 format packs 8 pixels per byte, MSB first:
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -109,6 +109,10 @@ avif-native = ["dep:mp4parse", "dep:dav1d"] # Enable native dependency libdav1d
 benchmarks = [] # Build some inline benchmarks. Useful only during development (requires nightly Rust)
 serde = ["dep:serde"]
 
+[[example]]
+name = "bmp_1bit_encoding"
+required-features = ["bmp"]
+
```
RunDevelopment, line 115: No other example has this, so could you please remove this for consistency?

examples/bmp_1bit_encoding.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    // Example 3: Convert existing image to 1-bit
+    println!("\nTo convert an existing image to 1-bit BMP:");
+    println!("  let img = image::open(\"input.png\")?;");
+    println!("  let gray = img.into_luma8();");
+    println!("  let l1_data: Vec<u8> = gray.as_raw().iter()");
+    println!("      .map(|&v| if v > 127 {{ 1 }} else {{ 0 }})");
+    println!("      .collect();");
+    println!("  BmpEncoder::new(&mut output)");
+    println!("      .encode(&l1_data, gray.width(), gray.height(), ExtendedColorType::L1)?;");
```
RunDevelopment: What's the purpose of printing this code?

RunDevelopment (changes requested):
Hi @libesz!

Before I go to the review, I'd like to ask a question: was this code, entirely or substantial parts of it, generated by an AI agent? This isn't an accusation and won't prevent your PR from being merged (I think. I'm not aware of an AI policy, but it's probably something along the lines of "human in the loop"). I'd just like to know, because this code reads like AI to me.

As for the review: You misunderstood the `ExtendedColorType::L1` color. Our pixels are packed, so we stuff 8 pixels into a single byte for L1. For a more detailed description of the bit layout of L1, see the comments in #1646 and #2425. I don't fault you for missing this. It's undocumented after all. Unfortunately, this does mean that the feature is implemented incorrectly.

libesz:
Hey! 
Yea, sorry for just opening this without any context or description, including statements about AI usage. Yes this change is AI assisted. As it does not include any design decisions or broader changes to the crate, I thought I would just attempt to close this gap quickly for everyone's benefit.

Will rework the L1 part, thanks for the review! Let me know if there's anything else thats too AIy, happy to redo it.

examples/bmp_1bit_encoding.rs
libesz: Essentially just presenting my use-case and maybe others'. I need to convert PNGs to this exact format.
Removing.

Cargo.toml
libesz: I couldn't come up with a way to force 1bit palette usage without using the encoder directly, and so the BMP feature. Would there be a way? Or rather should I remove the whole example?

RunDevelopment: `bpm` is in the set of default features, so I think you don't need to explicitly require it. I think you can just use the decoder directly.

libesz: I removed the example from cargo toml to show how CI checks are failing this way.

RunDevelopment: Sorry for the delay!

My bad. I assumed that it'd be fine since all other examples work, but apparently not. So we either have to use `required-features` or `#[cfg(...)]`-gate everything that uses BMP. Probably best to `required-features`.

src/codecs/bmp/encoder.rs
```diff
@@ -458,4 +540,178 @@ mod tests {
...
     }
+
+    #[test]
+    fn round_trip_1bit() {
+        // 8x2 image with alternating pattern
+        // Row 1: [0,1,0,1,0,1,0,1] = 0b01010101 = 0x55
+        // Row 2: [1,0,1,0,1,0,1,0] = 0b10101010 = 0xAA
+        // Packed format: MSB = pixel 0, LSB = pixel 7
+        let image = vec![0x55, 0xAA];
+        let decoded = round_trip_image(&image, 8, 2, ExtendedColorType::L1);
```
RunDevelopment, line 523: Could you please also test an image that is 13x3?

Right now, there are no images that test both `width % 8 != 0` and `height > 1`.

src/codecs/bmp/encoder.rs
```diff
@@ -290,6 +324,48 @@ impl<W: Write> BmpEncoder<W> {
...
+        let bytes_per_row = width.div_ceil(8) as usize;
+        for row in (0..height).rev() {
+            // from the bottom up
+            let row_start = (row as usize) * bytes_per_row;
+            let row_end = row_start + bytes_per_row;
+
+            // Write the packed bytes directly
+            self.writer.write_all(&image[row_start..row_end])?;
+            self.write_row_pad(row_padding)?;
+        }
```
RunDevelopment: Since everything is in the right format already, we can simplify.

```suggestion
        for row in image.chunks_exact(bytes_per_row).rev() {
            // from the bottom up
            self.writer.write_all(row)?;
            self.write_row_pad(row_padding)?;
        }
```

src/codecs/bmp/encoder.rs
```diff
@@ -320,14 +396,20 @@ impl<W: Write> ImageEncoder for BmpEncoder<W> {
     }
 }
 
-/// Returns a tuple representing: (dib header size, written pixel size, palette color count).
+/// Returns a tuple representing: (dib header size, written pixel size in bits, palette color count).
+/// Note: For 1-bit images, written pixel size represents bits per pixel, not bytes.
```
RunDevelopment: I don't like this. This treats L1 as a special case, which then causes the many `if color == L1` in `encode_with_palette`.

If we change this to return a `Result<(u32, u16, u32), _>` where the `u16` is **bits** per pixel, everything is uniform. Then the entire calculation for row padding and image size can be simplified to:

```rs
        let (dib_header_size, bits_per_pixel, palette_color_count) =
            written_pixel_info(color_type, palette)?;

        let natural_row_bytes: u64 = (width as u64 * bits_per_pixel as u64).div_ceil(8);
        // each row must be padded to a multiple of 4 bytes
        let padded_row_bytes: u64 = natural_row_bytes.next_multiple_of(4);
        let row_padding = (padded_row_bytes - natural_row_bytes) as u32;

        let image_size = u32::try_from(padded_row_bytes)
            .ok()
            .and_then(|v| v.checked_mul(height))
            .ok_or_else(|| {
                ImageError::Parameter(ParameterError::from_kind(
                    ParameterErrorKind::DimensionMismatch,
                ))
            })?;
```

and a 

```rs
self.writer.write_u16::<LittleEndian>(bits_per_pixel)?;
```

further down.

RunDevelopment (commented):
Again, sorry for the delay! I looked a little more closely at the code this time and found some opportunities to simplify.

src/codecs/bmp/encoder.rs
libesz: I agree.
Came up with a slightly different version. Please check.

libesz:
> Again, sorry for the delay! I looked a little more closely at the code this time and found some opportunities to simplify.

No worries at all. I addressed the comments.

src/codecs/bmp/encoder.rs
```diff
@@ -74,13 +76,23 @@ impl<W: Write> BmpEncoder<W> {
...
+        // Calculate row size in bytes: (width * bits_per_pixel) / 8, rounded up
+        let row_bytes = (width as u64)
+            .checked_mul(bits_per_pixel as u64)
+            .map(|v| v.div_ceil(8))
+            .and_then(|v| u32::try_from(v).ok())
+            .ok_or_else(|| {
+                ImageError::Parameter(ParameterError::from_kind(
+                    ParameterErrorKind::DimensionMismatch,
+                ))
+            })?;
```
RunDevelopment: No reason to used checked mul for a multiplication that can never overflow. Using normal mul also makes it obvious what is being calculated, so the comment isn't necessary.

```suggestion
        let row_bytes = u32::try_from((width as u64 * bits_per_pixel as u64).div_ceil(8))
            .ok()
            .ok_or_else(|| {
                ImageError::Parameter(ParameterError::from_kind(
                    ParameterErrorKind::DimensionMismatch,
                ))
            })?;
```

RunDevelopment (approved):
LGTM from my side. Thank you for the work @libesz!

@197g Could you please take a look too?

Issue #2954, opened 2026-05-16
Claude Code Review Report


qarmin:
Recently, I ran Claude Code against `image-rs` to look for potential issues.

I previously did something similar for `lofty`, and a surprisingly large portion of the findings turned out to be legitimate, ranging from incorrect comments to actual logic bugs.

Looking through the `image-rs` results with a non-expert eye, my rough impression is that around half of the findings are probably nitpicks or false positives, while the other half may point to real problems and likely deserve manual human review.

<details><summary>Example of findings</summary>

### C2. AVIF SMPTE 240M `kr`/`kb` coefficients are swapped

- **File:** `src/codecs/avif/yuv.rs:226-229`
- **Confirmed in code.**

```rust
YuvStandardMatrix::Smpte240 => YuvBias {
    kr: 0.087f32,
    kb: 0.212f32,
},
```

SMPTE 240M defines `Y = 0.212·R + 0.701·G + 0.087·B`. The literal for `kr` (red coefficient) is `0.087`, which is actually the blue coefficient. As a result, any AVIF tagged with `MatrixCoefficients::ST170M` or `ST240M` (mapped to this matrix in `decoder.rs:321-326`) decodes with red and blue swapped in the inverse YUV→RGB transform — i.e. a heavy magenta/teal tint.

----

### L2. AVIF YCgCo macro doc says "420" for 422 codec

- **File:** `src/codecs/avif/ycgco.rs:324-331`

```rust
define_ycgco_half_chroma!(
    ycgco422_to_rgba8, yuv422_to_rgbx_invoker,
    u8, RGBA_CN, 8,
    "Converts YCgCo 420 8-bit planar format to Rgba 8-bit"
);
```

Function and invoker are 4:2:2 but doc says 4:2:0.

----

### M15. Format-detection hooks accept empty signatures (matches every input)

- **File:** `src/hooks.rs:152-162` and `:165-183`

```rust
pub fn register_format_detection_hook(mut extension: OsString, signature: &'static [u8],
                                       mask: Option<&'static [u8]>) {
    extension.make_ascii_lowercase();
    GUESS_FORMAT_HOOKS.write().unwrap().push((signature, mask.unwrap_or(&[]), extension));
}
// matcher (:177-180):
if mask.is_empty() {
    if start.starts_with(signature) { return Some(extension.clone()); }
}
```

[...]

PR #2956, opened 2026-05-17
ICO: Fix silent overflow and truncation for large frames

RunDevelopment:
Fixes M13 from #2954.

I found that not does `image.encoded_image.len() as u32` truncate, `offset += image.encoded_image.len() as u32` can also overflow. This PR fixes both issues and adds a test for the overflow, since the truncation is hard to test without allocating 4 gigs of memory.

### DIFF
--- a/src/codecs/ico/encoder.rs
+++ b/src/codecs/ico/encoder.rs
@@ -3,8 +3,8 @@ use std::borrow::Cow;
 use std::io::{self, Write};
 
 use crate::codecs::png::PngEncoder;
-use crate::error::{ImageError, ImageResult, ParameterError, ParameterErrorKind};
-use crate::{ExtendedColorType, ImageEncoder};
+use crate::error::{EncodingError, ImageError, ImageResult, ParameterError, ParameterErrorKind};
+use crate::{ExtendedColorType, ImageEncoder, ImageFormat};
 
 // Enum value indicating an ICO image (as opposed to a CUR image):
 const ICO_IMAGE_TYPE: u16 = 1;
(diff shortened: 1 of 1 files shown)


src/codecs/ico/encoder.rs
```diff
@@ -101,21 +101,33 @@ impl<W: Write> IcoEncoder<W> {
...
+                data_size,
             )?;
 
-            offset += image.encoded_image.len() as u32;
+            offset = offset.checked_add(data_size).ok_or_else(|| {
+                ImageError::Encoding(EncodingError::new(
+                    ImageFormat::Ico.into(),
+                    "the total size of the ICO file must be at most 4 GiB",
+                ))
+            })?;
```
197g, line 138: This is stricter than necessary for the last file but unlikely to be an issue in practice. I'm not an expert on the format though it should be consistent with our decoder at least. I'm wondering how other decoders handle that in practice.

RunDevelopment: True. Our decoder supports files >4 GiB and [one of `SixLabors.ImageSharp`](https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Formats/Icon/IconDecoderCore.cs) does too. Then let's support it.

197g (approved):

RunDevelopment:
@197g Anything that still needs to be done here?

197g:
Pushing the merge button ^^ By the way if there's an approval on it you can also do so yourself (as part of trust of the members team of image-rs).

RunDevelopment:
Thank you for giving me the go ahead, but I'm still a little cautious with merging. One day I'll hit that button when I feel confident that none of the other members would object :)

PR #2959, opened 2026-05-18
PNM: Fix potential underflow and panic

RunDevelopment:
This PR addresses C6 from https://github.com/image-rs/image/issues/2954.

I fixed the panic using an early return. This also fixed the underflow, but I used `div_ceil` for clearer code.


197g (approved):
Good, also thought `div_ceil` here when I saw the pattern.

PR #2962, opened 2026-05-18
Panic when tiling an empty image

RunDevelopment:
This PR addresses C12 from #2954.

Instead of doing nothing when given an empty image, I decided that panicking is the better option. Put simply, tiling an empty image does not make any sense.


197g (approved):
Yes, same argument as `step_by` applies so we simply lift that error into a comprehensible message. 👍

PR #2965, opened 2026-05-18
Fast blur: Hoist weight computation

RunDevelopment:
Addresses L25 from #2954.

I benchmarked `fast_blur` for both u8 and f32 images. This change does not impact performance at all on my system. LLVM probably already moved the computation outside the loop.

So, yeah. Merge if you like it, close if you don't. It really doesn't matter.


197g (approved):
This will maybe give you just a little more pause if one were to modify it _not_ to be a loop constant. I consider it a little code cleanliness improvement hence.

PR #2968, opened 2026-05-19
Cleanup ICO alpha mask application

197g:
Addresses C7 of #2954 , although on hindsight the criticality maybe should have been even lower. It's a bad code smell that we were doing very manual indexing here (and we can probably do even less) but overall the bounds verification was already in the contract of the layout vs. the length of `buf` and verified further up in the function. The new implementation makes this much clearer though.

### DIFF
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -371,9 +371,15 @@ impl<R: BufRead + Seek> ImageDecoder for IcoDecoder<R> {
                 // the mask is not required. Unfortunately, Wikipedia does not have a citation
                 // for that claim, so we can't be sure which is correct.
                 if data_end >= image_end + mask_length {
+                    let rgba = buf.as_chunks_mut::<4>().0;
+                    let rows = rgba.chunks_exact_mut(width as usize);
+                    debug_assert_eq!(rows.len(), height as usize, "entry matches dimension");
+
                     // If there's an AND mask following the image, read and apply it.
-                    for y in 0..height {
+                    // This from the bottom up (in terms of our coordinates).
+                    for row in rows.rev() {
                         let mut x = 0;
+
                         for _ in 0..mask_row_bytes {
                             // Apply the bits of each byte until we reach the end of the row.
                             let mask_byte = r.read_u8()?;
@@ -381,10 +387,12 @@ impl<R: BufRead + Seek> ImageDecoder for IcoDecoder<R> {
                                 if x >= width {
                                     break;
                                 }
+
                                 if mask_byte & (1 << bit) != 0 {
                                     // Set alpha channel to transparent.
-                                    buf[((height - y - 1) * width + x) as usize * 4 + 3] = 0;
+                                    row[x as usize][3] = 0;
                                 }
+
                                 x += 1;
                             }
                         }
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -371,9 +371,16 @@ impl<R: BufRead + Seek> ImageDecoder for IcoDecoder<R> {
                 // the mask is not required. Unfortunately, Wikipedia does not have a citation
                 // for that claim, so we can't be sure which is correct.
                 if data_end >= image_end + mask_length {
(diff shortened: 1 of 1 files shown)


src/codecs/ico/decoder.rs
```diff
@@ -371,20 +371,28 @@ impl<R: BufRead + Seek> ImageDecoder for IcoDecoder<R> {
                 // the mask is not required. Unfortunately, Wikipedia does not have a citation
                 // for that claim, so we can't be sure which is correct.
                 if data_end >= image_end + mask_length {
+                    let rgba = buf.as_chunks_mut::<4>().0;
+                    let rows = rgba.chunks_exact_mut(width as usize);
```
RunDevelopment, line 379: This will panic for empty images.

Maybe we should make a wrapper for `chunks_exact{,_mut}` that only panics if chunk_size == 0 and slice.len() > 0. The case slice.len() == 0 and chunk_size == 0 is pretty common for us, so that may be worth it.

src/codecs/ico/decoder.rs
```diff
@@ -371,20 +371,28 @@ impl<R: BufRead + Seek> ImageDecoder for IcoDecoder<R> {
...
+                    debug_assert_eq!(rows.len(), height as usize, "entry matches dimension");
```
RunDevelopment: I don't think we ever verified that `buf.len() == w*h*c`. So this may be false for valid files.

197g: Above we pass `buf` to `read_image`. I could also turn that into an error return of course.

RunDevelopment: We don't though. It's passed to `read_image_data`, which does a bunch of things but it doesn't enforce the length of `buf`.

I just realized that this isn't related to what you're trying to fix, but ICO's `read_image` is a little wrong for BMP entries. Any `read_image` should panic for `buf.len() != w*h*c`, but ICO's doesn't. The PNG path is covered by PNG's `read_image`, but the BMP path isn't.

I think we should put a big `assert_eq!(buf.len(), ...);` at the start of this function like we do for other decoders.

197g: > Any read_image should panic for buf.len() != w*h*c, but ICO's doesn't. 

I'm not certain we necessarily should with the new interface. Previously, the reason behind suggesting it to implementations was that `read_image` would be regularly called directly and this was the only guidance to ensure uniform behavior. With the new interface you're nudged towards using `ImageReader::read_image` as the main  entry point which can verify the  buffer layout before calling any of the decoder's methods for it.

RunDevelopment: Suggesting? It's part of the contract of `ImageDecoder`, no? A `buf` with an incorrect length is an invalid input, and the interface says that the implementation should panic.

I'm not a fan of assuming that inputs are always valid, so I think an explicit check is the way to go.

197g: Oh yeah, that was said from the standpoint of verification. We do not `unsafe`-ly depend on it (and it is hard to validate comprehensively), I'm trying to avoid the C-like wording that practical seems to imply higher confidence in the check. We could maybe drop that contract, as long do you do something sound we'll now run `ImageReader` code before you and can hence take over the contract part (with _actual_ guarantees, so an error is now more useful).

RunDevelopment: AFAIK, `unsafe`ly depending on a safe trait is unsound, since there's nothing guaranteeing that the contract of the trait has been implemented correctly. (Unless you own all implementations, but that's not the case for `ImageDecoder`.) I'm also not sure what "C-like wording" is, since I never worked in C. To me, "if A then panic" reads like a normal API contract.

I don't really care what we do when the buffer length is wrong, as long as we do something for the sake of local reasoning and erroring early. I'm only saying "panic" because that's what the trait says should happen. If you want to change it to an error, then I think we should change it everywhere. (And make a dedicated error variant for it. You're using an error that means something else in this PR.)

Also, since the current check is always the same for all decoders, maybe we could make that a util method on `ImageLayout`/`DecoderPreparedImage`?

PR #2969, opened 2026-05-19
Avoid overflow panic dithering tiny images

197g:
The correct behavior is a no-op, or for images with one direction 1 to dither only along that one direction. Notably since we special case the first and last pixel, which would be the same for dimN=1, that needs careful attention.

Fixes C8 in #2954


RunDevelopment (approved):
Good stuff!

PR #2971, opened 2026-05-19
Make JpegDecoder not read the whole stream to memory

197g:
With the decoder trait changed to take a mutable reference as well as `zune-jpeg` providing an IO-based cursor we can now implement this trait in terms of iteration over a file; instead of buffering the whole stream into memory once at the start.

### DIFF
--- a/src/codecs/jpeg/decoder.rs
+++ b/src/codecs/jpeg/decoder.rs
@@ -14,15 +14,19 @@ type ZuneColorSpace = zune_core::colorspace::ColorSpace;
 
 /// JPEG decoder
 pub struct JpegDecoder<R> {
-    input: zune_jpeg::JpegDecoder<R>,
-    orig_color_space: ZuneColorSpace,
+    decoder: zune_jpeg::JpegDecoder<R>,
     spec_compliance: SpecCompliance,
-    width: u16,
-    height: u16,
     limits: Limits,
+    header: Option<HeaderData>,
     // For API compatibility with the previous jpeg_decoder wrapper.
     // Can be removed later, which would be an API break.
     phantom: PhantomData<R>,
+}
+
+struct HeaderData {
+    orig_color_space: ZuneColorSpace,
+    width: u16,
+    height: u16,
 }
 
 impl<R: BufRead + Seek> JpegDecoder<R> {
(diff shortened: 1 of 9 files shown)


src/codecs/jpeg/decoder.rs
```diff
@@ -16,73 +14,105 @@ type ZuneColorSpace = zune_core::colorspace::ColorSpace;
...
-    orig_color_space: ZuneColorSpace,
+    decoder: zune_jpeg::JpegDecoder<R>,
     spec_compliance: SpecCompliance,
-    width: u16,
-    height: u16,
     limits: Limits,
+    header: Option<HeaderData>,
     // For API compatibility with the previous jpeg_decoder wrapper.
     // Can be removed later, which would be an API break.
     phantom: PhantomData<R>,
```
RunDevelopment: This `PhantomData<R>` probably isn't necessary anymore

src/codecs/jpeg/decoder.rs
```diff
@@ -16,73 +14,105 @@ type ZuneColorSpace = zune_core::colorspace::ColorSpace;
 
 /// JPEG decoder
 pub struct JpegDecoder<R> {
-    input: Vec<u8>,
-    orig_color_space: ZuneColorSpace,
+    decoder: zune_jpeg::JpegDecoder<R>,
     spec_compliance: SpecCompliance,
```
RunDevelopment: This field isn't necessary anymore. It's only ever written to but never read.

src/codecs/jpeg/decoder.rs
```diff
@@ -16,73 +14,105 @@ type ZuneColorSpace = zune_core::colorspace::ColorSpace;
...
+
+        this.spec_compliance = spec;
+        this.decoder.set_options({
+            this.decoder
+                .options()
+                .set_strict_mode(matches!(spec, SpecCompliance::Strict))
+        });
+
+        this
+    }
```
RunDevelopment: This is nitpicky, but wouldn't it be easier to implement `new` and `new_with_spec_compliance` the other way around? I.e. `new` to be implemented as `Self::new_with_spec_compliance(r, SpecCompliance::default())`.

Also, I remember you disliking the naming schema `new_with_*`. Should this be just `with_spec_compliance`?

src/codecs/jpeg/decoder.rs
```diff
@@ -16,73 +14,105 @@ type ZuneColorSpace = zune_core::colorspace::ColorSpace;
 
 /// JPEG decoder
 pub struct JpegDecoder<R> {
-    input: Vec<u8>,
-    orig_color_space: ZuneColorSpace,
+    decoder: zune_jpeg::JpegDecoder<R>,
     spec_compliance: SpecCompliance,
-    width: u16,
-    height: u16,
     limits: Limits,
```
RunDevelopment, line 17: Same here. It's only ever written to but never read.

RunDevelopment (commented):
Love how much simpler this is now. I just have a few nits, nothing major.

src/codecs/jpeg/decoder.rs
197g: I think I'll keep the limit around, with a note. There may be some unused part of it and we do run `Option<Vec>::cloned` in some of the metadata retrieval methods.

RunDevelopment (approved):
Nice!

PR #2979, opened 2026-05-22
Add some debug asserts to help with local reasoning

RunDevelopment:
Addresses L46 and L47 from #2954.

### DIFF
--- a/src/codecs/avif/yuv.rs
+++ b/src/codecs/avif/yuv.rs
@@ -172,6 +172,8 @@ pub(crate) struct YuvChromaRange {
 
 impl YuvIntensityRange {
     pub(crate) const fn get_yuv_range(self, depth: u32) -> YuvChromaRange {
+        debug_assert!((8..=16).contains(&depth));
+
         match self {
             YuvIntensityRange::Tv => YuvChromaRange {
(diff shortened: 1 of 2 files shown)


src/codecs/avif/yuv.rs
```diff
@@ -172,6 +172,8 @@ pub(crate) struct YuvChromaRange {
 
 impl YuvIntensityRange {
     pub(crate) const fn get_yuv_range(self, depth: u32) -> YuvChromaRange {
+        debug_assert!(matches!(depth, 8..=16));
```
197g, line 175: We seem to be checking the bit depth in all individual callers:

```
    assert!(
        (8..=16).contains(&BIT_DEPTH),
        "Invalid bit depth is provided"
    );
```

With this being an internal but supposedly local reasoning maybe we should use a simple wrapper type for it instead of `u32` in this case?

RunDevelopment: Might be tricky. Before this function, bit depth is passed around with const generics everywhere. So switching to a parameter might make perf worse. I can't test this though, since `avif-native` doesn't work on my machine (out of the box).

PR #2982, opened 2026-05-24
docs(encoder): correct write_image panic condition for ExtendedColorType

### LINKED ISSUE #2425: Panic documentation for `ImageEncoder::write_image` is outdated

SAY-5:
Fixes #2425.

The `# Panics` doc on `ImageEncoder::write_image` and the per-codec `encode` methods stated `Panics if width * height * color_type.bytes_per_pixel() != buf.len()`. These functions take an `ExtendedColorType`, which has no `bytes_per_pixel()` method (only `bits_per_pixel()`), and the formula is wrong for sub-byte color types like `L1`.

This corrects the documented condition to match the actual requirement: the buffer must hold exactly the bytes for the given dimensions and color type, with each row padded to a whole number of bytes (`height * ((width * bits_per_pixel + 7) / 8)`), which is what `ExtendedColorType::buffer_size` already computes.

As noted in the issue, the same stale text was copied across several files; updated all six occurrences in `io/encoder.rs`, `codecs/tiff.rs`, `codecs/jpeg/encoder.rs`, `codecs/tga/encoder.rs`, `codecs/webp/encoder.rs`, and `codecs/bmp/encoder.rs`.

`cargo doc --no-deps --lib` builds with no new warnings.

### DIFF
--- a/src/codecs/bmp/encoder.rs
+++ b/src/codecs/bmp/encoder.rs
@@ -26,7 +26,9 @@ impl<W: Write> BmpEncoder<W> {
     ///
     /// # Panics
     ///
-    /// Panics if `width * height * c.bytes_per_pixel() != image.len()`.
+    /// Panics if the buffer does not hold exactly the number of bytes required for the given
+    /// `width`, `height`, and color type, accounting for rows padded to whole bytes for
+    /// sub-byte color types: `height * ((width * color_type.bits_per_pixel() as u32 + 7) / 8)`.
     #[track_caller]
     pub fn encode(
         &mut self,
@@ -43,7 +45,9 @@ impl<W: Write> BmpEncoder<W> {
     ///
     /// # Panics
     ///
-    /// Panics if `width * height * c.bytes_per_pixel() != image.len()`.
+    /// Panics if the buffer does not hold exactly the number of bytes required for the given
+    /// `width`, `height`, and color type, accounting for rows padded to whole bytes for
+    /// sub-byte color types: `height * ((width * color_type.bits_per_pixel() as u32 + 7) / 8)`.
     #[track_caller]
     pub fn encode_with_palette(
         &mut self,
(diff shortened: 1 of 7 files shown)


197g:
We haven't made it public previously out of concern that the interpretation of it as bits-per-line times height is not sensible for all of them, and that we might silently be incorrect / have unexpected results in some cases. Consider for instance a block layout with a minimum alignment may want to round up to 4 byte boundaries instead of just the byte. (On WebGPU the row alignment would be as high as 256).

However, this implies that the documentation needs _more_ fixing. Making it public with an explicit mention of the computation's rounding mode being to the nearest byte and nothing else would be my preferred way to see this handled. We can always add different calculation methods later (and with the new metadata in the decoding pipeline a decoder may choose this appropriately).

197g (approved):

PR #2983, opened 2026-05-24
Fix expanding bit depth in AVIF

awxkee:
Rotate left replicates bits, with assumption that our data is 16 bit, by amount shifted left, so for 10 bit it shift left by 6 and it replicates 0 bits to the lower half, but we need 6.
Hence rotate left is `(value << n) | (value >> (16 - n))` but we need `(value << n) | (value >> (bit_depth - n))`.

### DIFF
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -691,9 +691,10 @@ impl<R: Read> AvifDecoder<R> {
         }
 
         // Expand current bit depth to target 16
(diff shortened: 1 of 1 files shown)


RunDevelopment (approved):
LGTM.

Using bitshifts only approximates a proper unorm conversion, but it should be good enough here. The conversion from 8 bits is exact, and the maximum absolute error for 10 and 12 bits is 1 bit (or $2^{-16}$).

PR #2999, opened 2026-05-29
Fix AVIF signature mask

RunDevelopment:
Addresses L36 from #2954.

The first 4 bytes should be ignored according to [this](https://stackoverflow.com/questions/68314227/what-is-an-avif-file-signature) and [this](https://github.com/fdintino/pillow-avif-plugin/issues/5).


Shnatsel (approved):
The references are a bit muddled - this doesn't add support for matching `ftypmif1` images as described in https://github.com/fdintino/pillow-avif-plugin/issues/5

But this does correctly mask size bytes as per the spec

PR #3001, opened 2026-05-29
Add lenient mode to ICO decoder

mpospelova:
Add lenient mode to ICO decoder

Ignore dimension mismatch in lenient mode and decode the image using the embedded image's dimensions


RunDevelopment:
I also tested your example file in a few programs, and the results are interesting:

1. Windows Explorer decodes it as a 2x2 image and correctly decodes color (I tested this by modifying the colors of the BMP).
2. Gimp decodes it as a 3x3 image. The 2x2 BMP is placed at the top left corner and everything else is transparent.
3. Chrome decodes it as a 3x3 image, but all colors are white even when I hex-edit the BMP to have different colors.
4. Firefox refuses to open it.

When I hex-edit the ICO file to report a size of 1x2 (while the BMP stays 2x2), results are similar. The only difference is that Gimp decodes it as full transparent.

Since Windows is the authority when it comes to ICO files, I think your behavior is correct. Windows seems to ignore the size reported by ICO files and use the underlying size directly.

So I am generally on board with this change.

However, I think we need to change the implementation a little.

1. Only the BMP path is truly lenient. The PNG path sets a dimension limit on the decoder, so that needs to be changed to behave the same as BMP. I think a hard-coded max size of 256x256 should be fine.
2. I think we should still reject images with a dimension >256. It's just very unexpected for ICO files to be larger than that.

What do you think?

mpospelova:
@RunDevelopment  thank you for the thorough review and for testing this with different programs! :)

I've updated my changes to make the PNG path lenient as well. Though I decided not to enforce strict dimension limits for PNGs in lenient mode since it's lenient mode, and this matches the behavior of the BMP decoder which doesn't enforce size restrictions either.

I also found that there exist some tools like https://redketchup.io/icon-converter, where it is possible to generate custom ICO images containing 512x512 PNGs. So even if those images are rare, they still might exist in the wild, and with no hard dimension restrictions the rust decoder can decode them as well

RunDevelopment:
I think you're right. Basically all Microsoft products support ICOs with image data >256x256. I tested this with the tool you linked to create some 1024x1024 icos. So I think it's okay to allow them in lenient mode.

src/codecs/ico/decoder.rs
```diff
@@ -256,17 +267,22 @@ impl DirEntry {
...
+            if spec == SpecCompliance::Strict {
+                let limits = crate::Limits {
+                    max_image_width: Some(self.real_width().into()),
+                    max_image_height: Some(self.real_height().into()),
+                    max_alloc: Some(256 * 256 * 4 * 2), // width * height * 4 bytes per pixel * safety factor of 2
+                };
+                Ok(Png(Box::new(PngDecoder::with_limits(r, limits))))
+            } else {
+                Ok(Png(Box::new(PngDecoder::new(r))))
+            }
```
RunDevelopment: I added this in #2682 to prevent OOMs but this is no longer necessary since we `ImageDecoder::set_limits` now. So let's just remove it instead of making things more complex than they have to be.

```suggestion
            Ok(Png(Box::new(PngDecoder::new(r))))
```

197g (commented):
It seems fine to me to use the interpretation that the ICO meta entry's dimensions is mainly used for _selection_ and in a lenient decoding we should probably always use the factual underlying PNG. For Bitmaps.. have a read: https://learn.microsoft.com/en-us/previous-versions/ms997538(v=msdn.10) and let me know your interpretation of it.

> Only the following members are used: biSize, biWidth, biHeight, biPlanes, biBitCount, biSizeImage. All other members must be 0. The biHeight member specifies the combined height of the XOR and AND masks.

It implies that the inner bitmap may store multiple planes of image data that are post-processed to form the actual color data for the ICO, right? More context, well I'm using the term more loosely as it is very little, [in Raymond Chen](https://devblogs.microsoft.com/oldnewthing/?p=12503)'s post.

RunDevelopment:
> It implies that the inner bitmap may store multiple planes of image data

According to the docs of [`BITMAPINFOHEADER`](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapinfoheader#:~:text=Specifies%20the%20number%20of%20planes%20for%20the%20target%20device.%20This%20value%20must%20be%20set%20to%201.), `biPlanes` must be set to 1. We also [verify this in strict mode](https://github.com/image-rs/image/blob/1063f55cef361fc55ab2e805030d07d11337dad3/src/codecs/bmp/decoder.rs#L119-L122).

RunDevelopment (approved):
LGTM; Thank you, @mpospelova!

Could you please resolve the merge conflict?

PR #3002, opened 2026-05-29
ICO: Ignore AND mask for 32bpp images with native alpha

mpospelova:
2bpp BMPs already have a native alpha channel, so applying the 1-bit AND mask is destructive, as it overwrites real alpha values.


RunDevelopment:
Could you please show that this is correct? Just because applying the AND mask is destructive doesn't mean we can just ignore it. So could you please find relevant documentation that supports your change and/or show that other widely used decoders behave equivalently?

I think an example image would really help here. Then we could just see what other programs are doing.

mpospelova:
@RunDevelopment thank you for the comment! As I've understood it from [here](https://devblogs.microsoft.com/oldnewthing/20101021-00/?p=12483), Windows looks at the alpha channel for 32bpp images during standard rendering and does not apply the AND mask

197g (approved):
Thank you, behavior checks against Chromium, Firefox, feh, eog also check out with the test image.

PR #3003, opened 2026-05-29
ICO: Zero all channels for AND-mask-transparent pixels

mpospelova:
Zero out all four channels (R, G, B, A) instead of only the alpha channel when the AND mask marks a pixel as transparent.

Previously, transparent pixels retained their original RGB values, which left meaningless data in the output buffer. Now they are fully zeroed.

Also updated reference images to reflect the new output.


RunDevelopment:
Is this correct? If I'm reading [this](https://devblogs.microsoft.com/oldnewthing/20101019-00/?p=12503#:~:text=pixel%20%3D%20(screen%20AND%20mask)%20XOR%20image) and [this](https://learn.microsoft.com/en-us/previous-versions/ms997538(v=msdn.10)#:~:text=The%20icAND%20member%20contains) correctly, then the AND mask is applied *before* the XOR mask (=color). So the AND mask shouldn't be able to remove color, right?

mpospelova:
@RunDevelopment Thank you for the review! You are right that the AND mask only determines the transparency of the pixel. As I've understood the code, here we are decoding the image to an RGBA buffer, and when the AND mask marks a pixel as transparent, the previous code already set alpha to 0 which makes the color invisible, so this change additionally zeros the RGB channels to clean up that unused data

RunDevelopment:
I think you misunderstood what I said. My comment is talking about how ICO files are *supposed to be* decoded according to the linked resources. I know how we're doing it and what your PR changes. I'm asking whether those changes are correct.

Hmmm, I tested it with GIMP and that does leave the transparent white in the two test images as is. However, GIMP did decode various image formats incorrectly in the past, so take this with a grain of salt.

mpospelova:
@RunDevelopment  I checked Chrome, and Chrome does set the pixels to 0 if I understood it correctly ([link to code](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/platform/image-decoders/bmp/bmp_image_reader.cc#L991-L993)). But Chrome does that inside the BMP decoder, which is not something that we can do in image-rs, so I selected this location. Maybe we could also move this functionality to lenient mode, what do you think? Also do you have any other concerns about this change?

RunDevelopment:
Thanks for looking into this! 

I looked at the code and Chrome seems to behave the way you describe. I also looked at Firefox and [it behaves like Chrome](https://github.com/mozilla-firefox/firefox/blob/6a252a86c5bb0fdd7a197d1c244f8a28e0827b3a/image/decoders/nsICODecoder.cpp#L639-L642). 

So it's probably fine. I'm still not 100% sure whether this is correct, but doing the same thing as Chrome and Firefox shouldn't be too wrong.

So let's go ahead with this change. Please resolve the merge conflict and then we can merge this.

> Maybe we could also move this functionality to lenient mode, what do you think?

I'm super against that. IMO, lenient spec compliance should accept more inputs, not change what accepted inputs get decoded as.

RunDevelopment (approved):

RunDevelopment:
I quickly resolved the conflict and updated the now-correct reference. This is good to go now.

CI fails because clippy got a new lint. But this is no reason to hold back this PR IMO. I'll make a PR fixing the new warnings shortly.

PR #3005, opened 2026-05-30
Only change image pixels in `{from,into}_raw_{bgr,bgra}`

RunDevelopment:
Addresses C4 and L7 from #2954.

This changes `{from,into}_raw_{bgr,bgra}` to only perform the BGR <-> RGB conversion on pixels in the image. This has 2 effects:

1. If the buffer is too large, then data outside the image is no longer modified. We don't guarantee this, but I think users would expect that we don't do unnecessary work.
2. If the buffer is too small (i.e. invalid), then the `into_raw_*` methods will now panic. This can only happen if the user breaks the invariant of the type. I think panicking is preferable to returning data that does not comply with the invariant of the `ImageBuffer`.

### DIFF
--- a/src/images/buffer.rs
+++ b/src/images/buffer.rs
@@ -1434,13 +1434,13 @@ where
     /// Construct an image by swapping `Bgr` channels into an `Rgb` order.
     pub fn from_raw_bgr(width: u32, height: u32, container: Container) -> Option<Self> {
         let mut img = Self::from_raw(width, height, container)?;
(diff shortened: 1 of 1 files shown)


197g (approved):
As noted in my review comments on C4, L7, I was not convinced this is a change we _want_ to do. The panic in particular, hm.. But it is more consistent to handle both in the same manner as other constructions where we do _not_ touch superfluous data.

PR #3008, opened 2026-05-31
TIFF: Add support for f32 grayscale images

RunDevelopment:
I added support for encoding and decoding f32 grayscale TIFF images. Encoding is supported directly (thanks to #2793), but decoding is awkward. Since `ColorType` doesn't support L32F, I convert the grayscale data to RGB. This works but is a rather ugly workaround (#2934 would be nice). 

I also added a test for roundtrip encoding decoding for important color types.


RunDevelopment:
Thanks to #2934, this PR is now 2 lines of code + tests :)

197g (approved):

Issue #2966, opened 2026-05-18
About memory allocations in decoders


RunDevelopment:
#2954 uncovered a few issues regarding memory allocation in decoders and encoders. This made me think about the subject a bit more, and I've come to think that we're doing it "wrong".

Especially for decoders and `Limits`. Almost all allocations in decoders follow a pattern of (1) calculate the amount of memory to allocate, and (2) allocate it. This creates multiple problems:

1. Enforcing memory limits is optional and off by default. It's easy to just forget to check the limit.
2. Memory amount calculations are similar and non-trivial. It's easy for these calculations to accidentally overflow, and the inputs to those calculations come from the file being decoded.
3. There may be distance between step 1 and 2.
4. Allocations are assumed to succeed, because doing something like `try_reserve` would be too annoying.

To give an example of what I mean: `HdrDecoder`. It does a bunch of [allocations that ignore limits](https://github.com/image-rs/image/blob/3c02e9ed774dc2429d35ccb4c0ca40fe591421b5/src/codecs/hdr/decoder.rs#L728), and the [one check](https://github.com/image-rs/image/blob/3c02e9ed774dc2429d35ccb4c0ca40fe591421b5/src/codecs/hdr/decoder.rs#L317) for an allocation is only triggered if the user explicitly sets a limit. I don't mean to pick on `HdrDecoder` here. I think it's a symptom of a flawed system.

### Potential solution

Introduce a memory manager/allocator responsible for enforcing the memory limit, allocating memory, creating vecs, and so on. Basically, decoders are **not** allowed to allocate directly; they must go to the manager to request memory.

This immediately solves problems 1 and 3. Problem 4 can be solved more easily, since allocations are centralized. Lastly, problem 2 can be solved by adding convenience methods to the manager.

That's exactly what I did for `dds` and it worked quite well. The manager there is called `DecodeContext` and the most important method is `::alloc<T>(&mut self, len: usize) -> Result<Box<[T]>, _>` ([link](https://github.com/image-rs/image-dds/blob/d4af5831e17e40bf9f50925bdfd1e15ee58c89f9/src/decode/decoder.rs#L27)).

---

I don't like how loosely memory limits are currently enforced, and I think this would be a good out. I'd like to know what you think about this.

197g:
[...]

PR #3009, opened 2026-05-31
PNG: Account for memory limit when decompressing text

### LINKED ISSUE #2993: PNG iTXt XMP metadata decompression expands small file into large decoded buffer
This happens in pre-commit cf89a940096d7bfdf64899d99cbb4010acb38d2c

## Expected

`PngDecoder::xmp_metadata()` should respect the caller's `Limits.max_alloc` budget when decompressing iTXt/tEXt chunks. The underlying png crate already exposes `decompress_text_with_limit`, but image-rs calls the unbounded decompress_text instead.

## Actual behaviour
`xmp_metadata()` calls `itx_chunk.decompress_text()` with no size argument, then copies the result via `text.as_bytes().to_vec()`. The caller's `Limits.max_alloc` is never consulted on this path. A caller who sets Limits { max_alloc: Some(500_000) } still receives a `Vec<u8>` larger than 500,000 bytes if the iTXt chunk decompresses to more. 

## Reproduction steps

[poc.zip](https://github.com/user-attachments/files/28327699/poc.zip)
Local repro: a 1,118-byte PNG containing a compressed `XML:com.adobe.xmp` iTXt chunk returns a 1,000,002-byte buffer.

run with:
```bash
bash ./poc/run.sh
```
expected output:
```text
XMP_LEN=1000002
```

`XMP_LEN=1000002` means image-rs returned a decoded XMP buffer of 1,000,002 bytes from the compressed iTXt metadata in the supplied PNG.

RunDevelopment:
Resolves #2993 

I kept the old limit set by the `png` as a maxium. So the limit is 2 MB or the memory allocation limit, whichever is smaller. 

This may not be ideal. There are legitimate use cases for very large text chunks. Maybe we could raise the maximum depending on the memory limit. Maybe like this: `min(mem_limit, max(2 MB, mem_limit * 5%)`. The idea is that we don't want to blow the entire memory budget on decompressing text, but using a small percentage should be okay. This would support huge text chunks with the default memory limit, while still preventing decompression bombs. Alternatively, we could also just raise the hard-coded limit to something like 10 to 16 MB and call it a day. Thoughts?

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -223,6 +223,13 @@ impl<R: BufRead + Seek> PngDecoder<R> {
             (png::ColorType::Indexed, png::BitDepth::Sixteen) => ExtendedColorType::Unknown(16),
         }
     }
+
+    /// The maximum number of bytes iTXt and zTXt are allowed to decompress to.
+    /// This guards against decompression bombs.
+    fn text_decompress_limit(&mut self) -> usize {
+        let max = png::text_metadata::DECOMPRESSION_LIMIT as u64;
+        self.limits.max_alloc.unwrap_or(max).min(max) as usize
+    }
 }
 
 fn attributes_from_info(info: &png::Info<'_>) -> DecodedImageAttributes {
@@ -304,6 +311,7 @@ impl<R: BufRead + Seek> ImageDecoder for PngDecoder<R> {
     }
 
     fn xmp_metadata(&mut self) -> ImageResult<Option<Vec<u8>>> {
+        let decompression_limit = self.text_decompress_limit();
         let reader = self.ensure_reader_and_header()?;
 
         if let Some(mut itx_chunk) = reader
@@ -313,7 +321,9 @@ impl<R: BufRead + Seek> ImageDecoder for PngDecoder<R> {
             .find(|chunk| chunk.keyword == XMP_KEY)
             .cloned()
         {
-            itx_chunk.decompress_text().map_err(ImageError::from_png)?;
+            itx_chunk
+                .decompress_text_with_limit(decompression_limit)
+                .map_err(ImageError::from_png)?;
             return itx_chunk
                 .get_text()
                 .map(|text| Some(text.into_bytes()))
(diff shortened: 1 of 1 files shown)


mstoeckl:
~As per an old issue from a different project (https://github.com/opencv/opencv/issues/22551), people _do_ sometimes create images which may embed large files in text chunks, so I do not think this hard limit will suffice in the long term.~

~This might be OK as a short term patch, in which case I think you should add a TODO comment. The cleanest solution might be to introduce some way to share the memory limit of the underlying `image-png` decoder; you've suggested a custom allocator in https://github.com/image-rs/image/issues/2966.~

(Edit: oops, I don't think anyone will embed a file into XMP or IPTC, so this will probably be OK long term.)

> min(mem_limit, max(2 MB, mem_limit * 5%)

For the memory limit to remain accurate (assuming negligible allocation overhead), you'd need to reduce the memory limit passed to the PNG decoder accordingly.

RunDevelopment:
> > min(mem_limit, max(2 MB, mem_limit * 5%)
> 
> For the memory limit to remain accurate (assuming negligible allocation overhead), you'd need to reduce the memory limit passed to the PNG decoder accordingly.

We are passing the entire memory budget to the PNG decoder, so I think we'd need to reduce it by how much memory the PNG decoder is currently using. The PNG decoder should be tracking that internally (like we do), but I don't think there's any way to get that.

197g:
The current implementation is fine using it as the maximum allocation size for most purposes; there are two PRs open sketching changes to the limit system to be able to do this but we'd need every decoder to agree on a scheme here. Since allocators are being [stabilized soon (tm)](https://github.com/rust-lang/rust/pull/156882) I'd say that mechanism best wait on this. In the meantime, there is no consistent mechanism for splitting the budget and that's still better than not adhering to a limit. In particular, lowering the limit still has the effect of reducing the maximum resident size even if not precise.

RunDevelopment:
Hmmm, true. In that case this PR will be unnecessary soon (tm). Should I close it?

Btw, is there already a crate that implements an allocator like we'd need it? I did a quick search but found nothing that fits the bill.

197g (approved):
I'll take it with allocation duplication, we'll figure out the allocator independently. One final note that applies here in particular: The resulting allocation(`Vec<u8>`) is returned to the caller with the default global allocator which will remain the interface. But then there is no resource to be tracked by the allocator, the decoder can never dealloc it (to be clear, I'm not saying this is a needs-to-be-changed problem with the interface). Arguably, it should not count towards the budget use at that point and only intermittently check that budget would have been available.

PR #3015, opened 2026-06-08
Tiff: Support decoding f16 images

### LINKED ISSUE #2795: Support decoding f16 TIFF images following the `tiff` crate
The TIFF crate added f16 decoding support in https://github.com/image-rs/image-tiff/pull/257

It would be nice to have it in `image` as well. This is needed for broader image format support in e.g. my own wondermagick and (I assume) GNOME's Glycin.

The `half` crate exposes batched conversion APIs from `f16` to `f32` that make use of SIMD intrinsics under the hood, and `half` is already a transitive dependency via `tiff` and `exr` crates.

A test file to verify this functionality can be found in https://github.com/image-rs/image-tiff/issues/239

RunDevelopment:
Resolves #2795
Based on #3014

Changes:
- Tiff decoder now decodes Gray/Rgb/Rgba f16 images by converting them to f32.
- Added `ExtendedColorType::{L16F, La16F, Rgb16F, Rgba16F}`.
- Fixed a bug in testing infra

I added support for decoding f16 TIFF images by converting them to f32. This is a lossless conversion, if a bit inefficient. Importantly, this does **not** add new dependencies. I used the methods on `half::f16` exposed by the `tiff` crate to avoid adding `half` to our direct dependencies.

This correctly decodes the f16 test images from https://github.com/image-rs/image-tiff/pull/257. (Interestingly, a lot of programs don't seem to decode f16 correctly and interpret it as u16. From the programs I tested, only GIMP and Photoshop decode them correctly.)

TODO:
- [ ] I left a TODO on `ExtendedColorType`, because I probably messed up the order of variants, which breaks serde backwards compat.
- [ ] The f16 -> f32 conversion itself is less efficient than necessary. The `half` crate has optimized methods for converting a slice of f16s to a slice of f32s, but that requires taking `half` as a direct dependency. We need to decide whether we want that.


src/codecs/tiff.rs
```diff
@@ -534,6 +540,18 @@ impl<R: BufRead + Seek> ImageDecoder for TiffDecoder<R> {
...
+                if matches!(
+                    info.original_color_type,
+                    ExtendedColorType::L16F
+                        | ExtendedColorType::Rgb16F
+                        | ExtendedColorType::Rgba16F
+                ) =>
+            {
+                for (half, out) in v.iter().zip(buf.as_chunks_mut::<4>().0.iter_mut()) {
+                    *out = half.to_f32().to_ne_bytes();
+                }
```
mhils, line 537: Maybe a bit more intuitive:

```suggestion
for (out, f) in bytemuck::cast_slice_mut::<u8, f32>(buf).iter_mut().zip(v) {
                    *out = f.to_f32();
                }
```

src/codecs/tiff.rs
```diff
@@ -135,6 +136,8 @@ where
             (tiff::ColorType::RGBA(16), Uint) => ColorType::Rgba16,
             (tiff::ColorType::CMYK(8), Uint) => ColorType::Rgb8,
             (tiff::ColorType::CMYK(16), Uint) => ColorType::Rgb16,
+            (tiff::ColorType::CMYK(16), IEEEFP) => ColorType::Rgb32F,
+            (tiff::ColorType::RGBA(16), IEEEFP) => ColorType::Rgba32F,
```
mhils: Nit: Consistently sort the IEEEFP entries in this list?

mhils (commented):
Thank you for this PR! I was about to send one with pretty much the same changes, because I somehow missed this one initially. Just leaving minor comments on where we diverged, feel free to disregard them all. 😃

src/codecs/tiff.rs
RunDevelopment: This is wrong, unfortunately. A slice of `f32` must have an alignment requirement of 4, but the slice of `u8` we have doesn't guarantee that. So `bytemuck::cast_slice_mut` will panic if the slice doesn't happen to be aligned to a multiple of 4. Users are allowed to give us slices of any alignment, so we have to be careful.

Issue #2398, opened 2024-12-29
Make the `Frame` struct `Pixel`-agnostic?


moulins:
The `Frame`s returned by `AnimationDecoder` are hardcoded to contain `Rgba8Image`s, which is generally suboptimal when working with animated PNG or WEBP images:
- if the source image has no alpha channel, this imposes an extra conversion;
- 16-bit APNGs are unsupported because of this.

Would it make sense to have `Frame` contain a `DynamicImage` instead of a `Rgba8Image`, despite the fact that this is a breaking change? Existing users of the API can easily upgrade by doing the conversion to Rgba8 manually with `.into_rgba8()` or similar.

kornelski:
It would be nice to have. 

It might be doable in a backwards-compatible way by adding a new method next to `into_frames`, like `into_frames_dynamic()`.

However, I don't think it has practical utility besides optimization for opaque RGB8. AFAIK WebP doesn't have 16-bit coding — lossless is 8-bit-only, and VP8 implementations don't even use full range in 8 bits. PNG compression is pretty bad in 16 bits.

moulins:
You're right that this would be mostly for optimization purposes; another theoretical use-case would be grayscale APNGs, but I'd be surprised if they actually exist in the wild ^^

197g:
For unsupported color layouts, #2522 should be the general direction. It's probably _fine_ to convert the color while encoding. At least truncating the depth was a sought after compatibility adjustment to be made. Color space treatment such as rgba->luma might be a different beast but it's not a likely problem for most common formats. Event though some formats can not encode a CICP indication, they are currently just _wrong_ and I imagine fidelty with conversion would be better than wrong.

PR #3020, opened 2026-06-08
Change `Frame` from `RgbaImage` to `DynamicImage`

### LINKED ISSUE #2996: Add support for different image types for `into_frames`
I would like to be able to use `into_frames` whilst choosing the image format I'm converting to.

My specific use case for this functionality is converting animations to grayscale, and therefore converting to rgba8 where the image wasn't already encoded like that is wasteful.

This is more generally applicable to any time you want to specify the format you're changing the image into whilst processing.

## Draft

One of the following approaches would make sense to me:
- Adding a type parameter to `Frame` for the image format, and `into` support from DynamicImage to the various formats.
- Changing `Frame` to be over `DynamicImage` instead of `RgbaImage`
- Alternatively, enabling users to re-implement `into_frames` using only public methods (`decode_to_dynimage` and `reader.inner` are private)

RunDevelopment:
Fixes #2996.
Might also fix (idk) #2398.

This PR changes `Frame` to use an `DynamicImage` internally. This allows `Frames` and co. to represent animations that aren't RGBA8. As a side effect, this removes a forced DynImg -> Rgba8 conversion in `ImageReader::into_frames`, but also adds one in `GifEncoder`.

Not sure if we super duper want/need this.

Also, we may want to rename `Frame::buffer` to `Frame::image` (same for related methods), but I'm not sure if it's worth it.

### DIFF
--- a/src/codecs/gif.rs
+++ b/src/codecs/gif.rs
@@ -579,7 +579,7 @@ impl<W: Write> GifEncoder<W> {
         // get the delay before converting img_frame
         let frame_delay = img_frame.delay().into_ratio().to_integer();
         // convert img_frame into RgbaImage
-        let mut rbga_frame = img_frame.into_buffer();
+        let mut rbga_frame = img_frame.into_buffer().to_rgba8();
         let (width, height) = self.gif_dimensions(rbga_frame.width(), rbga_frame.height())?;
 
         // Create the gif::Frame from the animation::Frame
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -2,7 +2,7 @@ use std::cmp::Ordering;
 use std::time::Duration;
 
 use crate::error::ImageResult;
-use crate::RgbaImage;
+use crate::DynamicImage;
 
 /// An implementation dependent iterator, reading the frames as requested
 pub struct Frames<'a> {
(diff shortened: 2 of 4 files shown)


src/codecs/gif.rs
```diff
@@ -579,7 +579,7 @@ impl<W: Write> GifEncoder<W> {
         // get the delay before converting img_frame
         let frame_delay = img_frame.delay().into_ratio().to_integer();
         // convert img_frame into RgbaImage
-        let mut rbga_frame = img_frame.into_buffer();
+        let mut rbga_frame = img_frame.into_buffer().to_rgba8();
```
197g, line 582: This is odd. While consistent with automatically converting for other encoding of `DynamicImage`, this can only happen here since it _is_ `DynamicImage` and can not be another input that we'd reject automatically converting.

This feels like a _right_ change but I do find it remarkable regardless.

197g (commented):
Yeah, probably? I can't tell if I'm missing something here but then again it is rather conceptually simple.

RunDevelopment:
The change is rather simple, plus we're not really using the `Frame` API anywhere. Animations aren't that well tested, and GIF is the only animation encoder we have. Or put another way, almost everything that uses `Frame` did change.

PR #3022, opened 2026-06-09
BMP: Remove unnecessary memory allocation

RunDevelopment:
Found an unnecessary alloc in BMP that is made once when reading the header. I replaced it with a little stack memory.

Does this matter? No. Do I still want it? Yes.


197g (approved):
^^

PR #3027, opened 2026-06-10
ICO: Allow non-Rgba8 PNGs in lenient mode

### LINKED ISSUE #1590: ICO files support more color types than implemented
Decoder for ISO files contains [a check](https://github.com/image-rs/image/blob/v0.23.14/src/codecs/ico/decoder.rs#L313-L317) that all "embedded PNG images can only be of the 32BPP RGBA format":

```rs
// Embedded PNG images can only be of the 32BPP RGBA format.
// https://blogs.msdn.microsoft.com/oldnewthing/20101022-00/?p=12473/
if decoder.color_type() != ColorType::Rgba8 {
    return Err(DecoderError::PngNotRgba.into());
}
```

If I understand that article from Microsoft from 2010, its purpose is (was?) to prevent various programs from crashing too badly when trying to parse some ICO files with some different formats.

The problem is that I'm trying to parse and re-encode [this](https://mingle-cloudidentities.eu1.inforcloudsuite.com/assets/images/favicon.ico) ICO file, but it fails because of this exact check. It seems this happens because that ICO uses Rgb8 format instead of Rgb**a**8. However, when I completely removed that check, everything was fine and ICO (both original and re-encoded) is correctly displayed on Windows and in programs.

Is that check still needed or can be completely removed? Or maybe it just needs another condition for Rgb8 format?

RunDevelopment:
Resolves #1590.

Since `SpecCompliance` can configured for the ICO decoder now, I added support for PNGs that aren't Rgba8. Such ICOs are already widely supported. My 2 test images opened fine in Chrome, Firefox, GIMP, and Windows Explorer. Interestingly, Windows Image Viewer (Win10) does not support them.

Regarding the test files, I made the images myself and created the ICOs using `image`'s encoder.

Note: The issue also [talked about an alpha mask](https://github.com/image-rs/image/issues/1590#issuecomment-954278635). I think this is supposed to be the AND mask, but that one is exclusive to BMP and therefore not relevant.


src/codecs/ico/decoder.rs
```diff
@@ -329,7 +329,9 @@ impl<R: BufRead + Seek> ImageDecoder for IcoDecoder<R> {
 
                 // Embedded PNG images can only be of the 32BPP RGBA format.
                 // https://blogs.msdn.microsoft.com/oldnewthing/20101022-00/?p=12473/
```
197g: This is a partially outdated comment afterwards, should reference where those not strictly compliant images are handled properly.

197g (commented):
Heh. It'd be funny if adoption of this through Chromium -> Edge -> Windows would be how this gets supported officially in some capacity in the end.

src/codecs/ico/decoder.rs
RunDevelopment: I updated the comment to explain the situation.

PR #3032, opened 2026-06-11
ICO: Allow >256 bpp in lenient mode

mpospelova:
For this change, I was looking at Chrome's behavior. This PR makes the header parsing more permissive in lenient mode. 

Seems like there are lots of images with faulty `bits_per_pixel` values in the ICO header, but they are still parsed successfully by other decoders. This PR moves the validation of the `bits_per_pixel` to strict mode.

I also generated the test image, and it opens successfully in Gimp and Chrome.


RunDevelopment:
Very unfortunate that we can't trust `bits_per_pixel`. Since that is the case, could you please also change the entry selection logic in `best_entry`? I think we should use `entry.bits_per_pixel.min(32)` for selection to guard against invalid entries at least a little.

mpospelova:
@RunDevelopment  Good point! I have looked at what Chrome is doing, and they are first selecting the best entry based on its area first, and then based on bpp. They do not perform any validation checks on the bpp values as far as I understood ([link to code](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/platform/image-decoders/ico/ico_image_decoder.cc#L108-L115)). Is there a reason of why the Rust decoder is looking at the bpp values first, and only then on the area?

RunDevelopment:
> Is there a reason of why the Rust decoder is looking at the bpp values first, and only then on the area?

I think it was inspired by [WinXP's selection algorithm](https://web.archive.org/web/20110630025042/http://support.microsoft.com/kb/307213), but that's just a guess.

I think Chrome's approach makes more sense, but I'm not entirely sure to be honest. Would you rather have a 1bpp 256x256 image or a 32 bpp 128x128 image? But given that we can't trust `bits_per_pixel`, it's probably better to pick an entry based on size (not that we trust that either...).

Chrome has some additional logic to [fill in missing bits per pixel values](https://github.com/chromium/chromium/blob/480cd12b6f4696cd0f347cce979cda6cd2f8b1d9/third_party/blink/renderer/platform/image-decoders/ico/ico_image_decoder.cc#L335-L347). `bits_per_pixel = 0` means unspecified, so it uses `color_count` to set `bits_per_pixel  = ceil(log2(color_count))`. We should probably do that too.

That said, a new selection algorithm should be its own PR.

RunDevelopment:
So LGTM;

Just one thing: the test image. Images in `tests/images` are supposed to have a reference image in `tests/reference`. Could you please create one? The process is unfortunately quite annoying. You need to:

1. Create a dummy file `tests/reference/ico/test suite/<file name.ext>.png`.
2. Run `cargo test`. This will fail and save the expected output as `tests/output/<file name.ext>.png`.
3. Replace the dummy file with the expected output.

mpospelova:
Thank you for the review! :) I added the reference image, and will work on a PR with the selection algorithm next :)

RunDevelopment (approved):
Thank you!

> will work on a PR with the selection algorithm next :)

And thank you!

PR #3033, opened 2026-06-11
ICO: Add header validation in strict mode

RunDevelopment:
This PR adds more validation for the ICO header in strict mode. 

Changes:
- The reserved first two bytes are now checked to be zero.
- The type is now checked to be either ICO or CUR.


src/codecs/ico/decoder.rs
```diff
@@ -175,8 +175,25 @@ impl<R: BufRead + Seek> IcoDecoder<R> {
 fn read_entries<R: Read>(r: &mut R, spec: SpecCompliance) -> ImageResult<Vec<DirEntry>> {
     let mut header = [0u8; 6];
     r.read_exact(&mut header)?;
-    // header[0..2] = reserved, header[2..4] = type, header[4..6] = count
-    let count = u16::from_le_bytes(header[4..6].try_into().unwrap());
+    let mut header = header.as_slice();
+
+    let reserved = header.read_u16::<LittleEndian>()?;
```
197g, line 180: I'd use `split_first_chunk` here to instead of an IO-facing library. Alternatively, but only since this is uniformly `u16`, you might turn `header` into `[[u8; 2]; 3]` and read it with `as_flattened`? Lots of alternatives that do not involve an IO error for a statically sized array and constant condition.

197g (commented):
Appreciate the idea, more validation to keep strict mode from changing behavior silently on reserved data 👍

src/codecs/ico/decoder.rs
RunDevelopment: TBH, I just copied what TGA is doing. Should we not do that there too?

I also considered the `u16` approach, but didn't like it because it makes it really easy to mess up endian. E.g.

```rs
let mut header = [0_u16; 3];
reader.read_exact(bytemuck::cast_slice_mut(&mut header))?;
let [reserved, id_type, entries] = header; // only correct on LE systems
```

But I agree with you that the IO interface is a little overkill. Maybe we could just have a 2 or 3 util methods like this:

```rs
// Panics when OOB
fn read_u16<Endian>(slice: &[u8], offset: usize) -> u16;
```

I saw this pattern in Chrome, so that might be an idea.

RunDevelopment:
> Appreciate the idea, more validation to keep strict mode from changing behavior silently on reserved data 👍

That, and I want to lay the groundwork for supporting CUR files eventually.

PR #3042, opened 2026-06-15
ICO: Fallback to entry size when header is invalid

RunDevelopment:
Changes;
- Error if `biHeight` is odd in strict mode.
- If `biWidth` or `biHeight / 2` are zero:
    - Error in strict mode
    - Use the size specified in ICON entry as a fallback.

Integrating the fallback into the BMP decoder was a bit of a pain. I initially wanted to do in `read_metadata_in_ico_format`, but that doesn't work, because `read_metadata` (crudely) verifies that neither width nor height are zero. So the fallback logic had to be integrated more deeply.

I added 2 test images. `bmp-biHeight=1.ico` reproduces the bug from #3041. The only difference is that I made the 1x1 image red (black isn't a good color to check whether something decoded correctly). `bmp-32bpp-biWidth=biHeight=0.ico` has both width and height set to 0. Both decode fine in lenient mode and error in strict mode.


src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,125 @@ impl<'a> IcoFrame<'a> {
...
+
+    /// Construct a new `IcoFrame` by encoding `buf` as a BMP
+    ///
+    /// The `width` and `height` must be between 1 and 256 (inclusive)
+    pub fn as_bmp(
+        buf: &[u8],
+        width: u32,
+        height: u32,
+        color_type: ExtendedColorType,
+    ) -> ImageResult<Self> {
```
197g: Can this be non-pub? I'm thinking we should be trying to reduce interfaces that take width, height, color, buf as separate parameters.

src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,125 @@ impl<'a> IcoFrame<'a> {
...
+            )));
+        }
+
+        // ICO format doesn't allow 24-bit RGB, so we have to convert it to the 32-bit format with an unused alpha channel.
+        let converted_color_type = match color_type {
+            ExtendedColorType::Rgb8 => ExtendedColorType::Rgba8,
+            other => other,
+        };
+
+        let mut image_data: Vec<u8> = Vec::new();
```
197g: I have an allocation size concern here. It's not unbounded but we do not check against a limit that could be overflowed by a factor of `3` here, either. That's only a concern for the variable-size write happening after the header. We can predict the needed capacity ahead of time though.

src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,125 @@ impl<'a> IcoFrame<'a> {
...
+                    }
+                }
+
+                // AND mask
+                // Since there is no transparency, we can set all pixels to 0 (opaque) in the AND mask.
+                for _ in 0..height {
+                    let chunks_per_row = width.div_ceil(32);
+                    for _ in 0..chunks_per_row {
+                        image_data.extend_from_slice(&[0_u8; 4]);
+                    }
```
197g:
```suggestion
                    let chunks_per_row = width.div_ceil(32);
                    image_data.extend(core::iter::repeat(0u8, 4 * chunks_per_row as usize));
```

That is, only with the verification of capacity against a limit is this correct. Otherwise it could overflow `usize` of course.

src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,125 @@ impl<'a> IcoFrame<'a> {
...
+                    if !width.is_multiple_of(8) {
+                        image_data.push(mask_byte);
+                    }
+
+                    // Each row of the AND mask must be padded to a multiple of 4 bytes.
+                    let mask_row_len = width.div_ceil(8);
+                    let mask_row_padded = mask_row_len.div_ceil(4) * 4;
+                    let padding = &[0_u8; 4][..(mask_row_padded - mask_row_len) as usize];
+                    image_data.extend_from_slice(padding);
```
197g: nit: can this be simplified in any way? I'm slightly glazing over this when reading to be honest. A lot of the math happens inside the loop but is actually a loop constant. Some ideas in other comments. The padding is easily moved out.

src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,125 @@ impl<'a> IcoFrame<'a> {
...
+
+                // AND mask
+                // For 32-bit BMP the AND mask is typically ignored, but it should be provided nonetheless.
+                // We use a threshold of 128 for the alpha channel to determine whether a pixel is transparent or opaque in the AND mask.
+                for row in pixels.chunks_exact(width as usize).rev() {
+                    let mut mask_byte: u8 = 0;
+                    for (x, a) in row.iter().map(|pixel| pixel[3]).enumerate() {
+                        let pos = x % 8;
+                        let bit_pos = 7 - pos;
+                        mask_byte |= if a < 128 { 1 } else { 0 } << bit_pos;
```
197g:
```suggestion
                        mask_byte |= u8::from(a < 128) << bit_pos;
```

src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,125 @@ impl<'a> IcoFrame<'a> {
...
+                        image_data.extend_from_slice(&[*b, *g, *r, *a]);
+                    }
+                }
+
+                // AND mask
+                // For 32-bit BMP the AND mask is typically ignored, but it should be provided nonetheless.
+                // We use a threshold of 128 for the alpha channel to determine whether a pixel is transparent or opaque in the AND mask.
+                for row in pixels.chunks_exact(width as usize).rev() {
+                    let mut mask_byte: u8 = 0;
+                    for (x, a) in row.iter().map(|pixel| pixel[3]).enumerate() {
```
197g: I'd write this as a nested loop across `row.chunks(8)` (not chunks_exact) so that the push is not in a conditional. That should also take care of the fence post push on `is_multiple_of`. Also a comment maybe.

src/codecs/bmp/decoder.rs
```diff
@@ -1096,6 +1096,7 @@ pub struct BmpDecoder<R> {
     top_down: bool,
     no_file_header: bool,
     add_alpha_channel: bool,
+    fallback_size: Option<(u16, u16)>,
```
197g, line 1099: nit: I suppose ico and bmp are tied together pretty strongly so it makes sense to see this. Just wondering if it already makes sense to have a struct of context input to the BMP that would be filled by `ico` instead of a non-descript `u16` tuple. Your judgment call on whether that extra struct is already more mental overhead or not, a comment on the field suggesting this when future extensions pop up would also do.

RunDevelopment:
Uhm, you mostly review the ICO BMP encoder I was working on. I just messed up my commits. I'll incorporate your feedback into what I'm working on.

src/codecs/bmp/decoder.rs
RunDevelopment: Honestly, I was thinking of this mostly as a general modifier to how BMPs are parsed. Similar to `no_file_header`. I don't like the strong coupling between bmp and ico, so I tried to make something that isn't useful for ico alone. Not sure if I succeeded though...

So I'd hold the extra struct for now. If we add a second option that is exclusively used by ICO, then sure.

PR #3044, opened 2026-06-15
Add ICO BMP encoder

### LINKED ISSUE #2192: An invalid ICO image is generated when it including multiple BMP images
### Discussed in https://github.com/image-rs/image/discussions/2185

<div type='discussions-op-text'>

<sup>Originally posted by **sorairolake** March 26, 2024</sup>
I wrote a code to create a ICO file containing multiple BMP or PNG images:

https://github.com/sorairolake/favico/blob/1c5cb360350097900988bb407c81368d55512648/src/generate.rs#L40-L59

The generated ICO file containing multiple PNG images seems correct, but it containing multiple BMP images seems corrupted:

```text
$ identify png/favicon.ico
png/favicon.ico[0] PNG 16x16 16x16+0+0 8-bit sRGB 269B 0.010u 0:00.000
png/favicon.ico[1] PNG 32x32 32x32+0+0 8-bit sRGB 323B 0.010u 0:00.000
png/favicon.ico[2] PNG 48x48 48x48+0+0 8-bit sRGB 1027B 0.000u 0:00.000

$ identify bmp/favicon.ico
identify: improper image header `bmp/favicon.ico' @ error/icon.c/ReadICONImage/400.
```

How can I create a ICO file containing multiple BMP images which is not corrupted?</div>

RunDevelopment:
Resolves #2192

This PR adds an ICO BMP encoder. It supports Rgba8 and Rgb8 image data. Rgb8 is encoded as 32-bit 0RGB as 24-bit [24-bit is not allowed](https://devblogs.microsoft.com/oldnewthing/20101019-00/?p=12503#:~:text=Other%20restrictions%3A%20Supported%20color%20formats%20are%204bpp%2C%208bpp%2C%2016bpp%2C%20and%200RGB%2032bpp.%20Note%20that%2024bpp%20is%20not%20supported%3B%20you%E2%80%99ll%20have%20to%20convert%20it%20to%20a%200RGB%2032bpp%20bitmap.).

The test currently fails because the decoder doesn't support 0RGB yet (#3012).

---

Since @197g reviewed most of this PR in an accidental commit in #3042, I'll recreate her comments here. I won't recreate comments I already resolved.

### DIFF
--- a/src/codecs/ico/encoder.rs
+++ b/src/codecs/ico/encoder.rs
@@ -108,6 +108,18 @@ impl<'a> IcoFrame<'a> {
             )));
         }
 
+        // Check to see that the output buffer is not going to be larger than the 4 GiB that ICO can support.
+        let output_size = BITMAPINFOHEADER_SIZE as u64 // header
+            + (width as u64 * height as u64 * 4) // XOR mask (32 bits per pixel)
+            + (height as u64 * width.div_ceil(32) as u64 * 4); // AND mask (1 bit per pixel, padded to 32 bits per row)
+        if output_size > u32::MAX as u64 {
+            return Err(ImageError::Parameter(ParameterError::from_kind(
+                ParameterErrorKind::Generic(format!(
+                    "the encoded image data must be at most 4 GiB, but the provided image would encode to {output_size} bytes",
+                )),
+            )));
+        }
+
         let mut image_data: Vec<u8> = Vec::new();
 
         // https://learn.microsoft.com/en-us/previous-versions/ms997538(v=msdn.10)?redirectedfrom=MSDN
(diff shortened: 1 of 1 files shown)


src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,135 @@ impl<'a> IcoFrame<'a> {
...
+
+    /// Construct a new `IcoFrame` by encoding `buf` as a BMP
+    ///
+    /// The `width` and `height` must be between 1 and 256 (inclusive)
+    pub fn as_bmp(
+        buf: &[u8],
+        width: u32,
+        height: u32,
+        color_type: ExtendedColorType,
+    ) -> ImageResult<Self> {
```
RunDevelopment, line 97: > Can this be non-pub? I'm thinking we should be trying to reduce interfaces that take width, height, color, buf as separate parameters.

I mirrored `as_png`. I'm not against a nicer API, but I think it should be consistent for both.

src/codecs/ico/encoder.rs
```diff
@@ -80,6 +84,135 @@ impl<'a> IcoFrame<'a> {
...
+                            image_data.push(mask_byte);
+                            mask_byte = 0;
+                        }
+                    }
+                    if !width.is_multiple_of(8) {
+                        image_data.push(mask_byte);
+                    }
+
+                    image_data.extend(std::iter::repeat_n(0u8, mask_row_padding as usize));
+                }
```
RunDevelopment, line 171: > nit: can this be simplified in any way? I'm slightly glazing over this when reading to be honest. A lot of the math happens inside the loop but is actually a loop constant. Some ideas in other comments. The padding is easily moved out.

I simplified a little. Maybe that's enough?

RunDevelopment: I used your `chunks(8)` suggestion, and I think it's simple enough now. Hopefully.

src/codecs/ico/encoder.rs
197g: Yeah, probably. I'll note it down for the review list of `v1.0`, it's not an absolute stopper for doing that as it's easily compatible (just a little legacy code to carry around).

PR #3045, opened 2026-06-15
ICO: Change select best entry logic

mpospelova:
As discussed in https://github.com/image-rs/image/pull/3032, this PR updates the best entry selection logic to match the one from Chrome: first, compare the area, and then bpp.  This PR also adds logic to fill in missing bpp values (use `color_count` to set `bits_per_pixel  = ceil(log2(color_count))`).

### DIFF
--- a/src/codecs/ico/decoder.rs
+++ b/src/codecs/ico/decoder.rs
@@ -192,7 +192,19 @@ fn read_entry<R: Read>(r: &mut R, spec: SpecCompliance) -> ImageResult<DirEntry>
     }
 
     // buf[6..8]: may be bit depth (0 = unspecified) or vertical hotspot for CUR files
-    let bits_per_pixel = u16::from_le_bytes(buf[6..8].try_into().unwrap());
+    let mut bits_per_pixel = u16::from_le_bytes(buf[6..8].try_into().unwrap());
+    let color_count = buf[2];
+
+    // Some icons don't have a bit depth, only a color count. Convert the color count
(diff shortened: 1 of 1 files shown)


src/codecs/ico/decoder.rs
```diff
@@ -192,15 +192,27 @@ fn read_entry<R: Read>(r: &mut R, spec: SpecCompliance) -> ImageResult<DirEntry>
...
+        let count = match color_count {
+            0 => 256,
+            c => u16::from(c),
+        };
+        bits_per_pixel = count.next_power_of_two().trailing_zeros() as u16;
+    }
+
     if spec == SpecCompliance::Strict && bits_per_pixel > 256 {
         return Err(DecoderError::IcoEntryTooManyBitsPerPixelOrHotspot.into());
     }
```
RunDevelopment: Could you please move this back up to where `bits_per_pixel` is defined? This is a check for the raw data of the file, not on the value we infer.

RunDevelopment (commented):
Thanks for working on this! 

LGTM; Just one minor thing. Then this is ready to go.

src/codecs/ico/decoder.rs
mpospelova: Makes sense!

197g (approved):
LGTM. Motivation in https://github.com/image-rs/image/pull/2678, the prior picking implementation. was consistency with Chrome already so if this improves it, great!

RunDevelopment (approved):
Thanks!

PR #3049, opened 2026-06-21
Improve `Debug` formatting for `Delay`

RunDevelopment:
I'm working on something `Delay`'s horrible `Debug` formatting got in the way. I.e. here's what 10ms looked like before this PR:

```
Delay { ratio: Ratio { numer: 10, denom: 1 } }
```

Here's what it is now:

```
Delay(10 ms)
```

In general, delays that are whole milliseconds are formatted as just that. Delays that have fractional part are formatted as fractions. E.g. `Delay(100/3 ms)` for 33.333... ms.

### DIFF
--- a/src/animation.rs
+++ b/src/animation.rs
@@ -60,7 +60,7 @@ pub struct Frame {
 ///
 /// assert_eq!(delay_10ms, delay_10000us);
 /// ```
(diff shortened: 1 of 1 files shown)


197g (approved):
Sure, exact format subject to change but that seems like a good default.

PR #3056, opened 2026-06-26
Only reject >256 ICO color planes in strict mode (match bits_per_pixel)

mahirhir:
`read_entry()` in the ICO decoder (`src/codecs/ico/decoder.rs`) checks `num_color_planes > 256` unconditionally, even in lenient mode. Its sibling field `bits_per_pixel` was relaxed to strict-only by #3032, but the planes check was left applying in both modes. The two fields share identical CUR-hotspot semantics, so the planes check is the symmetric leftover.

The fix gates the planes check the same way #3032 gated `bits_per_pixel` (strict mode only). Verified both ways: a `lenient_too_many_planes` test passes (lenient accepts, strict still rejects); `cargo test --lib` 330 pass, 0 regression.


RunDevelopment (approved):
LGTM. Thanks for making this!

197g:
@RunDevelopment Thank you for merging this :)

PR #3064, opened 2026-07-05
Fix GIF disposal compositing for partial-width and partial-height frames

### LINKED ISSUE #3061: GIF disposal methods regression
See [this comment](https://github.com/image-rs/image/pull/2829#issuecomment-4719969777).

Between Mar 7 and Jun 16, the behavior of GIF disposal methods changed. Do a bisection to see which commit it was. The search for the bad commit shouldn't be hard. It's most likely #2914, #2924, or #2925. 

This wasn't caught by any tests, because the GIFs that were supposed to test this weren't actually being tested due to the previous test infra. So no new tests are required, since those GIFs are now actually being tested.

While you're at it, also verify that the disposal methods are actually implemented correctly (I trust neither the current nor the previous behavior). Use Chrome or Firefox as a reference decoder.

momomuchu:
Bisected the regression to #2672 (Remove AnimationDecoder trait, 0990a59e).
That refactor swapped the direction of two buffer copies in read_image:

- contiguous path (full canvas width): rows outside the frame's own height
  were zeroed instead of filled from the persistent disposal buffer
- row-based path (partial-width frames): the disposal buffer was overwritten
  with the freshly-zeroed output instead of the other way around

Same defect, two branches of the same function, same commit. Net effect: any
pixel inside a frame's row/column range but outside its own rect got reset to
transparent black instead of passing through the previous frame's composited
state. In the row-based path it also corrupted the persistent disposal buffer
itself, so subsequent frames stayed wrong too, not just the current one.

Verified by diffing decoded output against the pre-refactor code (last commit
before 0990a59e) on the existing anim fixtures; output now matches exactly.

The 10 changed reference PNGs are collateral of #2829/#3060 (merged a few days
ago), which regenerated references using the still-buggy decoder, so they baked
in this exact bug. any-disposal/mixed-disposal frames pick up real
(previously-zeroed) pixel data, which is also why large-gif-anim-combine's
anim_02 reference grew from ~1KB to ~6KB: solid zeroed regions compress a lot
better than real pixels.

Added two regression tests: one against border_touching_layers.gif for the
row-based path, one synthetic 4x6 GIF for the contiguous path.

This closes the regression-bisect part of #3061. It restores pre-refactor
parity but doesn't attempt the broader ask in that issue (auditing disposal
semantics against a browser reference), happy to open a follow-up for that.


RunDevelopment:
Thanks for looking into this!

I verified that the fix works. All animated GIF test images are now decoded correctly. The code also looks good.

Just one thing:

> Added two regression tests: one against border_touching_layers.gif for the
row-based path, one synthetic 4x6 GIF for the contiguous path.

The test for `border_touching_layers.gif` is unnecessary and can be removed. The reference images for it (`tests/reference/gif/anim/border_touching_layers.gif.anim_{01,02}.png`) already cover this and will make sure the regression doesn't repeat.

The second test doesn't have to be code. I think it would be better to save the synthetic GIF as a file (maybe `disposal_full_width_partial_height_frame_preserves_previous_state_outside_rows.gif`) and use the reference test infra to verify that it's decoded correctly (i.e. save it next to the other animated GIFs in `tests/images/gif/anim/`.)

src/codecs/gif.rs
```diff
@@ -683,4 +681,11 @@ mod test {
         let mut buf = vec![0u8; layout.total_bytes() as usize];
         assert!(decoder.read_image(&mut buf).is_ok());
     }
+
+    // Regression coverage for https://github.com/image-rs/image/issues/3061 (disposal compositing
+    // corrupted pixels outside a frame's own rectangle but still covered by that frame's row or
+    // column range) now lives entirely in the reference-image test suite: the row-based path is
+    // covered by `tests/reference/gif/anim/border_touching_layers.gif.anim_{01,02}.png`, and the
+    // contiguous fast path is covered by
+    // `tests/images/gif/anim/disposal_full_width_partial_height_frame_preserves_previous_state_outside_rows.gif`.
```
RunDevelopment: Please also remove this comment.

RunDevelopment:
Also, a question for my own sanity, @momomuchu. Did you use AI for this, and if so how much? It's fine either way. It's just that the comments feel a bit AI-y to me and so I'd like to know whether my sensors are off :)

momomuchu:
Done on the tests: dropped the `border_touching_layers` one and turned the synthetic test into a reference fixture (`disposal_full_width_partial_height_frame_preserves_previous_state_outside_rows.gif` under `tests/images/gif/anim/`), verified against the reference-image harness. Pushed.

And yeah, I use AI. I apply it and gate the quality with an ISO-style verify-and-validate pass: freeze what correct means, failing test first, then a separate review that reruns the real CI and negative-controls the test (revert the fix, confirm it goes red) so it actually has teeth. The review and the call to submit are mine.

Fair that the comments read a bit AI-y though, that's on me, I'll keep them plainer from here. Genuinely curious what tipped you off, the prose or something in the code?

RunDevelopment:
Please also remove the comment I mentioned before. Then I can merge this.

As for AI: That's a good process. I would just add that you mention AI usage in the PR description. That's usually required by most AI policies anyway (not that we currently have one), so it's a good habit.

> Genuinely curious what tipped you off, the prose or something in the code?

From my experience, AI-aided and AI-written PRs have a lot of test code and comments relative to the actual code changes. This, of course, comes down to the writing/coding style of the author, but generally purely or mostly human-made PRs focus on the code fix and have proportionally fewer comments and tests.

With this hint, I couldn't help but be biased to see the comments as a bit AI-ish. But I really wasn't sure and doubted myself, so I asked :)

RunDevelopment (approved):
Thank you @momomuchu!

PR #3066, opened 2026-07-07
exr: Support reading and writing `Luma<f32>` images

### LINKED ISSUE #3016: Exr: Support reading and writing `Luma<f32>` images
With `ColorType::Luma32F` being supported, we can finally have 32-bit single-channel (dynamic) images. However, our exr codec does not support reading and writing such images.

I.e. the following code fails.

```rs
let image = ImageBuffer::from_fn(16, 16, |x, y| Luma([(x * y) as f32 / 225.0]));
image.save("test.exr").unwrap();
```

## Description

Add support for both reading and writing single-channel luma exr images. (May be separate PRs.)

The OpenExr specification [recommends using `Y`](https://openexr.com/en/latest/TechnicalIntroduction.html#channel-names) as the channel name for luma. The encoder should create an image with only a `Y` channel. The decoder should look for a `Y` channel and use it. Since exr supports any number and combinations of channels (e.g. R+G+B+Y is possible), the decoder should prefer RGB/RGBA over pure luma.

The decoder implementation should be tested using a `.exr` file containing a `Y` channel. The encoder implementation should be tested using a roundtrip (=generate some image, encode, decode, compare decoded image to starting image).

Luma + Alpha encoding and decoding should not be a goal right now.

upuddu:
Closes #3016.

The OpenEXR codec previously only handled RGB/RGBA float images, so saving a `Luma<f32>` image (e.g. `ImageBuffer::from_fn(w, h, |x, y| Luma([...]))`) failed with an unsupported-color error.

This adds read and write support for single-channel luma. The encoder writes a single `Y` channel, as recommended by the OpenEXR specification. The decoder looks for a `Y` channel and decodes it as `Luma<f32>`; when a file contains both RGB and `Y` channels it still prefers RGB/RGBA, so existing behavior is unchanged.

Luma + alpha is intentionally out of scope, as noted in the issue.

Tested with a roundtrip (`roundtrip_luma`): generate a `Luma<f32>` image, encode it to an in-memory EXR, decode it back, and compare — mirroring the existing `roundtrip_rgb`/`roundtrip_rgba` tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@ Breaking Changes:
 Features:
  - Added `RbgImage::from_raw_bgr` and `RbgaImage::from_raw_bgra` constructors, which convert
    from `BGR(A)` with an optimized specialization for bytes of `BGRA`
+ - The OpenEXR codec can now read and write single-channel `Luma<f32>` images,
+   using the `Y` channel (#3016)
 
 Structural changes:
   - Increased MSRV to 1.88.0 (from 1.85.0)
--- a/src/codecs/openexr.rs
+++ b/src/codecs/openexr.rs
@@ -3,9 +3,11 @@
 //! OpenEXR is an image format that is widely used, especially in VFX,
 //! because it supports lossless and lossy compression for float data.
 //!
-//! This decoder only supports RGB and RGBA images.
-//! If an image does not contain alpha information,
+//! This decoder supports RGB and RGBA images, as well as single-channel
+//! luma (`Y`) images, which are decoded as `Luma<f32>`.
+//! If an RGB image does not contain alpha information,
 //! it is defaulted to `1.0` (no transparency).
+//! When a file contains both RGB and luma channels, the RGB channels are preferred.
 //!
 //! # Related Links
 //! * <https://www.openexr.com/documentation.html> - The OpenEXR reference.
@@ -38,8 +40,12 @@ use std::io::{BufRead, Seek, Write};
 pub struct OpenExrDecoder<R> {
     exr_reader: Option<exr::block::reader::Reader<R>>,
 
-    // select a header that is rgb and not deep
+    // select a header that is rgb or luma and not deep
     header_index: usize,
+
+    // the selected header only contains a `Y` (luma) channel, no rgb channels.
+    // such images are decoded as `L32F` and the alpha preference is ignored.
+    is_luma: bool,
 
     // decode either rgb or rgba.
     // can be specified to include or discard alpha channels.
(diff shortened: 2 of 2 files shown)


RunDevelopment:
Given that you made 9 PRs to 8 repos in the last hour, I'll assume that all of this is AI-generated, correct?

If so, then I'm not sure whether we should accept this. "Good first issue" issues aren't supposed to be fodder for AI agents but starting points for humans.

@197g What's your take on this?

197g:
Sure, I'd prefer them to be non-AI but it's not dogmatic either in that I suppose some newer programmers just do that nowadays so that _is_ the shape of their first contribution. I'd say the problem here is the LLM'ism of the code itself. The code needs a bit of structural thought but instead gets pure repetition.

Firstly, LumaA is maybe out-of-scope but that's different from more or less ignoring its existence. That is, if it's not supported to decode that plane and tack it on then there should be an error to avoid behavior change in the future when it gets added. (Note the exclusion is partly an artifact of missing `ColorType::LumaA32F` which is resolved, it's not a requirement _not_ to do it).

Secondly, I have a parse-don't-validate concern. While parsing the header there's a boolean flag that decides Rgba vs. Luma decoding based on available channels that was tacked on to the existing `is_alpha_present` flag. I'd rather we have a proper newtype `enum` representation of the *three* possible _decision_ instead, so that the one we made here that can be later used instead of re-computing the same in a potential different way and having an error. Planning for non-standard color of alternate names, it'd be interesting to have a type where we can retain information on the channel index when it is parsed.

upuddu:
Fair question, and yes I used AI assistance on these. I do review, test, and stand behind what I send, but I hear the concern, especially on good-first-issues. If AI-assisted PRs aren't welcome here, I completely understand - say the word and I'll close it.

On the substance, @197g:
- LumaA: agreed it shouldn't be silently dropped. I'll return an explicit unsupported-color error for a Y+A file, so adding LumaA<f32> later (now that ColorType::LumaA32F exists) is an additive change.
- parse-don't-validate: I'll replace the boolean-on-is_alpha_present with a small enum that captures the RGB/RGBA/Luma decision once at header-parse time (keeping the channel info), and have the reader consume that instead of re-deriving it.

Update incoming - happy to iterate.

upuddu:
@197g Sorry for the long silence - I pushed these changes in e93a978 the same day and then never actually said so here.

Both of your points are addressed:

**LumaA is no longer silently dropped.** A header carrying `Y` + `A` now returns `ImageError::Unsupported` with `UnsupportedErrorKind::Color(ExtendedColorType::La32F)`, instead of decoding as plain luma and discarding the alpha plane. That way, adding real `LumaA<f32>` support later is purely additive — no file that decodes today would change meaning.

**Parse, don't validate.** `is_luma` and `alpha_present_in_file` are both gone, replaced by a single `ExrColorLayout` enum covering exactly the three decisions — `Rgb`, `Rgba`, `Luma` — computed once in `with_alpha_preference`. The rest of the decoder consumes that decision rather than re-inspecting the channel list, so `prepare_image` and `read_image` can't interpret the same header in two different ways. Each variant also retains the resolved channel indices (`r`/`g`/`b`/`a`/`y`), per your note about non-standard names and alternate layouts, so future work can build on the parsed positions instead of rescanning.

Happy to rework it if the shape isn't what you had in mind.

PR #3068, opened 2026-07-09
Add a security policy

fintelia:
I tried writing up a security policy heavily inspired by https://opensourcesecurity.io/2026/06-rules-of-security.

The idea is to clearly exclude categories of issues that either don't need to be handled confidentially by maintainers, or which we've never promised in the first place.

Based on Google Project Zero's interactions with FFmpeg, I thought about also trying to add something like "if you're being paid by one of the corporate vulnerability finding programs, then you also have to write the patch/help us fix". But the concern seemed mostly hypothetical and couldn't find wording I liked.

This is very much intended to be a first draft that can be iterated on as we get experience.

---

_Note: rather than merging this PR, my plan is to create a `.github` repo and place the SECURITY.md file there. If the GitHub docs are to be believed, this should display the security policy for all repositories under the organization._

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -34,10 +34,11 @@ undefined-behavior due to the bug, then it doesn't qualify as a vulnerability.
 
 **Out-of-memory or other denial of service:** Resource limits are on a best-effort basis.
 
-**Vulnerabilities in tests, benchmarks, or examples:** Downsteam users do not run these directly so
+**Vulnerabilities in tests, benchmarks, or examples:** Downstream users do not run these directly so
 issues like path traversal, soundness bugs, and such are not a concern. The exception here is if
 merely running the code could compromise a developer's machine.
 
-**Vulnerabilities in GitHub Actions code:** We don't expose any secrets to our CI and don't use use
-[trusted publishing](https://crates.io/docs/trusted-publishing), so the worst that could happen here
-is wasting our (free) quota or getting inaccurate test results.
+**Code execution in GitHub Actions:** Our CI is configured to have read-only access and we don't use
+[trusted publishing](https://crates.io/docs/trusted-publishing) or self-hosted runners, so merely
+running attacker controlled code isn't a big risk. Escalating to repository write-access or similar
+should of course be treated as a vulnerability.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,43 @@
+
+<!--- This file should be copied to the root of github.com/image-rs/.github to be applied org wide --->
+
+# Security Policy
+
+ The purpose of this policy is to protect users of software built with _image-rs_ crates from harm
+ without placing an undue burden on the volunteers maintaining the project.
+
+## Reporting a vulnerability
+
+If you believe that you have discovered a security vulnerability in one of the _image-rs_ crates
+(and it doesn't fall under an out-of-scope category listed below) then please report it using
+GitHub's [private vulnerability
+reporting](https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/report-privately).
+
+Your report should be concise and human written. Otherwise, we may ask you to rewrite it before
+engaging further. We only need enough details to confirm that the vulnerability exists. Please do
+not fill out the "Affected products", "Severity", or "Weaknesses" sections (we cannot edit the
+template). Those can be addressed later.
+
+Once the vulnerability is confirmed, we'll work with you to remediate it, and then disclose.
+
+## Out-of-scope
+
+The following are examples of defects that are not considered security vulnerabilities and should be
+reported as normal bugs:
+
+**Crashes or panics (e.g. triggered from crafted input):** In Rust, panicking is considered safe.
+Our libraries extensively use assertions to detect implementation bugs and while we strive not to
+panic even on crafted inputs, it isn't guaranteed.
+
+**Contrived soundess bugs:** If there isn't likely to be downstream code that triggers
+undefined-behavior due to the bug, then it doesn't qualify as a vulnerability.
+
+**Out-of-memory or other denial of service:** Resource limits are on a best-effort basis.
+
+**Vulnerabilities in tests, benchmarks, or examples:** Downsteam users do not run these directly so
+issues like path traversal, soundness bugs, and such are not a concern. The exception here is if
+merely running the code could compromise a developer's machine.
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Out-of-memory or other denial of service:** Resource limits are on a best-effort basis.
+
+**Vulnerabilities in tests, benchmarks, or examples:** Downsteam users do not run these directly so
+issues like path traversal, soundness bugs, and such are not a concern. The exception here is if
+merely running the code could compromise a developer's machine.
+
+**Vulnerabilities in GitHub Actions code:** We don't expose any secrets to our CI and don't use use
+[trusted publishing](https://crates.io/docs/trusted-publishing), so the worst that could happen here
+is wasting our (free) quota or getting inaccurate test results.
```
197g: This is a little broad. It's true that data extraction is not a vulnerability due to the lack of secrets but a vulnerability that might cause silent modifications to the repository (main) branch state would still be a vector for worm attacks against developers or affect the released artifact indirectly.

197g (commented):
Overall very nice, thanks for putting it together.

197g:
> Based on Google Project Zero's interactions with FFmpeg, I thought about also trying to add something like "if you're being paid by one of the corporate vulnerability finding programs, then you also have to write the patch/help us fix". But the concern seemed mostly hypothetical and couldn't find wording I liked.

We could clarify our own process in a bit of a subversive manner. The total awarded sum is dependent on properties of the issue. From the manner in which we've handled these in the past, input from our side for these will not usually be complete until a resolution where we expect a discussion about how such an implementation covers (all) uncovered angles. This is of course descriptive and not prescriptive, certain situations might call for different handling. However the aim with such a paragraph is only to steer the incoming stream of bugs; and give bug programs some leverage in forwarding our processes. "Final confirmation of the scope of a security issue is expected to occur when a fix is merged".

However then another concern would be how to handle embargoed contributions. I'd be open to making a public _note_ of some existing bug and having implementation sent as e-mail if it's critical enough. As a concrete example from the past, Shnatsel had also opened branches in `imageproc` without a PR and forwarded the more detailed description and need for release privately. That certainly worked alright.

SECURITY.md
fintelia: I'll try to re-word. The lack of secrets _should_ mean that the CI doesn't have only read-access to our repository. I agree that an oversight that gave them write access would warrant confidential reporting.

fintelia:
I haven't used it, but based on [the docs](https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/report-privately) for private vulnerability reporting, it looks like there's a way to create a "temporary private fork" to work on fixes.

Also to clarify, Project Zero isn't a bug bounty program. They're a team of full-time Google employees charged with finding vulnerabilities in Google's dependencies, but at least in [their interactions with FFmpeg](https://thenewstack.io/ffmpeg-to-google-fund-us-or-stop-sending-bugs/) just sent a threat to disclose after 90 days rather than provide any help fixing the bugs they found.

RunDevelopment:
> I haven't used it, but based on [the docs](https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/report-privately) for private vulnerability reporting, it looks like there's a way to create a "temporary private fork" to work on fixes.

I have used it before and that's how it works. It's a private fork/branch that only the reporter and maintainers can see, plus a private discussion. Basically a private PR with the usual discussion, code review, and all. The main difference is that you also work on a CVE and security advisory alongside the patch.

SECURITY.md
```diff
@@ -0,0 +1,43 @@
...
+**Crashes or panics (e.g. triggered from crafted input):** In Rust, panicking is considered safe.
+Our libraries extensively use assertions to detect implementation bugs and while we strive not to
+panic even on crafted inputs, it isn't guaranteed.
+
+**Contrived soundess bugs:** If there isn't likely to be downstream code that triggers
+undefined-behavior due to the bug, then it doesn't qualify as a vulnerability.
+
+**Out-of-memory or other denial of service:** Resource limits are on a best-effort basis.
+
+**Vulnerabilities in tests, benchmarks, or examples:** Downsteam users do not run these directly so
```
Eh2406: Nit: `Downstream` not `Downsteam`

SECURITY.md
```diff
@@ -0,0 +1,43 @@
...
+## Out-of-scope
+
+The following are examples of defects that are not considered security vulnerabilities and should be
+reported as normal bugs:
+
+**Crashes or panics (e.g. triggered from crafted input):** In Rust, panicking is considered safe.
+Our libraries extensively use assertions to detect implementation bugs and while we strive not to
+panic even on crafted inputs, it isn't guaranteed.
+
+**Contrived soundess bugs:** If there isn't likely to be downstream code that triggers
```
Shnatsel:
```suggestion
**Contrived soundness bugs:** If there isn't likely to be downstream code that triggers
```

typo

SECURITY.md
Shnatsel: I don't think we should rule out Github Actions entirely. While the current setup seems fine, Github Actions have really subtle pitfalls with major security consequences.

Only ever using read-only tokens still allows escalating via the cache in some cases: https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/

Self-hosted / custom runners (like if we ever add POWER or s390x to CI) also pose a hazard regardless of the token:
https://adnanthekhan.com/2023/12/20/one-supply-chain-attack-to-rule-them-all/
https://johnstawinski.com/2024/01/11/playing-with-fire-how-we-executed-a-critical-supply-chain-attack-on-pytorch/

Shnatsel (commented):
I've left some nits about the details, but overall it seems well thought out!

fintelia:
Ok, I changed it to say that merely getting _code execution_ in GitHub Actions is out-of-scope.

197g (approved):

RunDevelopment:
Hmmm, I thought @fintelia didn't want to have this merged? Did the `.github` repo not work out?

> Note: rather than merging this PR, my plan is to create a .github repo and place the SECURITY.md file there. If the GitHub docs are to be believed, this should display the security policy for all repositories under the organization.

fintelia:
Just tried creating the `.github` repository, and it seems to have worked. 

Edit: Removed the SECURITY.md from this repository and it is still showing [here](https://github.com/image-rs/image/security/policy). Looks like a success

PR #3069, opened 2026-07-12
Provide defaults for more `Pixel` methods

RunDevelopment:
I thought it would be good idea to provide defaults for `Pixel` methods that can be implemented in terms of other methods.

This makes it easier to implement `Pixel` and reduces the complexity of the `define_colors` macro.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -445,22 +445,10 @@ impl<T: $($bound+)*> Pixel for $ident<T> {
         pix
     }
 
-    fn map<F>(& self, f: F) -> $ident<T> where F: FnMut(T) -> T {
-        let mut this = (*self).clone();
-        this.apply(f);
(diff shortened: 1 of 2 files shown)


197g (approved):
To explicitly rationalize the choice here in favor of defaulting `map` rather than `apply`: we require `Copy` to enable this but contrary to `std`'s iterator conundrum there are scenarios where in-place modification is highly expected and our naming makes this clear. It's usually simpler to write a modifying method as a base case where assignmented and compound assignment operators can be used with numerics of channels.

PS: also speculatively LLVM is sometimes picky, trying to default `apply` as read-modify-write creates more optimizer pressure. There's already a difference optimizing `*lpr++ = v` vs `*lpr = v; lpr++`.

PR #3070, opened 2026-07-13
Add method to expose CICP in image decoder

### LINKED ISSUE #2985: Expose `metadata::Cicp` in decoder
#2531 has introduced CICP metadata for color transformations.

Read this metadata from the decoder (e.g. `png`'s [`CodingIndependentCodePoints`](https://docs.rs/png/0.18.1/png/struct.CodingIndependentCodePoints.html)) and expose it in the API.

This is part of https://github.com/image-rs/image/issues/2222

mstoeckl:
This adds a method which exposes the image color space information as CICP values _when possible_.  This is immediately useful for decoders for formats like PNG, which directly supports CICP; but also for specialized formats like HDR (uses linear transfer function, although the official default color primaries are weird) and QOI (has a linear colorspace option that I expect is used for non-color data, assuming it gets used at all).

Edit: the current revision also updates the way ICC color profiles are communicated.

This PR is somewhat minimal; I've not added support for CICP in AVIF as that requires dependency upgrades; or tried to map the full set of PNG color chunks (gAMA, cHRM, sRGB) to CICP when possible; or looked at to what extent various other formats (BMP, JPEG, WEBP, Farbfeld, OpenEXR, TIFF) could have color space information implementable as CICP.  But this (along with synthesizing ICC profiles as needed, similar to the way the BMP decoder does it) is something I'd rather defer if there is no reason to expect it to break the core design.

This would close https://github.com/image-rs/image/issues/2985.

### DIFF
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -1,4 +1,5 @@
-use crate::io::DecoderPreparedImage;
+use crate::io::{DecodedColorProfile, DecodedMetadataHint, DecoderPreparedImage, FormatAttributes};
+use crate::metadata::Cicp;
 use crate::utils::vec_try_with_capacity;
 use std::cmp::{self, Ordering};
 use std::io::{self, BufRead, Seek, SeekFrom};
@@ -2448,6 +2449,14 @@ impl<R: BufRead + Seek> BmpDecoder<R> {
 }
 
 impl<R: BufRead + Seek> ImageDecoder for BmpDecoder<R> {
+    fn format_attributes(&self) -> FormatAttributes {
+        FormatAttributes {
+            icc: DecodedMetadataHint::InHeader,
+            color_profile: DecodedMetadataHint::InHeader,
+            ..FormatAttributes::default()
+        }
+    }
+
     fn prepare_image(&mut self) -> ImageResult<DecoderPreparedImage> {
         let color = if self.indexed_color {
             ColorType::L8
@@ -2465,23 +2474,28 @@ impl<R: BufRead + Seek> ImageDecoder for BmpDecoder<R> {
     }
 
     fn icc_profile(&mut self) -> ImageResult<Option<Vec<u8>>> {
+        Ok(self.icc_profile.clone())
+    }
+
+    fn color_profile(&mut self) -> ImageResult<Option<DecodedColorProfile>> {
         match self.color_space_info {
             Some(ColorSpaceInfo::CalibratedRgb(params)) => {
                 // Synthesize an ICC profile from the calibrated RGB parameters
                 // and store it directly — no file read needed.
-                Ok(Some(
+                Ok(Some(DecodedColorProfile::from_icc(
                     params
                         .to_color_profile()
                         .encode()
                         .expect("synthetic profile should always succeed"),
-                ))
+                )))
             }
-            Some(ColorSpaceInfo::EmbeddedIcc(_)) => {
-                Ok(Some(self.icc_profile.clone().expect("icc was loaded")))
+            Some(ColorSpaceInfo::EmbeddedIcc(_)) => Ok(Some(DecodedColorProfile::from_icc(
+                self.icc_profile.clone().expect("icc was loaded"),
+            ))),
+            Some(ColorSpaceInfo::Srgb) => {
+                Ok(Some(DecodedColorProfile::from_plain_cicp(Cicp::SRGB)))
             }
-            // LCS_sRGB / LCS_WINDOWS_COLOR_SPACE: the caller treats
-            // "no ICC profile" as sRGB, so nothing to store.
-            Some(ColorSpaceInfo::Srgb) | None => Ok(None),
+            None => Ok(None),
         }
     }
 
@@ -2608,6 +2622,7 @@ mod test {
         let mut decoder = BmpDecoder::new(f).unwrap();
         let profile = decoder.icc_profile().unwrap();
         assert!(profile.is_some());
+        assert!(profile == decoder.color_profile().unwrap().unwrap().to_icc().unwrap());
         let profile_data = profile.unwrap();
         assert_eq!(profile_data.len(), 3048);
         validate_icc_profile(
@@ -2622,6 +2637,7 @@ mod test {
         let mut decoder = BmpDecoder::new(f).unwrap();
         let profile = decoder.icc_profile().unwrap();
         assert!(profile.is_some());
+        assert!(profile == decoder.color_profile().unwrap().unwrap().to_icc().unwrap());
         let profile_data = profile.unwrap();
         assert_eq!(profile_data.len(), 540);
         validate_icc_profile(
@@ -2637,7 +2653,7 @@ mod test {
         // pal8v4.bmp has a V4 header with LCS_CALIBRATED_RGB — should synthesize an ICC profile.
         let data = std::fs::read("tests/images/bmp/images/pal8v4.bmp").unwrap();
         let mut decoder = BmpDecoder::new(Cursor::new(&data)).unwrap();
-        let profile = decoder.icc_profile().unwrap();
+        let profile = decoder.color_profile().unwrap().unwrap().to_icc().unwrap();
         assert!(
             profile.is_some(),
             "pal8v4: should have a synthesized ICC profile from calibrated RGB parameters"
--- a/src/codecs/bmp/decoder.rs
+++ b/src/codecs/bmp/decoder.rs
@@ -284,7 +284,7 @@ impl ParsedIccProfile {
 }
 
 /// Color space data parsed from V4/V5 BMP headers.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum ColorSpaceInfo {
     /// LCS_CALIBRATED_RGB: endpoint and gamma values specified in the header.
     CalibratedRgb(CalibratedRgb),
@@ -313,31 +313,26 @@ impl ColorSpaceInfo {
                     i32::from_le_bytes(buffer[offset..offset + 4].try_into().unwrap())
                 };
 
-                // FXPT2DOT30 (2.30 fixed-point) → f32.
-                let fxpt2dot30 = |val: i32| -> f32 { val as f32 * (1.0 / (1u64 << 30) as f32) };
-                // FXPT16DOT16 (unsigned 16.16 fixed-point) → f32.
-                let fxpt16dot16 = |val: u32| -> f32 { val as f32 / 65536.0 };
-
                 // CIEXYZTRIPLE: 9 FXPT2DOT30 values at offsets 60-95 from header
                 // start (56-91 from after size field). Layout:
                 //   RedX, RedY, RedZ, GreenX, GreenY, GreenZ, BlueX, BlueY, BlueZ
                 // We read only X and Y per primary (Z is implicit: Z = 1 - X - Y
                 // for chromaticity, but BMP stores raw CIE XYZ values).
-                let rx = fxpt2dot30(read_i32(56));
-                let ry = fxpt2dot30(read_i32(60));
-                let gx = fxpt2dot30(read_i32(68));
-                let gy = fxpt2dot30(read_i32(72));
-                let bx = fxpt2dot30(read_i32(80));
-                let by = fxpt2dot30(read_i32(84));
+                let rx = Fxpt2Dot30(read_i32(56));
+                let ry = Fxpt2Dot30(read_i32(60));
+                let gx = Fxpt2Dot30(read_i32(68));
+                let gy = Fxpt2Dot30(read_i32(72));
+                let bx = Fxpt2Dot30(read_i32(80));
+                let by = Fxpt2Dot30(read_i32(84));
 
                 // Gamma values at offsets 96-107 from header start (92-103 from after size).
-                let gamma_r = fxpt16dot16(read_u32(92));
-                let gamma_g = fxpt16dot16(read_u32(96));
-                let gamma_b = fxpt16dot16(read_u32(100));
+                let gamma_r = Fxpt16Dot16(read_u32(92));
+                let gamma_g = Fxpt16Dot16(read_u32(96));
+                let gamma_b = Fxpt16Dot16(read_u32(100));
 
                 // Validate: Y values must be non-zero (used as denominators in
                 // XYZ→chromaticity conversion by color management libraries).
-                if ry == 0.0 || gy == 0.0 || by == 0.0 {
+                if ry == Fxpt2Dot30(0) || gy == Fxpt2Dot30(0) || by == Fxpt2Dot30(0) {
                     return None;
                 }
 
@@ -362,41 +357,61 @@ impl ColorSpaceInfo {
     }
 }
 
+/// FXPT2DOT30, 2.30 fixed-point
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct Fxpt2Dot30(i32);
+
+impl Fxpt2Dot30 {
+    fn to_float(self) -> f32 {
+        self.0 as f32 * (1.0 / (1u64 << 30) as f32)
+    }
+}
+
+/// FXPT16DOT16, unsigned 16.16 fixed-point
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct Fxpt16Dot16(u32);
+
+impl Fxpt16Dot16 {
+    fn to_float(self) -> f32 {
+        self.0 as f32 / 65536.0
+    }
+}
+
 /// Calibrated RGB color space parameters from a BMP V4/V5 header.
 ///
 /// When the header's `bV4CSType` is `LCS_CALIBRATED_RGB`, these fields
 /// carry the CIE XYZ endpoint coordinates for the RGB primaries and
-/// per-channel gamma values, parsed from the FXPT2DOT30 / FXPT16DOT16
+/// per-channel gamma values, stored as the raw FXPT2DOT30 / FXPT16DOT16
 /// fixed-point fields in the header.
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct CalibratedRgb {
     /// Red primary CIE X coordinate (FXPT2DOT30).
-    rx: f32,
+    rx: Fxpt2Dot30,
     /// Red primary CIE Y coordinate (FXPT2DOT30).
-    ry: f32,
+    ry: Fxpt2Dot30,
     /// Green primary CIE X coordinate (FXPT2DOT30).
-    gx: f32,
+    gx: Fxpt2Dot30,
     /// Green primary CIE Y coordinate (FXPT2DOT30).
-    gy: f32,
+    gy: Fxpt2Dot30,
     /// Blue primary CIE X coordinate (FXPT2DOT30).
-    bx: f32,
+    bx: Fxpt2Dot30,
     /// Blue primary CIE Y coordinate (FXPT2DOT30).
-    by: f32,
+    by: Fxpt2Dot30,
     /// Red channel gamma (unsigned FXPT16DOT16).
-    gamma_r: f32,
+    gamma_r: Fxpt16Dot16,
     /// Green channel gamma (unsigned FXPT16DOT16).
-    gamma_g: f32,
+    gamma_g: Fxpt16Dot16,
     /// Blue channel gamma (unsigned FXPT16DOT16).
-    gamma_b: f32,
+    gamma_b: Fxpt16Dot16,
 }
 
 impl CalibratedRgb {
     /// Build a moxcms `ColorProfile` from the calibrated RGB primaries and gamma.
     fn to_color_profile(self) -> moxcms::ColorProfile {
         let primaries = moxcms::ColorPrimaries {
-            red: moxcms::Chromaticity::new(self.rx, self.ry),
-            green: moxcms::Chromaticity::new(self.gx, self.gy),
-            blue: moxcms::Chromaticity::new(self.bx, self.by),
+            red: moxcms::Chromaticity::new(self.rx.to_float(), self.ry.to_float()),
+            green: moxcms::Chromaticity::new(self.gx.to_float(), self.gy.to_float()),
+            blue: moxcms::Chromaticity::new(self.bx.to_float(), self.by.to_float()),
         };
 
         let mut profile = moxcms::ColorProfile::new_srgb();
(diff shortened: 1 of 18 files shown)


197g:
How does this handle different CICP priorities vs. other color information? For PNG the block is by specification supposed to supersede other color information so that an ICC profile becomes a mere auxiliary hint whereas in other formats the role is reversed. I'm also not sure how to deal with errors in these cases. The ICC can itself contain a CICP which _must_ be consistent with the profile itself. While that is consistently specified the situation of encountering unsupported by well-formed data here is odd. We don't want to error but the ImageReader could conceivably:

- fallback to the complicated ICC data and indicate our own `Cicp` as `undefined`
- attempt to further fallback to deprioritized CICP information.

mstoeckl:
> How does this handle different CICP priorities vs. other color information?

I've updated the `ImageDecoder` documentation to make it clear that if both `cicp_values()` and `icc_profile()` return Ok(Some(...)), they should match. 

The decoder implementation should ensure this is the case; if not, then users of the trait (like ImageReader) risk choosing arbitrarily.

> For PNG the block is by specification supposed to supersede other color information so that an ICC profile becomes a mere auxiliary hint whereas in other formats the role is reversed. 

The PNG decoder returns Ok(None) for the ICC profile when a cICP tag is present; it could also return a synthetic ICC profile derived from the cICP tag. (And for `cicp_values()`, if there is no `cICP`, and the ICC profile has a `cicp` tag, the decoder could return the contents of that tag.)

For a format where the ICC profile takes priority, if any ICC profile is present, then the decoder could just return an UnsupportedError in `cicp_values` because it cannot convert the color space of the image (determined by the ICC profile) to CICP.

197g:
> For a format where the ICC profile takes priority, if any ICC profile is present, then the decoder could just return an UnsupportedError

I find that a bit odd. A major incentive for the structural change of `ImageDecoder` to me was resolving that decoders could say what factually exists based on the file instead of having to cover something up for the sake of some policy of `ImageReader`. Imo such erasure of information should happen as part of the algorithm consuming the available information instead.

mstoeckl:
> A major incentive for the structural change of ImageDecoder to me was resolving that decoders could say what factually exists based on the file instead of having to cover something up for the sake of some policy of ImageReader. 

1. For AVIF and PNG have different policies on which of ICC vs CICP-derived color spaces is used when both chunks are available. (https://github.com/AOMediaCodec/libavif/wiki/CICP#aside-icc-profiles, vs https://www.w3.org/TR/png-3/#color-chunk-precendence)
2. Many formats support color space information that is not directly encoded using ICC or CICP; mostly these specify color primaries and maybe also custom gamma (see BMP's CalibratedRgb, PRIMARIES for HDR, gAMA+cHRM for PNG, the chromaticities for OpenEXR), but there is also HDR metadata (mDCV, cLLI for PNG) and other parameters (adoptedNeutral for OpenEXR). Mapping these to CICP and ICC when possible could make it is easier to establish a uniform interface.

That being said, I'm certainly not certain that this PR is the right way to handle the issue. One alternative I've been thinking of is a `colorspace() -> ColorSpace` method for ImageDecoder which provides the decoder's best estimate at the color space information of the image, where ColorSpace is an enum containing any of (CICP, ICC, gamma + primaries) plus whatever else proves necessary int the future. Then `icc_profile()` could be left as a largely decorative function to be used if someone needs the exact bytes or ICC profile metadata (author, etc.) contained in the image, and `cicp_values` would likely not be needed.

> Imo such erasure of information should happen as part of the algorithm consuming the available information instead.

This is tricky, because image format specs may conflict with metadata specs. [DPX](https://pub.smpte.org/doc/st268-2/) explicitly permits external metadata but recommends that the main file metadata (like orientation or CICP codes) override any present in external metadata (which presumably includes XMP and EXIF), except when left undefined; [JXL](https://github.com/libjxl/libjxl/blob/main/doc/format_overview.md#metadata-versus-image-data) files always ignore EXIF orientation, as should [AVIF](https://issues.chromium.org/issues/40197587).

197g:
It is indeed tricky, which is why we should be doing it right. Regarding non-icc/cicp color space information in combination with standard ICC data we could have a separate method for any profile synthesized by a decoder. At least that would fit with `ImageReader` choosing between available options and hopefully clarify the role for a re-encoding loop. (We could also hack it as an Abstract ICC Profile with specific copyright data, a mandatory tag, if we _must_ have in-band information). Regarding application of metadata we should improve the compliance and add flags if there's a difference between formats.

I continue to be weirded out by JXL and heif directions, this time for explicitly refusing to push complexity through dedicated standards as if that removes said complexity from reality, but reading through the ''justification'' given in the AVIF thread I seem to at least understand what collective mind parasite got us there—so now we have more complexity, yay. Since: "Metadata of type Exif as specified by JEITA CP-3451 may be present in the file and shall be output by the MIAF reader" but we must always heed "irot and imir" and it's definitely wrong to apply twice. Requiring "no image transformations […] indicated in the Exif metadata" as opposed to presuming equivalence is absolutely asinine. So now re-encoding such an image into another format is no longer simple, requiring processing of of metadata to scrub orientation; and suddenly a re-encoder must deeply understand *every* kind of metadata to modify it accordingly but of course it never does, evidently imagmagick does not, and then you get broken images. (Sidenote that CSS's [`image-orientation: from-image`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/image-orientation) also doesn't add to clarity by yet another need for information; is imir a from-image orientation or an inherent part of the pixel matrix? That kind of switch, if well-defined at all, would be easier to support if it were placed in `ImageReaderOptions` than something for every decoder to do).

mstoeckl:
> Regarding non-icc/cicp color space information in combination with standard ICC data we could have a separate method for any profile synthesized by a decoder.

I've updated the design of the MR to expose ICC profiles and CICP value through new `color_profile_to_icc` and `color_profile_to_cicp` functions that make it clear the results may be synthetic. I _also_ extended `color_profile_to_cicp` to return a struct through which additional HDR metadata can be passed; it looks like PQ's tone mapping rules may require additional maxCLL/maxFALL parameters to implement, and future video color standards might get even more complicated.

(I've made the changes somewhat quickly to get the core ideas out, so the decoder implementations may need a bit more work.)

mstoeckl:
> (I've made the changes somewhat quickly to get the core ideas out, so the decoder implementations may need a bit more work.)

I was indeed missing a few implementations; now every format which could expose an ICC profile should now have the `color_profile_to_*` methods implemented. I've also updated `color_profile_to_cicp` to return Result<Option<>> like `color_profile_to_icc`.

(Edit: I may tweak decoder implementations over the next few days to help make it clearer how this might play out, but the current change to the ImageDecoder API is certainly ready to review.)

src/codecs/bmp/decoder.rs
```diff
@@ -363,37 +363,42 @@ impl ColorSpaceInfo {
...
 /// carry the CIE XYZ endpoint coordinates for the RGB primaries and
-/// per-channel gamma values, parsed from the FXPT2DOT30 / FXPT16DOT16
+/// per-channel gamma values, stored as the raw FXPT2DOT30 / FXPT16DOT16
 /// fixed-point fields in the header.
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct CalibratedRgb {
     /// Red primary CIE X coordinate (FXPT2DOT30).
-    rx: f32,
+    rx: u32,
```
197g: I'd suggest a newtype here.

src/metadata/cicp.rs
```diff
@@ -111,6 +111,31 @@ impl CicpColorPrimaries {
     }
 }
 
+impl TryFrom<u8> for CicpColorPrimaries {
+    type Error = ();
```
197g: Side note: this inspired me to check how `std` implements `Try` for options and I found that its residual (the 'error' value type) is `Option<Infallible>` which is certainly odd 😂 In any case this is a note that we want some error type here, it can be shared between the components. In the worst case we make dedicated methods for detailed errors.

src/io/decoder.rs
```diff
@@ -115,14 +115,48 @@ pub trait ImageDecoder {
...
+    /// values (which could easily occur for images with HDR metadata), or if
+    /// the decoder simply has not implemented conversion logic for it,
+    /// this returns [`UnsupportedError`](`crate::error::UnsupportedError`). This allows a
     /// reader to continue while differentiating from missing metadata.
-    fn icc_profile(&mut self) -> ImageResult<Option<Vec<u8>>> {
+    ///
+    /// In the event that both this function and [`ImageDecoder::color_profile_to_cicp`]
+    /// do not return [`UnsupportedError`](`crate::error::UnsupportedError`), the returned
+    /// color profile informations should be consistent with each other.
+    fn color_profile_to_icc(&mut self) -> ImageResult<Option<Vec<u8>>> {
```
197g: I'd rather this not need to be implemented by each individual decoder, but rather similar to the conversion ICC -> CICP happens by our own very rough parser on the ICC byte object. That is this should only be necessary for decoders with specific representations of color that are not covered by the supported color space specifications (complication regarding interface design: this set may expand over time?).

197g:
I do like the new sketch better than the previous for sure, just the duplication in conversion between different color profile representations is rather annoying. It's of course plausible for (many older) formats with much more special representations but will just happen to be a lot of close-but-not-duplicate code otherwise especially for `image-ext`. Unless we make it accessible to downstream but that's a bigger API suggestion.

src/io/decoder.rs
mstoeckl: Sorry, I don't quite follow what you mean. 

The combination of the default Ok(None) return and consistency requirement for the `color_profile_to_icc` and `color_profile_to_cicp` methods implies they need to be implemented for any decoder for a format which may provide color profile information. A decoder that only has logic to provide one of the two methods (for example, because the format's color space is a subset of CICP) could return UnsupportedError on the other and leave the question of how to convert from CICP to ICC (or vice-versa) to the user.

(If a new `color_profile_to_X` is added in the future, the method should default to returning UnsupportedError, and individual decoders should override as they gain support for it.)

src/metadata/cicp.rs
mstoeckl: > In the worst case we make dedicated methods for detailed errors.

I've gone ahead and used separate error types; as tedious as it is, it _does_ make for better error messages.

src/codecs/bmp/decoder.rs
mstoeckl: Done.

src/io/decoder.rs
197g: What I'm saying is that _when_ you implement it the internals all look very similar. Currently every decoder that has ICC support implements `color_profile_to_cicp` with a variant that looks at the ICC itself—and then returns unsupported (but with a TODO and wording suggesting it should do something here). That's a lot of duplication with potential for inconsistency. (Edit: unless there's a priority for CICP as with PNG).

The conversion to cicp should only need to happen by the decoder if the applied color information is _not_ in a standard form. If there's an ICC profile with CICP information inside of it the extract should be handled by `ImageReader`.

197g: What I do appreciate here is separating ICC metadata, i.e. included ICC profiles, from synthesized ones as in the BMP decoder. That is convincingly better and makes the PNG case nice. But this, too, should not have to report a profile that's already returned from the ICC metadata.

mstoeckl: > The conversion to cicp should only need to happen by the decoder if the applied color information is not in a standard form. If there's an ICC profile with CICP information inside of it the extract should be handled by ImageReader.

Makes sense, there are also edge cases with old ICC profiles that encode sRGB or sRGB-with-display-P3 primaries which could require a lot of special case logic to determine are practically indistinguishable to some CICP combination.

I've updated the `color_profile_to_*` documentations again and added specific error codes to point users to use the other function.

(The fact that the `color_profile_to_icc` and `color_profile_to_cicp` implementations always have similarish logic is unfortunate, but the main ways I see to avoid this are either to define specific enums for each format's color profile (e.g. PngColorProfile, AvifColorProfile, etc.), which end up making the code quite a bit more verbose and may not be worth it yet; or to define the default color_profile_to_* implementations to _recursively_ call each other so that medium complexity formats could just define exactly one of the two, which has its own major pitfalls.)

src/io/decoder.rs
```diff
@@ -236,6 +282,23 @@ pub struct DecodedImageAttributes {
...
+/// Color profile information based on coding independent code points (see [`Cicp`])
+/// and additional HDR tone mapping parameters.
+///
+/// This is an alternative system to ICC-profile based mapping of colors from the
+/// image's numerical values to whatever space the user has. HDR continues to be
+/// actively standardized so new properties may be added to this struct over time
+/// as image formats add them.
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct CicpProfile {
```
197g: To be useful to external decoders this non-exhaustive struct must have a public constructor. Here, probably, `from_sdr_profile(Cicp)`

src/io/decoder.rs
197g: Okay, the default implementation is however strange. The documentation says `Ok(None)` is definitive evidence for the absence of a color profile, yet, the usual assumption when `icc_profile` returns data should be to use it which conflicts with this default implementation when a `Decoder` only implements the latter.

197g: > or to define the default color_profile_to_* implementations to recursively call each other so that medium complexity formats could just define exactly one of the two, which has its own major pitfalls.

Yes, clearly recursive calls here are super dangerous for partially defaulted implementations.

197g: The role for this `color_profile_to` return value is for a profile that's used with the `DynamicImage` representation, i.e. the current compute context. I'm not so quick to discard conflating them to one method though. The alternative I do wonder about would have an opaque `struct` return type:

```rust
/// A software-defined color profile.
pub struct ColorProfile { … }

impl ColorProfile {
    /// Use a color profile from a possibly synthetic ICC.
    pub fn from_icc(…) -> Self { … }

    pub fn from_cicp(_: Cicp) -> Self { … }

    /// Indicate that there is no well-defined profile that can be used.
    ///
    /// The image has no (currently) supported representation for the color profile
    /// but explicitly does not want you to fall back to the ICC profile that is
    /// included as a binary representation only.
    pub fn skip_icc() -> Self { … }

    /// Returns a Cicp equivalent of this profile if one can be determined.
    pub fn as_cicp(&self) -> Option<Cicp> { … }

    /// TBD: I wouldn't want to include this now but this would roundtrip at least to an encoder of the same format. Plus some query.
    pub fn set_native_representation(&mut self, Box<dyn Any>) { … }
}
```

That would allow us to extend it later with hints for the cicp conversion by storing that in the profile. The shape of the HDR interaction is more vague but it seems somewhat possible as an extension point there, too. Then this could be one method. As for the guidance it would:

```rust
trait ImageDecoder {
    /// An explicit color profile to be used with the image data. / / When
    /// `None` is returned (the default) then an auxiliary `icc_profile` metadata
    /// can be consulted. Formats can deny this default, for instance since they
    /// encode such a chunk but disallow its use, by returning an explicit other
    /// profile or by returning `ColorProfile::unsupported`.
    fn color_profile(&self) -> ImageResult<Option<ColorProfile>> { Ok(None) }

    /// Return the data of an encoded ICC metadata file contents.
    ///
    /// A synthetic profile or color data indicated in the file by means other
    /// than a fully encoded ICC profile should be returned from
    /// [`Self::color_profile`] instead.
    fn icc_profile(&self) -> ImageResult<Option<Vec<u8>>> { … }
}
```

Then for formats:
- png returns `Ok(Some(ColorProfile::from_cicp)))` if there's a CICP chunk
- otherwise png returns `Ok(None)` if there's an ICC chunk otherwise
- otherwise png returns `Ok(Some(…))` if there are other chunks that describe a profile [in priority order](https://www.w3.org/TR/png-3/#color-chunk-precendence).
- bmp/hdr return a synthetic profile instead of its current `icc_profile`
- jpeg and webp would not need any adjustment to current main
- heif returns cicp for `nclr`, or `from_icc` for a `rICC` with the ''reduced icc'' clusterfuck, but otherwise `skip_icc`. We still always get the full `prof` ICC profile through `icc_profile` when one such chunk exists

mstoeckl: After implementing it, I do think you are right about a ColorProfile type being cleaner, and have updated this PR to use one. (Although the design I've used is a bit different.) The old multifunction approach has been put into https://github.com/mstoeckl/image/tree/cicp-split. 

To keep the logic a bit cleaner, I've made the new `DecodedColorProfile` hold the entire color profile, whatever it is. (If we want to avoid copying the ICC profile or future HDR gain curves more than once, DecodedColorProfile could hold a reference to the original ImageDecoder, requiring users to evaluate it immediately, and making the return value for `to_icc` use `Cow`. But this may be somewhat less usable, and ICC profiles are usually small, and even the print profiles generally aren't more than a few MB, last I checked.)

src/io/decoder.rs
mstoeckl: This type was replaced by DecodedColorProfile.

src/codecs/bmp/decoder.rs
```diff
@@ -2448,6 +2460,31 @@ impl<R: BufRead + Seek> ImageDecoder for BmpDecoder<R> {
...
+            }
+            Some(ColorSpaceInfo::EmbeddedIcc(_)) => Ok(Some(DecodedColorProfile::from_icc(
+                self.icc_profile.clone().expect("icc was loaded"),
+            ))),
+            Some(ColorSpaceInfo::Srgb) => Ok(Some(DecodedColorProfile::from_plain_cicp(Cicp {
+                primaries: CicpColorPrimaries::SRgb,
+                transfer: CicpTransferCharacteristics::SRgb,
+                matrix: CicpMatrixCoefficients::Identity,
+                full_range: CicpVideoFullRangeFlag::FullRange,
+            }))),
```
RunDevelopment:
```suggestion
            Some(ColorSpaceInfo::Srgb) => Ok(Some(DecodedColorProfile::from_plain_cicp(Cicp::SRGB))),
```

src/codecs/qoi.rs
```diff
@@ -42,6 +63,34 @@ impl<R: Read> ImageDecoder for QoiDecoder<R> {
...
+        // like normal maps; the former requires different texture filtering due
+        // to a nonlinear transfer function.
+        Ok(Some(DecodedColorProfile::from_plain_cicp(
+            match self.decoder.header().colorspace {
+                ColorSpace::Srgb => Cicp {
+                    primaries: CicpColorPrimaries::SRgb,
+                    transfer: CicpTransferCharacteristics::SRgb,
+                    matrix: CicpMatrixCoefficients::Identity,
+                    full_range: CicpVideoFullRangeFlag::FullRange,
+                },
```
RunDevelopment:
```suggestion
                ColorSpace::Srgb => Cicp::SRGB,
```

src/codecs/hdr/decoder.rs
```diff
@@ -345,6 +361,36 @@ impl<R: Read> ImageDecoder for HdrDecoder<R> {
...
+                    ),
+                ),
+            ));
+        }
+        Ok(Some(DecodedColorProfile::from_plain_cicp(Cicp {
+            primaries: CicpColorPrimaries::Unspecified,
+            transfer: CicpTransferCharacteristics::Linear,
+            matrix: CicpMatrixCoefficients::Identity,
+            full_range: CicpVideoFullRangeFlag::FullRange,
+        })))
```
RunDevelopment, line 392: Why are the primaries unspecified here? This is different from `Cicp::SRGB_LINEAR`. Same for QOI.

I'm going to run this with DDS (it has sRGB and "linear"), so some guidance on what implementors should return (maybe in the method docs) would be nice.

RunDevelopment (commented):
Just some thoughts. I don't know a lot about color, so please take what I say with a grain of salt.

src/codecs/bmp/decoder.rs
```diff
@@ -359,41 +358,61 @@ impl ColorSpaceInfo {
     }
 }
 
+/// FXPT2DOT30, 2.30 fixed-point
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct Fxpt2Dot30(u32);
```
awxkee: It seems to me microsoft followed its own convention where the sign is declared as part of the fixed point format, so instead of Q1.30 this is Q2.30, as these seem to be [signed types](https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.16299.0/um/wingdi.h#L705) in microsoft SDK. That's actually makes sense, since those coordinates might be negative.

src/codecs/hdr/decoder.rs
awxkee: Seems the primaries are actually specified in the format as sRGB

Image

ofc if this URL is a correct one https://radsite.lbl.gov/radiance/refer/filefmts.pdf

src/codecs/bmp/decoder.rs
mstoeckl: That makes things simpler, done.

src/codecs/qoi.rs
mstoeckl: Done.

src/codecs/bmp/decoder.rs
mstoeckl: The CIE XYZ color space is constructed to ensure X,Y,Z are always nonnegative, see https://en.wikipedia.org/wiki/CIE_1931_color_space#CIE_XYZ_color_space, so I don't think this is likely. The definition of xyY also ensures x+y+z=1, so I expect x and y to only be valid if they are in the range [0,1]. Within this range, the high bit of 2.30 fixed point will never be set, so for valid color primaries it shouldn't make a difference whether FXPT2DOT30 is signed or not. So I think it is safe to make this change, and have done so.

src/codecs/hdr/decoder.rs
mstoeckl: This is the correct spec for Radiance HDR. Unfortunately, the spec does _not_ use sRGB primaries, but the BT470 system B primaries, whose green coordinate is slightly different (x=0.290, y=0.600) from sRGB (x=0.300, y=0.600). 

There is also the issue that the white point of 1/3, 1/3 doesn't match either System B or sRGB primary specifications. I don't know enough color management to say whether the white point is a major issue for the types of images I've seen HDR used: mainly environment maps/light sources, for which I'm not convinced white point adaptation is useful.

In practice, I expect everyone just converts data with sRGB primaries into HDR, and assumes sRGB primaries when decoding, and ignores this part of the spec.

src/codecs/bmp/decoder.rs
197g: Out-of-gamut colors are well-defined, algebraically at least, even though they may have no physical meaning. Afterall, some "subtractive" part is pretty much how the experimental setup for color matching CIE XYZ was defined in the first place and why the observer functions are in part negative. It's not even true that all colors lie in 1931's XYZ given that the blue and red monochromatic light sources were rather narrower than appropriate. (Should I be buying CIE 170:2-2015 for 190€? Idk).

At the end of the day, primaries are just base points so that the linear interpolation region covers more the physically possible color space sensibly. If you want to reduce clipping of computational artifacts it's not absurd to have negative x or y coordinates. And clearly the winsdk considers out-of-gamut definitions to be possible.

PR #3071, opened 2026-07-17
feat: impl Display for FilterType

Its-Just-Nans:
Hello

What do you think about impl display for FilterType ?

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -116,7 +116,7 @@ pub enum FilterType {
 
 impl std::fmt::Display for FilterType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-       match self {
+        match self {
             FilterType::Nearest => write!(f, "Nearest"),
             FilterType::Triangle => write!(f, "Triangle"),
             FilterType::CatmullRom => write!(f, "CatmullRom"),
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -114,6 +114,18 @@ pub enum FilterType {
     Lanczos3,
 }
 
+impl std::fmt::Display for FilterType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+       match self {
+            FilterType::Nearest => write!(f, "Nearest"),
+            FilterType::Triangle => write!(f, "Triangle"),
+            FilterType::CatmullRom => write!(f, "CatmullRom"),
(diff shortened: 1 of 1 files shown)


RunDevelopment:
What's the motivation for this?

As is, this implementation of `Display` behaves the same as the derived `Debug`-printing with `{:?}`.

Its-Just-Nans:
> What's the motivation for this?
> 
> As is, this implementation of `Display` behaves the same as the derived `Debug`-printing with `{:?}`.

Yes, it will be the same but it will use the Display Trait instead of the Debug which is more correct in case of printing it

RunDevelopment:
> which is more correct in case of printing it

What do you mean by "correct"? Could you please explain when and why the `Display` trait should be preferred over the `Debug` trait.

Its-Just-Nans:
> > which is more correct in case of printing it
> 
> What do you mean by "correct"? Could you please explain when and why the `Display` trait should be preferred over the `Debug` trait.

Debug is for developers and Display for end user. Therefore I prefer to use Display

RunDevelopment:
> Debug is for developers and Display for end user.

That much the documentation of those traits says...

I'm asking why *you* chose `Display` *here*. I'm still asking for motivation, because you still haven't given any. I don't see any upsides to having a redundant `Display` impl, so what do you see that I don't?

src/imageops/sample.rs
```diff
@@ -114,6 +114,18 @@ pub enum FilterType {
     Lanczos3,
 }
 
+impl std::fmt::Display for FilterType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            FilterType::Nearest => write!(f, "Nearest"),
+            FilterType::Triangle => write!(f, "Triangle"),
+            FilterType::CatmullRom => write!(f, "CatmullRom"),
```
197g: There's different requirements here, end-User display implementations should prefer plain english and avoid _code_ formatting. We might certainly consider if `Display` is applicable but making it work like `Debug` is not going to happen as I find that simply misleading. Since I can't really convince myself of any particular capitalization, I'm afraid that `FilterType` has too narrow a scope for implementing `Display`.

The `OsString::display()` method demonstrates another alternative to be _considered_ (not done, considered!). Maybe there's context in which all the above ambiguity is resolved and an expected (english) representation exists?

197g (changes requested):

src/imageops/sample.rs
Its-Just-Nans: Okay I've edited the MR, feel free to close it if needed

PR #3074, opened 2026-07-18
Remove unnecessary `#[allow(clippy::chunks_exact_to_as_chunks)]`

RunDevelopment:
I fixed the bug in clippy two weeks ago, so I *think* our CI should already have the fix since it uses nightly (toolchain).

Follow-up to #3058.


197g (approved):
You fixed the bug? Nice! Thank you, very much.

Issue #3076, opened 2026-07-19
AvifDecoder::new() fully decodes the picture via dav1d before Limits are ever applied — configured limits give zero protection


scadastrangelove:
## Summary

`AvifDecoder::<R>::new()` doesn't just parse a header — it feeds the entire AV1 payload to `dav1d::Decoder` and drains it to completion (`read_until_ready`), fully decoding the picture (and alpha picture, if present) into YUV plane buffers sized from the bitstream's own header. Only *after* this full decode does `dimensions()` become available. Since `AvifDecoder` never overrides `ImageDecoder::set_limits`, and `ImageReader::decode()`'s dispatch (`make_decoder`) constructs the decoder *before* ever calling `set_limits`/`reserve`, a caller who does everything the API asks for (`reader.limits(Limits{max_image_width: Some(N), ..})` then `.decode()`) gets **zero protection** against oversized AVIF frames — the memory/CPU cost is already spent by the time the limit check can even see the dimensions.

## Details

Confirmed by direct source reading against `image` 0.25.10 (not a hypothesis — this is the exact current code):

- `src/codecs/avif/decoder.rs`, `AvifDecoder::new()`: constructs `dav1d::Decoder`, calls `send_data`, and calls `read_until_ready(&mut primary_decoder)` — this blocks until the full picture is decoded — before returning `Self { picture, alpha_picture, .. }`. `dimensions()` only reads `self.picture.width()/height()` — post-decode data.
- `AvifDecoder`'s `impl ImageDecoder` block has no `set_limits`/`total_bytes` override, so it falls through to the default trait impl (`src/io/decoder.rs`), which calls `check_dimensions` against `dimensions()` — necessarily after the decode already happened.
- `src/io/image_reader_type.rs`'s `decode()`: `make_decoder(...)` (which for `ImageFormat::Avif` is literally `avif::AvifDecoder::new(reader)?`) is called at line 315, **before** `limits.reserve(decoder.total_bytes())` (line 319) and `decoder.set_limits(limits)` (line 320).
- The crate's own doc comment on `make_decoder` explicitly concedes this: *"For all formats except PNG, the limits are ignored and can be set with `ImageDecoder::set_limits` after calling this function."* PNG is special-cased (`PngDecoder::with_limits`) specifically because its constructor isn't cheap — but AVIF's constructor is by far the *most* expensive of all ~15 backends (it does the entire pixel decode), and wasn't given the same treatment.

## Impact

[...]

PR #3085, opened 2026-07-19
avif: reject color/alpha picture dimension mismatch instead of silent corruption

### LINKED ISSUE #3084: AVIF alpha-plane compositing has no dimension-equality check before zip — mismatch silently corrupts alpha instead of erroring
## Summary

`AvifDecoder::read_image`'s alpha-plane compositing loop `zip`s the main picture's output buffer against the independently-decoded alpha picture's plane data, with no check that the two pictures actually have matching dimensions — `Iterator::zip`'s silent-truncation-to-the-shorter-iterator semantics mean a dimension mismatch produces incomplete/incorrect alpha compositing rather than an error.

## Details

Confirmed by direct source reading against `image` 0.25.10, `src/codecs/avif/decoder.rs`:

```rust
if let Some(picture) = self.alpha_picture {
    if picture.pixel_layout() != PixelLayout::I400 {
        return Err(...);  // checks chroma layout only
    }
    let stride = picture.stride(PlanarImageComponent::Y) as usize;
    let plane = picture.plane(PlanarImageComponent::Y);
    for (buf, slice) in Iterator::zip(
        buf.chunks_exact_mut(width as usize * 4),   // sized by the MAIN picture's width
        plane.as_ref().chunks_exact(stride),         // sized by the ALPHA picture's own stride
    ) {
        ...
    }
}
```

The primary picture and the alpha picture are decoded via two entirely separate `dav1d::Decoder` instances (`AvifDecoder::new`), each from its own coded data (`ctx.primary_item_coded_data()` / `ctx.alpha_item_coded_data()`). There is no check anywhere that `alpha_picture`'s width/height match the primary `picture`'s width/height before this loop. `Iterator::zip` silently stops at the shorter of its two input iterators, so a mismatch doesn't error — it just applies alpha values to only part of the image (or leaves remaining rows with whatever default/prior alpha was in the buffer).

## Impact

A crafted AVIF file whose alpha auxiliary image has different dimensions than its primary image produces silently incorrect (partially-composited or entirely default-alpha) output rather than a decode error — a data-integrity bug, not a crash, in a format specifically used for images requiring correct transparency (UI assets, compositing pipelines).

## Suggested direction

Not submitting a PR — the fix is a straightforward dimension-equality check (`alpha_picture.width()/height() == picture.width()/height()`) before the compositing loop, returning a decode error on mismatch, but flagging for maintainer confirmation since I haven't constructed an actual crafted-AVIF PoC to confirm dav1d doesn't already reject mismatched auxiliary-image dimensions somewhere in its own validation (in which case this code is unreachable dead-code-for-safety rather than a live gap) — this is a source-level reachability argument, not an empirically-triggered one.

scadastrangelove:
Fixes #3084.

`AvifDecoder::read_image` (both the 8-bit and 16-bit paths) composited the alpha plane into the output buffer with `Iterator::zip`, but the primary picture and the alpha picture are decoded independently — each via its own `dav1d::Decoder`, from its own coded data (`AvifDecoder::new`). Nothing checked that the two `dav1d::Picture` results actually had matching width/height before the `zip`.

`Iterator::zip` silently stops at the shorter of its two inputs, so a mismatch didn't error — it corrupted the composited image instead.

**Verified against current `main` with a real dav1d-decoded crafted AVIF** (color item 8x8, alpha item's own independently-encoded AV1 bitstream 4x4, re-muxed via `avif-serialize`):
- **Pre-fix**: `read_image()` returned `Ok`. Rows 0-3 correctly got alpha values from the 4x4 alpha picture, but rows 4-7 got stray alpha bytes not sourced from the visible alpha image at all — `dav1d`'s picture-buffer padding beyond the coded height happened to satisfy the zip's row count. This is silent corruption, not simple truncation.
- **Post-fix**: the same file now returns `Err("Format error decoding Avif: Alpha plane dimensions (4x4) do not match color plane dimensions (8x8)")`.

The original issue only flagged the 8-bit path; while fixing it I found the identical unchecked `zip` in the 16-bit path (`process_16bit_picture`) and fixed both.

## Changes
- Add a dimension-equality check ahead of both zip loops (8-bit and 16-bit), returning a new `AvifDecoderError::AlphaPlaneDimensionMismatch` instead of silently compositing.
- Add a regression test (`tests/avif_alpha_dimension_mismatch.rs`) that crafts a real, dav1d-decodable dimension-mismatched AVIF at test time (via the crate's own `ravif`-backed encoder + `mp4parse` + `avif-serialize`, no external binary fixture needed) and asserts decoding now errors. Confirmed the test fails on unpatched `main` and passes with this fix.

## Testing
Full suite (`cargo test --features avif,avif-native,default-formats,rayon --lib --tests`): all green except one pre-existing, unrelated failure in `tests/limits.rs::avif` (`set_limits` not enforcing width/height before dav1d decode) — reproduces on unmodified `main` too, already tracked as #3076.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,8 @@ serde = { version = "1.0.214", optional = true, features = ["derive"] }
 pic-scale-safe = "0.1.5"
 
 [dev-dependencies]
+mp4parse = "0.17.0"
+avif-serialize = "0.8.9"
 num-complex = "0.4"
 glob = "0.3"
 quickcheck = "1"
--- a/src/codecs/avif/decoder.rs
+++ b/src/codecs/avif/decoder.rs
@@ -43,6 +43,10 @@ pub struct AvifDecoder<R> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum AvifDecoderError {
     AlphaPlaneFormat(PixelLayout),
+    AlphaPlaneDimensionMismatch {
+        color: (u32, u32),
+        alpha: (u32, u32),
+    },
     YuvLayoutOnIdentityMatrix(PixelLayout),
     UnsupportedLayoutAndMatrix(PixelLayout, YuvMatrixStrategy),
     InvalidStride(u32),
@@ -69,6 +73,12 @@ impl Display for AvifDecoderError {
                 }
                 PixelLayout::I444 => unreachable!("This option must be handled correctly"),
             },
+            AvifDecoderError::AlphaPlaneDimensionMismatch { color, alpha } => {
+                f.write_fmt(format_args!(
+                    "Alpha plane dimensions ({}x{}) do not match color plane dimensions ({}x{})",
+                    alpha.0, alpha.1, color.0, color.1
+                ))
+            }
             AvifDecoderError::UnsupportedLayoutAndMatrix(layout, matrix) => f.write_fmt(
                 format_args!("YUV layout {layout:?} on matrix {matrix:?} is not supported"),
             ),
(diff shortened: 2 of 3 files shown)


197g (commented):
Looks like libavif also has an error code for this, so this is reasonable (with the clippy complaint fixed).

PR #3090, opened 2026-07-26
avif: stop read_until_ready spinning when no picture can arrive

FrederickStempfle:
`read_until_ready` loops until dav1d hands back a picture:

```rust
Err(dav1d::Error::Again) => match decoder.send_pending_data() {
    Ok(()) => {}
    Err(dav1d::Error::Again) => {}
```

`send_pending_data` returning `Ok` means there is no pending data left to submit, so if
`get_picture` still yields `Again` the decoder is asking for data that cannot arrive, and
neither call changes any state on the next pass.

The attached 1509-byte file reaches exactly that state and spins there indefinitely. I
instrumented the loop to count branches: `send_pending_data` returned `Ok` on every one of
~300 million iterations in 25 seconds, and `Again` zero times. This is inside
`AvifDecoder::new`, so `load_from_memory_with_format` simply never returns.

Bail out once the decoder has been drained and still has no picture. One retry is kept,
since the flush itself can be what lets a picture through.

Found with a libFuzzer target for AVIF. There isn't one in `fuzz/` at the moment — happy to
send that separately if it would be useful.

The regression test decodes on a worker thread with a timeout, so that a regression fails
the suite rather than hanging it. The file is also placed under `tests/bad/`, but note
`bad_images.rs` will not currently pick it up: it walks `./bad`, while the images live in
`tests/bad`, so the directory it scans doesn't exist and the test passes vacuously (adding a
valid image there doesn't fail it either). I left the file in place so it is covered once
that is sorted, but didn't want to fold an unrelated fix into this PR.


197g (approved):
Thanks, that's not the most intuitive required code flow 🙂

197g:
> Found with a libFuzzer target for AVIF. There isn't one in fuzz/ at the moment — happy to
send that separately if it would be useful.

Though fuzzing of formats that we can't quite match in bandwidth are always in a rough spot. Better than them not existing though so go ahead.

PR #3092, opened 2026-08-02
Move small-sigma Gaussian blur return to a helper

fyrsta7:
## Summary

This moves the small-sigma identity return used by `GaussianBlurParameters::new_from_sigma` into a cold helper, leaving the normal positive-sigma path as the straight-line path.

## Why this can improve performance

The `sigma <= 0.0` case returns an identity kernel and is not the normal blur construction path. Keeping that path in a cold helper lets the common Gaussian-kernel setup stay a little smaller and more direct.

## Validation

- `git diff --check`
- `cargo check`

## Benchmark

Current machine, release microbenchmark, 9 samples, median:

| Benchmark | Base | This change | Delta |
|---|---:|---:|---:|
| 1024 `GaussianBlurParameters::new_from_sigma` calls | 5,228.287 ns | 5,214.858 ns | 0.26% faster |

### DIFF
--- a/src/imageops/sample.rs
+++ b/src/imageops/sample.rs
@@ -1231,12 +1231,7 @@ impl GaussianBlurParameters {
             // Any (normalized) kernel of size 1 is the identity, so sigma
             // doesn't matter. However, we pick sigma=1 to avoid  potential
             // issues with NaN, infinities, and subnormals.
-            return GaussianBlurParameters {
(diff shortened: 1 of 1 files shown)


fyrsta7:
Closing this draft because follow-up benchmarking did not show a large enough performance gain to justify this as an upstream performance PR.

PR #3100, opened 2026-08-18
Add OpenEXR support for single-channel Luma32F images

### LINKED ISSUE #3016: Exr: Support reading and writing `Luma<f32>` images

ChrisJr404:
Fixes #3016.

Now that `ColorType::Luma32F` exists, this teaches the OpenEXR codec to read and write single-channel 32-bit float luma images, which previously failed.

## Behavior

Following the [OpenEXR channel-naming recommendation](https://openexr.com/en/latest/TechnicalIntroduction.html#channel-names), luma is stored in a channel named `Y`.

- **Decoder**: layer selection now prefers a non-deep RGB(A) layer (unchanged), and falls back to a non-deep `Y`-only layer, which is decoded as `Luma32F`. When a file contains both RGB and `Y` channels, RGB is still preferred.
- **Encoder**: `ExtendedColorType::L32F` buffers are written to a single `Y` channel.

So the example from the issue now works:

```rust
let image = ImageBuffer::from_fn(16, 16, |x, y| Luma([(x * y) as f32 / 225.0]));
image.save("test.exr").unwrap();
```

Luma + alpha is intentionally left out of scope, as noted in the issue.

## Tests

Added `roundtrip_luma` (generate a `Luma32F` image, encode, decode, compare), mirroring the existing `roundtrip_rgb`/`roundtrip_rgba` tests. The decoder is exercised on a real `Y`-channel `.exr` produced by the new encoder path. All existing OpenEXR tests still pass, confirming RGB/RGBA decoding is unchanged.

`cargo test --features exr,hdr --lib codecs::openexr` — 6 passed. `cargo clippy --features exr` — no warnings. `cargo fmt --check` — clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@ Breaking Changes:
 Features:
  - Added `RbgImage::from_raw_bgr` and `RbgaImage::from_raw_bgra` constructors, which convert
    from `BGR(A)` with an optimized specialization for bytes of `BGRA`
+ - The OpenEXR codec can now read and write single-channel luma (`Luma32F`) images
+   via the `Y` channel (#3016)
 
 Structural changes:
   - Increased MSRV to 1.88.0 (from 1.85.0)
--- a/src/codecs/openexr.rs
+++ b/src/codecs/openexr.rs
@@ -3,18 +3,21 @@
 //! OpenEXR is an image format that is widely used, especially in VFX,
 //! because it supports lossless and lossy compression for float data.
 //!
-//! This decoder only supports RGB and RGBA images.
-//! If an image does not contain alpha information,
-//! it is defaulted to `1.0` (no transparency).
+//! This decoder supports RGB, RGBA, and single-channel luminance images.
+//! RGB(A) images use the `R`, `G`, `B` (and optional `A`) channels, while
+//! luminance images use a single `Y` channel. If an RGB image does not contain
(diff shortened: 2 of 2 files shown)


RunDevelopment:
https://github.com/image-rs/organization/blob/main/CONTRIBUTING.md#llm-policy

Given that you made >30 PRs to >15 repos yesterday and the LLM-speak in your PR description, I'm going to assume an LLM was involved substantually.

PR #3102, opened 2026-08-26
PNG: Avoid allocations when disposing APNG frames

pnn64:
This PR changes APNG background and previous-frame disposal to operate directly on the underlying image rows. Previously, these paths created temporary buffers before clearing or restoring a region, resulting in unnecessary allocations and memory copies. The new implementation performs the same work directly in the frame buffers without changing the public API or using unsafe code.

The existing six WPT APNG reference tests pass. Local measurements produced the following results:
| Disposal mode | Allocations | Allocated bytes | Reallocations | Latency |
| --- | ---: | ---: | ---: | ---: |
| Background | 20 → 19 | 331,504 → 306,928 | 10 → 1 | ~30% lower |
| Previous | 20 → 19 | 315,120 → 306,928 | 1 → 1 | ~16% lower |

The benchmarks were run locally and are not included in this PR.

Tested with `cargo test --test reference_images -- wpt` and `cargo clippy --lib -- -D warnings`.

AI disclosure: I used Codex with GPT-5.6 Sol at `xhigh` reasoning effort to help identify the allocation, implement the optimization, and construct the local benchmarks. I reviewed and tested the resulting change.

### DIFF
--- a/src/codecs/png.rs
+++ b/src/codecs/png.rs
@@ -25,8 +25,8 @@ use crate::math::Rect;
 use crate::metadata::LoopCount;
 use crate::utils::vec_try_with_capacity;
 use crate::{
-    DynamicImage, GenericImage, GenericImageView, ImageDecoder, ImageEncoder, ImageFormat,
-    ImageLayout, Limits, Luma, LumaA, Rgb, Rgba,
+    DynamicImage, ImageDecoder, ImageEncoder, ImageFormat, ImageLayout, Limits, Luma, LumaA, Rgb,
+    Rgba,
 };
 
 // http://www.w3.org/TR/PNG-Structure.html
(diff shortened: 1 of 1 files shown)


src/codecs/png.rs
```diff
@@ -528,10 +521,12 @@ impl<R: BufRead + Seek> ApngDecoder<R> {
...
-                let region_previous = previous.sub_image(rect);
-                current
-                    .copy_from(&region_previous.to_image(), rect.x, rect.y)
-                    .unwrap();
+                copy_pixel_region(
+                    current.as_mut_bytes(),
+                    &layout.layout,
+                    previous.as_bytes(),
+                    &rect,
+                );
```
197g: I'm slightly surprised this would be much faster than the current implementation given that we have a fast path for this copy now (through `GenericImage::copy_from_samples`). It should only be necessary to remove the explicit `to_image` clone of the input image, the internal dispatch should recognize the matrix regularity and take care of the rest with only a few more comparisons.

pnn64: Good catch, that's fair. I compared removing only `to_image()` with the specialized row-copy helper. The helper was around 2.5% faster, but Criterion considered the difference within noise. I've switched to copying directly from the subimage and removed the extra helper; this still avoids the allocation.

197g (approved):

PR #3103, opened 2026-08-26
CI churn cleanup

197g:
Should cleanup CI once more. One clippy fix, and the reference image in Catmul interpolation was no longer accurate after an update to pic-scale-safe very slightly modified a coefficient.


tests/regression.rs
```diff
@@ -225,18 +225,25 @@ fn resizing_with_catmul() {
...
     resizable.resize_exact(nw, nh, FilterType::CatmullRom);
-    assert_eq!(resizable, expected);
+    assert_eq!(resizable, expected, "Output (resized, target): {:?}", {
+        let path = base.join("output/image/resize-with-alpha-original-half-size.png");
+        save_image(&path, &resizable).unwrap();
+        (
+            path.display().to_string(),
+            target_path.display().to_string(),
+        )
+    });
```
RunDevelopment, line 246: This looks hacky to me but whatever. Not a blocker.

RunDevelopment (approved):

tests/regression.rs
197g: I'd reuse the abstractions in the regression tests but it's always a bit of a hassle sharing modules between them (and that one has its own harness, so we can't move them over easily either).

PR #3105, opened 2026-08-28
Bound Pixel iterator for empty images

197g:
The approach here is to fully avoid any modification to its internal state when an empty image is provided. By slightly strengthening its invariants, x < self.width except for empty images, we can achieve this without additional branches in its `next` method. This also avoids the peculiar code path of incrementing `y` after the last pixel when the Iterator is not used as a fuse'd one.

### DIFF
--- a/src/images/generic_image.rs
+++ b/src/images/generic_image.rs
@@ -63,7 +63,13 @@ pub trait GenericImageView {
     where
         Self: Sized,
     {
-        let (width, height) = self.dimensions();
+        let (width, mut height) = self.dimensions();
+
+        // Ensures `height` can be used as a consistent bound so that only one check is required in
+        // the iteration of the iterator.
+        if width == 0 || height == 0 {
+            height = 0;
+        }
 
         Pixels {
             image: self,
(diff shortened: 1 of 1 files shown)


RunDevelopment:
This is done weirdly IMO.

I think maintaining the following invariants would be simpler:

1. If the image is empty, $w = h = 0$. (new)
2. $w = 0 \implies x = 0$ and $w \ne 0 \implies x < w$.
3. $y \le h$
4. $y = h \implies x = 0$ (new)

This basically just describes a well-formed walk over all pixel positions with $(x=0,y=h)$ being the stop condition. IMO, this is simpler, because there aren't any edge cases like "what if the image is empty, but one dimension is freakishly large?" to consider.

These invariants maintain the correctness of your new `next` implementation, but they also simplify `size_hint` since they guarantee `total >= done`. So no `checked_sub` required.

197g:
I think that _is_ equivalent for the invariants I'm maintaining? The `w = 0` isn't necessary as `h = 0` would ensure that the position is never incremented (and both products with `width` in `total` and `done` will be zero). The `checked_sub` is there out of an abundance of caution so to say, not due to the iteration.

RunDevelopment:
It is, but only if the implementation of `next` is part of the invariants. I don't like that the reasoning currently flows as "`next` is implemented like this, therefore I can assume X" instead of "the type maintains these invariants, therefore `next` does Y."

But it's also not like this would change much apart from the flow of reasoning. Implementing my suggestion would mostly come down to (re-)writing a few comments.

197g:
Still sounds good if you're up for it. Feel free to open a PR with documentation.

RunDevelopment:
Prob won't. Too many of my PRs I care about lead to nowhere, so I just don't feel motivated to write code for this code base anymore. Maybe this will change, but I'm probably not going to work on any PRs for a couple of months.

Issue #2908, opened 2026-04-11
Formats to keep/remove for 1.0


fintelia:
Another topic we should discuss ahead of 1.0 is what formats we want to include. Now that there's `image-extras` and decoding hooks, we can remove formats from this main crate without totally removing support if a small number of people are still relying on them.

We also have the option of making some formats behind a feature flag but not part of the default feature set.[^1]

---
Looking at individual formats...

* **dds:** Already moved to `image-extras`.

* **farbfeld:** Another we should probably move to `image-extras`. I haven't come across anyone using this format who didn't add it because we did.

* **exr:** This format is implemented via an external crate and has a relatively heavy dependency tree. We've gotten very little in the way of complaints / issues about this one, so I'm not sure how much it gets used? Another candidate to remove.

* **pnm:** Not sure about this one. There's definitely a lot of references to PNM in older software. But I don't have a good sense of usage. And #2841 raised the question of how to properly handle subformats which doesn't seem like it has a good option besides treating them as five different formats. 

* **qoi:** I'd somewhat lean towards keeping this one simply because it isn't too complex and I like the format. But maybe doesn't need to be enabled by default.

* **avif:** Our current codecs have a pretty heavy dependency tree, and the decoder isn't enabled by default yet (since we're still waiting to switch to a Rust implementation). I think this could make sense behind a feature flag.
 
* **tiff:** I think we'll want to keep support. Though as we add more functionality upstream like ZStandard compression and such, we might want to consider a feature flag?

* **bmp / gif / hdr / ico / jpeg / png / tga / webp:** I think we want to keep all these. 

[^1]: I've been reluctant to allow disabled-by-defaults formats in the past because I feared it would lead to pressure to add tons of obscure/rarely used formats. But I think that's less likely now that there's other avenues for them.

mstoeckl:
While decoding hooks are supported, _encoding_ hooks are not, so moving any format to image-extras would affect the behavior of `image::save_buffer` or `DynamicImage::write_to`, which could be given arbitrary user-supplied paths. I would recommend caution in removing any of the current format set. Except maybe Farbfeld.

> pnm. [...] But I don't have a good sense of usage. [...]

[...]

Issue #1765, opened 2022-07-27
Support JPEG XL


sorairolake:
This format is preliminary supported by many major web browsers such as Firefox ~~and Chromium~~ (<https://caniuse.com/jpegxl>). Also, some graphics softwares already supports it, so I think it would be useful to support it.

- [Official website](https://jpeg.org/jpegxl/)
- [Community website](https://jpegxl.info/)

Edited, candidates for inclusion:
- https://github.com/tirr-c/jxl-oxide

lifthrasiir:
J40, the JPEG XL decoder implementation currently written in C, is considering a parallel Rust version that may (hopefully, positively) affect this issue. For more information please refer to lifthrasiir/j40#10.

Quackdoc:
https://github.com/tirr-c/jxl-oxide dual licenced with MIT and BSD

fintelia:
Interesting. Do you know if that crate has been published to crates.io yet or whether there's plans to do so?

Quackdoc:
crates are slowly being published, it should be best to wait for the `jxl-render` crate, currently it has not yet been published author's page is [here ](https://crates.io/users/tirr-c)

EugeneVert:
`jxl-oxide 0.1.0` has been published to crates.io
https://crates.io/crates/jxl-oxide

mirh:
Speaking of rust and jxl, you might be interested to https://github.com/tirr-c/jxl-oxide/issues/335

ypurpl:
is this still being worked on?

fintelia:
https://github.com/image-rs/image/issues/1979#issuecomment-1826870586 has the latest status that I know of

Camille019:
[jxl-rs](https://github.com/libjxl/jxl-rs) is also a project to follow.

Some highlights from the jxl-oxide author: https://github.com/tirr-c/jxl-oxide/issues/411#issuecomment-2613849735

justin13888:
By any chance, has anyone made a decision on what decoder/encoder implementation would be used for image-rs? What were the decision factors?

I'm really hoping this progress because otherwise I might just write a fork implementation to support several unsupported formats for my own use case.

fintelia:
Closing now that #2372 is merged

PR #3108, opened 2026-09-02
Add JPEG XL decoding with jxl-rs

hjanuschka:
## Summary

- add optional JPEG XL decoding through the pure-Rust `jxl-rs` decoder
- recognize `.jxl`, `image/jxl`, raw codestream, and container signatures
- support grayscale/RGB output, alpha, 8-bit, 16-bit, float, ICC profiles, and animation
- integrate `ImageFormat::JpegXl` with the high-level reader APIs

`jxl` is intentionally not part of `default-formats`. The dependency's SIMD
features are disabled because they currently require Rust 1.89 while this
crate's MSRV is Rust 1.88.

this is created ai assisted!


RunDevelopment:
See #1765 and #2908. The current direction is that `image` will only support a few built-in formats, and more image formats are added in the form of hooks.

Issue #3104, opened 2026-08-27
[AI] Batch of bugs found by Claude


qarmin:
I'm experimenting with autonomous bug finding to simplify this very time consuming step. Unlike my previous report(https://github.com/image-rs/image/issues/2954), which was based only on checking the source code, this report verifies the findings by compiling, applying patches, and running tests to check whether the implementation is correct, what crashes, which patch fixes the problem, etc. Because of this, the number of false positives/hallucinations that AI tends to produce should be much lower.

I also checked Slint(https://github.com/slint-ui/slint/issues/13007) and pic-scale(https://github.com/awxkee/pic-scale-safe/issues/19), and it looks like it found many interesting bugs(and optimizations, but that is a topic for a different issue - here I'm only posting the bugs).

Reproducible projects, example patches and bug explanations - [image_bughunt_20260827_143543.zip](https://github.com/user-attachments/files/31514800/image_bughunt_20260827_143543.zip)

| # | Class | Candidate | Fix lines | Root cause | Introduced by |
|---|---|---|---|---|---|
| 001 | crash | `GenericImageView::Pixels::next()` zero-width row-wrap guard | +4/-0 | Row-wrap check (`x >= width`) fires before a pixel is ever emitted when `width == 0`, advancing `y` and then indexing a zero-column row | d0b2b7b |
| 003 | logic | `imageops::sample::vertical_sample` resampling accumulator seed | +1/-1 | Per-pixel accumulator seeded `Rgba([1.0; 4])` instead of `[0.0; 4]` before summing sum-to-1.0 normalized filter weights, adding a flat +1.0 bias to every resized pixel | 20a2308 |
| 006 | crash | `filter_1d.rs` Gaussian-blur arena column-padding, zero-height guard | +6/-0 | `make_columns_arenas` slices `&image[..stride]` out of an empty row buffer whenever `image_size.height == 0` in the wide-kernel arena path, instead of no-op'ing on an empty image | 159d03e |
| 007 | crash | BMP-backed ICO decoding, post-halving zero-height guard | +6/-0 | `read_metadata_in_ico_format` halves the already-validated doubled height, so an odd doubled height reintroduces an unvalidated `height == 0`, which the pixel readers' unconditional `(height - 1).try_into::<u32>().unwrap()` then panics on | 281b5e6 |
[...]

PR #3111, opened 2026-09-14
Start the vertical resize accumulator at zero

yuefdev:
`vertical_sample` starts every output pixel at `Rgba([1.0; 4])` and then adds the weighted source samples. The weights are normalized to sum to 1, so every channel ends up 1.0 too high after the vertical pass, and the horizontal pass carries that through.

With `imageops::resize` on a uniform image (any filter, 16x16 → 7x9):

| input | before | after |
|---|---|---|
| `RgbImage` `[100, 150, 200]` | `[101, 151, 201]` | `[100, 150, 200]` |
| `Rgb32FImage` `[0.25, 0.5, 0.75]` | `[1.0, 1.0, 1.0]` | `[0.25, 0.5, 0.75]` |

The seed was `(0.0, 0.0, 0.0, 0.0)` until 20a2308 replaced the tuple with a pixel, so this is a regression on `main` only; 0.25.9 isn't affected. `DynamicImage::resize` goes through `pic-scale-safe` and doesn't hit this path, but `imageops::resize` on any `GenericImageView` does.

The fix is the one-character seed change. The new test resizes uniform u8 and f32 images up and down with every filter and checks the color is unchanged; it fails on `main` and passes with the fix.

This is candidate 003 from #3104. #2937 rewrites `vertical_sample` and drops this line anyway, but it has been open for a while, so this is the minimal fix in the meantime.


fintelia (changes requested):
Please review our [LLM policy](https://github.com/image-rs/organization/blob/main/CONTRIBUTING.md#llm-policy) and update your PR as necessary

PR #3112, opened 2026-09-14
Divide thumbnail sub-block sums by the number of sub-blocks

yuefdev:
When a thumbnail block covers more than 256 * 256 source pixels, `thumbnail_sample_block` averages it as 256x256 sub-blocks, sums those averages, and divides by the sub-block count. The count is computed with floor division:

```rust
((right - left) as u64 / STEP_SIZE as u64) * ((top - bottom) as u64 / STEP_SIZE as u64)
```

but `step_by(256)` also visits the smaller sub-blocks at the right and bottom edges. For a 300x300 block that is 2x2 = 4 sub-blocks, divided by 1.

For example, `thumbnail` of a uniform 600x600 `Luma([100])` image to 2x2 returns `Luma([255])` on `main`. Any block whose side isn't a multiple of 256 gets too bright. In practice that means shrinking a large photo to a tiny thumbnail, e.g. 4000x3000 → 15x11. `DynamicImage::thumbnail` goes through this function too.

This PR rounds the count up so it matches the sub-blocks that were actually summed, and adds a test that fails on `main`. The sub-block averages are still weighted equally, as before, so edge sub-blocks count a bit more than their size. That keeps the overflow headroom the split is there for, and a uniform image now comes out exact.

This was introduced in 61c7622 (#2994). It isn't in 0.25.9. It's candidate 020 from #3104.


fintelia (changes requested):
Please review our [LLM policy](https://github.com/image-rs/organization/blob/main/CONTRIBUTING.md#llm-policy) and update your PR as necessary

PR #3113, opened 2026-09-15
BmpDecoder: Clip encoded RLE4 runs at row boundaries

telecos:
## Summary

Adds regression coverage for encoded-mode RLE4 runs that extend beyond the remaining width of a row.

The decoder already clips these runs in lenient mode, writing only the pixels that fit while consuming the complete encoded command. Strict mode continues to reject the row overflow as corrupt RLE data.

This PR does not change absolute-mode RLE4, delta commands, RLE8, RLE24, or other BMP decoding behavior.

## Regression test

The test constructs a synthetic 1x1, 4-bpp BI_RLE4 bitmap with a two-entry black/red palette and the following RLE stream:

`[2, 0x10, 0, 1]`

The encoded run declares two pixels although only one fits in the row. Lenient decoding produces one red RGB pixel, while strict decoding rejects the overflow.

This PR is incorporating part of Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/8400331


telecos:
Not required as this was fixed in #2917

PR #3114, opened 2026-09-15
BmpDecoder: Decode the first image in an OS/2 Bitmap Array

telecos:
## Summary

Recognize a 14-byte OS/2 Bitmap Array `BA` header followed by an embedded `BM` file header, and decode the first embedded bitmap. The embedded `bfOffBits` value remains an absolute stream offset. Magic-byte detection now recognizes `BA` as BMP.

Support is bounded to the first image: linked later entries are deliberately not traversed. Truncated array headers and arrays whose first embedded header is not `BM` remain rejected.

## Regression test

The tests construct bytes inline for a `BA` header, an embedded `BM` header, a 12-byte BITMAPCOREHEADER, and one padded 24-bpp red pixel. They cover direct BMP decoding, guessed-format decoding, truncation, and an invalid embedded signature.

This PR is bring parts of Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/8400331


telecos:
Discarding this change after confirming that other decoders (i.e. Safari and Firefox) do not open this file either

PR #3115, opened 2026-09-15
BmpDecoder: Accept lower declared bit depths for RLE4

telecos:
## Summary

In lenient mode, accept BI_RLE4 with a declared bit depth of 1 or 2. RLE commands remain nibble-based, while the declared depth and `colors_used` continue to resolve the palette size.

Acceptance is bounded by requiring the complete resolved palette to end at or before the absolute pixel-data offset. Strict mode, declared bit depth 3, palette overlap, and lower bit depths for other compression methods remain rejected.

## Regression test

The tests construct 1-bpp and 2-bpp BI_RLE4 bitmaps inline with a black/red palette and the stream `[2, 0x11, 0, 1]`, producing two red RGB pixels. Additional cases cover strict rejection, depth 3, and a palette that overlaps pixel data.

This PR is incorporating part of Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/8400331


telecos:
Discarding change after confirming that other decoders (i.e. Safari and Firefox) cannot open this file either.

